PR #5, opened 2020-01-20
Basic Heroku CI configuration (no nginx yet)

rpiaggio:


rpiaggio:
It's normal that Heroku CI is still failing, it doesn't have node/npm provisioning yet. It should start working after this PR is merged.

cquiroz:
How does it know to call `fullOptJS`?

rpiaggio:
> How does it know to call `fullOptJS`?

For CI, it doesn't. It calls `sbt test`. For deployment as a static resource, that's coming in the next PR ;)

rpiaggio:
BTW, we can change the command we run for test in CI.

cquiroz:
ok, then we could remove the GH Action

PR #42, opened 2020-03-10
Test proxying + invoking Grackle demo

rpiaggio:
No need to merge, just testing things out.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,35 +1,35 @@
 {
-    "buildpacks": [
+  "buildpacks": [
+    {
+      "url": "heroku/nodejs"
+    },
+    {
+      "url": "heroku/scala"
+    },
+    {
+      "url": "https://github.com/hone/heroku-buildpack-static"
+    }
+  ],
+  "environments": {
+    "test": {
+      "formation": {
+        "test": {
+          "quantity": 1,
+          "size": "performance-l"
+        }
+      },
+      "buildpacks": [
         {
           "url": "heroku/nodejs"
         },
         {
           "url": "heroku/scala"
-        },
-        {
-          "url": "https://github.com/hone/heroku-buildpack-static"
         }
       ],
-    "environments": {
-      "test": {
-        "formation": {
-          "test": {
-            "quantity": 1,
-            "size": "performance-l"
-          }
-        },
-        "buildpacks": [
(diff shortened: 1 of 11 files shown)


app.json
```diff
@@ -1,35 +1,35 @@
 {
-    "buildpacks": [
+  "buildpacks": [
```
cquiroz: This probably took a long time to get right 👍

common/src/main/scala/explore/model/RootModel.scala
```diff
@@ -14,6 +14,7 @@ import monocle.macros.Lenses
 import cats.effect.concurrent.Ref
 import japgolly.scalajs.react._
 import diode.data._
+import explore.util.Pot._
```
cquiroz: Did you internalize `Pot`?

common/src/main/webpack/dev.webpack.config.js
```diff
@@ -37,7 +37,19 @@ const Web = Merge(
       host: "0.0.0.0",
       hot: true,
       contentBase: [__dirname, parts.rootDir],
-      historyApiFallback: true
+      historyApiFallback: true,
+      proxy: {
+        "/api/tasks": {
```
cquiroz: nice 👍

static.json
```diff
@@ -3,5 +3,13 @@
     "clean_urls": false,
     "routes": {
       "/**": "index.html"
+    },
+    "proxies": {
+      "/api/tasks": {
+        "origin": "https://todo-mongo-graphql-server.herokuapp.com"
+      },
+      "/api/grackle-demo": {
+        "origin": "https://${GRACKLE_DEMO_BASE}"
```
cquiroz: Does this come from heroku?

rpiaggio: We define it as an environment variable in Heroku. This allows us to use the same app pointing to different servers in staging and production.

common/src/main/scala/explore/model/RootModel.scala
rpiaggio: Not yet. For the moment in that object we have some utils for working with it.

rpiaggio:
I'll go ahead and merge so we can reuse the mechanism.

PR #65, opened 2020-03-27
[WIP] Edit conditions

rpiaggio:


common/src/main/scala/explore/components/forms/EnumSelect.scala
```diff
@@ -0,0 +1,65 @@
...
+import japgolly.scalajs.react.vdom.html_<^._
+import react.common.ReactProps
+import react.semanticui.modules.dropdown._
+import scala.scalajs.js.JSConverters._
+import crystal.react.io.implicits._
+
+/**
+  * Produces a dropdown menu, similar to a combobox
+  */
+final case class EnumSelect[A: Enumerated: Show](
```
cquiroz, line 20: I guess this will go to gsp-ui

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
```diff
@@ -0,0 +1,89 @@
...
+  protected def componentBuilder[D, A] =
+    ScalaComponent
+      .builder[Props[D, A]]("SubscriptionRender")
+      .initialState(State[D, A]())
+      .render { $ =>
+        <.div(
+          $.state.renderer.whenDefined(
+            _ { resultsPot =>
+              <.div(
+                resultsPot.renderPending(_ => Icon(name = "spinner", loading = true, size = Large)),
```
cquiroz, line 56: Nice

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
```diff
@@ -0,0 +1,89 @@
...
+  type Props[D, A] = SubscriptionRender[D, A]
+
+  final case class State[D, A](
+    subscription: Option[
+      GraphQLStreamingClient[IO]#Subscription[D]
+    ] = None,
+    renderer: Option[StreamRenderer[Pot[A]]] = None
+  )
+
+  implicit def propsReuse[D, A]: Reusability[Props[D, A]] = Reusability.always
```
cquiroz, line 44: I suppose this is `always` because both fields in props can't really be checked

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
```diff
@@ -0,0 +1,89 @@
...
+                resultsPot.render(results => $.props.valueRender(results))
+              )
+            }
+          )
+        )
+      }
+      .componentWillMount { $ =>
+        $.props.subscribe
+          .flatMap { subscription =>
+            implicit val cs = $.props.cs
```
cquiroz, line 66: Nice, the context shift goes with the props

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
```diff
@@ -0,0 +1,89 @@
...
+import cats.effect.ContextShift
+
+final case class SubscriptionRender[D, A](
+  subscribe:      IO[GraphQLStreamingClient[IO]#Subscription[D]],
+  streamModifier: fs2.Stream[IO, D] => fs2.Stream[IO, A] = identity[fs2.Stream[IO, D]] _
+)(
+  val valueRender: A => VdomNode,
+  val onNewData:   IO[Unit] = IO.unit
+)(
+  implicit val cs: ContextShift[IO]
```
cquiroz, line 28: Do we need it implicit here? we are not going to use it inside props or do we?

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
```diff
@@ -0,0 +1,89 @@
...
+) extends ReactProps {
+  override def render: VdomElement =
+    SubscriptionRender.component(this.asInstanceOf[SubscriptionRender[Any, Any]])
```
cquiroz, line 31: Why's the casting needed?

common/src/main/scala/explore/components/graphql/SubscriptionRenderMod.scala
```diff
@@ -0,0 +1,95 @@
...
+import cats.effect.Timer
+
+final case class SubscriptionRenderMod[D, A](
+  subscribe:      IO[GraphQLStreamingClient[IO]#Subscription[D]],
+  streamModifier: fs2.Stream[IO, D] => fs2.Stream[IO, A] = identity[fs2.Stream[IO, D]] _
+)(
+  val valueRender: (A, StreamRendererMod.ModState[A]) => VdomNode,
+  val onNewData:   IO[Unit] = IO.unit
+)(
+  implicit val cs: ContextShift[IO],
```
cquiroz, line 31: I'm wondering about these being `implicit` but I guess it makes constructing these easier?

common/src/main/scala/explore/components/graphql/SubscriptionRenderMod.scala
```diff
@@ -0,0 +1,95 @@
...
+              State(
+                subscription.some,
+                StreamRendererMod
+                  .build(
+                    $.props
+                      .streamModifier(subscription.stream)
+                      .map[Pot[A]](a => Ready(a))
+                      .flatTap(_ => fs2.Stream.eval($.props.onNewData))
+                      .cons1(Pending()),
+                    holdAfterMod = (2 seconds).some
```
cquiroz, line 83: What's this?

common/src/main/scala/explore/components/graphql/SubscriptionRenderMod.scala
```diff
@@ -0,0 +1,95 @@
...
+                  )
+                  .some
+              )
+            )
+          }
+      }
+      .componentWillUnmount($ => $.state.subscription.map(_.stop).orEmpty)
```
cquiroz, line 90: `componentWillUnmount(_.state...)`?

common/src/main/scala/explore/components/undo/Restorer.scala
```diff
@@ -0,0 +1,37 @@
...
+// We don't use a case class to avoid the type parameter on T
+trait Restorer[M] { // M = (Local) Model
+  type T // T = Value type
+
+  val value: T // Value that will be restored upon undo/redo
+  val getter: Getter[
+    M,
+    T
+  ] // How to refresh the value from the model. Used when going from undo=>redo or viceversa.
+  val setter: T => IO[Unit] // Modify the model
```
cquiroz, line 18: Should we parametrize `IO`?

common/src/main/scala/explore/components/undo/Restorer.scala
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.components.undo
+
+import cats.effect.IO
+import monocle.Getter
+
+// We don't use a case class to avoid the type parameter on T
+trait Restorer[M] { // M = (Local) Model
```
cquiroz, line 10: This is looking great, thanks for the comments

cquiroz, line 10: Should it be `sealed`?

common/src/main/scala/explore/components/undo/Undoer.scala
```diff
@@ -0,0 +1,128 @@
...
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import cats.effect.IO
+import cats.implicits._
+import monocle.macros.Lenses
+import crystal.react.io.implicits._
+import monocle.Lens
+import monocle.Getter
+
+final case class Undoer[M](
```
cquiroz, line 16: I guess this will wrap an undo context?

common/src/main/scala/explore/model/Conditions.scala
```diff
@@ -0,0 +1,56 @@
...
+
+package explore.model
+
+import cats._
+import cats.implicits._
+import monocle.macros.Lenses
+import enum._
+
+@Lenses
+final case class Conditions(
```
cquiroz, line 12: These should be available in gsp-core

common/src/main/webpack/dev.webpack.config.js
```diff
@@ -39,6 +39,12 @@ const Web = Merge(
       contentBase: [__dirname, parts.rootDir],
       historyApiFallback: true,
       proxy: {
+        /*"/api/conditions": {
```
cquiroz, line 42: I really like that we can proxy the backends this way

conditions/src/main/scala/explore/conditions/ConditionsPanel.scala
```diff
@@ -0,0 +1,253 @@
...
+    image_quality: "Any",
+    sky_background: "Any",
+    water_vapor: "Any"
+  }]) {
+    affected_rows
+  }
+}
+ */
+
+final case class ConditionsPanel(
```
cquiroz, line 59: How's having this in a separate module working? The idea was that we could do development in one part at a time but I don't know how it works in practice

conditions/src/main/scala/explore/conditions/ConditionsPanel.scala
```diff
@@ -0,0 +1,253 @@
...
+      .builder[ConditionsPanel]("ConditionsPanel")
+      .render { $ =>
+        SubscriptionRenderMod[Subscription.Data, Conditions](
+          AppState.clients.conditions
+            .subscribe(Subscription)(
+              Subscription.Variables($.props.observationId.format).some
+            ),
+          _.map(Subscription.Data.conditions.composeOptional(headOption).getOption _).unNone
+        ) { (conditions, modState) =>
+          Undoer[Conditions] { ctx =>
```
cquiroz, line 215: I suppose this is the boundary of the undo stack, so you'd undo other parts independently?

common/src/main/scala/explore/components/forms/EnumSelect.scala
rpiaggio: Yes, we shouldn't have duplicate code in different projects (this is copied from `ocs3`).

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
rpiaggio: Yes, but also the Props parameters are initialization parameters to set up the subscription. We don't expect them to change in the caller.

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
rpiaggio: No, but having it as implicit also means the caller doesn't need to specify it if it has an implicit instance in scope.

common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
rpiaggio: It is needed because the component has type parameters, and therefore has to be defined in a `def`. But we don't want it to be created in each rerender, so we store in a `val` an instance parametrized on `[Any, Any]`, which can be reused with the casting. I'm open to more elegant ways of solving this problem.

common/src/main/scala/explore/components/graphql/SubscriptionRenderMod.scala
rpiaggio: Yes, that's the idea.

common/src/main/scala/explore/components/graphql/SubscriptionRenderMod.scala
rpiaggio: That's the mechanism to avoid flicker.

Crystal's `StreamRendererMod` is a stateful component bound to a `Stream`, whose state is updated either by:
- receiving a new value in the `Stream`
- executing the `modState` effect that it provides to children.

When `modState` is executed repeatedly and the modification sent to the server, then flicker would happen when the server sent intermediate values in the stream. To avoid this, `StreamRendererMod` has this optional "hold" mechanism in which it will not pass stream values to children during some time after receiving a `modState`.

The amount of time it should wait before re-enabling the stream is that parameter `holdAfterMod`.

During the hold, the last value received in the stream is remembered and it is passed down the line when the hold ends. If another `modState` is received during the hold, the timer is reset.

common/src/main/scala/explore/components/graphql/SubscriptionRenderMod.scala
rpiaggio: Good catch

common/src/main/scala/explore/components/undo/Restorer.scala
rpiaggio: Great idea. I'm doing it in a bunch of other places too (`Undoer` backend, `SubscriptionRender`, `SubscriptionRenderMod`).

common/src/main/scala/explore/components/undo/Restorer.scala
rpiaggio: Yes! Thank you

common/src/main/scala/explore/components/undo/Undoer.scala
rpiaggio: Exactly. We can have different undo regions if we want.

common/src/main/webpack/dev.webpack.config.js
rpiaggio: Yes, it's nice since we can use the same `/api` endpoints as in Heroku, which uses its own proxy.

However, I'm not sure if we will be using this moving forward since it seems we are going to use mainly WebSockets, which aren't subject to CORS and therefore lifts the requirements for a proxy.

It might make sense to just use environment variables for the WebSocket addresses, with defaults to development values.

conditions/src/main/scala/explore/conditions/ConditionsPanel.scala
rpiaggio: Sorry, I haven't tested it much yet. To be honest, I keep forgetting that we have this capability.

common/src/main/scala/explore/model/Conditions.scala
rpiaggio: Yes, I agree that we should move them to `gsp-core` and share it with `ocs3`. I've just copied them here temporarily.

rpiaggio:
Included in #75

PR #75, opened 2020-04-07
[WIP] New global state and context management. Includes conditions editor.

rpiaggio:
This is an experiment of an alternate approach to manage application state and global context (ie: clients, interpreters).

Crystal has been published as `v0.1.0` with this proposed approach, which basically:
* Does away with the old `Model` and `View`.
* There's a new definition of a `View` which holds a value `A` and an effectful modifier. This is similar to a `StateSnapshot`, but it differs in that the modifier/setter is not applied to the encapsulated value but the function is instead forwarded to an external source of truth (in this case the `State` of the `AppRoot`).
* A `ViewCtx` which combines a `View` and a context (which can be anything).
* Provides an `AppRoot` component which receives the initial state, context and an application renderer. It holds the global state in its `State`, and passes a `ViewCtx` to the application renderer, whose children can modify the state effectfully through it.

Since this approach doesn't store anything in a global `object`, it does require to pass context down through the component tree. Personally, I found that to be less of a hassle than I initially thought, especially since it is encapsulated in `ViewCtx` in most of the places, and where not, it can be passed implicitly.


rpiaggio:
> I'm ok with merging this. I'd like to play with the code to get a better understanding of the details

Before merging, I'd like to have a better solution for the Router/ViewCtx chicken and egg conundrum. Please feel free to work on this branch if you want to experiment, I'd appreciate input.

PR #82, opened 2020-04-22
Upgrade to crystal 0.1.4.

rpiaggio:
New crystal version has simplified `AppRoot` and more precise effect conversion methods (eg: `startInCB` instead of `toCB`.

### DIFF
--- a/explore/src/main/scala/explore/Explore.scala
+++ b/explore/src/main/scala/explore/Explore.scala
@@ -3,12 +3,10 @@
 
 package explore
 
-import cats.effect.IO
 import scala.scalajs.js
 import js.annotation._
 import japgolly.scalajs.react.extra.router._
 import explore.model.RootModel
-import crystal.react.AppRoot
 import japgolly.scalajs.react.vdom.VdomElement
 import explore.Routing
 import japgolly.scalajs.react.vdom.html_<^._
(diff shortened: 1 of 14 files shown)


explore/src/main/scala/explore/Explore.scala
```diff
@@ -17,22 +15,19 @@ import gem.Observation
 @JSExportTopLevel("Explore")
 object ExploreMain extends AppMain {
 
-  override def rootComponent(
-    WithModelCtx: AppRoot.Component[IO, explore.AppContextIO, RootModel]
-  ): VdomElement =
-    WithModelCtx { viewCtx =>
-      val routing = new Routing(viewCtx) // !!! This creates a new router on each render.
+  override def rootComponent(viewCtx: ViewCtxIO[RootModel]): VdomElement = {
+    val routing = new Routing(viewCtx) // !!! This creates a new router on each render.
```
cquiroz, line 19: Is there a big drawback on creating a new router each time?

rpiaggio: Yes, huge. This actually remounts the whole application on a state change, since it creates a new `Router` component. This won't be acceptable behavior moving forward.

For the moment, the solution I found is to tweak the `Router` to allow properties going through it (PR submitted to `scalajs-react`), which allows us to use the same component when state changes.

PR #88, opened 2020-04-28
Abstract Undoer logic from component so we can test it.

rpiaggio:
- Crystal update, which now supports Prisms and Optionals.
- The changes removing node shims are to better support cats-effect in Scala.js 1.x, but they don't work with 0.6 so I left them commented out.

### DIFF
--- a/common/src/main/scala/explore/components/undo/UndoRegion.scala
+++ b/common/src/main/scala/explore/components/undo/UndoRegion.scala
@@ -1,0 +1,68 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.components.undo
+
+import react.common.ReactProps
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import cats.effect.IO
+import cats.implicits._
+import monocle.macros.Lenses
+import crystal.react.implicits._
+import cats.effect.Async
+import cats.kernel.Monoid
+
+final case class UndoRegion[M](
+  renderer: Undoer.Context[IO, M] => VdomElement
+) extends UndoRegion.Props[IO, M]
+    with ReactProps {
+  @inline override def render: VdomElement =
+    UndoRegion.component(this.asInstanceOf[UndoRegion.Props[IO, Any]])
+}
+
+object UndoRegion {
+  protected trait Props[F[_], M] {
+    val renderer: Undoer.Context[F, M] => VdomElement
+  }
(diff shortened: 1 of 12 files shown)


common/src/main/scala/explore/components/undo/UndoRegion.scala
```diff
@@ -0,0 +1,68 @@
...
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import cats.effect.IO
+import cats.implicits._
+import monocle.macros.Lenses
+import crystal.react.implicits._
+import cats.effect.Async
+import cats.kernel.Monoid
+
+final case class UndoRegion[M](
```
cquiroz, line 16: IS this is a component that goes around the component where you want to undo?

common/src/main/scala/explore/components/undo/Undoer.scala
```diff
@@ -44,94 +32,77 @@ object Undoer {
...
 
-    private val popRedo: F[Option[Restorer[F, M]]] =
-      pop(State.redoStack)
+  private def pop(lens: Lens[Stacks, Undoer.Stack[F, M]]): F[Option[Restorer[F, M]]] =
+    FlatMap[F].flatMap(getStacks) { s =>
+      lens.get(s) match {
+        case head :: tail =>
+          modStacks(lens.set(tail)).as(head.some)
+        case _ =>
+          Sync[F].delay(None)
```
cquiroz, line 59: This can be `none.pure[F]`

common/src/test/scala/explore/undo/TestUndoable.scala
```diff
@@ -0,0 +1,82 @@
...
+import monocle.Getter
+import cats.effect.concurrent.Ref
+
+@Lenses
+case class TestStacks[F[_], A](
+  undoStack: Undoer.Stack[F, A] = List.empty,
+  redoStack: Undoer.Stack[F, A] = List.empty
+)
+
+class TestUndoer[F[_]: Sync, M](stacks: Ref[F, TestStacks[F, M]])(
```
cquiroz, line 21: Brilliant

common/src/main/scala/explore/components/undo/UndoRegion.scala
rpiaggio: Yes, it's what was called `Undoer` before. But now `Undoer` is a trait independent of React, and `UndoRegion` implements it.

common/src/main/scala/explore/components/undo/Undoer.scala
rpiaggio: Looks nicer, thank you.

PR #89, opened 2020-04-28
Undo and redo reposition, delete, insert in lists and trees and keep modifications to elements

rpiaggio:


common/src/main/scala/explore/components/undo/ListItem.scala
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.components.undo
+
+import monocle._
+import cats.implicits._
+
+object ListItem {
+  def indexGet[A](idF: A => Boolean): Getter[List[A], Option[(A, Int)]] =
```
cquiroz, line 10: I guess you get the first item to match the predicate? What if you want to get all that match?

common/src/test/scala/explore/undo/UndoerSpec.scala
```diff
@@ -9,40 +9,118 @@ import cats.effect.IO
...
 import monocle.Iso
-import scala.concurrent._
-import ExecutionContext.Implicits.global
+import explore.components.undo.ListItem
 
 object UndoerSpec extends TestSuite {
 
   def id[A] = Iso.id[A].asLens.asGetter
 
+  def eqByRef[A]: A => A => Boolean = (a1: A) => (a2: A) => a1 == a2
```
cquiroz, line 18: if it is by Ref, should it be `a1 eq a2`? What about an `eqByEq[A: Equal]`

common/src/main/scala/explore/components/undo/ListItem.scala
rpiaggio: The idea here is that elements of the `List` can be uniquely identified (eg by some `.id` field). An element with the same identity should not be in the list more than once.

I was hoping we could use this mechanism to generate lenses into elements of the list by using this uniqueness, without knowing their position in the list.

Also, I was thinking of generalizing this so that it works with any collection `C[A]`, not only `List`s. Then we could use other structures, like a `ListMap` or `TreeMap`. In that case instead of passing a `A => Boolean` function we would probably have to pass a `C[A] => A` function.

I agree though that it would be useful to zoom into several elements in some cases. This can probably be done by providing a `C[A] => List[A]` function instead. 

I'd really appreciate input in all this.

common/src/test/scala/explore/undo/UndoerSpec.scala
rpiaggio: Yup, good catch, it's misnamed. And yes, `==` was just for tests, in real code we should use `Eq`, but I'll change the tests to that so that it sets a good example ;)

rpiaggio:
I added an extremely simple tree structure. We can reposition, delete and add elements and undo/redo.

PR #97, opened 2020-05-03
More undo

rpiaggio:
- Clearer DSL for generating `At` lenses into indexed collection elements or positions.
- New crystal provides context abstraction (`Ctx`) independent of `View`.
- Migrate tests to MUnit.


build.sbt
```diff
@@ -94,34 +96,39 @@ lazy val commonSettings = gspScalaJsSettings ++ Seq(
...
+    "io.chrisdavenport"                 %%% "log4cats-core"        % "1.0.1",
+    "io.chrisdavenport"                 %%% "log4cats-log4s"       % "0.4.0-M1",
+    "io.github.cquiroz.react"           %%% "react-semantic-ui"    % "0.4.12",
+    "com.github.julien-truffaut"        %%% "monocle-core"         % monocle,
+    "com.github.julien-truffaut"        %%% "monocle-macro"        % monocle,
+    "com.rpiaggio"                      %%% "crystal"              % "0.2.0",
+    "com.rpiaggio"                      %%% "clue-scalajs"         % "0.0.6",
+    "io.circe"                          %%% "circe-generic-extras" % "0.13.0",
+    "io.suzaku"                         %%% "diode-data"           % "1.1.7",
+    "io.suzaku"                         %%% "diode-react"          % "1.1.7.160"
```
cquiroz, line 116: Why do we need `diode-react`?

common/src/test/scala/org/typelevel/discipline/munit/Discipline.scala
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package org.typelevel.discipline
+package munit
+
+import _root_.munit.Location
+import _root_.munit.ScalaCheckSuite
+
+trait Discipline extends ScalaCheckSuite {
```
cquiroz, line 10: We may want to move this to tyelevel in the future

common/src/main/scala/explore/undo/Undoer.scala
```diff
@@ -22,17 +22,17 @@ object Undoer {
 
   trait Setter[F[_], M] {
     def set[A](
-      m:    M,
-      lens: Lens[M, A],
-      setM: M => F[Unit]
-    )(v:    A): F[Unit]
+      m:        M,
+      lens:     Lens[M, A],
+      onChange: M => F[Unit]
```
cquiroz, line 27: wHy this name change?

build.sbt
rpiaggio: IIRC, that provides the `.render...` methods for `Pot`. The usage of `diode-*` is temporary, we are only using `Pot`. We should extract and adapt the needed logic.

common/src/test/scala/org/typelevel/discipline/munit/Discipline.scala
rpiaggio: Ok, I'm not sure if I'm missing something here but it works for now. I'll start by moving it to a separate project.

common/src/main/scala/explore/undo/Undoer.scala
rpiaggio: Just to make the API friendlier.

PR #103, opened 2020-05-05
Remove GraphQL demo components and fixes.

rpiaggio:
- Remove GraphQL demo components.
- Fix Undoer flicker and change API. Now it only needs getter function.
- IndexedColMod no longer produces at. What it produced was not a lawful Lens, so now it just produces Getter and Setter for a given id. There's a wrapping class to encapsulate these for a nice DSL.


model/src/main/scala/explore/undo/GetSet.scala
```diff
@@ -0,0 +1,22 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.undo
+
+import monocle.Lens
+import monocle.Getter
+import monocle.Setter
+
+// This is not a Lens. It doesn't abide by any laws. It's just convenience to group a getter and setter functions.
+final case class GetSet[T, A](getter: Getter[T, A], setter: Setter[T, A]) {
```
cquiroz, line 11: 👍

PR #122, opened 2020-05-14
Upgrade to Scala.js 1.0/React 16.13.

rpiaggio:
Besides adapting the logic for the new React, this PR:

* Removes component names.
* Removes deprecated `why-did-you-update` in favor of `scalajs-react`'s new global `ReusabilityOverlay`. From preliminary use it seems to fit our needs.
* Temporarily uses `log4cats-noop` since `log4cats-log4s` is not available for Scala.js 1.0 yet (because `log4s` isn't, but a PR has been submitted). We could alternatively deploy our own JS backend for `log4cats`.
* Copies a class from `diode` (`ReactPot`) which provides convenience methods to integrate `diode`'s `Pot` with React, since this class is in the `diode-react` project which hasn't been published yet for Scala.js 1.0/React 16.13. This is also temporary, since we are planning to roll out our own `Pot`-like structure.

I purposefully left commented out dependencies in place regarding the last 2 points, so that we address them soon.


build.sbt
```diff
@@ -175,32 +176,32 @@ lazy val commonWDS = Seq(
...
+    "less-loader"                        -> "6.1.0",
     "webpack-merge"                      -> "4.2.2",
-    "mini-css-extract-plugin"            -> "0.8.0",
-    "webpack-dev-server-status-bar"      -> "1.1.0",
+    "mini-css-extract-plugin"            -> "0.9.0",
+    "webpack-dev-server-status-bar"      -> "1.1.2",
     "cssnano"                            -> "4.1.10",
-    "uglifyjs-webpack-plugin"            -> "2.2.0",
-    "html-webpack-plugin"                -> "3.2.0",
+    "terser-webpack-plugin"              -> "3.0.1",
```
cquiroz, line 195: Did you try with and without uglyfiyng just to check it really makes a difference. A few times `uglifyjs` hasn't worked as expected

build.sbt
```diff
@@ -175,32 +176,32 @@ lazy val commonWDS = Seq(
...
-    "why-did-you-update"                 -> "1.0.6",
+    "favicons-webpack-plugin"            -> "3.0.1",
     "@packtracker/webpack-plugin"        -> "2.2.0"
   ),
   npmDependencies in Compile ++= Seq(
     "react"            -> reactJS,
     "react-dom"        -> reactJS,
-    "semantic-ui-less" -> SUI,
+    "semantic-ui-less" -> SUILess,
     "prop-types"       -> "15.7.2"
```
cquiroz, line 205: Do we still need on react 16.13?

build.sbt
rpiaggio: Never occurred to me not to uglify (should we be saying terserify now?). Let me try and compare.

cquiroz: It is just some older version of uglyfyjs wouldn't work with scala.js so I took it away, A later version did work and I restored it

cquiroz:
.js size decreased 10%, that's nice

build.sbt
rpiaggio: prop-types? I'm assuming we use it because it's required by some dependency. It's independent from react now. A quick google search turned up an example of usage with react 16.13.1, so I would think it's still relevant.

cquiroz: 👍

build.sbt
rpiaggio: ok, uglify seems to be deprecated now in favor of terser, which seems to do a much better job

cquiroz: I wish best in kind JS tools would last for a few weeks, uglify is probably the third minifier I've replaced of a saga of obsoleted tool

PR #124, opened 2020-05-15
Use router with pass-through props.

rpiaggio:
New router now allows to pass through global state to components.

To test it out, I modified the target editor so that it sets a global property with the selected target.

This is mostly working except for a flicker the first time state is modified. This seems to be caused by react-grid-layout reusability logic. It doesn't happen with version 0.17.1 of react-grid-layout.

We should either find a workaround or maybe isolate the case and report a bug?

### DIFF
--- a/common/src/main/scala/explore/AppMain.scala
+++ b/common/src/main/scala/explore/AppMain.scala
@@ -47,7 +47,7 @@ trait AppMain extends IOApp {
 
     implicit val gqlStreamingBackend: StreamingBackend[IO] = WebSocketJSBackend[IO]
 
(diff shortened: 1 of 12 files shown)


cquiroz:
Why is that there are "No javascript assets found" messages on the PR?

rpiaggio:
> Why is that there are "No javascript assets found" messages on the PR?

Ah the name of bundle changed from explore.hash.js to explore-opt.hash.js, and that's excluded from packtracker (`exclude_assets: [/explore-opt.*/]`). Let me see how to fix this.

PR #126, opened 2020-05-15
Fix flicker on first update of grid layout.

rpiaggio:


cquiroz:
I guess this worth to test in a few browsers, FF, safari and chrome at least

rpiaggio:
> I guess this worth to test in a few browsers, FF, safari and chrome at least

Good idea. I'll test and report here.

rpiaggio:
Works fine on Chrome, Safari and Firefox ESR latest versions.

PR #146, opened 2020-05-27
Nested subscriptions

rpiaggio:


targeteditor/src/main/scala/explore/targeteditor/Test.scala
```diff
@@ -23,7 +23,10 @@ object Test extends AppMain {
       Observation
         .Id(ProgramId.Science.fromString.getOption("GS-2020A-DS-1").get, Index.One)
 
-    <.div(^.height := "100vh", ^.width := "100%", TargetEditor(obsId, view.zoomL(RootModel.target)))
+    <.div(^.height := "100vh",
+          ^.width := "100%",
+          TargetEditor(obsId, view.zoomL(RootModel.target), null)
```
cquiroz, line 28: Why the `null`?

rpiaggio: That shouldn't have made it into the PR. I'm pushing another commit.

rpiaggio:
> Thanks, this is an important sanity check. Could you make a test app with this PR

https://explore-pipe-nested-sub-fms3uj.herokuapp.com/

PR #152, opened 2020-06-05
Sidetabs

cquiroz:
This is a first change to include the side buttons as in the mockup
Note the buttons aren't functional yet but that will come later on.

Turns out drawing buttons vertically, properly aligned and in a cross browser fashion is not trivial

The previous sidebar component was removed and the overall structure is now done using css grid

Tested on: FF, Chrome, Safari, Safari iOS and FF ESR

image

### DIFF
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -27,9 +27,33 @@ body {
   background: darkgray;
 }
 
-.sidetab {
-  // transform: rotate(-90deg);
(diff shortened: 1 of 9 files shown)


swalker2m:
Looks great to me!  Do you know how you'll indicate the selected tab?

arturog8m:
Great. So Explore is done now! Get Chronicle next and we'll be ready! Good!

cquiroz:
It will change colors a bit like in

PR #169, opened 2020-06-16
User EnumZipper. Organize imports.

rpiaggio:


common/src/main/scala/explore/AppMain.scala
```diff
@@ -67,7 +68,7 @@ trait AppMain extends IOApp {
         Observation // TODO Remove this, it's here termporarily for testing URL automatic derivation.
           .Id(ProgramId.Science.fromString.getOption("GS-2020A-DS-1").get, Index.One)
           .some,
-      tabs = Zipper.fromNel(AppTab.all)
+      tabs = EnumZipper.of[AppTab]
```
cquiroz, line 71: Neat

common/src/main/scala/explore/model/reusability.scala
```diff
@@ -3,29 +3,32 @@
...
-  implicit val expTargetReuse: Reusability[ExploreSiderealTarget]  = Reusability.derive
-  implicit val expObsReuse: Reusability[ExploreObservation]        = Reusability.derive
-  implicit def enumReuse[A: Enumerated]: Reusability[A]            =
+  implicit val statusReuse: Reusability[StreamingClientStatus]             = Reusability.derive
+  implicit val durationReuse: Reusability[Duration]                        = Reusability.by(_.getSeconds)
+  implicit val obsIdReuse: Reusability[Observation.Id]                     = Reusability.by(_.format)
+  implicit val siderealTargetReuse: Reusability[SiderealTarget]            = Reusability.byEq
+  implicit val expTargetReuse: Reusability[ExploreSiderealTarget]          = Reusability.derive
+  implicit val expObsReuse: Reusability[ExploreObservation]                = Reusability.derive
+  implicit def enumReuse[A: Enumerated]: Reusability[A]                    =
```
cquiroz, line 28: Do we have a summon method to just do `Enumerated[A]`?

explore/src/main/scala/explore/Routing.scala
```diff
@@ -27,16 +27,26 @@ object Routing {
...
       import dsl._
 
       (emptyRule
         | staticRoute(root, HomePage) ~> renderP(view => HomeComponent(view))
-        | staticRoute("/constraints", ConstraintsPage) ~> render(UnderConstruction())
         | dynamicRouteCT(("/obs" / string("[a-zA-Z0-9-]+")).pmapL(obsIdP)) ~> renderP(view =>
           HomeComponent(view)
-        ))
+        )
+        | dynamicRouteCT(("/target" / string("[a-zA-Z0-9-]+")).pmapL(targetObsIdP)) ~> renderP(
```
cquiroz, line 45: I don't recall if the order is important

explore/src/main/scala/explore/Routing.scala
```diff
@@ -27,16 +27,26 @@ object Routing {
...
+          view => HomeComponent(view)
+        )
+        | staticRoute("/configurations", ConfigurationsPage) ~> render(UnderConstruction())
+        | staticRoute("/constraints", ConstraintsPage) ~> render(UnderConstruction()))
         .notFound(redirectToPage(HomePage)(SetRouteVia.HistoryPush))
         .verify(HomePage, ObsPage(Observation.Id.unsafeFromString("GS2020A-Q-1")))
```
cquiroz, line 51: Maybe we should verify `ConstraintsPage` and `ConfgurationsPage`

common/src/main/scala/explore/model/reusability.scala
rpiaggio: Good catch! Done.

explore/src/main/scala/explore/Routing.scala
rpiaggio: I don't think it is. I'm just keeping the same order everywhere (`AppTab`, `RootModelRouting` and here) for mental sanity.

explore/src/main/scala/explore/Routing.scala
rpiaggio: Yes, good idea! I'm adding that. And `TargetPage`.

PR #170, opened 2020-06-17
Make the tree area resizable

cquiroz:
This PR updates the layout to support an area on the right to place a tree eg (the targets tree). This area is as of now just a *placeholder* but it maybe used in the future
The tree area can be resized horizontally and it will always be 100% on height.
If the inner area is too big it will add a scroll bar

screencast 2020-06-16 23-39-35


cquiroz:
Tested on FF, FF ESR, Chrome, Safari and Safar iOS

PR #172, opened 2020-06-18
Navigate targets with tree

rpiaggio:


common/src/main/scala/explore/AppMain.scala
```diff
@@ -64,11 +65,13 @@ trait AppMain extends IOApp {
...
     val initialModel = RootModel(
-      obsId =
-        Observation // TODO Remove this, it's here termporarily for testing URL automatic derivation.
-          .Id(ProgramId.Science.fromString.getOption("GS-2020A-DS-1").get, Index.One)
-          .some,
-      tabs = EnumZipper.of[AppTab]
+      tabs = EnumZipper.of[AppTab],
+      focused = // TODO Remove this, it's here termporarily for testing URL automatic derivation.
+        FocusedObs(
+          UUID.randomUUID
```
cquiroz, line 71: Generally a randomUUID is considered a side effect. This is temporal so it should be ok

explore/src/main/scala/explore/HomeComponent.scala
```diff
@@ -57,76 +58,115 @@ object HomeComponent {
...
+        <.div(
+          GPPStyles.RGLArea,
+          SizeMe() { s =>
+            val treeWidth = state.treeWidth.toDouble
+            val coreWidth = s.width.toDouble - treeWidth
+
+            // Tree area
+            val tree =
+              <.div(^.width := treeWidth.px, GPPStyles.Tree)(
+                <.div(GPPStyles.TreeBodyOuter)(
```
cquiroz, line 102: You may not need it here but the reason for the tree inner element was to leave some space for the scroll bar. We can iterate on that later on

targeteditor/src/main/scala/explore/targeteditor/TargetEditor.scala
```diff
@@ -30,22 +30,35 @@ object TargetEditor {
 
   protected implicit val propsReuse: Reusability[Props] = Reusability.derive
 
+  // class Backend($ : BackendScope[Props, Unit]) {
+  class Backend() {
+    private val targetBodyRef = Ref.toScalaComponent(TargetBody.component)
```
cquiroz, line 35: Nice, this lets us reach the Aladdin component

common/src/main/scala/explore/AppMain.scala
rpiaggio: That's good to know

explore/src/main/scala/explore/HomeComponent.scala
rpiaggio: Ah, sorry, I misunderstood that, I thought it was only there for the example.

rpiaggio:
> This is looking great, thanks for the tests
> How’s the structure con sub projects working for you? I want to avoid the extra maintenance if there are no real benefits

To develop the observations trees, I found the setup really useful to test things out. As the application grows, I envision that it will be more desirable to be able to develop isolated components without deploying the full fledged app for each small change.

rpiaggio:
This should be deployed to https://explore-staging.herokuapp.com/ in a couple of minutes.

The functionality includes:
* When you click on an observation or target:
  * The URL is updated (not working for targets)
  * The target or the observation's target is displayed.
* When you drag'n'drop an observation to another target:
  * Its target is updated.
  * If it was the selected observation, the new target is displayed.
  * If you undo/redo and the affected observation is selected, the previous target is displayed.

PR #173, opened 2020-06-18
Targets with UUID ids. New crystal. Visual improvements.

rpiaggio:


common/src/main/resources/less/style.less
```diff
@@ -130,8 +130,25 @@ body {
...
+  width: 100%;
+}
+
+.obs-tree-item {
+  margin-left: 25px;
+  padding: 5px;
+}
+
+.obs-badge {
+  width: 100% !important;
```
cquiroz, line 153: It would be nice to avoid `!important` maybe we need a more specific selector

common/src/main/resources/theme/elements/button.variables
```diff
@@ -0,0 +1 @@
+@activeBackgroundColor: @blue;
```
cquiroz, line 1: neat

explore/src/main/scala/explore/HomeComponent.scala
```diff
@@ -120,7 +132,7 @@ object HomeComponent {
                 axis = Axis.X,
                 width = treeWidth,
                 height = Option(s.height).getOrElse(0),
-                minConstraints = (295: JsNumber, 0: JsNumber),
+                minConstraints = (250: JsNumber, 0: JsNumber),
```
cquiroz, line 135: Yeah we should change these to be just `Int` upstream

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -151,7 +152,7 @@ object TargetObsList {
       val observations = props.observations.get
       val obsByTarget  = observations.groupBy(_.target)
 
-      <.div(^.width := "270px", ^.marginTop := "30px")(
+      <.div(GPPStyles.ObsTree, ^.marginTop := "30px")(
```
cquiroz, line 155: Maybe we should put the top margin in the css

common/src/main/resources/less/style.less
rpiaggio: It overrides the `width` set within Semantic UI's `Card`. We can pass a `^.width` to the `Card` within the code and it will take it, but if we want to set it via CSS, it wasn't being prioritized without the `!important`. I ran out of ideas but I might be missing something.

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: I wasn't sure about leaving the margin. I just put it there so that the reusability counters wouldn't block visuals in development. But we might leave it, it didn't look bad without the counters.

PR #193, opened 2020-06-26
Views in forms. Simplify undo logic. Undo buttons.

rpiaggio:


common/src/main/scala/explore/components/undo/UndoButtons.scala
```diff
@@ -0,0 +1,51 @@
...
+  protected def componentBuilder[F[_], A] =
+    ScalaComponent
+      .builder[Props[F, A]]
+      .render_P { p =>
+        implicit val effect = p.effect
+
+        <.div(
+          Button(onClick = p.undoCtx.undo(p.value).runInCB,
+                 size = p.size,
+                 disabled = p.undoCtx.undoEmpty
```
cquiroz, line 36: What's `undoEmpty`? That the undo stack is empty?

rpiaggio: Yes. There's also `redoEmpty`.

model/src/main/scala/explore/data/Optics.scala
```diff
@@ -0,0 +1,28 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.data
+
+import cats._
+import cats.implicits._
+import monocle._
+
+object Optics {
+  def filteredTraversal[T](predicate: T => Boolean): Traversal[List[T], T] =
```
cquiroz, line 11: Should we propose some of these back to Monocle? Or put them in gsp-cors/gpp-ui

rpiaggio: I'm on the fence about this.

`zip` has been proposed in https://github.com/optics-dev/Monocle/issues/545, but it's unlawful if the `Lens`es overlap (hence the name `disjoint` here). But it does offer some convenience to have this instead of defining the zip manually each time we need it (it could be used here for example: https://github.com/gemini-hlsw/explore/blob/master/model/src/main/scala/explore/model/ModelOptics.scala#L44). I guess it could be added to Monocle's `unsafe` module, but I seem to remember that's going to be discarded in Monocle 3.

About `filteredTraversal`: TBH, we are not using it at the moment. I haven't checked if it's lawful.

PR #197, opened 2020-06-26
Store targets and observations in DB and sync clients with DB

rpiaggio:
Also, renamed conditions->constraints.

Animated GIF screen capture is a bit large to upload to GitHub, but here it is in MP4: https://drive.google.com/file/d/1F_phHQYh6bb1qoWsPTHd7LsCVn-4-K2j/view?usp=sharing

### DIFF
--- a/common/src/main/scala/explore/model/decoders.scala
+++ b/common/src/main/scala/explore/model/decoders.scala
@@ -1,0 +1,83 @@
+package explore.model
+
+import cats.implicits._
+import gem.util.Enumerated
+import io.circe.Decoder
+import io.circe.DecodingFailure
+import io.circe.Encoder
+import io.circe.HCursor
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveDecoder
+import io.circe.generic.semiauto.deriveEncoder
+import gsp.math.RightAscension
+import gsp.math.Declination
+import gsp.math.Coordinates
+import gsp.math.Epoch
+import gsp.math.ProperMotion
+import explore.model.enum.ObsStatus
+import java.time.Duration
+
+object decoders {
+
+  implicit def enumDecoder[E: Enumerated]: Decoder[E] =
+    new Decoder[E] {
+      final def apply(c: HCursor): Decoder.Result[E] =
+        // TODO Obtain the failure CursorOp list from c.
+        c.as[String]
+          .flatMap(s =>
+            Enumerated[E]
+              .fromTag(s)
+              .toRight(DecodingFailure(s"Invalid Enumerated value [$s] on [$c].", List.empty))
+          )
+    }
+
+  implicit val raDecoder = new Decoder[RightAscension] {
+    final def apply(c: HCursor): Decoder.Result[RightAscension] =
+      for {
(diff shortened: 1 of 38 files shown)


cquiroz:
The capture looks amazing

common/src/main/scala/explore/model/decoders.scala
```diff
@@ -0,0 +1,67 @@
...
+import gsp.math.Declination
+import gsp.math.Coordinates
+import gsp.math.Epoch
+import gsp.math.ProperMotion
+import explore.model.enum.ObsStatus
+import java.time.Duration
+
+object decoders {
+
+  implicit def enumDecoder[E: Enumerated]: Decoder[E] =
```
cquiroz, line 25: This is great. We should consider whether these should go to gsp-core eventually

db.sql
```diff
@@ -0,0 +1,88 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
cquiroz, line 1: Is this here just to keep a record? or do we use it to build the db?

db.sql
```diff
@@ -0,0 +1,88 @@
...
+  'Proposed',
+  'Approved',
+  'ForReview',
+  'Ready',
+  'Ongoing',
+  'Observed'
+);
+
+CREATE TABLE constraints (
+  id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
```
cquiroz, line 48: We don't need to change this but there is a requirement for "memorable" urls. we may want to use a human readable id eventually/aditionally

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -32,12 +32,12 @@ import explore.model.Focused
...
+import TargetObsQueries._
 
 final case class TargetObsList(
-  targets:        List[SiderealTarget],
-  observations:   View[List[ExploreObservation]],
-  focused:        ViewOpt[Either[SiderealTarget.Id, ExploreObservation.Id]],
-  onTargetSelect: SiderealTarget.Id => Callback
+  targets:      List[SiderealTarget],
+  observations: View[List[ExploreObservation]],
+  focused:      View[Option[Either[SiderealTarget.Id, ExploreObservation.Id]]]
```
cquiroz, line 40: Why the Either? Is this to indicate we either clicked on a given target or one of the obs attached to the target?

rpiaggio: Exactly. And the `Option` is for the case where we are in this page but no target or obs has been clicked yet.

db.sql
rpiaggio: Ok. We can change to anything as long as it's unique.

db.sql
rpiaggio: It provides the `uuid_generate_v4` function.

common/src/main/scala/explore/model/decoders.scala
rpiaggio: Yes, good idea. We will probably use this from other apps too.

PR #200, opened 2020-06-30
Simplified, working, routing. Fix layout.

rpiaggio:
Landing on a target/obs URL was not working.

Also, the layout got messed up when deploying to staging, but it kept "working on my machine", so I'm trying changing it a bit. This worries me though, is it possible this is due to some JS library versions not exactly matching up in both deploys?

### DIFF
--- a/model/src/main/scala/explore/model/RootModel.scala
+++ b/model/src/main/scala/explore/model/RootModel.scala
@@ -7,20 +7,14 @@ import cats.effect._
 import cats.implicits._
 import cats.kernel.Eq
 import clue._
+import explore.model.Focused.FocusedObs
 import explore.model.enum.AppTab
-import gem.Observation
 import gem.data.EnumZipper
 import io.chrisdavenport.log4cats.Logger
-import monocle.Optional
-import monocle.Prism
-import monocle.function.Possible.possible
-import monocle.macros.GenPrism
+import monocle.Lens
 import monocle.macros.Lenses
-import monocle.std.option.some
 import sttp.model.Uri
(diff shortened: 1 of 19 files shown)


model/src/main/scala/explore/model/RootModel.scala
```diff
@@ -16,9 +16,11 @@ import monocle.Prism
 import monocle.function.Possible.possible
 import monocle.macros.GenPrism
 import monocle.macros.Lenses
+import monocle.std.option.some
```
cquiroz, line 19: Are you using this?

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -33,11 +33,14 @@ import gem.Observation
...
 import TargetObsQueries._
+import explore.model.Focused.FocusedTarget
+import react.semanticui.modules.popup.PopupOn.Focus
+import explore.model.Focused.FocusedObs
 
 final case class TargetObsList(
   targets:      List[SiderealTarget],
   observations: View[List[ExploreObservation]],
-  focused:      View[Option[Either[SiderealTarget.Id, ExploreObservation.Id]]]
+  focused:      View[Option[Focused]]
```
cquiroz, line 43: Much better :D

model/src/main/scala/explore/model/RootModel.scala
rpiaggio: Not anymore. I'm sending another commit with organized imports.

PR #201, opened 2020-06-30
Add targets (with undo/redo). Visual improvements.

rpiaggio:
Screen Capture: 

https://drive.google.com/file/d/1BWaGpTO6XfHt4329dN90QVa_dkzSuzEG/view?usp=sharing

### DIFF
--- a/common/src/main/resources/theme/globals/site.variables
+++ b/common/src/main/resources/theme/globals/site.variables
@@ -1,3 +1,0 @@
-@floatingShadow:
-  0px 0px 1px 2px rgba(221, 219, 217, .8)
-;
--- a/common/src/main/resources/theme/globals/site.variables
+++ b/common/src/main/resources/theme/globals/site.variables
@@ -1,0 +1,3 @@
+@floatingShadow:
+  0px 0px 1px 2px rgba(221, 219, 217, .8)
+;
--- a/model/src/main/scala/explore/model/ExploreSiderealTarget.scala
+++ b/model/src/main/scala/explore/model/ExploreSiderealTarget.scala
@@ -11,6 +11,11 @@ import gem.Target
 import gsp.math.ProperMotion
 import monocle.Lens
 import monocle.macros.Lenses
+import gsp.math.Coordinates
+import gsp.math.RightAscension
+import gsp.math.Epoch
+import gsp.math.Declination
+import cats.effect.Sync
 
 /**
   * A refinement of gem.Tracker meant for sidereal targets
@@ -20,6 +25,22 @@ final case class SiderealTarget(id: SiderealTarget.Id, name: String, track: Prop
 
 object SiderealTarget {
   type Id = UUID
+
+  def New[F[_]: Sync]: F[SiderealTarget] =
+    Sync[F]
+      .delay(UUID.randomUUID())
+      .map(id =>
+        SiderealTarget(
+          id,
+          "New Target",
+          ProperMotion(Coordinates(RightAscension.Zero, Declination.Zero),
+                       Epoch.J2000,
+                       none,
+                       none,
+                       none
+          )
+        )
+      )
 
   implicit val siderealTargetEq: Eq[SiderealTarget] = Eq.by(x => (x.name, x.track))
 }
--- a/model/src/main/scala/explore/model/ExploreSiderealTarget.scala
+++ b/model/src/main/scala/explore/model/ExploreSiderealTarget.scala
@@ -6,16 +6,16 @@ package explore.model
 import java.util.UUID
 
 import cats._
+import cats.effect.Sync
 import cats.implicits._
 import gem.Target
+import gsp.math.Coordinates
+import gsp.math.Declination
+import gsp.math.Epoch
 import gsp.math.ProperMotion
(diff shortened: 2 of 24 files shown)


common/src/main/resources/theme/globals/site.variables
```diff
@@ -0,0 +1,3 @@
+@floatingShadow:
```
cquiroz, line 1: Is this overriding the theme right? I don't see it being used in the css

model/src/main/scala/explore/model/ExploreSiderealTarget.scala
```diff
@@ -21,6 +26,22 @@ final case class SiderealTarget(id: SiderealTarget.Id, name: String, track: Prop
 object SiderealTarget {
   type Id = UUID
 
+  def New[F[_]: Sync]: F[SiderealTarget] =
```
cquiroz, line 29: Seems odd to see a capitalized function name

model/src/main/scala/explore/model/ExploreSiderealTarget.scala
```diff
@@ -21,6 +26,22 @@ final case class SiderealTarget(id: SiderealTarget.Id, name: String, track: Prop
...
+    Sync[F]
+      .delay(UUID.randomUUID())
```
cquiroz, line 31: Nice

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -168,65 +212,85 @@ object TargetObsList {
...
-                                        Icon("sort")
-                                      )
+                            opIcon,
+                            <.span(
+                              // Segment(raised =
+                              //   props.focused.get
+                              //     .exists(_ === FocusedTarget(target.id))
+                              // )(
+                              target.name,
+                              <.span(^.float.right, s"$obsCount Obs")
```
cquiroz, line 246: This is low priority. Are we using the float to do layout, aligning to the right? Could we use flexbox instead?

observationtree/src/main/scala/explore/observationtree/TargetObsQueries.scala
```diff
@@ -60,23 +63,7 @@ object TargetObsQueries {
...
-                     } yield ExploreObservation(
-                       id,
-                       target,
-                       status,
-                       conf,
-                       constraintsName,
-                       duration
-                     )
-                   }
+        obsList <- Traverse[List].traverse(obsJson)(obsDecoder(target).apply)
```
cquiroz, line 66: I think this can be written as `obsJson.traverse(obsDecoder...`

project/Settings.scala
```diff
@@ -9,6 +9,7 @@ object Settings {
     val cats              = "2.1.1"
     val catsEffect        = "2.1.3"
     val circe             = "0.13.0"
+    val contextApplied    = "0.1.3"
```
cquiroz, line 12: What's this? I don't see it used in `build,sbt`

common/src/main/resources/theme/globals/site.variables
rpiaggio: This seems to be an oversight in the "slate" theme. This style is applied when you have a `Segment` that is `raised`. In the default theme, which uses light backgrounds, that's a dark shade around the segment. The "slate" theme didn't change this, and since its background is dark, it was not visible. I tried to make it look similar to the `Card` `raised` style.

However, this will stop working if we switch to a theme with a light background. Should we move this to the slate theme? Is that copy of the theme definitive or is there a chance that we may copy a newer version of the theme over?

model/src/main/scala/explore/model/ExploreSiderealTarget.scala
rpiaggio: Yes, I will change this. This started life as a `val` constant, but it then evolved and I should have renamed it.

common/src/main/resources/theme/globals/site.variables
cquiroz: I doubt we'll update themes

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: Good idea, I'll try that. Also deleting that commented out code.

observationtree/src/main/scala/explore/observationtree/TargetObsQueries.scala
rpiaggio: Yes, that's simpler. Thank you.

project/Settings.scala
rpiaggio: Briefly tried to used `context-applied` and forgot to remove the version declaration from here.

BTW, I rolled back on this because it caused linking errors on `SubscriptionRenderMod` regarding the `Reusability` type class.

common/src/main/resources/theme/globals/site.variables
rpiaggio: Ok, I'll move this to the slate theme then.

PR #202, opened 2020-07-01
Remove targets. Add/remove/undo/redo buttons at the bottom.

rpiaggio:


observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -43,9 +43,11 @@ import react.semanticui.elements.button.Button
 import react.semanticui.elements.icon.Icon
 import react.semanticui.elements.segment.Segment
 import react.semanticui.modules.popup.PopupOn.Focus
+import react.semanticui.sizes._
 
 import scalajs.js.|
 import TargetObsQueries._
+import gpp.svgdotjs.std.stdBooleans.`true`
```
cquiroz, line 50: oh? What's this?

rpiaggio: I'm asking myself the same question.

PR #204, opened 2020-07-03
KeyedIndexedList and Adjuster. Performance and soundness.

rpiaggio:
- New `KeyedIndexedList` structure. Stores a list where each element must have a unique key. Provides O(1) access by key to the element and to the position of the element in the list.
- New `Adjuster` optic. Like a `Setter` but does not abide by the "compose modify" law, since it is allowed to adjust the values before setting them. We were using unlawful `Setter`s before.
- Better code organization in `TargetObsList`.
- New `KeyedIndexedTree` structure, with similar mechanics to `KeyedIndexedList`.


model/src/main/scala/explore/data/KeyedIndexedList.scala
```diff
@@ -0,0 +1,78 @@
...
+import scala.collection.immutable.ListMap
+
+import cats.implicits._
+import cats.kernel.Eq
+import monocle.Lens
+
+// Each element has a unique Key.
+// Elements can be accessed by Key in O(1).
+// Position in the list (index) can be accessed by Key in O(1).
+case class KeyedIndexedList[K, A] private (private val list: ListMap[K, (A, Int)]) {
```
cquiroz, line 15: Does `ListMap` give you `O(1)` access time?

model/src/main/scala/explore/data/KeyedIndexedList.scala
```diff
@@ -0,0 +1,78 @@
...
+              case (id, (a, i)) if i > idx => (id, (a, i - 1))
+            })
+        )
+      )
+
+  def contains(key: K): Boolean = list.contains(key)
+
+  def exists(p: A => Boolean): Boolean = elements.exists(p)
+
+  def take(n: Int): KeyedIndexedList[K, A] = KeyedIndexedList.unsafeFromListMap(list.take(n))
```
cquiroz, line 43: Should these be unsafe too?

model/src/main/scala/explore/optics/Adjuster.scala
```diff
@@ -0,0 +1,136 @@
...
+import cats.Contravariant
+import cats.Functor
+import cats.arrow.Choice
+import cats.arrow.Profunctor
+import cats.syntax.either._
+import monocle._
+
+// An Adjuster is just like a Setter, but weaker: it does not abide by the "compose modify" law.
+// It allows the modify/set functions to adjust the result (eg: not going out of bounds), which breaks composability.
+abstract class PAdjuster[S, T, A, B] extends Serializable { self =>
```
cquiroz, line 15: Monocle is going to use only 2 types in the future, basically T=S and B=A but feel free to use this encoding. Julien also changed the type names to make them more descriptive, so it would be like `Adjuster[From, To]` I think

model/src/main/scala/explore/optics/package.scala
```diff
@@ -0,0 +1,103 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore
+
+import cats._
+import cats.implicits._
+import monocle._
+import monocle.std.option.some
+
+package object optics {
```
cquiroz, line 11: nice

model/src/main/scala/explore/optics/package.scala
```diff
@@ -0,0 +1,103 @@
...
+          optA.flatMap[A] { a =>
+            modOptB(other.get(a).some).map(b => other.set(b)(a))
+          }
+        }
+      }
+  }
+
+  // Lenses must be disjoint (not overlap), or the result will be unsafe.
+  // See https://github.com/optics-dev/Monocle/issues/545
+  def disjointZip[S, A, B](l1: Lens[S, A], l2: Lens[S, B]): Lens[S, (A, B)] =
```
cquiroz, line 94: Is it possible to ensure the lenses don't overlap?

model/src/main/scala/explore/undo/TreeMod.scala
```diff
@@ -1,84 +1,84 @@
...
-      (idx._1.forall(id => parentId.exists(pid => Eq[Id].eqv(id, pid)))).fold(
-        children.take(idx._2) ++ (node +: children.drop(idx._2)),
-        children.map(node => Node(node.value, goChildren(node.children, idLens.get(node.value).some)))
-      )
-
-    Tree(goChildren(tree.children))
-  }
-  // format: on
-}
+// package explore.undo
```
cquiroz, line 4: Will you uncomment this on the next PR?

model/src/test/scala/explore/optics/AdjusterLaws.scala
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.optics
+
+import monocle.internal.IsEq
+
+case class AdjusterLaws[S, A](setter: Adjuster[S, A]) {
```
cquiroz, line 8: Great to have some laws

cquiroz:
This seems to go in the right direction. I wonder if we could ask on the Monocle site what they think about the `Adjuster` optics

model/src/main/scala/explore/data/KeyedIndexedList.scala
rpiaggio: Apparently not, it stores just a `List` internally. I thought it used a mixed structure, which is what `VectorMap` and `TreeSeqMap` do. I'll switch to the latter, which seems to be more performant in the general case.

model/src/main/scala/explore/data/KeyedIndexedList.scala
rpiaggio: I don't think so. The unsafe constructor is unsafe because there's no guarantee that the indices are correct. But we can keep track of that here.

That being said, the `drop` method is wrong: indices should be adjusted.

model/src/main/scala/explore/optics/Adjuster.scala
rpiaggio: Yup, this class is mostly copied and adapted from Monocle 2, and I left it like that to preserve the current approach, but maybe it's a good idea to adopt the new mechanics.

model/src/main/scala/explore/optics/package.scala
rpiaggio: No. That's why zip is unsafe in the general case. Should we add `unsafe` to the name? In that case I would just name them `unsafeZip`.

model/src/main/scala/explore/undo/TreeMod.scala
rpiaggio: Yes, the idea is to come up with a `KeyedIndexedTree` and adapt the code in this PR.

rpiaggio:
> This seems to go in the right direction. I wonder if we could ask on the Monocle site what they think about the `Adjuster` optics

That's a good idea. Maybe someone already went through this.

cquiroz:
It seems some file needs formatting

model/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
```diff
@@ -0,0 +1,153 @@
...
+      case false => this
+    }
+
+  def inserted(key: K, node: Node[A], idx: Index[K]): KeyedIndexedTree[K, A] = {
+    def insertInChildren(
+      children:  List[Node[IndexedElem[K, A]]],
+      parentKey: Option[K] = None
+    ): List[Node[IndexedElem[K, A]]] =
+      (parentKey === idx.parentKey) match {
+        case true  =>
```
cquiroz, line 76: Maybe we can do this with a simple if rather than pattern matching on true/false

model/src/main/scala/explore/data/tree/tree.scala
```diff
@@ -3,18 +3,29 @@
...
+@Lenses final case class Node[A](value: A, children: List[Node[A]] = List.empty) {
+  def map[B](f: A => B): Node[B] =
+    Node(f(value), children.map(_.map(f)))
+}
 
 object Tree {
   def empty[A]: Tree[A] = Tree(List.empty)
   def apply[A](children: Node[A]*): Tree[A] = Tree(children.toList)
+
+  def eqTree[A]: Eq[Tree[A]] = Eq.fromUniversalEquals
```
cquiroz, line 22: Would it be better to request an `Eq[A]` and compare that way?

model/src/main/scala/explore/optics/Adjuster.scala
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.optics
+
+import monocle._
+
+// An Adjuster is just like a Setter, but less strict: it does not abide by the "compose modify" law.
+// It allows the modify/set functions to adjust the result (eg: not going out of bounds), which breaks composability.
+trait Adjuster[From, To] { self =>
```
cquiroz, line 10: Looks great

model/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
rpiaggio: Ok let's do that.

model/src/main/scala/explore/data/tree/tree.scala
rpiaggio: Good, idea. Done.

PR #207, opened 2020-07-06
Update to scala 2.13.3

cquiroz:
Updates scala, the PR is mostly removal of unusued imports


common/src/main/scala/explore/components/graphql/SubscriptionRender.scala
```diff
@@ -91,7 +90,7 @@ object SubscriptionRender {
       .componentWillUnmount { $ =>
         implicit val ce = $.props.ce
 
-        $.state.subscription.map(_.stop.runInCB).getOrEmpty
+        $.state.subscription.map(_.stop().runInCB).getOrEmpty
```
rpiaggio, line 93: Calling `stop` does not perform any side effects. Maybe I should remove the parenthesis in `clue`?

cquiroz: I guess that could be a good alternative

PR #217, opened 2020-07-08
Decouple constraints from target tab.

rpiaggio:
Start work on observations tab.


constraints/src/main/scala/explore/constraints/ConstraintsQueries.scala
```diff
@@ -151,7 +151,7 @@ object ConstraintsQueries {
         .void
     )
 
-  def constraintsSubscription(
+  def ConstraintsSubscription(
```
rpiaggio, line 154: We are treating this like a component, thus the upper casing,  but we can go back.

PR #219, opened 2020-07-08
Partial models

rpiaggio:
- Start using partial models when possible.
- Basic Observations tab functionality: can select, edit target, and edit constraints. Note that several observations can share the same constraints. The name of the constraint set is shown in the constraints panel.


observationtree/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -0,0 +1,61 @@
...
+        )
+      )
+    }
+
+  }
+
+  protected val component =
+    ScalaComponent
+      .builder[Props]
+      .backend(_ => new Backend())
```
cquiroz, line 58: Mabye we can use `renderBackend[Backend]`

rpiaggio: I think the macro expects the Backend constructor to have the scope (`$`) parameter.

PR #227, opened 2020-07-13
Switch to FomanticUI

cquiroz:
This PR switches to the SemanticUI fork. On the target editor checkboxes we can see the effect as they were marked as mini but that was not available on SemanticUI's css

image

The PR is fairly large but it is mostly about copying the updated themes


cquiroz:
Ups, big increase on css size

swalker2m:
This is the alternative that came up in the review?

cquiroz:
Yes, this is a fork that is actively
Maintained, they've fixed a few bugs which is why the checkboxes look smaller in the screenshot and added some extras which explains the increase on css size

rpiaggio:
Maybe it's possible to (easily) leave out the big parts that we don't use?

cquiroz:
This looks fine in heroku, I'll merge

PR #228, opened 2020-07-15
Basic elevation plot.

rpiaggio:
Code organization is not the final one yet. Computation of elevations will be moved to gsp-math.

### DIFF
--- a/common/src/main/resources/css/charts.css
+++ b/common/src/main/resources/css/charts.css
@@ -1,0 +1,68 @@
+@import 'https://code.highcharts.com/css/highcharts.css';
+/* TODO Can we import from the facade package? */
+
+/* Using custom classes for series works, but messes up tooltip border colors.
+See https://github.com/highcharts/highcharts/issues/13320 */
+
+/* Target Elevation */
+.highcharts-color-0 {
+  fill: #f45b5b;
+  stroke: #f45b5b;
+}
+
+/* Parallatic angle */
+.highcharts-color-1 {
+  fill: #8bbc21;
+  stroke: #8bbc21;
+}
+
+/* Sky Brightness */
+.highcharts-color-2 {
+  fill: #7cb5ec;
+  stroke: #7cb5ec;
+}
+
+/* Lunar Elevation */
+.highcharts-color-3 {
+  fill: #f7a35c;
+  stroke: #f7a35c;
+}
+
+.plot-axis-parallatic-angle {
+  stroke-width: 0;
+  fill: #8bbc21;
+}
+
+.plot-axis-parallatic-angle text {
+  fill: #8bbc21;
+}
+
+.plot-axis-sky-brightness {
+  stroke-width: 0;
+  fill: #7cb5ec;
+}
+
+.plot-axis-sky-brightness text {
+    fill: #7cb5ec;
+}
+
+.plot-band-twilight-nautical {
+  fill-opacity: 0.05;
+}
+
+.plot-band-twilight-nautical-end {
+  fill-opacity: 0;
+}
+
+.highcharts-plot-band-label {
+  fill: #666666;
+}
+
+.highcharts-xaxis-grid .highcharts-grid-line {
+	stroke-width: 1px;
+}
+
+.highcharts-minor-grid-line {
+  stroke: #e6e6e6;
+  stroke-dasharray: 2, 2;
+}
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -209,3 +209,36 @@ body {
 .dragging-over {
   background-color: gray;
 }
+
+.sky-plot {
+  height: 350px;
+}
+
+.sky-plot-section {
+  height: 390px;
+  background-color: white;
+  color: black;
+}
+
+.sky-plot-section .ui.checkbox + label {
+  color: black;
+}
+
+.sky-plot-controls {
+  display: flex;
+  justify-content: space-around;
+}
+
+.plot-toggle {
+  margin-left: 10px;
+  margin-right: 10px;
+  padding-top: 6px;
+}
+
+.plot-toggle.ui.slider.checkbox .box:before, .plot-toggle.ui.slider.checkbox label:before {
+  background-color: #545454 !important;
+}
+
+.plot-toggle.ui.slider.checkbox input:focus:checked ~ label:before {
+  background-color: #545454 !important;
+}
--- a/common/src/main/scala/explore/implicits.scala
+++ b/common/src/main/scala/explore/implicits.scala
@@ -3,12 +3,47 @@
 
 package explore
 
+import scala.annotation.unused
+
+import cats._
 import cats.effect.ContextShift
 import cats.effect.Timer
+import cats.implicits._
 import explore.model.AppContext
 import io.chrisdavenport.log4cats.Logger
+import shapeless._
 
-object implicits extends ShorthandTypes {
+trait ListImplicits {
+
+  // Adapted from https://stackoverflow.com/a/21444327/5808801
+  implicit object hnilMonoid extends Monoid[HNil] {
+    val empty = HNil
+    def combine(@unused f1: HNil, @unused f2: HNil) = HNil
+  }
+
+  implicit def hconsMonoid[H: Monoid, T <: HList: Monoid] =
+    new Monoid[H :: T] {
+      val empty = Monoid[H].empty :: Monoid[T].empty
+      def combine(f1: H :: T, f2: H :: T) =
+        (f1.head |+| f2.head) :: (f1.tail |+| f2.tail)
+    }
+
+  private object singleton extends Poly1 { implicit def anything[A] = at[A](List(_)) }
+
+  // def unzipN[L <: HList, Out <: HList](hlists: List[L])(implicit
+  //   mapper:                                    ops.hlist.Mapper.Aux[singleton.type, L, Out],
+  //   monoid:                                    Monoid[Out]
+  // ): Out = hlists.map(_.map(singleton)).combineAll
+
+  implicit class UnzipListOpts[L <: HList](hlists: List[L]) {
+    def unzipN[Out <: HList](implicit
+      mapper: ops.hlist.Mapper.Aux[singleton.type, L, Out],
+      monoid: Monoid[Out]
+    ): Out = hlists.map(_.map(singleton)).combineAll
+  }
+}
+
+object implicits extends ShorthandTypes with ListImplicits {
   implicit def appContext2ContextShift[F[_]](implicit ctx: AppContext[F]): ContextShift[F] = ctx.cs
   implicit def appContext2Timer[F[_]](implicit ctx:        AppContext[F]): Timer[F]        = ctx.timer
   implicit def appContext2Logger[F[_]](implicit ctx:       AppContext[F]): Logger[F]       = ctx.logger
--- a/common/src/main/scala/explore/implicits.scala
+++ b/common/src/main/scala/explore/implicits.scala
@@ -30,11 +30,6 @@ trait ListImplicits {
 
   private object singleton extends Poly1 { implicit def anything[A] = at[A](List(_)) }
 
-  // def unzipN[L <: HList, Out <: HList](hlists: List[L])(implicit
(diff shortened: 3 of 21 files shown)


common/src/main/resources/less/style.less
```diff
@@ -207,3 +207,21 @@ body {
...
+}
+
+.site-toggle {
+  margin-left: 10px;
+  margin-right: 10px;
+  padding-top: 6px;
+}
+
+.site-toggle.ui.slider.checkbox .box:before, .site-toggle.ui.slider.checkbox label:before {
+  background-color: #545454 !important;
```
cquiroz, line 222: Even being that specific you need an `!important`?

targeteditor/src/main/scala/explore/targeteditor/ElevationPlot.scala
```diff
@@ -0,0 +1,142 @@
...
+  class Backend( /*$ : BackendScope[Props, State]*/ ) {
+    def render(props: Props) = {
+
+      val start      = Instant.now()
+      val end        = start.plus(PlotPeriod)
+      val elevations =
+        ElevationCalc.elevationsForInterval(props.site, start, end, PlotEvery, _ => props.coords)
+
+      val options = Options()
+        .setChart(ChartOptions().setHeight(280))
```
cquiroz, line 68: I suppose this come from the generated facade. if we were to do this by hand, the height would be a field right?

targeteditor/src/main/scala/explore/targeteditor/ElevationPlot.scala
```diff
@@ -0,0 +1,142 @@
...
+              .setData(
+                elevations.collect {
+                  case (instant, angle) if angle.toSignedDoubleDegrees <= 90 =>
+                    PointOptionsObject()
+                      .setX(instant.toEpochMilli.toDouble)
+                      .setY(angle.toSignedDoubleDegrees)
+                      .asInstanceOf[Chart.Data]
+                }.toJSArray
+              )
+          ).map(_.asInstanceOf[SeriesOptionsType]).toJSArray
```
cquiroz, line 128: Maybe we can use `collect` here?

targeteditor/src/main/scala/explore/targeteditor/plotPeriod.scala
```diff
@@ -0,0 +1,39 @@
...
+import java.time.Instant
+import java.time.LocalDate
+
+import gem.Semester
+import gem.enum.Site
+import gem.math.LocalObservingNight
+import gem.math.ObservingNight
+
+sealed trait PlotPeriod {
+  def start(site: Site): Instant
```
cquiroz, line 16: Does this mean we could create a PlotPeriod with start on GN and end on GS?

common/src/main/resources/less/style.less
rpiaggio: The setting we are overriding has `!important` in SUI, so I couldn't find a way to override it otherwise.

targeteditor/src/main/scala/explore/targeteditor/ElevationPlot.scala
rpiaggio: Highcharts is supposed to adapt to the height of the container, but I can't get that work yet in our setup and it was showing it's default height of 400px. This is meant to be temporary for demo purposes. Worst case, I'm counting on `SizeMe`.

common/src/main/resources/less/style.less
cquiroz: ah ok, that makes sense

targeteditor/src/main/scala/explore/targeteditor/plotPeriod.scala
rpiaggio: The PlotPeriod is independent of the site. You can ask it its start and end instants in a Site. For example: for a NightPlot, the idea is to get the period from dusk till dawn, which yields different start and end times in each site. This is not working yet since we have to migrate the code to compute sunset and sunrise.

targeteditor/src/main/scala/explore/targeteditor/ElevationPlot.scala
rpiaggio: With the current typings generated by ScalablyTyped, I don't see a way to avoid the casting. This may even be an issue in the TS declarations.

`setSeries` is defined as `setSeries(value: js.Array[SeriesOptionsType])` and `SeriesOptionsType` is unrelated at all to `SeriesLineOptions` (or options for series of other types). Maybe we can hide the castings in the facade.

rpiaggio:
We are now computing things from gsp-math and gsp-core.

There are a few things missing but maybe we should merge so as not to diverge too much:
* Ability to choose dates.
* Ability to choose semesters.
* Draw moon phase.
* Better solve issues with height.

common/src/main/resources/css/charts.css
```diff
@@ -0,0 +1,68 @@
+@import 'https://code.highcharts.com/css/highcharts.css';
```
cquiroz, line 1: Will this be loaded at build time? or be included when loading the page?

common/src/main/scala/explore/implicits.scala
```diff
@@ -3,12 +3,42 @@
 
 package explore
 
+import scala.annotation.unused
```
cquiroz, line 6: Oddly, this seems unused

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -226,91 +226,90 @@ object TargetObsList {
...
-                  Droppable(target.id.toString) {
-                    case (provided, snapshot) =>
-                      <.div(
-                        provided.innerRef,
-                        provided.droppableProps,
-                        getListStyle(snapshot.isDraggingOver)
+              props.targetsWithObs.get.targets.elements.toTagMod { target =>
+                val targetId = target.id
+
+                val targetObs = obsByTarget.getOrElse(targetId, List.empty)
```
cquiroz, line 232: does `orEmpty` work?

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -226,91 +226,90 @@ object TargetObsList {
...
+
+                val opIcon =
+                  targetObs.nonEmpty.fold(
+                    Icon(
+                      "chevron " + state.collapsedTargetIds
+                        .exists(_ === targetId)
+                        .fold("right", "down")
+                    )(^.cursor.pointer,
+                      ^.onClick ==> { e: ReactEvent =>
+                        e.stopPropagationCB >> toggleCollapsed(targetId)
```
cquiroz, line 243: The other day I discovered
https://github.com/japgolly/scalajs-react/blob/master/core/src/main/scala/japgolly/scalajs/react/Callback.scala#L758

Don't know if it is useful

common/src/main/scala/explore/implicits.scala
```diff
@@ -3,12 +3,42 @@
...
+
+trait ListImplicits {
+
+  // Adapted from https://stackoverflow.com/a/21444327/5808801
+  implicit object hnilMonoid extends Monoid[HNil] {
+    val empty = HNil
+    def combine(@unused f1: HNil, @unused f2: HNil) = HNil
+  }
+
+  implicit def hconsMonoid[H: Monoid, T <: HList: Monoid] =
```
cquiroz, line 24: uh, galaxy brain level

targeteditor/src/main/scala/explore/targeteditor/SkyPlot.scala
```diff
@@ -0,0 +1,281 @@
...
+    def tag(a: ElevationSeries) = a.name
+
+    val all = List(Elevation, ParallacticAngle, SkyBrightness, LunarElevation)
+  }
+
+  private val seriesDataGen = Generic[SeriesData]
+
+  val dateTimeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("HH:mm")
+
+  def formatAngle(degs: Double): String = {
```
cquiroz, line 88: Can/should we use a `Formatter` from the angle class?

targeteditor/src/main/scala/explore/targeteditor/SkyPlotSection.scala
```diff
@@ -0,0 +1,114 @@
...
+
+object SkyPlotSection {
+  type Props = SkyPlotSection
+
+  protected sealed trait TimeDisplay
+  protected object TimeDisplay {
+    final case object UTC  extends TimeDisplay
+    final case object Site extends TimeDisplay
+
+    implicit val TimeDisplayEq: Eq[TimeDisplay]             = Eq.fromUniversalEquals
```
cquiroz, line 37: I suppose this can be `Eq.byRef` as they are singleton objects, but don't worry just a detail

targeteditor/src/main/scala/explore/targeteditor/SkyPlotSection.scala
```diff
@@ -0,0 +1,114 @@
...
+        case TimeDisplay.Site => site.timezone
+      }
+  }
+
+  implicit val propsReuse: Reusability[Props] = Reusability.derive
+  implicit val stateReuse: Reusability[State] = Reusability.derive
+
+  class Backend($ : BackendScope[Props, State]) {
+    val toggleSite: Callback =
+      $.modStateL(State.site) {
```
cquiroz, line 55: I really like `modStateL`

common/src/main/resources/css/charts.css
rpiaggio: It's resolved at page load time.

I'd like to eventually extract the css from the facade artifacts and resolve at build time, but I'm not sure it's making it into the facade artifacts or we have to do something to include it there.

common/src/main/scala/explore/implicits.scala
rpiaggio: It's used here: https://github.com/gemini-hlsw/explore/blob/ea8ae6a7cb0ab6cf517283bf33ee868073a48800/common/src/main/scala/explore/implicits.scala#L21

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: Yes, it does, thank you!

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: Interesting. Good tip.

common/src/main/scala/explore/implicits.scala
rpiaggio: Yes, but not mine. Original author is attributed in a comment ;)

targeteditor/src/main/scala/explore/targeteditor/SkyPlot.scala
rpiaggio: The existing fomatter: a) Uses `:` to separate all components instead of the symbols for degress, arcminutes and arcseconds. b) Doesn't round the decimals, showing quite a bit of them. It's appropriate for exact, human readable, string representation but for UI it wasn't looking so nice.

targeteditor/src/main/scala/explore/targeteditor/SkyPlotSection.scala
rpiaggio: Yes, good catch.

targeteditor/src/main/scala/explore/targeteditor/SkyPlot.scala
cquiroz: Ok, we may want to move these (later) to either math or `gpp-ui` to share them across projects

targeteditor/src/main/scala/explore/targeteditor/SkyPlotSection.scala
cquiroz: Oh right, there should be one :D

targeteditor/src/main/scala/explore/targeteditor/SkyPlot.scala
rpiaggio: That's a good idea.

PR #243, opened 2020-07-26
Update react-grid-layout and allow moving tiles

cquiroz:
This PR upgrades react grid layout and restore the ability to relocate tiles

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabContents.scala
@@ -99,9 +99,7 @@ object ObsTabContents {
                   margin = (5, 5),
                   containerPadding = (5, 5),
                   rowHeight = 30,
-                  draggableHandle = ".tileTitle",
-                  useCSSTransforms =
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -99,9 +99,7 @@ object ObsTabContents {
                   margin = (5, 5),
                   containerPadding = (5, 5),
                   rowHeight = 30,
-                  draggableHandle = ".tileTitle",
-                  useCSSTransforms =
```
rpiaggio, line 103: Does `react-grid-layout` 1.0 fix the problem? It happened when initializing the app on a multi-tile page directly, like with the tab `Observations`.

cquiroz: I tested it and didn't see any flicker but could test a bit more. Maybe I can deploy a test app to check without the reusability indicators

rpiaggio: I tested locally and it's working. Thank you!

PR #261, opened 2020-08-03
Traget Panning and some ui touches

cquiroz:
This PR makes the explore target not change when we do panning in aladin. Instead there is an internal target that we track and visualizations are drawn referring to the internal target.
FOV and current cursor position is also displayed now outside aladin and a center target button is also included


targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -37,11 +54,54 @@ final case class AladinContainer(
...
+    * On the state we keep the svg to avoid recalculations during panning
+    */
+  @Lenses
+  final case class State(svg: Option[Svg], fov: Fov, current: Coordinates)
+
+  object State {
+    val Zero: State = State(None, Fov(Angle.Angle0, Angle.Angle0), Coordinates.Zero)
+  }
+
+  // TODO: We may want to move these to gsp-math
```
rpiaggio, line 67: We should also probably unify with how it's shown in the elevation plots. Right now, in the plots I'm showing with the degrees, seconds and minutes symbols. What's the usual format for angles?

targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -95,68 +155,194 @@ object AladinContainer {
...
+        CallbackTo
+          .pure(
+            visualization
+              .shapesToSvg(GmosGeometry.shapes(p.options.posAngle),
+                           GmosGeometry.pp,
+                           v.pixelScale,
+                           GmosGeometry.ScaleFactor
+              )
+          )
+          .flatTap(svg => $.modState((s: State) => s.copy(svg = svg.some, fov = v.fov)))
```
rpiaggio, line 185: Is the type ascription required?

targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -37,11 +54,54 @@ final case class AladinContainer(
 object AladinContainer {
   type Props = AladinContainer
 
+  /**
+    * On the state we keep the svg to avoid recalculations during panning
```
rpiaggio, line 58: Nice

targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
cquiroz: In this case they are not really angles, they are coordinates so using : seems right
FOV is special as we want to show in degrees if it is a large FOV while going down to mins/seconds/etc when they get smaller

targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
cquiroz: Yeah, it didn't compile for me otherwise

PR #262, opened 2020-08-03
Moon phase and illumination in plot. Datepicker.

rpiaggio:
Also: SASS processing in webpack for configuring highcharts and datepicker.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -105,6 +105,7 @@ lazy val targeteditor = project
     libraryDependencies ++=
       GeminiLocales.value ++
         ReactAladin.value ++
+        ReactDatepicker.value ++
         ReactHighcharts.value ++
         ReactSizeMe.value
   )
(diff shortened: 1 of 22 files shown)


build.sbt
```diff
@@ -156,6 +157,13 @@ lazy val commonSettings = Seq(
   homepage := Some(url("https://github.com/geminihlsw/explore")),
   licenses := Seq("BSD 3-Clause License" -> url("https://opensource.org/licenses/BSD-3-Clause")),
   scalacOptions += "-Ymacro-annotations",
+  scalacOptions ~= (_.filterNot(
+    Set(
+      // By necessity facades will have unused params
+      "-Wdead-code",
+      "-Wunused:params"
+    )
+  )),
```
rpiaggio, line 166: This is because we are including a simple facade for the react-moon library. We should move it to gpp-ui eventually.

common/src/main/resources/css/charts.scss
```diff
@@ -1,5 +1,4 @@
-@import 'https://code.highcharts.com/css/highcharts.css';
-/* TODO Can we import from the facade package? */
+@use "~highcharts/css/highcharts.scss";
```
cquiroz, line 1: Is use, import in sass?

rpiaggio: @use is a "new @import" (@import is also supported but deprecated). @use imports things only one even if used multiple times, imported things are namespaced, plus other features for better modularization. More here: https://css-tricks.com/introducing-sass-modules/#import-files-with-use

PR #277, opened 2020-08-24
Move Tile.scala to common and fix Show[Constraints] format string.

toddburnside:


model/src/main/scala/explore/model/Constraints.scala
```diff
@@ -27,6 +27,6 @@ object Constraints {
     Eq.by(x => (x.id, x.name, x.cc, x.iq, x.sb, x.wv))
 
   implicit val showConstraints: Show[Constraints] =
-    Show.show(x => s"$name (${List(x.cc, x.iq, x.sb, x.wv).mkString(", ")}")
+    Show.show(x => s"${x.name} (${List(x.cc, x.iq, x.sb, x.wv).mkString(", ")})")
```
rpiaggio, line 30: Shall we make it explicit in the string that this is a `Constraints`? Maybe we can get away with`Show.fromToString`? (which I admit I didn't know existed when I wrote this) i

toddburnside: I think `Show.fromToString` is probably sufficient and would explicitly indicate it is a Constraint, but I guess that depends on how Show is going to be used. However, it might be even nicer if the CloudCover, ImageQuality, etc. indicated what they were, instead of just showing `Percent50`, or `Any`. One way to fix that would be to override the toString methods in CloudCover, etc., to print something like `CloudCover: 70%/Cirrus` (use the label of the enums).

Typelevel/kittens also allows you to derive Show instances that include field names and uses existing Show instances for the parameters, but we might not want to drag in that dependency.

rpiaggio: I think the idea is to use `Show` for debugging. I.e. never show that output to users. (@cquiroz can you confirm this?)

Anyway, this is still in prototype stage and we don't know yet how constraints are going to look like. There's a chance the final model will hold numeric values instead of enums.

I didn't know about that functionality of kittens, that's nice. I don't know if it's feasible in this case, but just for future reference: Scala.js allows us to define code that is only linked in when in development mode, and is shaken off when linking to prod.

cquiroz: I don't expect we should rely on `Show` (nor `toString` for that matter) to show users. `Show` could be more for internal debugging

toddburnside: Should we just just stick with the current, simple implementation until such time as the model matures?

swalker2m: Agree that `Show` and `toString` should not be used for anything other than debugging.  In fact, I think we should always make them the same wherever we define `Show`.  I've taken to writing them in a way that I can copy and paste the output directly into the repl (or whatever).

If there is a particular `String` format that we wish to actually use for something, we should consider adding a `Format` instance.

In the end, I do think the constraints we show to the user are going to be based on real values and not percentage buckets though we might have to map them back to buckets for ITC.

rpiaggio: > Should we just just stick with the current, simple implementation until such time as the model matures?

I think it would still be nice to show all the info (id and name).

toddburnside: OK. I'll make this a little nicer - correlating toString with Show and formatting the entire thing better.

toddburnside: @swalker2m With respect to writing `toString` such that the output can be copied into a REPL, do you have a way of doing that which is refactor safe? For the Constraints I'd need things like `UUID.fromString("58f4b80b-c816-46cc-a98d-e5c9198fb303")` and `CloudCover.Percent50` in the output. For example, I don't know how to safely get the `CloudCover` part of `CloudCover.Percent50`.

swalker2m: You're right, but maybe that's not too terrible.  If you're debugging something that has been refactored such that the `toString` doesn't correspond you can update it then.  I wouldn't put a lot of machinery into it.  It's just something that, for me, has made debugging a bit easier sometimes.

toddburnside: Pushed change that makes the show output pastable into a REPL and makes toString and show output the same.

model/src/main/scala/explore/model/Constraints.scala
```diff
@@ -18,15 +18,22 @@ final case class Constraints(
   iq:   ImageQuality,
   sb:   SkyBackground,
   wv:   WaterVapor
-)
+) {
+  override def toString: String =
+    s"""$productPrefix(UUID.fromString("$id")
```
rpiaggio, line 23: Nice.

PR #293, opened 2020-09-07
Upgrade fomantic ui 

cquiroz:
This is an upgrade of fomantic ui. I didn't check to much details for changes on the UI but at first glance it seems ok
This PR aims to test the fixes for packtracker and integration with clubhouse too


rpiaggio (approved):
LGTM

PR #315, opened 2020-09-15
Add initial Proposal Details model.

toddburnside:
Just the elements needed for initial draft of UI. Does not include co-investigators or time allocations.

### DIFF
--- a/model/src/main/scala/explore/model/ProposalDetails.scala
+++ b/model/src/main/scala/explore/model/ProposalDetails.scala
@@ -1,0 +1,24 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats._
+import explore.model.enum._
+import lucuma.core.model.StandardUser
+import monocle.macros.Lenses
+
+@Lenses
+final case class ProposalDetails(
+  title:         String,
+  pi:            StandardUser,
+  proposalClass: ProposalClass,
+  category:      TacCategory,
+  toOActivation: ToOActivation,
+  keywords:      List[Keyword],
+  abstrakt:      String
+)
+
+object ProposalDetails {
+  implicit val equalProposalDetails: Eq[ProposalDetails] = Eq.fromUniversalEquals
+}
--- a/model/src/main/scala/explore/model/ProposalDetails.scala
+++ b/model/src/main/scala/explore/model/ProposalDetails.scala
@@ -15,7 +15,7 @@ final case class ProposalDetails(
   proposalClass: ProposalClass,
   category:      TacCategory,
   toOActivation: ToOActivation,
-  keywords:      List[Keyword],
+  keywords:      Set[Keyword],
   abstrakt:      String
 )
 
--- a/model/src/main/scala/explore/model/enum/ProposalClass.scala
+++ b/model/src/main/scala/explore/model/enum/ProposalClass.scala
@@ -1,0 +1,21 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model.enum
+
+import lucuma.core.util.Enumerated
+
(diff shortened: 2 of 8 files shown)


model/src/main/scala/explore/model/ProposalDetails.scala
```diff
@@ -0,0 +1,24 @@
...
+import monocle.macros.Lenses
+
+@Lenses
+final case class ProposalDetails(
+  title:         String,
+  pi:            StandardUser,
+  proposalClass: ProposalClass,
+  category:      TacCategory,
+  toOActivation: ToOActivation,
+  keywords:      List[Keyword],
```
cquiroz, line 18: I guess this could be a `Set`?

model/src/main/scala/explore/model/enum/ProposalClass.scala
```diff
@@ -0,0 +1,21 @@
...
+sealed abstract class ProposalClass(val label: String) extends Product with Serializable
+
+object ProposalClass {
+  case object LargeProgram   extends ProposalClass("Large Program Observing at Gemini")
+  case object FastTurnaround extends ProposalClass("Fast Turnaround Observing at Gemini")
+  case object Queue          extends ProposalClass("Queue Observing at Gemini")
+  case object Classical      extends ProposalClass("Classical Observing at Gemini")
+  case object Exchange       extends ProposalClass("Exchange Observing at Keck/Subaru")
+  case object Intensive      extends ProposalClass("Intensive Program Observing at Subaru")
+  case object Other          extends ProposalClass("Other Proposal Types")
```
cquiroz, line 17: Maybe we should expand Other into `DT`, `SV`, etc

cquiroz (approved):
It's looking great. I added a few small comments

swalker2m (commented):
Please forgive my ignorance on this stuff.  Does `explore.model` work like a staging ground for `lucuma.core.model`?  The enums at least would probably make sense down in lucuma?  We'll need to store this stuff in the database and generate corresponding GraphQL enums for example.

cquiroz:
That is the idea, it is a playground to work on the UI eventually moving to core if needed
It is kind of unclear yet if there is going to be a rich model in core, is it?

swalker2m:
> That is the idea, it is a playground to work on the UI eventually moving to core if needed
> It is kind of unclear yet if there is going to be a rich model in core, is it?

Yeah and that makes sense to me.  The enums though will surely be needed or we'll need to duplicate them.  It's fine here for now and this all looks great. :+1:

model/src/main/scala/explore/model/ProposalDetails.scala
toddburnside: Good point. I think I need to make some changes to `EnumViewMultipleSelect` to handle that. It would make sense that any time we use a multiple select that the result is a `Set`.

model/src/main/scala/explore/model/enum/ProposalClass.scala
toddburnside: I'd be happy to. These are just the values I found in the Proposal Class dropdown in the PIT.

toddburnside:
> > That is the idea, it is a playground to work on the UI eventually moving to core if needed
> > It is kind of unclear yet if there is going to be a rich model in core, is it?
> 
> Yeah and that makes sense to me. The enums though will surely be needed or we'll need to duplicate them. It's fine here for now and this all looks great. 👍

When the enums are moved to lucuma core we'll want to add explicit tags for type safety.

model/src/main/scala/explore/model/enum/TacCategory.scala
```diff
@@ -0,0 +1,32 @@
...
+sealed abstract class TacCategory(val label: String) extends Product with Serializable
+
+object TacCategory {
+  case object PlanetarySystems         extends TacCategory("Planetary systems")
+  case object StarAndPlanetFormation   extends TacCategory("Star and planet formation")
+  case object StarsAndStellarEvolution extends TacCategory("Stars and stellar evolution")
+  case object FormationOfCompactObjects
+      extends TacCategory("Formation and evolution of compact objects")
+  case object ResolvedStellarPopulations
+      extends TacCategory("Resoloved stellar populations and their environments")
```
cquiroz, line 17: `Resolved`

rpiaggio (approved):
LGTM

PR #317, opened 2020-09-17
Modify the way the Side Tab buttons are created.

toddburnside:
When I started adding the Proposal tab button, I found it was difficult to add a new button given the current implementation. I started needing things like `tabsL.tail.head`, etc.

I came up with this method of specifying tab groups, which will also allow filtering out tabs that shouldn't be shown. For example, the Proposal tab should only be shown if the user is logged in.

This is just one option. Let me know what you think.

### DIFF
--- a/explore/src/main/scala/explore/SideTabs.scala
+++ b/explore/src/main/scala/explore/SideTabs.scala
@@ -58,15 +58,14 @@ object SideTabs {
           case AppTab.Constraints    => 2
         }
 
-        def tabButton(style: Css)(tab: AppTab): Button =
+        def tabButton(tab: AppTab): Button =
           Button(active = tab === focus,
-                 clazz = style,
                  onClick = p.tabs.mod(z => z.findFocus(_ === tab).getOrElse(z)).runInCB
           )(tab.title)
 
         def makeButtonSection(tabs: List[AppTab]): TagMod = tabs match {
-          case justOne :: Nil => VerticalSection()(tabButton(Css.Empty)(justOne))
-          case _              => VerticalSection()(ButtonGroup(tabs.reverse.map(tabButton(Css.Empty)).toTagMod))
+          case justOne :: Nil => VerticalSection()(tabButton(justOne))
+          case _              => VerticalSection()(ButtonGroup(tabs.reverse.map(tabButton).toTagMod))
         }
 
         val buttonSections: List[TagMod] =
--- a/explore/src/main/scala/explore/SideTabs.scala
+++ b/explore/src/main/scala/explore/SideTabs.scala
@@ -3,7 +3,7 @@
 
 package explore
 
-import cats.syntax.all._
+import cats.implicits._
 import crystal.react.implicits._
 import explore.components.ui.GPPStyles
 import explore.model.enum.AppTab
@@ -40,13 +40,17 @@ object SideTabs {
 
         def makeButtonSection(tabs: List[AppTab]): TagMod = tabs match {
           case justOne :: Nil => VerticalSection()(tabButton(justOne))
-          case _              => VerticalSection()(ButtonGroup(tabs.reverse.map(tabButton).toTagMod))
+          case _              =>
+            VerticalSection()(
+              ButtonGroup(tabs.sortBy(_.groupOrder.value).reverse.map(tabButton).toTagMod)
+            )
         }
 
         val buttonSections: List[TagMod] =
           tabsL.toList
             .groupBy(_.buttonGroup)
             .toList
+            .sortBy(_._1)
             .map(tup => makeButtonSection(tup._2))
 
         <.div(
--- a/explore/src/main/scala/explore/SideTabs.scala
+++ b/explore/src/main/scala/explore/SideTabs.scala
@@ -42,7 +42,7 @@ object SideTabs {
           case justOne :: Nil => VerticalSection()(tabButton(justOne))
           case _              =>
             VerticalSection()(
-              ButtonGroup(tabs.sortBy(_.groupOrder.value).reverse.map(tabButton).toTagMod)
+              ButtonGroup(tabs.reverse.map(tabButton).toTagMod)
             )
         }
 
--- a/explore/src/main/scala/explore/SideTabs.scala
+++ b/explore/src/main/scala/explore/SideTabs.scala
@@ -25,6 +25,21 @@ object SideTabs {
 
   implicit val propsReuse: Reusability[Props] = Reusability.derive
 
+  // This is a protected method on cats.Foldable. If we add
+  // fs2 to the project, we can switch to intersperse on fs2.Stream
+  def intersperseList[A](xs: List[A], x: A): List[A] = {
+    val bld = List.newBuilder[A]
+    val it  = xs.iterator
+    if (it.hasNext) {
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/SideTabs.scala
```diff
@@ -33,26 +48,41 @@ object SideTabs {
...
 
+        def makeButtonSection(tabs: List[AppTab]): TagMod = tabs match {
+          case justOne :: Nil => VerticalSection()(tabButton(justOne))
+          case _              => VerticalSection()(ButtonGroup(tabs.reverse.map(tabButton).toTagMod))
+        }
+
+        val buttonSections: List[TagMod] =
+          tabsL.toList
+            .map(t => (tabGrouping(t), t))
+            .groupMap(_._1)(_._2)
```
rpiaggio, line 74: Would `groupBy(tabGrouping)` work here?

explore/src/main/scala/explore/SideTabs.scala
```diff
@@ -25,6 +25,21 @@ object SideTabs {
 
   implicit val propsReuse: Reusability[Props] = Reusability.derive
 
+  // This is a protected method on cats.Foldable. If we add
+  // fs2 to the project, we can switch to intersperse on fs2.Stream
+  def intersperseList[A](xs: List[A], x: A): List[A] = {
```
cquiroz, line 30: We can probably add fs2

explore/src/main/scala/explore/SideTabs.scala
```diff
@@ -33,26 +48,41 @@ object SideTabs {
         val tabsL = p.tabs.get.toNel
         val focus = p.tabs.get.focus
 
-        def tabButton(style: Css)(tab: AppTab): Button =
+        // group the tab buttons. Setting the value to 0 will
+        // cause the button to NOT be displayed.
+        def tabGrouping(tab: AppTab): Integer = tab match {
+          case AppTab.Overview       => 1
```
cquiroz, line 54: Could we name groups using some enum rather than an int?

explore/src/main/scala/explore/SideTabs.scala
```diff
@@ -33,26 +48,41 @@ object SideTabs {
...
+          case AppTab.Observations   => 2
+          case AppTab.Targets        => 2
+          case AppTab.Configurations => 2
+          case AppTab.Constraints    => 2
+        }
```
rpiaggio, line 59: Let's put this info in `AppTab`, so that it's in the model and not in the display logic.

Let's also (eventually?) put there the condition required to display each tab (maybe it's just a list of roles, or a `User => Boolean`?). We can then filter here. We can use that info to validate when a user lands on a tab URL too. But I don't know if we are integrated with sso yet, so I wouldn't worry about this until we do have a model for users/roles.

For filtering, I **think** we can use `Zipper.unsafeSelect` traversal, or add a `filter` method to `Zipper`.

explore/src/main/scala/explore/SideTabs.scala
rpiaggio, line 30: I **think** that after we have a `List[List[A]]`, we can use cat's `intercalate` with a `List(x)` to achieve this.

Also, `fs2` is available since it's imported transitively. But we could add it explicitly if we are going to build streams. IDK what the right criteria would be for explicitly declaring libs

rpiaggio:
Thank you for this initiative, it's definitely something we need to improve. I left some comments inline.

cquiroz (commented):
I don't recall that proposals con be displayed only for logged in users.
Should we let anonymous users sketch a proposal

Regardless this seems like a generically useful feature

Added a few comments inline

explore/src/main/scala/explore/SideTabs.scala
toddburnside: Yes. Good catch.

explore/src/main/scala/explore/SideTabs.scala
toddburnside: At the point where we call intersperse, we have a `List[TagMod]`, not a `List[List[TagMod]]`. I could make it work by having `makeButtonSection` return `List[TagMod]` instead of a `TagMod`, bit that feels wrong to me.

I didn't think about fs2 being pulled in transitively.

toddburnside:
I think we should discuss some of the details of implementation in our "UI team" meeting today.

explore/src/main/scala/explore/SideTabs.scala
toddburnside: Button grouping moved to AppTab. We will had filtering when we come to needing it.

explore/src/main/scala/explore/SideTabs.scala
toddburnside: `mkTagMod` was right there in the VDOM documentation. Thanks for thinking of it, Carlos. That cleaned things up nicely.

rpiaggio (approved):
LGTM

model/src/main/scala/explore/model/enum/AppTab.scala
```diff
@@ -4,20 +4,42 @@
...
 import cats.data.NonEmptyList
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.PosInt
 import lucuma.core.util.Enumerated
 
-sealed abstract class AppTab(val title: String) extends Product with Serializable
+sealed abstract class AppTab(
+  val title:       String,
+  val buttonGroup: ButtonGroup,
+  val groupOrder:  PosInt
```
rpiaggio, line 14: If I understand correctly, `groupOrder` is being used to define the order of tabs within a group. Since order is preserved in groups after `groupBy`, we might not need this.

cquiroz: Is it so that `groupOrder` is ordering inside the group? While the ordering of groups is defined by
the `ButtonGroup` enumerated?

toddburnside: Correct, groupOrder provides ordering within groups, while the ordering of groups is provided by the Enumerated `Order` instance.

@rpiaggio Yes, ordering is preserved, so technically we don't need this. But, the direction of comments on the PR seemed to be towards explicitly stating groups and orders so I included it.

Personally, this commit seems like overkill to me. The button group/order will seldom be changed and, when it is, it should be verified by the developer. It seems reasonable to require the AppTabs be `Order`ed in the order they will be displayed, and use a simple method of grouping them. I think that even including the `ButtonGroup` Enumerated is extra machinery that does little to provide safety. The developer can still assign a button to the wrong `ButtonGroup` and the compiler can't catch that. But, that is just one opinion. 😄

toddburnside: @rpiaggio Carlos and I were discussing this, and think going back to the "int in AppTab" is probably a good balance between simplicity and control. If we do this, I'll add a comment for extra clarification. What do you think?

rpiaggio: Ok, sounds good for this use case.

cquiroz: Yeah it seems it is getting to complex for a small gain in safety

PR #322, opened 2020-09-24
Initial draft of the Proposal Details UI.

toddburnside:
Does not include changing the PI, adding co-investigators, or allocating requested time to partners. These will be part of upcoming stories.


common/src/main/scala/explore/components/FormStaticData.scala
```diff
@@ -0,0 +1,45 @@
...
+) extends ReactProps[FormStaticData](FormStaticData.component) {
+  def apply(mods: TagMod*): FormStaticData = copy(modifiers = modifiers ++ mods)
+}
+
+object FormStaticData {
+  type Props = FormStaticData
+
+  val component =
+    ScalaComponent
+      .builder[Props]("FormStaticData")
```
rpiaggio, line 32: We can omit the name in the newer `scalajs-react`.

common/src/main/scala/explore/components/FormStaticData.scala
```diff
@@ -0,0 +1,45 @@
...
+      .render_P { props =>
+        <.div(
+          props.modifiers.toTagMod,
+          (GPPStyles.HideLabel).when(props.hideLabel),
+          ^.cls := "field",
+          <.label(props.label, ^.htmlFor := props.id),
+          <.div(^.cls := "ui input",
```
rpiaggio, line 39: Would it be possible to use SUI controls? `Label` and `Input`?

toddburnside: > Would it be possible to use SUI controls? `Label` and `Input`?

Probably. I'll see what I get.

rpiaggio (approved):
LGTM

cquiroz:
Great to see this. Could you make a review app

common/src/main/scala/explore/components/FormStaticData.scala
toddburnside: >  I might be able to use an SUI `Label`.
It turns out the using a Label complicates things more than I thought. It adds the `ui label` classes to the label, which styles it unlike the labels for the other Form* controls. Apparently, if you pass a string (not a Label) to a Form* control in the `label` parameter, it doesn't actually use a an SUI label. I could make the `label` parameter here be a `String | Label` union to allow usage of an SUI Label if desired.

toddburnside:
Here's a screenshot.
image

toddburnside:
> Great to see this. Could you make a review app

Review app at https://explore-pipe-ch216-init-pvruxz.herokuapp.com/

cquiroz:
Tested the review app and it is looking great. A tiny nitpick is to start focused on the first field (title)

cquiroz:
Tested the review app and it is looking great. A tiny nitpick is to start focused on the first field (title)

toddburnside:
> Tested the review app and it is looking great. A tiny nitpick is to start focused on the first field (title)

Thanks. I'll set the focus.

I'm also working on what I believe is an improved method of laying out the form that eliminates the SUI `FormGroup` and uses CSS Grid behind the scenes. I was having issues with the responsive behavior of the layout with nested `FormGroup`. And, I found it hard to have good control over it. This new layout method will be easily re-usable for other CSS grid layouts both inside and outside of forms.

cquiroz:
That's a great idea, SUI is pre css grid and we can use the native grid if it suits us better

common/src/main/resources/less/gpp-grid.less
```diff
@@ -0,0 +1,91 @@
...
+
+.gpp-grid {
+    display: grid;
+    grid-gap: 1em;
+
+    /*
+     You do not not to specify the row and column of each cell 
+     for most simple layouts, such as most forms. It will fill 
+     in each cell in a row and then go to the next column.
+    */
```
toddburnside, line 12: If you like this concept, I can add classes for specifying rows and columns, as well as row and column spans. This would be useful for more complex layouts.

proposal/src/main/scala/explore/proposal/ProposalDetailsEditor.scala
```diff
@@ -0,0 +1,118 @@
...
+                )
+              ),
+              EnumViewMultipleSelect(
+                details.zoom(ProposalDetails.keywords),
+                label = "Keywords",
+                search = true
+              ),
+              <.div(
+                GPPStyles.CellWrapper,
+                FormStaticData(value = "7.50h", id = "requested", label = "Requested Time"),
```
toddburnside, line 67: I could make a CellWrapper component to simplify this.

toddburnside (commented):
If you like the general method, I can address the issues I placed in the comments.

common/src/main/resources/less/gpp-grid.less
```diff
@@ -0,0 +1,91 @@
+@two-column-break: 60em;
+@three-column-break: 90em;
```
toddburnside, line 2: We can play with these column breaks, but they will of necessity be inexact until CSS provides a way of querying container size, not just the viewport size.

proposal/src/main/scala/explore/proposal/ProposalTabContents.scala
```diff
@@ -0,0 +1,58 @@
...
+    def render() =
+      AppCtx.withCtx { implicit appCtx =>
+        val ref: SignallingRef[IO, ProposalDetails] =
+          SignallingRef.in[SyncIO, IO, ProposalDetails](proposalDetails).unsafeRunSync()
+
+        val component = StreamRendererMod.build(ref.discrete)
+
+        component(_ match {
+          case Ready(view) => ProposalDetailsEditor(view)
+          case _           => <.div("Ruh-Roh")
```
cquiroz, line 50: I guess this never happens because the ref is initialized right away?

cquiroz (approved):
This is looking great, I especially like the idea of using css grid for the forms.
I added some comments inline

common/src/main/resources/theme/collections/form.overrides
```diff
@@ -0,0 +1,77 @@
+/*
+  Style a readonly FormInput as static data. 
+  The FormInput requires a className of 'static-data'
+*/
+.ui.form .static-data input[readonly] {
```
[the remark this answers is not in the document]
toddburnside: `static-data` is not part of the SUI, but the styling makes use of the SUI less variables. I'm not sure we can move it entirely into explore's css. I might be wrong, though.

common/src/main/scala/explore/components/FormStaticData.scala
```diff
@@ -0,0 +1,45 @@
...
+    ScalaComponent
+      .builder[Props]
+      .render_P { props =>
+        <.div(
+          props.modifiers.toTagMod,
+          (GPPStyles.HideLabel).when(props.hideLabel),
+          ^.cls := "field",
+          <.label(props.label, ^.htmlFor := props.id),
+          <.div(^.cls := "ui input",
+                <.data(props.value, GPPStyles.StaticData, ^.id := props.id, ^.tabIndex := 0)
```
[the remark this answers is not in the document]
toddburnside: For the gpp-grid.css comment, I think that what is currently there works fine for simple forms, and in conjunction with the line break stuff makes for a responsive UI. But, I have some ideas for making more general grid layout functionality that would work for page layouts, and other stuff outside of simple forms. I'm going to do a quick feasibility check on that.

About the tab index, a tab index of 0 means to use the html source order of fields for the tabbing. This is the default for elements like `input`s. Generally, the default behavior is what is desired, so this makes setting up tabbing easier. But, other elements, such as `data` have it set (I think) to -1, which means it isn't accessible by tabbing. I set the tab index explicitly here to allow behavior consistent with a readonly `input`. My hope is that by using an element with some semantic meaning (`data`) in conjunction with a label and ability to tab into, will make it more valuable to screen readers. But, I'm not completely certain of that.

proposal/src/main/resources/dev.js
```diff
@@ -0,0 +1,9 @@
+import "resources/theme/semantic.less";
+import "resources/less/style.less";
+
```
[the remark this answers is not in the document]
toddburnside: I'm importing `gpp-grid.less` into `style.less`, following the precedent set by `verticalbuttons.less`.

proposal/src/main/scala/explore/proposal/ProposalDetailsEditor.scala
```diff
@@ -0,0 +1,118 @@
...
+import react.semanticui.collections.form._
+import react.semanticui.addons.textarea.TextArea
+
+final case class ProposalDetailsEditor(proposalDetails: View[ProposalDetails])
+    extends ReactProps[ProposalDetailsEditor](ProposalDetailsEditor.component)
+
+object ProposalDetailsEditor {
+  type Props = ProposalDetailsEditor
+
+  class Backend() {
```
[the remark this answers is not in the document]
toddburnside: Yep, you're right. I'll change it.

proposal/src/main/scala/explore/proposal/ProposalTabContents.scala
toddburnside: Regarding focusing, it should be receiving focus via tabbing as indicated by the border changing to a light blue color, just like it does for the readonly FormInput. But, again like the readonly FormInput, there is no cursor because it isn't editable. If the border doesn't change color to indicate focus for you, let me know.

For ProposalTabContents, I believe that is true. We're not dealing with any real latency from a network or anything.

PR #323, opened 2020-09-28
Fix for new GraphQL Schema

rpiaggio:
I reset the DB and Hasura slightly changed the schema.

This PR makes most things work again, except for editing conditions.

I'm sending it anyway so as not to block anyone, and things are working anyway in the not-yet-ready-to-merge branch that uses the new `clue`.


cquiroz (approved):
LGTM

PR #324, opened 2020-09-29
ODB endpoint in env variable.

rpiaggio:
The environment variable, `ODB_ENDPOINT`, is read at _build time_, so if it is changed the app must be rebuilt.


swalker2m (approved):
Makes sense to me. :+1:

toddburnside:
LGTM

PR #325, opened 2020-09-29
Improvements to CSS grid layout.

toddburnside:
Also allows the flex container to be used outside of a grid.


common/src/main/resources/less/gpp-grid.less
```diff
@@ -1,6 +1,15 @@
 @two-column-break: 60em;
 @three-column-break: 90em;
 
+/*
+ Overrides the default values of fields inside an SUI form.
+ These values interfere with the gpp-grid layout.
+ */
+.ui.form.gpp-grid .field {
+    margin: 0;
```
toddburnside, line 9: I was following prior art, but should we change all the `gpp` to `explore` or `lucuma`, and rename `GPPStyles` similarly?

cquiroz: I suppose we should, maybe explore considering it is explore specific but if it is going to be used more broadly, lucuma?

common/src/main/resources/less/gpp-grid.less
```diff
@@ -31,61 +40,42 @@
...
+ nature of the grid - they will remain grouped together even
+ when the grid goes down to one column.
+*/
+.gpp-flex-container {
+    display: flex;
 
-            &-three {
-                flex: 3;
-            }
+    .ui.form & > * + *, // override SUI
```
cquiroz, line 57: Cool, I didn't know this construct

common/src/main/scala/explore/components/ui/GPPStyles.scala
```diff
@@ -71,15 +73,12 @@ object GPPStyles {
...
   val TwoColumnGrid: Css   = Grid |+| Css("gpp-two-columns")
   val ThreeColumnGrid: Css = Grid |+| Css("gpp-three-columns")
-  val CellWrapper: Css     = Css("gpp-cell-wrapper")
 
-  val GrowOne: Css   = Css("gpp-grow-one")
-  val GrowTwo: Css   = Css("gpp-grow-two")
-  val GrowThree: Css = Css("gpp-grow-three")
-  val GrowFour: Css  = Css("gpp-grow-four")
+  val FlexContainer: Css = Css("gpp-flex-container")
+  def Grow(i: Int Refined Interval.Closed[1, 4]): Css = Css(s"gpp-grow-$i")
```
cquiroz, line 78: Neat

cquiroz (approved):
This is looking great, some good use of less constructs

common/src/main/resources/less/gpp-grid.less
rpiaggio: 🤯

rpiaggio (approved):
It's very nice to have these unified layout styles. LGTM!

PR #330, opened 2020-10-01
Fix for odd shifts when hovering in FF

cquiroz:
I noted a bit of flickering when hovering on the vertical buttons on FF

This PR fixes it. Tested in FF, Chrome and Safari

### DIFF
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -62,6 +62,19 @@ body {
   background: darkgrey;
 }
 
+// Taken from the definitíon of ui.button transition but removed background-color and background
+// When animating those two properties you see some flickering in FF
+.ui.button.side-button {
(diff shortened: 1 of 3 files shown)


toddburnside:
I don't see any flickering in FF, but it LGTM.

common/src/main/resources/less/style.less
```diff
@@ -62,6 +62,15 @@ body {
   background: darkgrey;
 }
 
+// Taken from the definitíon of ui.button transition but removed background-color and background
+// When animating those two properties you see some flickering in FF
```
rpiaggio, line 66: Does this apply only to side buttons or to all buttons?

cquiroz: Only side buttons, it looks fine in regular buttons

PR #331, opened 2020-10-01
WIP Tooltips + clue websocket connection status indicator

rpiaggio:
WIP wrapper component that shows "WIP" tooltip on hover.

Connection status indicator on top right corner, with tooltip providing details, useful in case of error.

image

image

image

image

### DIFF
--- a/common/src/main/scala/explore/components/ConnectionsStatus.scala
+++ b/common/src/main/scala/explore/components/ConnectionsStatus.scala
@@ -1,0 +1,50 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.components
+
+import clue.StreamingClientStatus._
+import crystal._
+import explore.AppCtx
+import explore.components.ui.ExploreStyles._
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import react.common.ReactProps
+import react.semanticui.elements.icon._
+import react.semanticui.modules.popup._
+
+final case class ConnectionsStatus()
+    extends ReactProps[ConnectionsStatus](ConnectionsStatus.component)
+
+object ConnectionsStatus {
+  type Props = ConnectionsStatus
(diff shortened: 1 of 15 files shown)


model/src/main/scala/explore/model/RootModel.scala
```diff
@@ -23,61 +17,3 @@ case class RootModel(
 object RootModel {
   implicit val eqRootModel: Eq[RootModel] = Eq.by(m => (m.tabs, m.focused))
 }
-
```
rpiaggio, line 26: This was pending cleanup.

common/src/main/scala/explore/components/ConnectionsStatus.scala
```diff
@@ -0,0 +1,50 @@
...
+                case Closing    => ("Closing...", ConnectionWarning)
+                case Closed     => ("Closed", ConnectionError)
+              }
+          }
+
+          Popup(
+            header = "ODB Connection Status",
+            content = message,
+            position = PopupPosition.BottomRight,
+            trigger = Icon(name = "circle", clazz = clazz)
```
cquiroz, line 43: Should we animate on `Connecting`?

cquiroz (approved):
Great idea, thanks!

common/src/main/scala/explore/model/AppContext.scala
```diff
@@ -6,19 +6,21 @@ package explore.model
...
 
-case class AppConfig(programsURL: Uri)
+case class AppConfig(odbURI: Uri)
 
 case class Clients[F[_]: ConcurrentEffect: Logger](
-  programs: GraphQLStreamingClient[F]
+  odb: GraphQLStreamingClient[F]
 ) {
-  // lazy val programsConnectionStatus =
-  // StreamRenderer.build(programs.statusStream)
```
toddburnside, line 18: Delete these?

common/src/main/scala/explore/components/ConnectionsStatus.scala
rpiaggio: It's a good idea, but in practice it happens so fast I doubt it would be noticeable.

PR #337, opened 2020-10-03
Removes unused theme and upgrades js dev libs

cquiroz:
An extra sui theme is present that we don't need and creates confusion


toddburnside (approved):
LGTM

PR #341, opened 2020-10-05
Modify Target Position tile to use CSS Grid

toddburnside:
This changes the `Target Position` tile, which appears on both the Targets and Observations tabs, to use a CSS grid for layout.

I made a `compact` grid layout which reduces the space between cells as compared to what is currently used on the Proposal tab and used it for the `CoordinatesForm`. I also increased the Form Size to `Small` which makes it more readable. WIth the compact grid it still takes up slightly more space than the previous version, but decreasing the padding in the inputs will maintain readability and reduce size. I have not modified the `CataloguesForm` but similar changes could be made there.

### DIFF
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -359,6 +359,7 @@ body {
 }
 
 .target-aladin-cell {
+  height: 100px;
   min-height: 300px; // don't collapse in single column mode.
 }
 
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -78,13 +78,13 @@ body {
 }
 
 .explore-tile-body {
-  height: ~"calc(100% - @{minHeight}/2)";
+  height: ~"calc(100% - @{minHeight})";
+  overflow-y: auto;
 }
 
 .explore-tile {
   .full-height-width();
   background: @tileBackground;
-  overflow-y: auto;
 }
 
 .proposal-tile {
@@ -344,7 +344,7 @@ body {
 }
 
 // The Target Position layout
-@target-grid-responsive-cutoff: 1350px;
+@target-grid-responsive-cutoff: 992px;
 
 .target-grid {
   display: grid;
@@ -371,8 +371,16 @@ body {
   }
 }
 
+.coordinates-form {
+  margin-bottom: 1em;
+}
+
+.catalogue-form {
+  height: max-content; // keep it from expanding to fill cell
+}
+
 // for the ra and dec inputs and the flex container
 // that holds them
 .target-ra-dec-min-width {
-  min-width: 7em;
+  min-width: 3em;
 }
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -71,10 +71,6 @@ body {
   box-shadow @defaultDuration @defaultEasing,
 }
 
-// Form properties
-.gpp-form {
-}
-
 .explore-tile-title {
   border-bottom-style: solid;
   border-bottom-width: 1px;
@@ -88,6 +84,7 @@ body {
 .explore-tile {
   .full-height-width();
   background: @tileBackground;
+  overflow-y: auto;
 }
 
 .proposal-tile {
@@ -266,9 +263,6 @@ body {
   padding-bottom: 0px;
 }
 
-.aladin-column {
-}
-
 .aladin-container-column {
   height: 100%;
   display: grid;
@@ -282,9 +276,6 @@ body {
 .aladin-container-body {
   grid-area: aladin;
   margin-bottom: 0.2em;
-}
-
-.aladin-container-status {
 }
 
 .aladin-status-fov {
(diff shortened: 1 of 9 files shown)


common/src/main/resources/less/style.less
```diff
@@ -350,4 +341,37 @@ body {
 
 .wip:hover .wip-warning {
   visibility: visible;
-}
\ No newline at end of file
+}
+
+// The Target Position layout
+@target-grid-responsive-cutoff: 1350px;
```
cquiroz, line 347: Where does the cutoff number come from.

SUIR used to have these:

```javascript
Responsive.onlyMobile = { minWidth: 320, maxWidth: 767 }
Responsive.onlyTablet = { minWidth: 768, maxWidth: 991 }
Responsive.onlyComputer = { minWidth: 992 }
Responsive.onlyLargeScreen = { minWidth: 1200, maxWidth: 1919 }
Responsive.onlyWidescreen = { minWidth: 1920 }
```
Bootstrap uses:

```css
// Small devices (landscape phones, 576px and up)
@media (min-width: 576px) { ... }

// Medium devices (tablets, 768px and up)
@media (min-width: 768px) { ... }

// Large devices (desktops, 992px and up)
@media (min-width: 992px) { ... }

// Extra large devices (large desktops, 1200px and up)
@media (min-width: 1200px) { ... }
```

common/src/main/resources/less/style.less
```diff
@@ -88,6 +84,7 @@ body {
 .explore-tile {
   .full-height-width();
   background: @tileBackground;
+  overflow-y: auto;
```
cquiroz, line 87: After testing I think this should go to `.explore-tile-body`

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
```diff
@@ -79,26 +83,28 @@ object CoordinatesForm {
...
-              label = "Dec"
+              label = "Dec",
+              clazz = ExploreStyles.Grow(1) |+| ExploreStyles.TargetRaDecMinWidth
             ),
-            FormButton(width = Two,
-                       size = Small,
-                       icon = true,
-                       label = "X",
-                       onClick = props.goToRaDec(Coordinates(state.raValue, state.decValue))
+            FormButton(
```
cquiroz, line 103: I think you had a way to remove the label keeping the alignment correct

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
```diff
@@ -100,33 +99,28 @@ object TargetBody extends ModelOptics {
...
             searchAndGo(modify.andThen(_.runInCB))
 
-          Grid(columns = Three,
-               clazz = ExploreStyles.FullHeightWidth,
-               stretched = true,
-               padded = GridPadded.Horizontally
-          )(
-            GridRow(stretched = true)(
-              GridColumn(stretched = true, computer = Four)(
+          <.div(
```
cquiroz, line 102: Can this be converted to `ReactFragment`?

cquiroz (commented):
This is looking great, I tested it only in FF and it would be interesting to check it on other
browsers and platforms.

The scrollbar seems it is reaching all the way into the tile title which should not include it I think

common/src/main/resources/less/style.less
toddburnside: Since this was specific to the targets grid, I just determined the width of the viewport when the form fields seemed to be getting too small and chose that number.

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
toddburnside: I can hide the label. Even if hidden, it seems like we should have a better label than `X`. Do you have a suggestion?

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
toddburnside: Sounds like it. I'll look into it.

common/src/main/resources/less/style.less
toddburnside: I originally had it on `.explore-tile-body`. For some reason, the scrolling body still extended about one title bar height below the bottom of the tile. I'll see if I can figure out what bit of CSS was causing that.

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
cquiroz: Yeah X is bad :D. it maybe `GoTo` but still hidden

common/src/main/resources/less/style.less
cquiroz: Today testing on my large screen I noted the grid stacking vertically even for a wide enough screen. Perhaps we could use the 992 cutoff

toddburnside: When I switch to 992, the RA/DEC inputs seem a bit small:
image
But, if I reduce the padding that might be OK. 

Are you running `exploreWDS`, or `targetEditorWDS` when you test? A major shortcoming of CSS responsiveness is that media queries only query the entire viewport size, so that will make a big difference.

cquiroz (approved):
This is great, thanks

PR #344, opened 2020-10-07
Search targets with simbad

cquiroz:
This PR changes the way we search targets issuing direct `simbad` searches using `lucuma-catalog` rather than through `aladin`
While aladin does also `simbad` searches it only retrieves coordinates while `lucuma-catalog` can give more info

This PR includes work on UI improvements including:
* Added a waiting indicator:
waiting indicator

* Show when nothing is found in the catalog
not found

* Ability to do searches with just the keyboard


targeteditor/src/main/scala/explore/targeteditor/SimbadSearch.scala
```diff
@@ -0,0 +1,46 @@
...
+  def search(term: NonEmptyString)(implicit cs: ContextShift[IO]): IO[Option[Target]] = {
+    val backend  = FetchBackend()
+    def response = basicRequest
+      .post(
+        uri"https://simbad.u-strasbg.fr/simbad/sim-id?Ident=${term}&output.format=VOTable"
+      )
+      .readTimeout(5.seconds)
+      .send(backend)
+
+    IO.fromFuture(IO(response))
```
rpiaggio, line 25: This is very very minor, but I think it would be clearer to me if this was `val response = IO(.....send(backend))` and then `IO.fromFuture(response)`.

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
```diff
@@ -38,7 +37,9 @@ final case class TargetBody(
 }
 
 object TargetBody extends ModelOptics {
-  type Props = TargetBody
+  type Props          = TargetBody
+  // search term, after, on empty, on error
+  type SearchCallback = (NonEmptyString, Callback, Callback, Throwable => Callback) => Callback
```
rpiaggio, line 42: Can we use a case class for this?

rpiaggio (approved):
Nice, LGTM!

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
cquiroz: Yes, it maybe better to do that and get better names for the params

targeteditor/src/main/scala/explore/targeteditor/SimbadSearch.scala
toddburnside: > This is very very minor, but I think it would be clearer to me if this was `val response = IO(.....send(backend))` and then `IO.fromFuture(response)`.

I think that would also be more correct. Because Futures evaluate eagerly, you're supposed to wrap the creation of the Future in IO to suspend it's creation. I think that's why `IO.fromFuture` takes an `IO[Future]` instead of just a `Future`, isn't it?

PR #345, opened 2020-10-07
Add usermap clubhouse<->github

cquiroz:
The clubhouse action is assigning all stories to `cquiroz`. This should be solved with the user map below but it will take some testing

### DIFF
--- a/.github/workflows/clubhouse.yml
+++ b/.github/workflows/clubhouse.yml
@@ -14,3 +14,10 @@ jobs:
           opened-state-name: In Development
           merged-state-name: Completed
           ignored-users: geminihlswadmin
+          user-map: |
+            {
+              "rpiaggio": "5f4a944d-faa8-48d6-8611-f565df2715d5",
+              "swalker": "5dc4137c-a307-4be9-b530-4516dc2bf0bf",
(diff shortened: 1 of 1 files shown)


toddburnside:
Thanks for doing this. I've been thinking of it, but never got around to it...

toddburnside (approved):

.github/workflows/clubhouse.yml
```diff
@@ -14,3 +14,10 @@ jobs:
           opened-state-name: In Development
           merged-state-name: Completed
           ignored-users: geminihlswadmin
+          user-map: |
```
rpiaggio, line 17: cquiroz is the default?

rpiaggio (approved):
LGTM

.github/workflows/clubhouse.yml
cquiroz: That's the format suggested in the action documentation
https://github.com/singingwolfboy/create-linked-clubhouse-story#user-map

cquiroz: We are using a token created by me. I think that's the reason all tickets endup under my name

PR #355, opened 2020-10-12
Add UI for setting time allocations to partners

toddburnside:
Any input on handling the PartnerSplitsEditor Modal is welcome. One thing I was not able to do is get the modal to cancel on Escape.

The relevant section in the proposal editor looks like this:
image
And the modal splits editor looks like this:
image


cquiroz:
I'm still not sure whether we should just let do editing in place.

Should we allow the sum of percentages not to add to 100% until we do the full proposal validation

Maybe you could send the screenshots to the CH issue and discuss it a bit with science

toddburnside:
Carlos and I discussed it, and decided to declutter the UI to make it more like the mockup. It now looks like:
image

common/src/main/resources/less/style.less
```diff
@@ -409,3 +409,48 @@ body {
...
 }
 
+
+.partner-split-data {
+  width: 6.5em;
+}
+
+.explore-text-in-form {
+  margin-top: auto;
+  margin-bottom: 3px;
```
cquiroz, line 419: I wonder if we should make all margins be defined in `em`. I don't think there is a hard rule though.

common/src/main/scala/explore/components/ui/PartnerFlags.scala
```diff
@@ -0,0 +1,48 @@
...
+
+package explore.components.ui
+
+import lucuma.core.model.Partner
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSImport
+
+object PartnerFlags {
+  @js.native
+  @JSImport("resources/images/flags/argentina-flag-icon-32.png", JSImport.Default)
```
cquiroz, line 12: I think this will be fine but we should check after we do full opt if there are any issues

cquiroz: I wonder if we should consider using WebP, there maybe some backwards compatibility issues but that could be solved using `picture`
https://www.html5rocks.com/en/tutorials/responsive/picture-element/#toc-file-type

proposal/src/main/scala/explore/proposal/PartnerSplitsEditor.scala
```diff
@@ -0,0 +1,135 @@
...
+    ModalActions(
+      <.div(
+        ExploreStyles.FlexContainer,
+        <.div("Total must be 100.",
+              ExploreStyles.PartnerSplitsEditorError,
+              ExploreStyles.FlexGrow(1)
+        )
+          .unless(addsUpTo100(p)),
+        Button(onClick = p.closeMe)(^.tpe := "button")("Cancel", ExploreStyles.FlexEnd),
+        Button(^.tpe := "submit")("OK", ^.disabled := !addsUpTo100(p))
```
cquiroz, line 45: Nice, we won't allow submission unless we add up to 100

model/src/main/scala/explore/model/ProposalDetails.scala
```diff
@@ -5,20 +5,27 @@ package explore.model
...
+  toOActivation:  ToOActivation,
+  keywords:       Set[Keyword],
+  abstrakt:       String,
+  partnerSplits:  List[PartnerSplit],
+  band1And2Hours: Double,
+  band3Hours:     Double
 )
 
+@Lenses
+final case class PartnerSplit(partner: Partner, percent: Int)
```
cquiroz, line 27: Can we have decimals in percentages?

proposal/src/main/scala/explore/proposal/PartnerSplitsEditor.scala
```diff
@@ -0,0 +1,135 @@
...
+        }
+      }
+      .fold(Callback.empty)(p.updateSplits)
+  }
+
+  private def makeTableRows(p: Props): TagMod =
+    p.splits.zipWithIndex.toTagMod { case (ps, idx) =>
+      val id = s"${ps.partner.tag}-percent"
+      React.Fragment(
+        <.tr(
```
cquiroz, line 71: Why are we using tables for layout that rather than e.g. grid layout?

proposal/src/main/scala/explore/proposal/PartnerSplitsEditor.scala
```diff
@@ -0,0 +1,135 @@
...
+          ),
+          toolbar(p)
+        )
+      )
+    )
+  }
+
+  val component = ScalaComponent
+    .builder[Props]
+    // .renderBackend[Backend]
```
cquiroz, line 131: Unused code

proposal/src/main/scala/explore/proposal/ProposalDetailsEditor.scala
```diff
@@ -26,11 +32,90 @@ final case class ProposalDetailsEditor(proposalDetails: View[ProposalDetails])
...
+  // temporary until abbreviation is in Partner
+  class Backend($ : BackendScope[Props, State]) {
+
+    def render(props: Props, state: State) = {
+      val details        = props.proposalDetails
+      val band1And2Hours = details.zoom(ProposalDetails.band1And2Hours).get
+      val band3Hours     = details.zoom(ProposalDetails.band3Hours).get
+
+      def updateStateSplits(splits: List[PartnerSplit]): Callback =
+        $.modState(State.splits.set(splits))
```
cquiroz, line 99: not very important, here you can also use `$.modStateL(State.splits)(splits)`. It doesn't make much difference here but it is nicer for nested state items

cquiroz (changes requested):
It's looking great. I wonder if the modal would be clearer if we show the percentage signs either in the text fields or next to it

Added some smallish comments inline

common/src/main/resources/less/style.less
toddburnside: I got this number empirically in the dev tools. But, you're right, this should be in `form.overrides` and use the appropriate less variables to make this more generally usable, for example in forms of different "sizes".

common/src/main/scala/explore/components/ui/PartnerFlags.scala
toddburnside: I'll take a look at that.

proposal/src/main/scala/explore/proposal/PartnerSplitsEditor.scala
toddburnside: Because this seems like a semantically appropriate use for the `<table>` element. There is a header, rows of semantically identical data, and a footer. 

The problem with `table` was that people were using it in non-semantic ways for layout because things like CSS flex and grid did not exist. But, this seems like a proper use to me.

proposal/src/main/scala/explore/proposal/ProposalDetailsEditor.scala
toddburnside: That's a good thing to know about. So many little hidden bits to find. :) In this case, it seems I need `setStateL`, though.

toddburnside (commented):
> It's looking great. I wonder if the modal would be clearer if we show the percentage signs either in the text fields or next to it

I thought about trying to do that, but decided that the fact that the heading was `Percent` was pretty clear. I can add it if you'd like.

model/src/main/scala/explore/model/ProposalDetails.scala
```diff
@@ -5,20 +5,27 @@ package explore.model
...
+  pi:             StandardUser,
+  proposalClass:  ProposalClass,
+  category:       TacCategory,
+  toOActivation:  ToOActivation,
+  keywords:       Set[Keyword],
+  abstrakt:       String,
+  partnerSplits:  List[PartnerSplit],
+  band1And2Hours: Double,
+  band3Hours:     Double
 )
```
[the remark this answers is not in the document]
cquiroz: I was going to ask exactly that but also thought it could be overkill. I guess it is up to you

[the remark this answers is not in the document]
toddburnside: I did this to see what it would be like, and I think it's OK. You just end up with some `field.value` and `field.value.value` some places in your code. I'll put in a commit with it after I finish up the current UX stuff I'm working on.

One question I did have is how using coulomb/refined will interact with graphQL. I'll start a conversation in slack.

toddburnside:
It now looks like:
image
as long as the screen is wide enough.

model/src/main/scala/explore/model/ProposalDetails.scala
```diff
@@ -4,21 +4,41 @@
...
+  toOActivation:  ToOActivation,
+  keywords:       Set[Keyword],
+  abstrakt:       String,
+  partnerSplits:  List[PartnerSplit],
+  band1And2Hours: ProposalDetails.NonNegHour,
+  band3Hours:     ProposalDetails.NonNegHour
 )
 
 object ProposalDetails {
+  type NonNegHour = Quantity[NonNegDouble, Hour]
```
toddburnside, line 33: If/when this data is added to lucuma-common, we'll need to find a home for this type and the ones in PartnerSplit. I'm not sure where the best place is until then.

proposal/src/main/scala/explore/proposal/PartnerSplitsEditor.scala
rpiaggio: Sorry I'm commenting so late. It's definitely a table. Could we use the SUI `Table` though? I'd like to SUI widgets whenever possible to ensure consistent CSS is applied throughout, and also to keep a uniform coding style.

rpiaggio:
It's looking great, it seems you really got the hang of React.

proposal/src/main/scala/explore/proposal/PartnerSplitsEditor.scala
toddburnside: semantic-ui-react has a component for the SUI table, but there is no facade for it in scalajs-reacte-semantic-ui. Since it is just adding classes, do we need to create a facade for it? I don't think there is any JS functionality in the Table.

rpiaggio: Yes, we should add facades for `Table` and related components. I had not realized we had't done it yet, but we should.

Even if there's no behavior added, using the SUIR components allows us to define the UI at the semantic level instead of at the HTML level. For example, we can use `basic`, `color`, `stackable`, `striped`, etc, and the behavior of these properties will be coherent with the overall theme for the app.

toddburnside: Yeah, I came to the same conclusion myself. I just added the classes to the table/row elements to test it out. I'll probably push that in a bit. Then, I'll add the facade and change it when it's available. I think I'll switch to working on the masked input stuff for Carlos first, though.

rpiaggio: Sounds good :+1:

rpiaggio (approved):
LGTM

PR #367, opened 2020-10-17
Merge slate and custom theme

cquiroz:
in explore the SUI theme is built as part of the build and it has three set of configurations:

1. It start taking configurations from the `default` theme coming directly from `FomanticUI`. `default` theme should be updated every time `FomanticUI` is updated
1. Then we use `slate` which is an old dark theme for `Semantic UI` which overrides some of the `default` configurations. I don't expect `slate` to get updates in the future.
1. Finally we have custom overrides which we could call the `explore` theme.

This is getting complicated to manage so I'm proposing to eliminate step 2 altogether. This PR moves all `slate` setting into the `explore` theme. `default` should remain clean and only be updated when `FomanticUI` changes.

For some reason this is failing in CI with the default settings and I had to increase the memory settings for `fullOptJS`


cquiroz:
Given the increase in memory usage in node it is no surprise that test apps cannot be built with this change

cquiroz:
I removed the experiments and I'll check that is another PR.
This is ready for review

toddburnside (approved):
LGTM

common/src/main/webpack/prod.webpack.config.js
```diff
@@ -11,7 +11,7 @@ const ci = process.env.CI === 'true'; // When on CI don't add hashes
 
 const Web = Merge(
   ScalaJSConfig,
-  // parts.noNode, // We will have to use this in Scala 1.x, but doesn't work with 0.6
+  parts.noNode, // We will have to use this in Scala 1.x, but doesn't work with 0.6
```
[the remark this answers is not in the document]
cquiroz: Right, I'll remove it

common/src/main/webpack/prod.webpack.config.js
```diff
@@ -52,6 +52,9 @@ const Web = Merge(
         persistentCache: false,
       }),
     ],
+    node: {
+      fs: "empty"
+    }
```
rpiaggio, line 57: Is this required by the new CSS plugin?

rpiaggio (approved):
LGTM

common/src/main/webpack/prod.webpack.config.js
cquiroz: Not really. I'm trying to remove a warning about the fsevents dependency while trying to fix the test app issue

cquiroz:
Ok I think I tracked the reason for the extra memory usage to a specific version of the less processor we are using.
With that change I was able to produce a review app

PR #373, opened 2020-10-22
Split some parts of the aladin container for performance reasons

cquiroz:
I noted there was high cpu usage while moving the cursor over the aladin control.
We are showing the cursor coordinates and that produces too many updates.
This PR changes the structure to do less computation and also reduces the frequency of mouse events.

### DIFF
--- a/targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
+++ b/targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
@@ -3,14 +3,12 @@
 
 package explore.targeteditor
 
-import scala.math.rint
+import scala.concurrent.duration._
 
 import cats.syntax.all._
 import crystal.react.implicits._
-import explore.Icons
 import explore.View
 import explore.components.ui.ExploreStyles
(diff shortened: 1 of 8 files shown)


targeteditor/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -217,7 +169,13 @@ object AladinContainer {
       v.onFullScreenToggle(recalculateView) *> // re render on screen toggle
         v.onZoom(onZoom(v)) *>                 // re render on zoom
         v.onPositionChanged(onPositionChanged(v)) *>
-        v.onMouseMove(s => $.setStateL(State.current)(Coordinates(s.ra, s.dec)))
+        v.onMouseMove(s =>
+          $.props.flatMap(
+            _.updateMouseCoordinates(Coordinates(s.ra, s.dec))
+              .debounce(100.millis)
```
rpiaggio, line 175: Nice

rpiaggio (approved):
LGTM

PR #374, opened 2020-10-22
Save name of the target on blur

cquiroz:
This PR lets the user save the name of a target independent of searching it via simbad.
It will not allow empty names though.

This only saves when focus is lost. We may want to investigate saving more often with a timer
but I'll wait for the new db libs to be in place

### DIFF
--- a/targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
+++ b/targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
@@ -10,6 +10,7 @@ import crystal.react.implicits._
 import eu.timepit.refined._
 import eu.timepit.refined.auto._
 import eu.timepit.refined.collection._
+import eu.timepit.refined.types.string.NonEmptyString
 import explore.AppCtx
 import explore.Icons
 import explore.components.ui.ExploreStyles
@@ -31,13 +32,13 @@ import react.common.implicits._
 import react.semanticui.collections.form._
 import react.semanticui.elements.icon.Icon
 import react.semanticui.elements.label._
-import react.semanticui.modules.dropdown.DropdownItem
 import react.semanticui.sizes._
 
 final case class CoordinatesForm(
   target:           TargetResult,
   searchAndGo:      SearchCallback => Callback,
-  goToRaDec:        Coordinates => Callback
+  goToRaDec:        Coordinates => Callback,
+  onNameChange:     NonEmptyString => Callback
 )(implicit val ctx: AppContextIO)
     extends ReactProps[CoordinatesForm](CoordinatesForm.component) {
   def submit(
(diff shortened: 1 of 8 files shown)


targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
```diff
@@ -120,6 +123,9 @@ object CoordinatesForm {
             label = nameLabel,
             focus = true,
             loading = state.searching,
+            error = state.searchTerm.isEmpty,
```
rpiaggio, line 126: `|| state.searchError.nonEmpty` ?

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
```diff
@@ -35,11 +35,13 @@ import react.semanticui.elements.icon.Icon
...
 import react.semanticui.modules.dropdown.DropdownItem
 import react.semanticui.sizes._
+import eu.timepit.refined.types.string.NonEmptyString
 
 final case class CoordinatesForm(
   target:           SiderealTarget,
   searchAndGo:      SearchCallback => Callback,
-  goToRaDec:        Coordinates => Callback
+  goToRaDec:        Coordinates => Callback,
+  nameChange:       NonEmptyString => Callback
```
rpiaggio, line 44: Maybe we can call this parameter `onNameChange` ?

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
```diff
@@ -85,6 +85,15 @@ object TargetBody extends ModelOptics {
               )
             }
           ) _
+          val modifyName = undoSet[NonEmptyString](
+            SiderealTarget.name,
+            n =>
+              TargetsSetInput(
+                name = n.value.some,
+                ra = none,
+                dec = none
```
rpiaggio, line 94: I think these values default to `None` for cases like these where we don't want to change them.

rpiaggio (approved):
LGTM

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
cquiroz: Actually I only want to show the reddish color when the search term is empty. In case of search errors they will be displayed next to the label

targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
```diff
@@ -73,6 +76,14 @@ object TargetQueries {
   val targetRA: Lens[TargetResult, RightAscension] =
     TargetResult.tracking ^|-> properMotionRA
 
+  /**
+   * Lens to the name of a sidereal target
+   */
+  val unsafeTargetName: Lens[TargetResult, NonEmptyString] =
```
cquiroz, line 82: This is unsafe lens to be used in undo, It will be protected of misuse by the ui while the api layer is updated to support refined scalars.

cquiroz:
Rebased to work with the new api. It should be ready for review

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
rpiaggio: 👍

rpiaggio:
Still LGTM 👍

PR #383, opened 2020-10-23
Modify action to opt-in to creating a new clubhouse story.

toddburnside:
I believe this change should only run the `create-linked-clubhouse-story` action if the body of the PR (which I believe is what I am writing now) contains the token `[ch-new]`, or the PR is given a `new clubhouse story` label. However, I don't know how to test this other than trying it. When I create this PR, I think it will use the new yaml, so that will be one test - no story should be created.

The label is a bit long but descriptive. We could name it anything, and add it to the labels for the repo.

Note that the action is only run on PR open or close, so we can't add a new story later. The story would have to be created in clubhouse and assigned via one of the clubhouse integration methods.

### DIFF
--- a/.github/workflows/clubhouse.yml
+++ b/.github/workflows/clubhouse.yml
@@ -7,6 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: singingwolfboy/create-linked-clubhouse-story@v1.5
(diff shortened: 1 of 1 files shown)


cquiroz (approved):
Thanks, let's test it

toddburnside:
> Thanks, let's test it

It appears my yaml is invalid. I'll need to fix it first. :)

PR #393, opened 2020-10-24
Fix opt-in check in clubhouse action

toddburnside:
This should create a new story [ch-new].

### DIFF
--- a/.github/workflows/clubhouse.yml
+++ b/.github/workflows/clubhouse.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: singingwolfboy/create-linked-clubhouse-story@v1.5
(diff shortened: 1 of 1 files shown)


rpiaggio:
Nice! Thank you!

Is it possible to have a trigger when [ch-new] is in a comment, for those cases where we forget to add it when the PR is created?

toddburnside:
> Is it possible to have a trigger when [ch-new] is in a comment, for those cases where we forget to add it when the PR is created?

 Unfortunately, that isn't possible without some changes to the action itself. It is designed to only run on PR creation and close. We **could** fork the action and make one better suited to us. Without that, we'll need to create the story in clubhouse and associate it via `[ch-####]`.

I have also put in a ticket with clubhouse for them to add this sort of thing as an integration feature, but I don't expect it will happen soon, if ever.

PR #397, opened 2020-10-27
Theme improvements

cquiroz:
This PR is the result of some experimentation on css variables.

The core idea is to replace some of the less variables that define colors by css-variables. Using variables would allow to dynamically change the theme.

There are many details on how this is organized, for more details check `THEMING.md`

Besides using css-variables I made some changes to theme after looking at the clubhouse ui and reading more about selecting color palettes for dark modes.

image

Note this is still a work in progress, e.g. there is some pending improvement around buttons and dropdown controls pending, but I'd rather get the ball rolling.

Another note, I'm hiding some parts off the UI that are not included in M1 like the elevation plot and the visualization controls.


toddburnside:
Very cool. Where are the values actually assigned to all of the variables?

cquiroz:
@toddburnside They are set in `styles.less` at the body level. Note they are linked to a class for `dark/light` theme which lets you flip themes with just changing the class at the top level

toddburnside:
> @toddburnside They are set in `styles.less` at the body level. Note they are linked to a class for `dark/light` theme which lets you flip themes with just changing the class at the top level

Doh. I didn't notice that it wasn't displaying the diff for `styles.less` because of the size of the diff. Thanks.

cquiroz:
The size of style file is a bit of an issue, using variables gets more verbose indeed

toddburnside:
We should probably break `styles.less` into some smaller files. The variables could go into one and be imported. We could probably also break up some of the other styles by area of the UI. I'm not saying it needs to be done in the PR, but at some point.

rpiaggio:
Very nice!

Is there an advantage for computing colors in CSS rather than using LESS for that?

cquiroz:
It gives you bragging rights for one.
It also lets you more easily derive colors from the existing themes, having said that, you could use less to derive the colors on dark/light

toddburnside:
> Is there an advantage for computing colors in CSS rather than using LESS for that?
Street cred, bro. 🤘

rpiaggio:
> It gives you bragging rights for one.


🤣 🤣 🤣 

It's certainly no small feat, but for the sake of readability/maintenance I was thinking we could go from this:
``` css
  --color-background-base-h: 0;
  --color-background-base-s: 0%;
  --color-background-base-l: 7.06%;
  --color-background-base-hsl: var(--color-background-base-h), var(--color-background-base-s),
    var(--color-background-base-l);
  --color-background: hsl(var(--color-background-base-hsl));

  --color-background-light-10: hsl(
    var(--color-background-base-h),
    var(--color-background-base-s),
    calc(var(--color-background-base-l) + 10%)
```

to this:

``` css
  @color-background-base: #121212;
  --color-background: @color-background-base;
  --color-background-light-10: lighten(@color-background-base, 10%);
```


> It also lets you more easily derive colors from the existing themes

How would that work?

cquiroz:
In the theme there are many instances of e.g.
```
@hoverItemBackground: lighten(@pageBackground, 10%);
```
There is no `lighten` in css yet (There is an spec on the works) but you can do the calculation in
pure css as shown in this PR.

This removes the need of less, though arguably we will continue using it in the foreseeable future.
In that sense we could do what @rpiaggio suggest and simplify the code above, though after learning about hsl doing the calculation seems fine to me, especially for dark themes.

Anyway I could switch to the suggested scheme, either in this PR or in future updates

rpiaggio (approved):
LGTM

Maybe in the future we can split the definitions from the computations. For the moment,  we can see how it grows as we move forward. It's definitely looking great.

toddburnside (approved):
LGTM

PR #401, opened 2020-10-29
Update queries to read/write proper motion

cquiroz:
This PR updates the graphql queries to include all `SiderealTracking` fields and also supports updating them.
When a simbad query is successful all fields will be populated.

This is also a chance to learn about the GraphQL backend and frontend implementations. It seems great, congrats @rpiaggio @swalker

### DIFF
--- a/common/src/main/scala/explore/utils/package.scala
+++ b/common/src/main/scala/explore/utils/package.scala
@@ -3,9 +3,23 @@
 
 package explore
 
+import cats.data.State
+import monocle.Lens
+import monocle.state.all._
+
 package object utils {
 
   def abbreviate(s: String, maxLength: Int): String =
     if (s.length > maxLength) s"${s.substring(0, maxLength)}\u2026" else s
 
+  implicit final class LensEditorOps[S, A](self: Lens[S, A]) {
+    def edit(a: A): State[S, A]         =
+      self.assign(a)
+    def :=(a:   A): State[S, A]         =
+      edit(a)
+    def edit(a: Option[A]): State[S, A] =
+      a.fold(self.st)(self.assign)
+    def :=(a:   Option[A]): State[S, A] =
+      edit(a)
+  }
 }
--- a/targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
+++ b/targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
@@ -3,7 +3,10 @@
 
 package explore.target
 
+import cats.Eval
+import cats.data.IndexedStateT
 import cats.effect.IO
+import cats.implicits._
 import clue.GraphQLOperation
 import clue.macros.GraphQL
 import eu.timepit.refined.collection.NonEmpty
@@ -14,9 +17,10 @@ import explore.GraphQLSchemas._
 import explore.implicits._
 import explore.model.decoders._
 import explore.undo.Undoer
+import explore.utils._
 import lucuma.core.math.Coordinates
-import lucuma.core.math.Declination
-import lucuma.core.math.RightAscension
+import lucuma.core.math.Epoch
+import lucuma.core.model.SiderealTracking
 import lucuma.core.model.Target
 import lucuma.ui.reusability._
 import monocle.Lens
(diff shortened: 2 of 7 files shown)


common/src/main/scala/explore/utils/package.scala
```diff
@@ -3,9 +3,23 @@
...
+  implicit final class LensEditorOps[S, A](self: Lens[S, A]) {
+    def edit(a: A): State[S, A]         =
+      self.assign(a)
+    def :=(a:   A): State[S, A]         =
+      edit(a)
+    def edit(a: Option[A]): State[S, A] =
+      a.fold(self.st)(self.assign)
+    def :=(a:   Option[A]): State[S, A] =
+      edit(a)
+  }
```
swalker2m, line 24: With the non-`Option` versions included, maybe it would make sense to share these in `core` syntax.

targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
```diff
@@ -145,4 +144,35 @@ object TargetQueries {
       )(value)
   }
 
+  /**
+   * Updates all the fields of sideral tracking
+   */
+  def updateSiderealTracking(
+    t: SiderealTracking
+  ): IndexedStateT[Eval, EditSiderealInput, EditSiderealInput, Unit] = {
```
swalker2m, line 152: Why not `State[EditSiderealInput, Unit]`?

swalker2m (commented):
Looks great to me, but I best leave it to the Explore experts to approve.

common/src/main/scala/explore/utils/package.scala
rpiaggio: Also, maybe we want the methods `@inline`, at least the aliases (`:=`).

cquiroz: BTW, this was copied from @swalker implementation. I just added an extra or non `Option`. We should move it to core.

targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
cquiroz: Right, I should change that

rpiaggio (approved):
Very nice. LGTM

targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
cquiroz: A downside of this method is that we change every field. This is fine for the response of a simbad query.
But if we edit a single field, should we only set that field on the save?

rpiaggio: If a single field is edited, it'd be nice to only set that field because then the undoer will weave that change with other possible remote changes. In other words: it will keep remote changes to other fields that can happen between the set and the undo. Otherwise, the undo will override those remote changes and take the whole target to the state it was before the initial local editing.

swalker2m: That's a good point, but I'm inclined to think it doesn't make too much of a difference in practice.  It's more a convenience for people using the API than anything that the non-updated fields can be empty.

swalker2m: > I'm inclined to think it doesn't make too much of a difference in practice

I was referring there to the backend and not taking into account the impact on undo that @rpiaggio mentions.  I see what you're saying but now I'm trying to decide if, from the user perspective, it is actually good that the entire target changes as a unit.  At least you would get a consistent collection of values?

cquiroz: I think it is in principle ok to send everything if only one field change, though I have to think about undo.
I was a bit more worried about the extra traffic and db activity on updating fields that don't change.

rpiaggio: I think we should send everything after a catalog query. After all, values could have changed remotely on the server since we last read them, and in this case we do want to override everything with the values from the catalog.

PR #416, opened 2020-11-02
Fix duplicate queries + tools for tracing

rpiaggio:


common/src/main/scala/explore/components/graphql/LiveQueryRenderMod.scala
```diff
@@ -89,67 +86,6 @@ object LiveQueryRenderMod {
             State.apply
           )
       )
-      .componentDidMount { $ =>
```
rpiaggio, line 92: All this was abstracted away to `Render` in the last refactor, but I forgot to remove it from here. This was causing duplicate queries.

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
```diff
@@ -54,14 +52,10 @@ object TargetBody {
 
   implicit val propsReuse = Reusability.derive[Props]
 
-  class Backend($ : BackendScope[Props, Unit]) {
+  class Backend {
     // Create a mutable reference
     private val aladinRef = Ref.toScalaComponent(AladinRef)
```
cquiroz, line 60: Odd we don't need this anymore

cquiroz (approved):
LGTM

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
rpiaggio: It was only being used in `setName`, which isn't used anymore.

PR #418, opened 2020-11-04
UI to edit the proper motion details

cquiroz:
This is a largish PR that lets you read and edit the Proper Motion fields.
One feature not supported yet in the backend is to unset values though the UI allows it

image

### DIFF
--- a/model/src/main/scala/explore/optics/package.scala
+++ b/model/src/main/scala/explore/optics/package.scala
@@ -4,6 +4,13 @@
 package explore
 
 import cats.syntax.all._
+import coulomb._
+import lucuma.core.math.ApparentRadialVelocity
+import lucuma.core.math.Constants._
+import lucuma.core.math.ProperVelocity
+import lucuma.core.math.RadialVelocity
+import lucuma.core.math.Redshift
+import lucuma.core.math.units._
 import monocle._
 import monocle.std.option.some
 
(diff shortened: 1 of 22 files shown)


targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
```diff
@@ -86,11 +103,61 @@ object TargetBody {
                 update.runS(input).value
             }
           ) _
+
           val modifyName = undoSet[NonEmptyString](
             unsafeTargetName,
             n => EditSiderealInput.name.set(n.value.some)
           ) _
 
+          val modifyEpoch = undoSet[Epoch](
```
cquiroz, line 112: @rpiaggio There is some amount of repetition here. any ideas on how to reduce it?

targeteditor/src/main/scala/explore/targeteditor/RVInput.scala
```diff
@@ -0,0 +1,144 @@
...
+import lucuma.ui.reusability._
+import monocle.macros.Lenses
+import react.common._
+import react.common.implicits._
+import react.semanticui.elements.label.LabelPointing
+
+final case class RVInput(
+  value:           ViewF[IO, Option[RadialVelocity]],
+  modify:          Option[RadialVelocity] => IO[Unit]
+)(implicit val cs: ContextShift[IO])
```
rpiaggio, line 36: There's a mechanism to access the `IO` typeclass instances from the context, so that we don't have to pass them around:

You can do inside the `render` method: `AppCtx.withCtx{ implicit ctx => ...}`. If you have `import explore.implicits._` then you will have implicit values for: `ContextShift[IO]`, `Timer[IO]`, `Logger[IO]`, and the GraphQL clients.

targeteditor/src/main/scala/explore/targeteditor/RVInput.scala
```diff
@@ -0,0 +1,144 @@
...
+        )
+    }
+  }
+
+  implicit def propsReuse: Reusability[Props] = Reusability.by(_.value)
+  implicit def stateReuse: Reusability[State] = Reusability.derive
+
+  class Backend($ : BackendScope[Props, State]) {
+    def render(props: Props, state: State) = {
+      val rvView = ViewF.fromState[IO]($)(Async[IO], props.cs).zoom(State.rvView)
```
rpiaggio, line 82: With the mechanism described above, you don't need to pass the implicits.

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
rpiaggio: Although they follow a similar pattern, I can't think of a way of reusing things: each invocation to `undoSet` needs its own `Lens` and `EditSiderealInput`.

model/src/main/scala/explore/optics/package.scala
```diff
@@ -107,4 +114,36 @@ package object optics {
...
+
+  val unsafePVDecLensO: Lens[Option[ProperVelocity], Option[ProperVelocity.Dec]] =
+    Lens[Option[ProperVelocity], Option[ProperVelocity.Dec]](_.map(ProperVelocity.dec.get))(s =>
+      a => (s, a).mapN(ProperVelocity.dec.set(_)(_))
+    )
+
+  val unsafePVRALensO: Lens[Option[ProperVelocity], Option[ProperVelocity.RA]] =
+    Lens[Option[ProperVelocity], Option[ProperVelocity.RA]](_.map(ProperVelocity.ra.get))(s =>
+      a => (s, a).mapN(ProperVelocity.ra.set(_)(_))
+    )
```
rpiaggio, line 126: If i understand correctly, setting RA or Dec to `None` on a `val pv = Some(properVelocity)`, will set `pv` to `None`. The other coordinate will be lost. Is this the desired behavior?

rpiaggio (approved):
I made a couple of comments but it definitely LGTM. I'm eager to see this working.

toddburnside (approved):
LGTM

model/src/main/scala/explore/optics/package.scala
cquiroz: I think there is a modelling issue here. Do we let PV to have only ra and not dec or viceversa. at the moment we asume we need both but the UI in principle lets you have only one.

The unsafe lens reflects that and would produce undesired results. We may rather remove the wrapper and have each component as a part of `SiderealTracking`.

I'd leave as it is now but we need to revisit this soon

targeteditor/src/main/scala/explore/targeteditor/RVInput.scala
cquiroz: Thanks I'll change that

model/src/main/scala/explore/optics/package.scala
rpiaggio: 👍

PR #419, opened 2020-11-04
Show target magnitudes from DB and catalog.

rpiaggio:
It seems we can't send magnitudes to the API yet, so it's flickering when querying the catalog (the new magnitudes come from the catalog, are shown in the UI, but a second later they are refreshed from the DB):

Kapture 2020-11-04 at 17 07 08

Therefore I'm opening the PR as draft since I don't know if we want to include this in the milestone. Or we could tag it with "WORK IN PROGRESS".

### DIFF
--- a/common/src/main/scala/explore/model/reusability.scala
+++ b/common/src/main/scala/explore/model/reusability.scala
@@ -6,14 +6,11 @@ package explore.model
 import java.time.Duration
 
 import clue.StreamingClientStatus
-import coulomb.Quantity
 import explore.data.KeyedIndexedList
 import io.circe.Json
 import japgolly.scalajs.react.CatsReact._
 import japgolly.scalajs.react.Reusability
 import japgolly.scalajs.react.raw.JsNumber
-import lucuma.core.math.MagnitudeValue
-import lucuma.core.model.Magnitude
 import lucuma.ui.reusability._
 import react.common.Size
 import react.common.implicits._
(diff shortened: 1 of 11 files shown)


common/src/main/scala/explore/model/reusability.scala
```diff
@@ -25,6 +27,8 @@ object reusability {
   // TODO: Move to lucuma-ui.
   implicit def quantityReuse[N: Reusability, U]: Reusability[Quantity[N, U]] =
     Reusability.by(_.value)
+  implicit val magnitudeValueReuse: Reusability[MagnitudeValue]              = Reusability.byEq
```
cquiroz, line 30: I thought we had these on lucuma-ui. If not we could move them there later on

cquiroz (approved):
It would be nice to include it but I agree that the revert to db values seems odd.
I wouldn't mind including it if we can fix that

common/src/main/scala/explore/model/reusability.scala
rpiaggio: I now moved them to lucuma-ui.

rpiaggio:
I hid the magnitudes editor for now and included the fix for when tracking info from the catalog is missing. Also, marked the proposal editor as WIP.

swalker2m:
Sorry guys, I dropped the ball on this by not including magnitudes in the editor input.  I just need to figure out how to indicate when magnitude values should be deleted.  I was planning on rolling it with the unset feature in general.

rpiaggio:
> Sorry guys, I dropped the ball on this by not including magnitudes in the editor input. I just need to figure out how to indicate when magnitude values should be deleted. I was planning on rolling it with the unset feature in general.

No worries.

I'm not sure that "resetting" a list is the same as "unsetting" it but I will continue the discussion in https://github.com/gemini-hlsw/lucuma-tmp-api/pull/62.

PR #425, opened 2020-11-05
Disable coordinates control and go to button

cquiroz:
Should we make the ra/dec fields read only for the current milestone

image


rpiaggio (approved):
LGTM

PR #429, opened 2020-11-06
Improvements to the search ui

cquiroz:
This adds a few small improvements to the search UI:

* A bug fix where you needed to click twice on the search icon
* Support doing a search when pressing enter
* Disable autofill which seems to create problems for some users
* Disable the text field while searching

### DIFF
--- a/targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
+++ b/targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
@@ -124,6 +124,8 @@ object CoordinatesForm {
             focus = true,
             loading = state.searching,
             error = state.searchTerm.isEmpty,
+            onTextChange = (u: String) =>
+              $.setStateL(State.searchTerm)(u) *> $.setStateL(State.searchError)(none),
             onBlur = (u: ValidatedNec[String, String]) =>
               u.toOption
                 .flatMap(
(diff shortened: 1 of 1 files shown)


rpiaggio (approved):
LGTM

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
```diff
@@ -124,22 +124,27 @@ object CoordinatesForm {
...
               u.toOption
                 .flatMap(
                   refineV[NonEmpty](_).toOption
                     .map(props.onNameChange(_))
                 )
                 .getOrEmpty,
-            onValidChange = (_: Boolean) => $.setStateL(State.searchError)(none),
+            disabled = state.searching,
             icon = Icons.Search
               .link(true)
```
toddburnside, line 138: It seems like the `action` in `FormInputEV` is meant for this sort of thing. But, I'm not sure how to use it because the examples don't actually do anything when it is clicked/enter is hit.

rpiaggio: An unrealated comment regarding this line: it seems `State.searchError` is not being set anywhere now. Maybe we can change it to `searchEnabled: Boolean = true` and have here `onValidChange = v => $.setStateL(State.searchEnabled)(v)`.

cquiroz: `searchError` is set in response to a failure querying `aladin` if you look above.

cquiroz: @toddburnside I'll look at `action` it may help to a bug I'm seeing and it is nicer than the hidden submit field

cquiroz: `action` doesn't help much here as it is supposed to contain the url you are submitting too.

toddburnside: Ah. Interesting.

toddburnside (commented):
LGTM

PR #433, opened 2020-11-06
Input improvements

rpiaggio:
* Keep partially input PVs in UI.
* Let FormInputEV call mutations as needed (skips when not changed).


targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
```diff
@@ -41,7 +41,7 @@ final case class CoordinatesForm(
   target:           TargetResult,
   searchAndGo:      SearchCallback => Callback,
   goToRaDec:        Coordinates => Callback,
-  onNameChange:     NonEmptyString => Callback
+  onNameChange:     NonEmptyString => IO[Unit]
```
rpiaggio, line 44: `IO` composes more easily with the setters in `View`, `Undo`, GraphQL clients. Besides, in the caller it was originally an `IO` that had been converted to `Callback` and now I needed to convert it back here to `IO` to compose with `View.set`.

Should we convert the other callbacks to `IO`? In general I find it more ergonomic to work with `IO`s and convert them to `Callback` at the last moment, which sometimes is not even necessary if we end up passing a `View` to a final control (like `FormInputEV`).

targeteditor/src/main/scala/explore/targeteditor/TargetBody.scala
```diff
@@ -167,36 +159,32 @@ object TargetBody {
                         clazz = ExploreStyles.FormSectionLabel |+| ExploreStyles.ColumnSpan(2)
                   ),
                   InputWithUnits(
-                    props.target.zoom(TargetQueries.pvRALens),
+                    props.target.zoom(TargetQueries.pvRALens).withOnMod(modifyProperVelocitRA),
```
rpiaggio, line 162: This makes the mutation part of the `View` setter and I think is the pattern we should use. `FormInputEV` will only call the whole thing if the value changed, whereas `onBlur` is always called.

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
cquiroz: Sound good, it would be nice to work with only one `Effect` but I guess that requires big changes in `scalajs-react`

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
```diff
@@ -59,7 +59,7 @@ object CoordinatesForm {
 
   @Lenses
   final case class State(
-    searchTerm:  String,
+    searchTerm:  NonEmptyString,
```
cquiroz, line 62: I didn't want to use eariler `NonEmptyString` as an empty input is valid though it is useless to do a search. I suppose with this change (validating in `FormInputEV`) we still let an empty string but we can't do a search?

targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
```diff
@@ -106,10 +107,10 @@ object TargetQueries {
    * Lens used to change name and coordinates of a target
    */
   val targetPropsL =
-    Lens[TargetResult, (String, SiderealTracking, List[Magnitude])](t =>
-      (t.name, TargetResult.tracking.get(t), t.magnitudes)
+    Lens[TargetResult, (NonEmptyString, SiderealTracking, List[Magnitude])](t =>
+      (NonEmptyString.from(t.name).getOrElse("<EMPTY>"), TargetResult.tracking.get(t), t.magnitudes)
```
cquiroz, line 111: Maybe this should match the `<New String>` used when creating a new target?

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
rpiaggio: Would it be OK with you to switch to `IO` internally and only convert to `CB` when interfacing with `scalajs-react` (via `.runInCB` ?

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
rpiaggio: Exactly, we use the new `FormInputEV` validation capabilities to trigger the error message. I will add another commit to  disable the search button when input is invalid (string empty) with this new logic.

targeteditor/src/main/scala/explore/targeteditor/TargetQueries.scala
rpiaggio: Good idea. I'll put it in a constant and reference from both places.

targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
cquiroz: Yes that would be fine

PR #442, opened 2020-11-10
Use View setter in RVInput.

rpiaggio:


model/src/main/scala/explore/optics/GetAdjust.scala
```diff
@@ -16,6 +16,6 @@ object GetAdjust {
   def apply[T, A](lens: Lens[T, A]): GetAdjust[T, A] =
     GetAdjust(lens.asGetter, lens.asAdjuster)
 
-  def apply[T, A](get: T => A, set: (A => A) => T => T): GetAdjust[T, A] =
-    GetAdjust(Getter(get), Adjuster(set))
+  def apply[T, A](get: T => A, mod: (A => A) => T => T): GetAdjust[T, A] =
+    GetAdjust(Getter(get), Adjuster(mod))
```
rpiaggio, line 20: This is an unrelated naming fix.

targeteditor/src/main/scala/explore/targeteditor/RVInput.scala
```diff
@@ -87,29 +98,21 @@ object RVInput {
             FormInputEV(
               id = state.rvView.tag,
               label = state.rvView.tag,
-              value = props.value.zoom(unsafeRVtoZLens),
+              value = props.value.zoom(rvToRedshiftGet)(rvToRedshiftMod),
```
rpiaggio, line 101: In cases where we don't have a `Lens` available, it's OK to zoom via get/mod functions.

Also, we allow setting to be done via `InputFormEV` calling the `View`'s `set` rather than doing it manually `onBlur`.

cquiroz: I like this, I was thinking to do it as the lens are unlawful.

cquiroz (approved):
LGTM

toddburnside (approved):
LGTM

PR #443, opened 2020-11-11
Keep searchTerm in sync.

rpiaggio:


targeteditor/src/main/scala/explore/targeteditor/CoordinatesForm.scala
```diff
@@ -129,15 +118,16 @@ object CoordinatesForm {
...
             validFormat = ValidFormatInput.nonEmptyValidFormat,
             label = "Name",
             error = state.searchError.orUndefined,
             loading = props.searching.get,
             disabled = props.searching.get,
             errorClazz = ExploreStyles.InputErrorTooltip,
             errorPointing = LabelPointing.Below,
-            onTextChange = _ => $.setStateL(State.searchError)(none),
+            onTextChange =
+              v => $.setStateL(State.searchTerm)(v) >> $.setStateL(State.searchError)(none),
```
cquiroz, line 130: Right, I was thinking this solution, to keep the search term independent of the props

cquiroz (approved):
LGTM, thanks

rpiaggio:
We should change tracking.ra and tracking.dec to a View mechanism in a future PR, when we enable editing again.

PR #444, opened 2020-11-11
Ch530 make the bottom toolbar static

cquiroz:
This is an UI change to move the targets toolbar to the top, as it is easier to reach than being on the top.
The delete target button has been moved next to the targets, IMHO this more clearly indicates what are we deleting and makes it easier to delete many targets quickly

image

I'm also changing the style of the buttons to get closer to the style in clubhouse, removing the background gradient and adding a small border


explore/src/main/scala/explore/Routing.scala
```diff
@@ -86,14 +86,17 @@ object Routing {
         )
         .onPostRenderP {
           case (prev, next, view)
-              if next.some =!= prev &&
+              if prev.exists(_ =!= next) &&
```
rpiaggio, line 89: If we do `prev.forAll(_ =!= next)` it will also match in case of `prev === None` and we only need one case clause.

cquiroz: 👍

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -176,24 +177,26 @@ object TargetObsList {
...
-    protected def deleteTarget(setter: Undoer.Setter[IO, TargetsWithObs]): Callback =
-      $.props >>= { props =>
-        deleteTargetEnabled(props.focused.get, props.targetsWithObs.get)
-          .map(targetId =>
-            targetMod(setter, props.targetsWithObs, props.focused, targetId)(
-              targetListMod.delete
-            ).runInCB
-          )
-          .getOrEmpty
+    protected def deleteTarget(
```
cquiroz, line 180: @rpiaggio I struggled a bit with this part, it maybe worth a review

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -230,12 +233,27 @@ object TargetObsList {
...
+                  Button(size = Mini, compact = true, onClick = newTarget(undoCtx.setter))(
+                    Icons.New.size(Small).fitted(true)
+                  )
+                ),
+                UndoButtons(props.targetsWithObs.get, undoCtx, size = Mini)
+              ),
+              targets.toTagMod { target =>
+                val targetId     = target.id
+                val currIdx      = targetIds.indexOf(targetId)
+                val targetIdxs   = targets.zipWithIndex
```
rpiaggio, line 254: Maybe this should go outside of the loop, so we only compute it once.

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -230,12 +233,27 @@ object TargetObsList {
...
+                val nextToSelect = targetIdxs.find(_._2 === currIdx + 1).map(_._1)
+                val prevToSelect = targetIdxs.find(_._2 === currIdx - 1).map(_._1)
```
rpiaggio, line 256: We could add functionality to `IndexedCollMod` to make this easier. I'll look into it.

rpiaggio: I wonder if we can use a `Zipper`?

rpiaggio:
Maybe I'm missing something, but is seems to me that:

1) The local model isn't being updated on delete. The target probably doesn't disappear until there's a server roundtrip (correct me if I'm wrong).
2) There's now some duplicated logic with `targetMod`, and there's code now in `setTargetWithIndex` that is never run (the first clause in the `fold`, which takes care of deletions). I wonder if we can weave back the flow to go through `targetMod` (and therefore `setTargetWithIndex`), which will not only not duplicate logic but take care of point (1) too. It seems to me that the only difference is correctly refocusing after a deletion. Maybe we can compute the new focused target inside `setTargetWithIndex` instead of the old logic of just setting it to `none`? This way the next and previous would only computed when needed (since it takes linear time to compute each of them).

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: All this superseded by https://github.com/gemini-hlsw/explore/pull/444#issuecomment-725609876

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: I admit all this original code lacks a lot of comments.

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
cquiroz: How could it go outside the loop considering that we refer to `target`?

cquiroz:
@rpiaggio Thanks for the comments. I did as you suggested passing the next to focus target as a param and it seems to work better.
It is correct that with the previous code, the item was deleted when the remote endpoint updated
Thanks

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: I was referring to `val targetIdxs   = targets.zipWithIndex`

rpiaggio (approved):
LGTM

PR #445, opened 2020-11-12
UI improvements fixing overflow issues on the target list

cquiroz:
Fixes some details when having many targets and long texts. It includes also some fixes towards the light theme


toddburnside (approved):
LGTM

PR #447, opened 2020-11-13
Improve search by coordinates

toddburnside:
Re-enable the RA/Dec inputs and the Go To button.

I tried to empirically determine the number of decimals for the inputs. Will add some comments in the code diffs.


cquiroz (commented):
Code wise it is looking great, I'll play with the test app to see how it works
We'll need to integrate changes with the api to save the targets in the backend but that can be a separate PR

rpiaggio (approved):
LGTM

PR #451, opened 2020-11-16
Environment-dependent configuration

rpiaggio:
This is achieved by having JSON-serialized versions of `AppConfig` for each environment. It is read from the server at application startup, always from the same URL (`/conf.json`), but the server decides which version to serve based on an environment variable called `ENVIRONMENT`, which can be `development`, `staging` or `production` (or other values if the respective `<environment>.conf.json` files are defined and included in the build.

In development, `development.conf.json` is always served.

For the moment, both `development` and `staging` point to the `staging` version of `lucuma-odb`.

~I'm opening as draft for the moment, since this needs to be tested as a review app before proceeding.~

### DIFF
--- a/common/src/main/resources/conf/production.conf.json
+++ b/common/src/main/resources/conf/production.conf.json
@@ -1,0 +1,5 @@
+{
+  "environment": "PRODUCTION",
+  "odbURI": "wss://lucuma-odb.herokuapp.com/ws",
+  "exploreDBURI": "wss://explore-hasura.herokuapp.com/v1/graphql"
+}
--- a/common/src/main/webpack/webpack.parts.js
+++ b/common/src/main/webpack/webpack.parts.js
@@ -215,6 +215,16 @@ exports.extraAssets = {
           name: "[name].[hash].[ext]",
         },
       },
+      {
+        test: /\.conf.json$/,
+        loader: 'file-loader',
+        type: 'javascript/auto',
+        options: {
+          name() {
+            return 'conf/[name].[ext]';
+          },
+        },
+      }
     ],
   },
 };
--- a/explore/src/main/resources/prod.js
+++ b/explore/src/main/resources/prod.js
@@ -5,6 +5,9 @@ import "resources/less/vendor/aladin.less";
 import "resources/css/charts.scss";
 import "resources/css/datepicker.scss";
 
+import "resources/conf/staging.conf.json";
+import "resources/conf/production.conf.json";
+
 var App = require("sjs/explore-opt.js");
 
 App.Explore.runIOApp();
--- a/model/src/main/scala/explore/model/encoders.scala
+++ b/model/src/main/scala/explore/model/encoders.scala
@@ -10,10 +10,12 @@ import io.circe.refined._
 import io.circe.syntax._
 import lucuma.core.math.Declination
 import lucuma.core.math.RightAscension
+import sttp.model.Uri
 
 import ModelOptics._
 
 object encoders {
+  implicit val uriEncoder: Encoder[Uri] = Encoder.encodeString.contramap[Uri](_.toString)
 
   implicit val raEncoder: Encoder[RightAscension] =
     Encoder.encodeString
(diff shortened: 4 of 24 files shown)


common/src/main/resources/conf/production.conf.json
```diff
@@ -0,0 +1,4 @@
+{
+  "odbURI": "wss://lucuma-odb-production.herokuapp.com/ws",
```
cquiroz, line 2: Should we use the `gemini.edu` domain?

common/src/main/scala/explore/model/AppConfig.scala
```diff
@@ -0,0 +1,17 @@
...
+
+package explore.model
+
+import sttp.model.Uri
+import io.circe._
+import io.circe.generic.semiauto._
+import explore.model.encoders._
+import explore.model.decoders._
+
+case class AppConfig(exploreDBURI: Uri, odbURI: Uri)
```
cquiroz, line 12: Maybe this should have something like `isStaging` to indicate in the UI

common/src/main/webpack/webpack.parts.js
```diff
@@ -215,6 +215,16 @@ exports.extraAssets = {
           name: "[name].[hash].[ext]",
         },
       },
+      {
+        test: /\.conf.json$/,
```
cquiroz, line 219: Looks like fun

explore/src/main/resources/prod.js
```diff
@@ -5,6 +5,9 @@ import "resources/less/vendor/aladin.less";
 import "resources/css/charts.scss";
 import "resources/css/datepicker.scss";
 
+import "resources/conf/staging.conf.json";
```
cquiroz, line 8: Do we need tests?

model/src/main/scala/explore/model/encoders.scala
```diff
@@ -12,8 +12,10 @@ import lucuma.core.math.Declination
 import lucuma.core.math.RightAscension
 
 import ModelOptics._
+import sttp.model.Uri
 
 object encoders {
+  implicit val uriEncoder: Encoder[Uri] = Encoder.encodeString.contramap[Uri](_.toString)
```
cquiroz, line 18: Do we need tests?

common/src/main/webpack/webpack.parts.js
rpiaggio: And that's not even the half of it.

common/src/main/scala/explore/model/AppConfig.scala
rpiaggio: Good idea. Or an `Environment` `enum` ?

common/src/main/resources/conf/production.conf.json
rpiaggio: And the `lucuma.xyz` for staging... I'll look into setting these up.

model/src/main/scala/explore/model/encoders.scala
rpiaggio: I'll look into it. Do you know if we have a `Gen[Uri]` ?

cquiroz: Maybe scalacheck includes one?

rpiaggio: This `Uri` is from `sttp-model` and I can't find `Arbitrary` or `Eq` instances for it.

We can look into implement circe's `CodecTests` for everything that we `Encode`/`Decode` in another PR, and implement a proper `Gen[Uri]` in that context.

toddburnside (approved):
LGTM

cquiroz (approved):
LGTM

static
```diff
@@ -0,0 +1 @@
+explore/target/scala-2.13/stage/static/
```
rpiaggio, line 1: This is a symlink.

It's used to have only one `static.json` configuration and for the app to work in both build situations: Heroku or remotely (GitHub).

rpiaggio:
This PR also fixes forcing https.

PR #453, opened 2020-11-16
Read auth token periodically

cquiroz:
This is the first PR related to authentication (thus going to the feature/auth branch). It can read the token from sso and redirect to do login via orcid

The token is refreshed before it expires according to the token info


toddburnside (approved):
LGTM

build.sbt
```diff
@@ -202,7 +203,8 @@ lazy val explore: Project = project
         ReactGridLayout.value ++
         ReactHighcharts.value ++
         ReactResizable.value ++
-        ReactSizeMe.value
+        ReactSizeMe.value ++
+        Sttp.value
```
rpiaggio, line 207: This seems to be duplicated (line 202). Also, I'm not sure we need it in `explore` project if it is already in `common`.

common/src/main/scala/explore/AppMain.scala
```diff
@@ -65,7 +69,21 @@ trait AppMain extends IOApp {
...
+        val sleepTime = SSOClient.refreshTimoutDelta.max(
+          (n.until(expiration, ChronoUnit.SECONDS).seconds - SSOClient.refreshTimoutDelta)
+        )
+        IO.timer.sleep(sleepTime)
+      } *> SSOClient
+        .whoami[IO](IO.fromFuture)
+        .flatTap((u: UserVault) => v.zoom(RootModel.vault).set(u))
+
+    def repeatTokenRefresh(expiration: Instant, v: View[RootModel]): IO[Unit]      =
+      refreshToken(expiration, v).flatMap(u => repeatTokenRefresh(u.expiration, v))
```
rpiaggio, line 83: This is a very small thing: `refreshToken` and `repeatTokenRefresh` don't need to access all of `RootModel`, they could just receive a `View[UserVault]` and have the `zoom` in the caller.

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -0,0 +1,69 @@
...
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import lucuma.ui.reusability._
+import react.common._
+import react.semanticui.collections.menu._
+import react.semanticui.modules.dropdown.Dropdown
+import react.semanticui.modules.dropdown.DropdownItem
+import react.semanticui.modules.dropdown.DropdownMenu
+
+final case class TopBar(vault: View[UserVault]) extends ReactProps[TopBar](TopBar.component)
```
rpiaggio, line 22: Nice. We were starting to need this.

rpiaggio (approved):
LGTM

model/src/main/scala/explore/model/UserVault.scala
```diff
@@ -0,0 +1,20 @@
...
+import eu.timepit.refined.types.string.NonEmptyString
+import lucuma.core.model.User
+import monocle.macros.Lenses
+import java.time.Instant
+
+@Lenses
+final case class UserVault(user: User, expiration: Instant, token: NonEmptyString)
+
+object UserVault {
+  implicit val eqUserVault: Eq[UserVault] = Eq.by(x => (x.user, x.expiration, x.token))
```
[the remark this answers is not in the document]
cquiroz: I wouldn't mind automatic derivation. we have generally avoided too much magic but we could reconsider.
@tpolecat maybe has some insights

build.sbt
cquiroz: 👍

common/src/main/scala/explore/AppMain.scala
cquiroz: Good call, You're right we may want to focus on the relevant field.

model/src/main/scala/explore/model/UserVault.scala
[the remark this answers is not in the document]
tpolecat: I would rather do this by hand for now and then switch to `derives` in Scala 3. Any machinery we add at this point will just have to be removed again in a few months.

PR #460, opened 2020-11-20
Allow zooming into coulomb Quantity

toddburnside:
Provides an easy way to zoom into the value contained in a coulomb Quantity.

ex: `props.zoom(<lens>).stripQuantity`

Having it as an ops method allows the types for the quantity to be inferred, rather than having to be specified at the call site.


rpiaggio (approved):
Nice. LGTM

PR #471, opened 2020-11-24
Ch560 allow login as guest

cquiroz:
This PR adds a welcome screen that lets you either continue as guest or login with orcid:

image


common/src/main/resources/suithemes/default/elements/button.variables
```diff
@@ -256,28 +256,28 @@
 @tertiaryLinePadding: 0.5em;
 @tertiaryLineHeight: 0.2em;
 @tertiaryTextColor: @textColor;
-@tertiaryLineColor: lighten(@tertiaryTextColor, 20%);
+// @tertiaryLineColor: lighten(@tertiaryTextColor, 20%);
```
rpiaggio, line 259: Shall we remove the commented lines?

cquiroz: Actually no, I have to comment these out to get css variables to work, but eventually when there is a fomantic update we'd need to replace the default theme and I want to track what was commented

common/src/main/scala/explore/common/SSOClient.scala
```diff
@@ -45,17 +48,37 @@ object SSOClient {
...
+             case Right(Some(u)) => d.complete(u)
+             case Right(None)    =>
+               Logger[F].debug("Unauthenticated go to login form") *> guestUI[F](d, fromFuture)
+             case Left(e)        =>
+               Logger[F].error(e)("Error attempting to login") *> guestUI[F](d, fromFuture)
+           }
+      v <- d.get
+    } yield v
+
+  def guestUI[F[_]: Sync: Effect](
```
rpiaggio, line 66: Effect is Sync already.

common/src/main/scala/explore/common/SSOClient.scala
```diff
@@ -45,17 +48,37 @@ object SSOClient {
       window.location.href = uri"$baseSSO/auth/v1/stage1?state=$returnUrl".toString
     }
 
-  def whoami[F[_]: Sync](
+  def vault[F[_]: ConcurrentEffect: Logger](
```
rpiaggio, line 51: Do we need the power of a ConcurrentEffect here? Are we cancelling?

cquiroz: The creation of a `Deferred` needs `Concurrent` and we also need `Effect` to call `runAsyncCB`

common/src/main/scala/explore/common/SSOClient.scala
```diff
@@ -75,11 +98,45 @@ object SSOClient {
...
+
+  private def whoami[F[_]: Sync](
+    fromFuture: FromFuture[F, Response[Either[String, String]]]
+  ): F[Option[UserVault]] = {
+    val backend  = FetchBackend(FetchOptions(RequestCredentials.include.some, none))
+    val httpCall =
+      Sync[F].delay(
+        basicRequest
+          .post(
+            uri"$baseSSO/api/v1/refresh-token"
```
rpiaggio, line 113: The `baseSSO` can go into `AppConfig` now.

cquiroz: Yes, I have that ready to go in the next PR

common/src/main/scala/explore/components/UserSelectionForm.scala
```diff
@@ -0,0 +1,101 @@
...
+import react.semanticui.modules.modal.ModalSize
+import react.semanticui.modules.modal.ModalContent
+import explore.components.ui.ExploreStyles
+import explore.Icons
+import react.semanticui.elements.image.Image
+import scalajs.js
+import scalajs.js.annotation._
+import react.semanticui.sizes.Big
+
+final case class UserSelectionForm[F[_]: FlatMap: Effect](
```
rpiaggio, line 31: `Effect` implies `FlatMap`.

cquiroz: 👍

common/src/main/scala/explore/components/UserSelectionForm.scala
```diff
@@ -0,0 +1,101 @@
...
+              )
+            )
+          ),
+          open = true
+        )
+      }
+      .configure(Reusability.shouldComponentUpdate)
+      .build
+
+  def launch[F[_]: Sync: Effect: FlatMap](
```
rpiaggio, line 86: `Effect` includes `Sync` and `FlatMap`

common/src/main/scala/explore/components/UserSelectionForm.scala
```diff
@@ -0,0 +1,101 @@
...
+    fromFuture: FromFuture[F, Response[Either[String, String]]]
+  ): F[Unit] = Sync[F].delay {
+    val container = Option(dom.document.getElementById("root")).getOrElse {
+      val elem = dom.document.createElement("div")
+      elem.id = "root"
+      dom.document.body.appendChild(elem)
+      elem
+    }
+
+    UserSelectionForm(d, fromFuture).renderIntoDOM(container)
```
rpiaggio, line 97: Is it possible to use the router's conditional routes to display the login page instead of mounting a different root app? https://github.com/japgolly/scalajs-react/blob/master/doc/ROUTER.md#conditional-routes

cquiroz: I thought about going that way but I want that the `RootModel` always have a user, be it orcid or guest.
When you're on the user selection part there is no user yet (That's why there is a `Deferrable` there.
So my solution was to split the waters and have this `no user` state only in that particular separate app.

rpiaggio:
It's looking good. Unless there's a strong reason not to do so, I think we should attempt to put the `UserVault` in the `RootModel` and have the router decide whether to display the login screen or the app. WDYT?

cquiroz:
The `UserVault` is in `RootModel`, to do what you suggest I'd probably need to replace it by an `Option[UserVault]` which I guess would be ok.
Another reason I was trying to do this is that I actually want to preserve the route. Let's say you get a link to the url
`<host>/a/b/c`
and you're not logged in. I want to remember that route so when you login you get back to that address (That is not yet implemented).
That maybe harder to do if I need to replace the route with e.g. `<host>/userselect`. I suppose I could put the original route as a param.

rpiaggio:
I see, yes, we would need an `Option[UserVault]` in the `RootModel`.

I think we don't need to redirect to `/userselect`. The conditional route will display the login screen in `/a/b/c` if there's no user logged in, and it should reevaluate and display the actual `/a/b/c` page when the user logs in (I'm not sure if we need to do something to trigger reevaluation of the app state, I think it will be automatic, since the router receives it as `Props`).

rpiaggio (approved):
LGTM. I'd like to experiment with the possibility of unifying the logic, but we can do that later.

PR #485, opened 2020-11-27
Magnitude editor. Unset. UndoViews.

rpiaggio:
* Allow unsetting fields, by using the new functionality of the ODB (sending `null`) together with clue 0.4.0's `Input[A]` type.
* Convenience definition of `UndoableView`, which weaves `Undoer` logic into `View`'s `set`/`mod`. This simplifies and generalizes logic previously found in `UndoSet`. It is also transparent for child components that receive `View`s. Also, this is the correct way to go, the previous way of doing things updated the local model twice.
* Magnitude editor supporting undo operations. It will be refactored into an `ag-grid` once we have the facade. For simplicity, the magnitude list is treated as a single field. If different users are both editing the magnitude list of the same target at the same time, they will overwrite each other's changes. It would require profound changes and added complexity to the undo logic to support treating each band separately, especially given that the level's primary key (the band) is editable.

~This PR is a draft since requires changes in lucuma-ui and lucuma-odb which are yet to be released. Otherwise, I consider it ready to go, modulo review.~

### DIFF
--- a/constraints/src/main/scala/explore/constraints/ConstraintsQueries.scala
+++ b/constraints/src/main/scala/explore/constraints/ConstraintsQueries.scala
@@ -5,11 +5,10 @@ package explore.constraints
 
 import cats.effect.ContextShift
 import cats.effect.IO
-import cats.syntax.all._
 import clue.GraphQLOperation
+import clue.data.syntax._
 import clue.macros.GraphQL
 import crystal.ViewF
-import explore.AppCtx
 import explore.GraphQLSchemas.ExploreDB.Types._
 import explore.GraphQLSchemas._
 import explore.components.graphql.SubscriptionRenderMod
@@ -24,7 +23,6 @@ import explore.model.reusability._
 import explore.undo.Undoer
 import japgolly.scalajs.react.vdom.html_<^._
 import monocle.Lens
-import monocle.function.Cons.headOption
 import monocle.macros.Lenses
 
 object ConstraintsQueries {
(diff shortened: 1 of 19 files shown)


common/src/main/scala/explore/components/ConnectionsStatus.scala
```diff
@@ -52,7 +52,7 @@ object ConnectionsStatus {
     .render(_ =>
       AppCtx.withCtx { ctx =>
         <.span(
-          ctx.clients.ExploreDBConnectionStatus(renderStatus("Hasura DB")).when(false),
+          // ctx.clients.ExploreDBConnectionStatus(renderStatus("Hasura DB")).when(false),
```
rpiaggio, line 55: References to Explore DB are left commented out on purpose. In the future we are planning to have a DB for user settings, etc.

constraints/src/main/scala/explore/constraints/ConstraintsQueries.scala
```diff
@@ -83,31 +81,32 @@ object ConstraintsQueries {
...
-    }
+    ???
+  // AppCtx.withCtx { implicit appCtx =>
+  //   SubscriptionRenderMod[Subscription.Data, Constraints](
+  //     Subscription.subscribe(id),
+  //     _.map(
+  //       Subscription.Data.constraints.composeOptional(headOption).getOption _
+  //     ).unNone
+  //   )(render)
+  // }
```
rpiaggio, line 111: All this needs to be reimplemented when ODB supports Constraints.

targeteditor/src/main/scala/explore/targeteditor/MagnitudeForm.scala
```diff
@@ -3,44 +3,205 @@
...
+
+                    val magnitudeBigDecimalView: View[BigDecimal] =
+                      magnitudeValueView.zoom(v => BigDecimal(v.toDoubleValue))(mod =>
+                        m => MagnitudeValue.fromDouble(mod(BigDecimal(m.toDoubleValue)).toDouble)
+                      )
+
+                    // TODO Add this to lucuma-ui?
+                    // val doubleValidFormat: ValidFormatInput[Double] = ???
+                    //
+                    // Actually we need something for an Int / 1000
```
rpiaggio, line 117: Ooops... This is an oversight. We need a good `ValidFormatInput` for `MagnitudeValue`s. I'll send this in another commit.

common/src/main/resources/conf/development.conf.json
```diff
@@ -1,5 +1,4 @@
 {
   "environment": "DEVELOPMENT",
-  "odbURI": "wss://odb.gpp.lucuma.xyz/ws",
-  "exploreDBURI": "wss://explore-hasura.herokuapp.com/v1/graphql"
+  "odbURI": "wss://lucuma-odb-p-unset-jlpak1dtmcd.herokuapp.com/ws"
```
cquiroz, line 3: Should we have a nae under `lucuma.xyz`

model/src/main/scala/explore/model/AppConfig.scala
```diff
@@ -12,7 +12,7 @@ import io.circe._
 import io.circe.generic.semiauto._
 import sttp.model.Uri
 
-case class AppConfig(environment: ExecutionEnvironment, exploreDBURI: Uri, odbURI: Uri)
+case class AppConfig(environment: ExecutionEnvironment, /*exploreDBURI: Uri,*/ odbURI: Uri)
```
cquiroz, line 15: I wonder if we should just keep it, for once it will make the merge with `auth` harder and OTOH we'll start using the userdb soon

cquiroz (commented):
Looks good, I'm looking forward to test it on a demo

model/src/main/scala/explore/model/AppConfig.scala
rpiaggio: Ok, I'll uncomment then. The ExploreDB we end up using will be very different to the one we were using until now, but let's just keep the connection open for now.

common/src/main/resources/conf/development.conf.json
rpiaggio: We already have it for staging: https://odb.gpp.lucuma.xyz. I was using a review app here since the required ODB changes weren't merged yet, but they are now.

rpiaggio:
Review app in https://explore-pipe-magnitudes-hohljn.herokuapp.com/

toddburnside (approved):
LGTM

PR #488, opened 2020-11-28
Ch355 logout button

cquiroz:
This PR adds a logout button which delets the token and cookie
It also notifies other tabs so they all logout once you logout of one
I'm using `BroadcastChannel` but safari isn't supporting so I had  to relay on an npm package to provide a workaround for this missing API


toddburnside (approved):
Cool.

PR #489, opened 2020-11-30
Switch cats imports

cquiroz:
Starting with cats 2.3.0 it is recommended to switch to import `cats.syntax.all`


toddburnside (approved):
LGTM

PR #490, opened 2020-11-30
Add button for guest users to switch to ORCID

cquiroz:
This PR adds a menu entry to let guest users switch to ORCID

Note that this only applies to switching the user, no data is being converted. this is irrelevant at this state because we all share the same data but will need to be fixed:

image


toddburnside (approved):
LGTM

PR #491, opened 2020-11-30
Fix position of RA/Dec error labels.

toddburnside:


common/src/main/resources/less/style.less
```diff
@@ -1007,10 +1007,15 @@ img.partner-split-flag {
   }
 }
 
-.explore-input-error-tooltip {
+// position all fields relative so that error labels position properly.
+.ui .field {
   position: relative;
```
cquiroz, line 1012: 👍

cquiroz (approved):
Looks great, thanks

PR #500, opened 2020-12-03
Clue 0.5.0. Actually 0.6.0, probably. Make that 0.7.0.

rpiaggio:
Adapts code to the new version of clue. In particular, this allows for an authorization token to be passed upon establishing an Apollo WS connection.

Build will fail until clue 0.5.0 is released, thus I'm marking this PR as draft until then.

### DIFF
--- a/common/src/main/scala/explore/AppMain.scala
+++ b/common/src/main/scala/explore/AppMain.scala
@@ -3,6 +3,8 @@
 
 package explore
 
+import java.util.concurrent.TimeUnit
+
 import scala.concurrent.duration._
 import scala.scalajs.js
 
@@ -11,6 +13,7 @@ import cats.effect.IO
 import cats.effect.IOApp
 import cats.syntax.all._
 import clue.Backend
+import clue.WebSocketReconnectionStrategy
 import clue.js.AjaxJSBackend
 import clue.js.WebSocketJSBackend
 import crystal.AppRootContext
@@ -21,7 +24,7 @@ import explore.model.RootModel
 import explore.model.enum.AppTab
 import explore.model.reusability._
 import io.chrisdavenport.log4cats.Logger
-import io.circe.JsonObject
+import io.circe.Json
 import io.circe.syntax._
 import japgolly.scalajs.react.vdom.VdomElement
 import log4cats.loglevel.LogLevelLogger
@@ -36,9 +39,6 @@ import sttp.client3.circe._
 import sttp.model.Uri
 
 import js.annotation._
-import clue.ReconnectionStrategy
-import org.scalajs.dom.raw.CloseEvent
-import java.util.concurrent.TimeUnit
 
 object AppCtx extends AppRootContext[AppContextIO]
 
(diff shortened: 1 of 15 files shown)


common/src/main/scala/explore/AppMain.scala
```diff
@@ -106,6 +108,7 @@ trait AppMain extends IOApp {
       appConfig <- fetchConfig
       _         <- logger.info(s"Config: ${appConfig.show}")
       ctx       <- AppContext.from[IO](appConfig)
+      _         <- ctx.clients.odb.connect(JsonObject("Authorization" -> "Bearer ...".asJson))
```
cquiroz, line 111: I guess we need to stick the token instead of `...`?

targeteditor/src/main/scala/explore/targeteditor/MagnitudeForm.scala
```diff
@@ -82,7 +81,7 @@ object MagnitudeForm {
               )
             )
 
-          Form(size = Small)(
+          React.Fragment(
```
cquiroz, line 84: Why this change? Seems unrelated

rpiaggio: Yes, it's unrelated. There was a `Form` inside of another `Form` (from `TargetBody`).

common/src/main/scala/explore/AppMain.scala
rpiaggio: Exactly. I just pushed a new commit with further changes to clue which allow us to specify a different token in each reconnection.

common/src/main/scala/explore/AppMain.scala
```diff
@@ -101,11 +105,30 @@ trait AppMain extends IOApp {
...
+
     for {
       _         <- logger.info(s"Git Commit: [${BuildInfo.gitHeadCommit.getOrElse("NONE")}]")
       appConfig <- fetchConfig
       _         <- logger.info(s"Config: ${appConfig.show}")
       ctx       <- AppContext.from[IO](appConfig)
+      _         <- ctx.clients.odb
+                     .connect(
+                       connectParameters,
+                       reconnectionStrategy.asInstanceOf[ReconnectionStrategy[IO, ctx.clients.odb.CE]].some
```
rpiaggio, line 130: I'm working of ways to improve the types in `clue` so that this cast is not necessary. I will investigate using `sttp` but it will take a while, so I'm pushing this in the meantime not to hold things up.

common/src/main/scala/explore/components/ConnectionsStatus.scala
```diff
@@ -54,8 +55,12 @@ object ConnectionsStatus {
         <.span(
           ctx.clients.ExploreDBConnectionStatus(renderStatus("Hasura DB")).when(false),
           ctx.clients.ODBConnectionStatus(renderStatus("ODB")),
-          Button(size = Tiny)(^.onClick --> ctx.clients.odb.close().runAsyncCB)("Close ODB")
-            .when(false)
+          Button(size = Tiny)(
+            ^.onClick --> ctx.clients.odb
+              .disconnect(WSJSCloseParams(code = 4000).asInstanceOf[ctx.clients.odb.CP])
```
rpiaggio, line 60: I'm working of ways to improve the types in `clue` so that this cast is not necessary. I will investigate using `sttp` but it will take a while, so I'm pushing this in the meantime not to hold things up.

common/src/main/scala/explore/AppMain.scala
```diff
@@ -55,7 +59,7 @@ trait AppMain extends IOApp {
   def runIOApp(): Unit = main(Array.empty)
 
   override final def run(args: List[String]): IO[ExitCode] = {
-    japgolly.scalajs.react.extra.ReusabilityOverlay.overrideGloballyInDev()
+    // japgolly.scalajs.react.extra.ReusabilityOverlay.overrideGloballyInDev()
```
cquiroz: Should we restore this?

common/src/main/scala/explore/AppMain.scala
```diff
@@ -101,11 +105,30 @@ trait AppMain extends IOApp {
       IO.fromFuture(httpCall).map(_.body)
     }
 
+    val fetchToken: IO[String] = IO("...")
+
+    val connectParameters: IO[JsonObject]                          =
+      fetchToken.map(token => JsonObject("Authorization" -> s"Bearer $token".asJson))
+
+    val reconnectionStrategy: ReconnectionStrategy[IO, CloseEvent] =
+      ReconnectionStrategy(maxAttempts = 100,
```
cquiroz, line 114: Any reason for a 100? Or is it reset once you reconnect?

cquiroz (approved):
LGTM

swalker2m:
Would either of you take a look at the [corresponding updates in the server for `0.5.0`](https://github.com/gemini-hlsw/lucuma-tmp-api/pull/96) to make sure it is correct and compatible?

common/src/main/scala/explore/AppMain.scala
rpiaggio: Good catch

common/src/main/scala/explore/AppMain.scala
rpiaggio: The attempt counter is indeed reset upon a successful reconnection.

I just put a high number there for now. We have to decide what reconnection strategy we will use, we probably want an exponential backoff instead of a linear one.

How does it sound to have a capped exponential backoff with infinite attempts?

cquiroz: Sure that sounds like what the seqexec does
Would you do that in a separate PR?

rpiaggio: Didn't think of copying what we do in Seqexec. I'm sending a new commit with changes for clue 0.6.0 and I'm including the same logic as there.

PR #502, opened 2020-12-05
Show logged in user status and public name

cquiroz:
This PR adds some improvements to the UI in particular when showing the public display name.
It also makes it more robust against errors renewing the ticket.


common/src/main/scala/explore/components/ConnectionsStatus.scala
```diff
@@ -9,54 +9,52 @@ import crystal.Error
...
 
 object ConnectionsStatus {
   type Props = ConnectionsStatus
 
-  private def renderStatus(name: String)(status: Pot[StreamingClientStatus]): VdomNode = {
-    val (message, clazz) = status match {
-      case Error(t)     => (t.getMessage, ConnectionError)
-      case Pending(_)   => ("Mounting...", ConnectionWarning)
+  private def renderStatus(name: String)(status: Pot[StreamingClientStatus]): VdomElement = {
+    val (message, clazz, show) = status match {
```
rpiaggio, line 29: Nice, thank you.

common/src/main/scala/explore/AppMain.scala
```diff
@@ -79,7 +79,11 @@ trait AppMain extends IOApp {
     def repeatTokenRefresh(ssoURI: Uri, expiration: Instant, v: View[UserVault]): IO[Unit] =
       SSOClient
         .refreshToken[IO](ssoURI, expiration, v.set, IO.fromFuture)
-        .flatMap(u => repeatTokenRefresh(ssoURI, u.expiration, v))
+        .attempt
+        .flatMap {
+          case Right(u) => repeatTokenRefresh(ssoURI, u.map(_.expiration).getOrElse(expiration), v)
+          case Left(_)  => repeatTokenRefresh(ssoURI, expiration, v)
```
rpiaggio, line 85: Is the `Left` case when there's no longer a valid token?

rpiaggio (approved):
LGTM

common/src/main/scala/explore/AppMain.scala
cquiroz: We need to differentiate (in another PR) when there is a regular failure which we are ignoring here and when the user was kicked out of SSO for some reason
I'll handle that case in another PR

PR #503, opened 2020-12-08
Send the token to the backend on the initial connect

cquiroz:
This should be the last PR on feature/auth and then we can merge it and do better testing

### DIFF
--- a/common/src/main/scala/explore/AppMain.scala
+++ b/common/src/main/scala/explore/AppMain.scala
@@ -119,10 +119,8 @@ trait AppMain extends IOApp {
       IO.fromFuture(httpCall).map(_.body)
     }
 
-    val fetchToken: IO[String] = IO("...")
-
-    val connectParameters: IO[Map[String, Json]]            =
-      fetchToken.map(token => Map("Authorization" -> s"Bearer $token".asJson))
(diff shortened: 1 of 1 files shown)


common/src/main/scala/explore/AppMain.scala
```diff
@@ -119,10 +119,8 @@ trait AppMain extends IOApp {
       IO.fromFuture(httpCall).map(_.body)
     }
 
-    val fetchToken: IO[String] = IO("...")
-
-    val connectParameters: IO[Map[String, Json]]            =
-      fetchToken.map(token => Map("Authorization" -> s"Bearer $token".asJson))
+    def connectParameters(vault: UserVault): IO[Map[String, Json]] =
+      Map("Authorization" -> s"Bearer ${vault.token.value}".asJson).pure[IO]
```
rpiaggio, line 123: I think we may want to re-read the current token here from the current vault, so that we have the latest one upon reconnection. But we can investigate a way to do that in another PR.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/AppMain.scala
cquiroz: Yes we need to do that, I wasn't sure how to send a new token to the connection

PR #507, opened 2020-12-10
Full Gids in URLs. Also, disable magnitudes when loading.

rpiaggio:
We now use the canonical form of `Gid`s in URLs. This creates consistency with the API and everywhere else. Also, it seems to resolve [ch-534] by sidestepping the issue with Safari.

Also, we now disable magnitude editing whenever the rest of of the target editor fields are disabled.


cquiroz (approved):
Looks Good, do you know if we can disable `verify` in production?

rpiaggio:
> Looks Good, do you know if we can disable `verify` in production?

That's a good idea, let me give it a try.

targeteditor/src/main/scala/explore/targeteditor/MagnitudeForm.scala
```diff
@@ -124,15 +126,17 @@ object MagnitudeForm {
...
                           )
                         )
                       ),
                       TableCell(
                         <.span(
                           EnumViewSelect(
                             id = magnitude.band.toString + "_SYSTEM",
-                            value = magnitudeView.zoom(Magnitude.system)
+                            value = magnitudeView.zoom(Magnitude.system),
+                            disabled = props.disabled
```
toddburnside, line 139: Should all the buttons on the form be disabled, too?

rpiaggio: Yes, that's a good idea.

PR #508, opened 2020-12-11
Detect if using older Safari and forbid access

cquiroz:
When logging in we detect the browser and can indicate if it isn't supported, so far we only blacklist older Safari versions

image (11)

Parsing UA is harder than I expected so I'm bringing in a little (well maintained) js library


swalker2m:
This is nitpicky but I think that message is technically a comma splice.

rpiaggio (approved):
LGTM

PR #519, opened 2020-12-15
Use monocle Index in PartnerSplitsEditor

toddburnside:
I based the original version on the code in MagnitudeForm. But, what do y'all think of using a monocle Index for zooming into the list?

### DIFF
--- a/common/src/main/scala/explore/implicits.scala
+++ b/common/src/main/scala/explore/implicits.scala
@@ -12,6 +12,7 @@ import cats.syntax.all._
 import clue._
 import coulomb.Quantity
 import crystal.ViewF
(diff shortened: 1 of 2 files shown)


cquiroz (commented):
Looks great, we deprecated a bunch of stuff for Monocle 3 but `index` remains.

cquiroz (approved):

rpiaggio:
This is cool. I wasn't aware of this. Maybe we can use it in magnitudes too.

toddburnside:
> This is cool. I wasn't aware of this. Maybe we can use it in magnitudes too.

If things go with ag-grid like I'm hoping, it will be used for magnitudes, but not directly.

PR #520, opened 2020-12-16
Show a banner in staging

cquiroz:
This  is a small improvement, it will show a banner to indicate we are in staging. I hope it makes it clear to users when they are entering data in staging

This adds a node and styles it via css. It puts it in an absolute position thus it should not alter the layout

image


toddburnside (approved):
LGTM

PR #529, opened 2020-12-22
Allow computing URLs.

rpiaggio:
Add a method `AppCtx.pageURL(AppTab, Option[Focused]): String` to allow building URLs from arbitrary (tab, focused) combinations.

This is useful if we want to provide navigation via links in the app.

Note that for subprojects, since there is no router, it will always return `#`.


cquiroz (approved):
Looks great, thanks

toddburnside (approved):
👍

PR #530, opened 2020-12-22
General rework of the main layout improving the mobile UI and a few

cquiroz:
This is a rework of the main layout to include:

* Cosmetic changes to the tabs
* Some changes to the tab moving them to the bottom in devices with a reduced width
* Rework of the style files splitting the theme variables to their own files
* Rework of the target list to work better on smaller devices:

https://user-images.githubusercontent.com/3615303/102926732-4df92200-4474-11eb-8678-f35a0be19439.mov

### DIFF
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -6,6 +6,9 @@
 @import "verticalbuttons.less";
 @import "vendor/react-resizable.less";
 @import "vendor/react-grid-layout.less";
+// theming variables
+@import "variables-dark.less";
+@import "variables-light.less";
 
 @import "explore-grid.less";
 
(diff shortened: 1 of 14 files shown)


common/src/main/resources/less/style.less
```diff
@@ -523,12 +39,15 @@ i.icon.explore-button-icon:focus {
     0 0 0 0.2rem var(--focus-outer-color);
 }
 
+html {
+  // Required for iOS to work with 100vh
+  // https://css-tricks.com/css-fix-for-100vh-in-mobile-webkit/
+  height: -webkit-fill-avilable;
```
toddburnside, line 45: I think this needs to be `-webkit-fill-available`. (an extra `a` in `available`)

cquiroz: Oh thanks for noting it

cquiroz: Maybe we should have a css linter as part of the build

toddburnside (approved):
Looks good. A lot of fancy CSS work. :) 

One thing I notice is that with the Magnitudes being in a table, on a narrow screen the table can collapse into a single column. If we use something like ag-grid, that won't happen (at least I think it can't happen). But, I guess the user can just scroll horizontally.

common/src/main/resources/less/style.less
toddburnside: That would be nice. It's pretty easy for a typo to get missed.

cquiroz:
I think on narrow screen we'll rework a bit the order, I didn't want to touch it in this PR though
We should get it into ag-grid before doing further work

rpiaggio (approved):
LGTM

rpiaggio:
The idea of sliding panels on small screens is very nice. As you mention in the code comments, it may be worth looking for a way to generalize it when we do the other views that have a tree and an edit area.

cquiroz:
Yes, I wanted to get this done now as we'll start working on the other screens and it will be harder to retrofit later

PR #531, opened 2020-12-22
Updates coordinates after search

cquiroz:
Just a small bug fix. After  search result is retrieved we should update the coordinates on the aladin toolbar. This PR does that


toddburnside (approved):
LGTM

PR #533, opened 2020-12-23
Converts the navigation items to links 

cquiroz:
This is a simple but IMHO very useful improvement.
The buttons right now are just plain buttons but we can convert them to proper links allowing actions like e.g. open in new tab or copy the url:

https://user-images.githubusercontent.com/3615303/103002747-3a01fe80-450e-11eb-89ad-c66f040205c0.mp4

### DIFF
--- a/explore/src/main/scala/explore/SideTabs.scala
+++ b/explore/src/main/scala/explore/SideTabs.scala
@@ -13,9 +13,11 @@ import lucuma.core.data.EnumZipper
 import lucuma.ui.reusability._
 import react.common._
 import react.semanticui.elements.button.Button
+import react.semanticui.elements.button.Button.ButtonProps
 import react.semanticui.elements.button.ButtonGroup
 import react.semanticui.elements.divider.Divider
 import react.semanticui.elements.label.Label
+import react.semanticui.elements.label.Label.LabelProps
 import react.semanticui.sizes._
 
 final case class SideTabs(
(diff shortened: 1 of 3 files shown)


swalker2m (commented):
I don't know anything about the code but that does seem quite useful.

explore/src/main/scala/explore/SideTabs.scala
```diff
@@ -32,53 +34,65 @@ object SideTabs {
...
-          )(tab.title)
+          def tabButton(tab:          AppTab): Button       =
+            Button(
+              as = <.a,
+              active = tab === focus,
+              clazz = ExploreStyles.SideButton,
+              onClickE = (e: ReactMouseEvent, _: ButtonProps) =>
+                e.preventDefaultCB *> p.tabs
+                  .mod(z => z.findFocus(_ === tab).getOrElse(z))
+                  .runAsyncCB
```
rpiaggio, line 49: I would have expected the `href` to be enough so that we could remove the `onClick`. It doesn't trigger the router?

rpiaggio (approved):
LGTM

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -269,76 +272,83 @@ object TargetObsList {
...
+                                .set(focusedTarget.some)
+                                .runAsyncCB
+                            )(
+                              <.span(ExploreStyles.ObsTreeGroupHeader)(
+                                <.a(
+                                  ^.href := ctx.pageUrl(AppTab.Targets, focusedTarget.some),
+                                  opIcon,
+                                  target.name.value,
+                                  ExploreStyles.TargetLabelTitle,
+                                  ^.onClick ==> {(e: ReactEvent) => e.preventDefaultCB}
```
toddburnside, line 302: I'm not sure I'm following this. What is having onClick prevent the default action on an anchor tag preventing?

explore/src/main/scala/explore/SideTabs.scala
toddburnside: If it's a link, does it even have to be a Button? Well, other than the fact that we are using ButtonGroup to group them. Ignore this if the onClickE is necessary.

cquiroz: Without the onClick it will go to that route but with a page reload. 
With respect to making it a plain link, it is possible but it would require extra css to restore the current look

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
cquiroz: If you don't prevent it will go to the correct url but via a page reload rather than an internal navegation

toddburnside: 👍

explore/src/main/scala/explore/SideTabs.scala
toddburnside: Oh, interesting. So, this combination allows opening in another tab, etc., but prevents a page reload when clicking. Good job figuring that out - it seems non-obvious to me.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/SideTabs.scala
rpiaggio: Isn't e.preventDefaultCB enough as in TargetObsList?

cquiroz: Thanks for spotting this, `preventDefaultCB` isn't enough

cquiroz: Fixed, and I realized I'm blocking the regular cmd+click shortcuts, so now it is handled as well.
We may want to refactor this into a shared method for handling these

PR #536, opened 2020-12-28
Add AgGrid for Magnitudes

toddburnside:
[ch568]
Checkpoint PR to look at implementation and ag-grid itself. There are still a number of issues (besides styling).
- Editing a cell won't really change the backing value. We need the grid `lifecyle` methods to accomplish that.
- The dropdown for the enum editor is cutoff at the cell boundary. I'm hoping that will be fixed by telling the grid to use a popup editor, but a `lifecycle` method is required for that.

I am using the `alpine-dark` theme provided by ag-grid. It is close, but doesn't quite match our theme.
image

### DIFF
--- a/common/src/main/scala/explore/components/AgGridHelpers.scala
+++ b/common/src/main/scala/explore/components/AgGridHelpers.scala
@@ -3,8 +3,6 @@
 
 package explore.components
 
-import scala.annotation.unused
-
 import cats._
 import cats.data.Validated._
 import cats.syntax.all._
@@ -64,7 +62,7 @@ object AgGridHelpers {
    */
   def editableViewColumn[A, B](
     lens:          Lens[A, B],
-    validFormat:   ValidFormatInput[B] = ValidFormatInput.id,
+    validFormat:   ValidFormatInput[B],
     changeAuditor: ChangeAuditor[B] = ChangeAuditor.accept[B],
     disabled:      Boolean = false
   )(implicit eq:   Eq[B]): AgGridColumn.ColDef.Builder =
@@ -116,9 +114,9 @@ object AgGridHelpers {
    * @return An AgGridColumn Builder.
    */
   def buttonViewColumn[A](
-    @unused button:   Button,
-    @unused onClick:  A => Callback,
-    @unused disabled: Boolean = false
+    button:   Button,
+    onClick:  A => Callback,
+    disabled: Boolean = false
   ): AgGridColumn.ColDef.Builder = {
     val component = ScalaComponent
       .builder[View[A]]
@@ -201,7 +199,7 @@ object AgGridHelpers {
             val len = i.value.length
             i.setSelectionRange(0, len)
           }
-        }.void.delayMs(50).completeWith(_ => Callback(()))
+        }.void.delayMs(1).toCallback
 
       // We have to wait until getValue is called before setting the value or
       // we get "kicked out of" editing due to re-rendering of the table.
--- a/common/src/main/scala/explore/components/AgGridHelpers.scala
+++ b/common/src/main/scala/explore/components/AgGridHelpers.scala
@@ -1,0 +1,258 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.components
+
+import scala.annotation.unused
+
+import cats._
+import cats.data.Validated._
+import cats.syntax.all._
+import cats.effect.IO
+import crystal.ViewF
+import crystal.react.implicits._
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.string._
+import explore.AppCtx
+import explore._
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import japgolly.scalajs.react.MonocleReact._
+import lucuma.core.util.Display
+import lucuma.core.util.Enumerated
+import lucuma.ui.forms._
+import lucuma.ui.optics._
+import monocle.Lens
+import monocle.macros.Lenses
+import org.scalajs.dom.document
+import org.scalajs.dom.ext.KeyCode
+import org.scalajs.dom.html
+import react.semanticui.elements.button.Button
+import reactST.agGridCommunity.colDefMod._
+import reactST.agGridCommunity.iCellEditorMod.ICellEditorParams
+import reactST.agGridCommunity.iCellRendererMod.ICellRendererParams
+import reactST.agGridReact.components._
+
+import scalajs.js
+
+object AgGridHelpers {
+
+  /**
+   * Returns a display only AgGridColumn Builder.
+   * Since this returns a Builder, additional column configuration methods can
+   * be chained onto the return value.
+   *
+   * @param lens A Lens to go between the Row value and the Cell value.
+   * @param formatter Function for formatting the value. This will often be the `reverseGet` of a Format or ValidFormatInput.
+   * @return An AgGridColumn Builder.
+   */
+  def viewColumn[A, B](lens: Lens[A, B])(formatter: B => String): AgGridColumn.ColDef.Builder =
+    AgGridColumn.ColDef
+      .valueGetter(getter(lens))
+      .cellRendererFramework(renderer(formatter))
+
+  /**
+   * Create a AgGridColumn Builder for a column which edits the value in a FormInputEv.
+   * Since this returns a Builder, additional column configuration methods can
+   * be chained onto the return value.
+   *
+   * @param lens Lens to go between the Row value and the Cell value.
+   * @param validFormat The ValidFormatInput for the editor.
+   * @param changeAuditor The ChangeAuditor for the editor.
+   * @param disabled Whether editing should be disabled or not.
+   * @return An AgGridColumn Builder
+   */
+  def editableViewColumn[A, B](
+    lens:          Lens[A, B],
+    validFormat:   ValidFormatInput[B] = ValidFormatInput.id,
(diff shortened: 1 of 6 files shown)


common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+import react.semanticui.elements.button.Button
+import reactST.agGridCommunity.colDefMod._
+import reactST.agGridCommunity.iCellEditorMod.ICellEditorParams
+import reactST.agGridCommunity.iCellRendererMod.ICellRendererParams
+import reactST.agGridReact.components._
+
+import scalajs.js
+
+object AgGridHelpers {
+
```
toddburnside, line 39: I haven't been able to figure out a way to make the public methods typesafe. The renders, etc. are doing a lot of `asInstanceOf[View[...]]`s, but I can't ensure that these are the correct types.

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+  /**
+   * Returns a display only AgGridColumn Builder.
+   * Since this returns a Builder, additional column configuration methods can
+   * be chained onto the return value.
+   *
+   * @param lens A Lens to go between the Row value and the Cell value.
+   * @param formatter Function for formatting the value. This will often be the `reverseGet` of a Format or ValidFormatInput.
+   * @return An AgGridColumn Builder.
+   */
+  def viewColumn[A, B](lens: Lens[A, B])(formatter: B => String): AgGridColumn.ColDef.Builder =
```
toddburnside, line 49: Two parameter lists to improve type inference. Not sure if this is the preferred approach here.

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+      // We need to add a little delay before looking for the input element. I think
+      // ag-grid is using some virtualization of it's own, and the input is not in
+      // the DOM immediately.
+      def setCursor(): Callback =
+        CallbackTo {
+          val inputOpt = Option(document.querySelector(s"#${id.value}").asInstanceOf[html.Input])
+          inputOpt.map { i =>
+            i.focus();
+            val len = i.value.length
+            i.setSelectionRange(0, len)
```
toddburnside, line 202: Currently selecting the entire field when starting editing. Could switch to putting the cursor at the end.

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+          // If the current text is invalid, we reset to the original to behave more
+          // like the inputs that are not in tables.
+          case Invalid(_) => $.state.flatMap(s => $.setStateL(State.current)(s.original))
+        }
+
+      // ag-grid lifecycle method.
+      // Note that it doesn't matter what `getValue` actually returns, since
+      // 'valueSetter' is a noop. We just need to set the value in the View
+      // in the props.
+      // Is there a way to avoid calling runNow()?
```
toddburnside, line 225: Is there a way to avoid calling runNow() here and in `isCancelAfterEnd`? Is calling runNow() safe in this instance?

common/src/main/scala/explore/implicits.scala
```diff
@@ -43,6 +44,33 @@ trait ListImplicits {
       monoid: Monoid[Out]
     ): Out = hlists.map(_.map(singleton)).combineAll
   }
+
+  implicit class ViewListOps[F[_], A](val viewList: ViewF[F, List[A]]) {
+    def toAgGridData[B](eqBy: A => B)(implicit eq: Eq[B]): List[ViewF[F, A]] =
```
toddburnside, line 49: Should probably have better names for these. Would calling them `sequence` and `sequenceSorted` be more clear or more confusing?

common/src/main/scala/explore/components/AgGridHelpers.scala
cquiroz: Very often when interacting with the js world you need to call `runNow()`, basically if this is to be sent to `ag-grid` you probably need to call `runNov` to execute the callback

common/src/main/scala/explore/implicits.scala
cquiroz: `sequence` could be confusing considering its meaning in `cats`

common/src/main/scala/explore/components/AgGridHelpers.scala
toddburnside: Thanks. I thought that was probably the case, and that it was being sent outside of react it was safe to call. But, I thought I should make sure.

cquiroz (commented):
Codewise seems good, It would be nice to see it in a test app :S

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+   *
+   * @param lens Lens to go between the Row value and the Cell value.
+   * @param validFormat The ValidFormatInput for the editor.
+   * @param changeAuditor The ChangeAuditor for the editor.
+   * @param disabled Whether editing should be disabled or not.
+   * @return An AgGridColumn Builder
+   */
+  def editableViewColumn[A, B](
+    lens:          Lens[A, B],
+    validFormat:   ValidFormatInput[B] = ValidFormatInput.id,
```
rpiaggio, line 67: I would recommend against using `ValidFormatInput.id`. I regret using it in `FormInputEV`. It has caused me some headeaches since it forces `B = String`.

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+   *
+   * @param button The button to put in the cell.
+   * @param onClick The onClick handler for the button.
+   * @param disabled Whether the button should be disabled.
+   * @return An AgGridColumn Builder.
+   */
+  def buttonViewColumn[A](
+    @unused button:   Button,
+    @unused onClick:  A => Callback,
+    @unused disabled: Boolean = false
```
rpiaggio, line 121: Leftover @unused annotations?

common/src/main/scala/explore/components/AgGridHelpers.scala
toddburnside: It's caught by the compiler, though, right? Still, it's not generally what we want, so it could be surprizing. I'll just remove the default and require the passing of a ValidFormatInput.

common/src/main/scala/explore/components/AgGridHelpers.scala
toddburnside: Oh, yeah. Good catch. Was trying to figure out why the whole thing was just blowing up. 💣

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+  ): AgGridColumn.ColDef.Builder = {
+    val component = ScalaComponent
```
rpiaggio, line 123: Can we try doing `button.addModifiers(...).render` here and see if that saves us from creating the wrapper Scala component?

rpiaggio: Sorry, it's not `render` that we need to call but `component`, which is actually protected in `GenericJsComponentAC`. Maybe we could make it public in react-common...

rpiaggio: Ok, that wouldn't work either... `Button` encapsulates other properties... Let's leave it as it is.

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+
+    class Backend($ : BackendScope[View[A], State])(implicit val eq: Eq[A]) {
+      val id = newId
+
+      // We need to add a little delay before looking for the input element. I think
+      // ag-grid is using some virtualization of it's own, and the input is not in
+      // the DOM immediately.
+      def setCursor(): Callback =
+        CallbackTo {
+          val inputOpt = Option(document.querySelector(s"#${id.value}").asInstanceOf[html.Input])
```
rpiaggio, line 198: Can we use a `Ref` for this?

common/src/main/scala/explore/components/AgGridHelpers.scala
```diff
@@ -0,0 +1,258 @@
...
+          inputOpt.map { i =>
+            i.focus();
+            val len = i.value.length
+            i.setSelectionRange(0, len)
+          }
+        }.void.delayMs(50).completeWith(_ => Callback(()))
```
rpiaggio, line 204: `delayMs(0)` might work here. It will still queue the `Callback` instead of executing it immediately. Also you can do `.toCallback` on the resulting `AsyncCallback` (which does the same thing as the current code).

common/src/main/scala/explore/components/AgGridHelpers.scala
toddburnside: I also tried that with the same sort of problems. The Ref was always empty. However, unlike the dom search, the Ref was still empty after waiting. I guess it was set to empty in `render()` because it wasn't actually in the dom, but never gets updated?

common/src/main/scala/explore/components/AgGridHelpers.scala
toddburnside: Zero ms doesn't seem to work, but 1 ms does. Thanks for the `toCallback` info. I was trying to find something like that but missed it.

rpiaggio (approved):
LGTM so far, and worth exploring more. Thank you for all the docs too.

toddburnside:
Switched to react-table.

PR #537, opened 2020-12-28
Setup stylelint

cquiroz:
I've toying with the idea of using linter the less/css is some way. This PR adds such check using `stylelint`.
It only checks our css and not SemanticUI which is out of our control.

the setup is a bit complex as we have the light and dark themes, it may break if we change conventions

Hopefully it doesn't add a burden as it reflects some personal preferences (e.g. using hsl over rgb)


rpiaggio (approved):
LGTM

toddburnside (approved):
Seems like a good idea to me.

PR #544, opened 2021-01-04
Fixes issue of deleting grid rows in the Magnitudes ag-grid.

toddburnside:
Also fixes issues with the MagnitudeBand dropdown list - the excluded values were not updated when the data changed.

I created this as a separate PR to emphasize what I needed to do. The ergonomics of using ag-grid seem to be a little lacking.

The basic issue is this: All of the settings, (including the column definitions and setting methods like `getRowNodeId` are only evaluated the first time the grid is created by a component. This means that any data that is closed over by an functions used in the above is essentially frozen since these functions will never be updated. In the case of the row deleting function, this meant that the data from which the rows were being deleted was always the original data from when the Magnitude form was mounted.

It took awhile to figure this out, and then I tried numerous approaches to try to force the columns to be updated. (One of these attempts led to the discovery that the problem extended to other settings such as `getRowNodeId`.) I thought maybe I could obtain the `gridApi` from the `gridOnReady` event and reset the column definitions. Unfortunately, only vanilla JS columns can be set via the gridApi - react components cannot be used.

The only solution I could come up with for the row deleting was to provide the entire dataset with the data for each row,  pass that back to the button onClick method, and use that for deleting the row and updating the data. 

The solution for the "excluded" values of the band dropdown was similar. The dataset provided with the row is used to determine the excluded values via a function that is passed in.

### DIFF
--- a/common/src/main/scala/explore/components/AgGridHelpers.scala
+++ b/common/src/main/scala/explore/components/AgGridHelpers.scala
@@ -13,6 +13,7 @@ import eu.timepit.refined.auto._
 import eu.timepit.refined.types.string._
 import explore.AppCtx
 import explore._
+import explore.utils._
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import japgolly.scalajs.react.MonocleReact._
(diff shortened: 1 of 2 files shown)


toddburnside:
Switched to react-table.

PR #546, opened 2021-01-06
Save field changes on Enter. Do not submit form.

rpiaggio:
- Avoid PM Form triggering on `Enter` by rendering as `<.div`.
- No need for shadowing of search term in target editor now that `FormInputEV` sets the value in the model on `Enter`.


toddburnside (approved):
LGTM

PR #552, opened 2021-01-12
Copy the structure of the target tree to the obs tree

cquiroz:
The target tree was changed in a previous PR to better support small devices. This PR brings an equivalent change for the obs tree.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -130,8 +130,11 @@ lazy val common = project
       scalaVersion,
       sbtVersion,
       git.gitHeadCommit,
-      BuildInfoKey.action("buildTime")(System.currentTimeMillis)
     ),
+    buildInfoKeys ++= {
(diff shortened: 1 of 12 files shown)


build.sbt
```diff
@@ -130,8 +130,11 @@ lazy val common = project
       scalaVersion,
       sbtVersion,
       git.gitHeadCommit,
-      BuildInfoKey.action("buildTime")(System.currentTimeMillis)
     ),
+    buildInfoKeys ++= {
```
cquiroz, line 134: We are getting extra compilation when e.g. the style files change. I tracked this to the `BuildInfo` that reads the time on every change.
With this change you can set an env variable `SBT_IGNORE_BUILDTIME` and it will set a fixed time instead

rpiaggio: Nice, thank you!

rpiaggio (approved):
LGTM

PR #567, opened 2021-01-14
Ch645 add observation notes tile

cquiroz:
This PR adds a placeholder for the observation notes tile.
It is mostly work on the observation layout. In particular it now supports maximizing and restoring a fixed size tile

screencast 2021-01-14 20-37-55

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -3,17 +3,30 @@
 
 package explore.components
 
+import cats.syntax.all._
 import explore.Icons
 import explore.components.ui.ExploreStyles
+import explore.model.enum.TileSizeState
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common._
+import react.common.implicits._
 import react.semanticui.collections.menu._
+import react.semanticui.elements.button.Button
 
 final case class TileButton(body: VdomNode)
 
-final case class Tile(title: String, movable: Boolean, back: Option[TileButton])
-    extends ReactPropsWithChildren[Tile](Tile.component)
+final case class Tile(
+  title:             String,
+  back:              Option[TileButton] = None,
+  canMinimize:       Boolean = false,
+  canMaximize:       Boolean = false,
+  state:             TileSizeState = TileSizeState.Normal,
+  sizeStateCallback: TileSizeState => Callback = _ => Callback.empty
+) extends ReactPropsWithChildren[Tile](Tile.component) {
+  def showMaximize: Boolean = canMaximize && state === TileSizeState.Minimized
+  def showMinimize: Boolean = canMinimize && state === TileSizeState.Normal
+}
 
 object Tile {
   type Props = Tile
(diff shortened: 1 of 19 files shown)


explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -35,46 +36,102 @@ import react.semanticui.elements.button.Button.ButtonProps
...
+      breakPointNote(BreakpointName.sm).composeLens(layoutItemHeight)
+    def notesHeightState(s: State): TileSizeState = notesHeight.getOption(s) match {
+      case Some(x) if x === 1 => TileSizeState.Minimized
+      case _                  => TileSizeState.Normal
+    }
+  }
+
+  type Props = ObsTabContents
   implicit val propsReuse: Reusability[Props] = Reusability.derive
+  implicit val stateReuse: Reusability[State] = Reusability.never
```
rpiaggio, line 127: Why can't we reuse the `State`?

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -3,17 +3,30 @@
...
 import react.common._
+import react.common.implicits._
 import react.semanticui.collections.menu._
+import react.semanticui.elements.button.Button
 
 final case class TileButton(body: VdomNode)
 
-final case class Tile(title: String, movable: Boolean, back: Option[TileButton])
-    extends ReactPropsWithChildren[Tile](Tile.component)
+final case class Tile(
```
rpiaggio, line 19: I understand that all our tiles are unmovable now, is that right?

explore/src/main/scala/explore/tabs/ObsTabContents.scala
cquiroz: Oh, I missed this. I had it as `never` to get it to compile but I need to make it properly derive

common/src/main/scala/explore/components/Tile.scala
cquiroz: I'd expect yes but we don't need to specify that in the tile but in the layout. The flag here was to put an icon to use as a handle
but that is not needed anymore

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -3,17 +3,30 @@
...
+  title:             String,
+  back:              Option[TileButton] = None,
+  canMinimize:       Boolean = false,
+  canMaximize:       Boolean = false,
+  state:             TileSizeState = TileSizeState.Normal,
+  sizeStateCallback: TileSizeState => Callback = _ => Callback.empty
```
rpiaggio, line 25: These two parameters can be conmbined into a `View` (and probably use `ViewF.fromState` at the call site).

rpiaggio (approved):
CSS linter is complaining, otherwise LGTM

common/src/main/scala/explore/components/Tile.scala
cquiroz: That's a good idea, let me try that

PR #568, opened 2021-01-15
Attempt to setup a user preferences backend with graphql and hasura

cquiroz:
I want to setup a user preferences db. It is a simple set of tables in postgres
to store things the user defines, like e.g. positions of movable elements
To avoid overhead this setups a hasura instance to provide a graphql interface.

I'm still learning all the moving parts and this PR is to get initial feedback, in particular of my use of `clue`

At the moment there are 2 tables defined, one with a list of user ids, and another with the width of some UI sections that we want to make persistent


common/src/main/scala/explore/AppMain.scala
```diff
@@ -115,6 +118,17 @@ trait AppMain extends IOApp {
...
     )
 
+    // Creates a "profile" for user preferences.
+    def createUserPrefs(vault: Option[UserVault], ctx: AppContext[IO]): IO[Unit] = {
+      implicit val db = ctx.clients.preferencesDB
+      (for {
+        u <- OptionT.fromOption[IO](vault)
+        i <- OptionT.pure[IO](u.user.id)
+        _ <- OptionT.liftF(UserInsertMutation.execute(Input(i.toString())))
+        _ <- OptionT.liftF(UserWidthsCreation.execute(Input(i.toString()), Input(10)))
```
cquiroz, line 128: I may remove this and initialize this only the first time there is a change

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -0,0 +1,51 @@
...
+          user_id
+          }
+        }
+    """
+  }
+
+  @GraphQL
+  object UserWidthsCreation extends GraphQLOperation[UserPreferencesDB] {
+    val document = """
+        mutation initial_area_width($userId: String, $width: Int) {
```
cquiroz, line 34: If I understand correctly, I maybe able to replace this by a single `WidthInsertInput` field though I'm not yet able to do so.

common/src/main/resources/conf/development.conf.json
```diff
@@ -1,7 +1,7 @@
 {
   "environment": "DEVELOPMENT",
   "odbURI": "wss://odb.gpp.lucuma.xyz/ws",
-  "exploreDBURI": "wss://explore-hasura.herokuapp.com/v1/graphql",
+  "preferencesDBURI": "wss://userprefs-hasura-staging.herokuapp.com/v1/graphql",
```
rpiaggio, line 4: Just a heads up not to forget to change `staging` and `production` conf files.

common/src/main/scala/explore/AppMain.scala
```diff
@@ -115,6 +118,17 @@ trait AppMain extends IOApp {
       }
     )
 
+    // Creates a "profile" for user preferences.
+    def createUserPrefs(vault: Option[UserVault], ctx: AppContext[IO]): IO[Unit] = {
+      implicit val db = ctx.clients.preferencesDB
```
rpiaggio, line 123: If you pass `AppContext[IO]` implicitly and `import explore.impicits._` then you don't need this line (the  `db` is implicitly derived from `ctx`).

common/src/main/scala/explore/AppMain.scala
```diff
@@ -123,6 +137,7 @@ trait AppMain extends IOApp {
       ctx       <- AppContext.from[IO](appConfig, reconnectionStrategy, pageUrl, IO.fromFuture)
       vault     <- ctx.sso.whoami
       _         <- AppCtx.initIn[IO](ctx)
+      _         <- createUserPrefs(vault, ctx).start
```
rpiaggio, line 140: Nice touch triggering this concurrently.

common/src/main/scala/explore/GraphQLSchemas.scala
```diff
@@ -64,4 +66,21 @@ object GraphQLSchemas {
         explore.model.encoders.siderealTargetEncoder
     }
   }
+
```
rpiaggio, line 69: Let's remove `object ExploreDB`.

common/src/main/scala/explore/GraphQLSchemas.scala
```diff
@@ -64,4 +66,21 @@ object GraphQLSchemas {
...
+  @GraphQLSchema(debug = false)
+  object UserPreferencesDB {
+    object Scalars {
+      type UserId        = User.Id
+      type ResizableArea = String
+    }
+
+    object Types {
+      type WidthInsertInput = ExploreResizableWidthInsertInput
+      type WidthInsertInputL = Array[ExploreResizableWidthInsertInput]
```
rpiaggio, line 79: Why `Array` instead of `List` ?

common/src/main/scala/explore/GraphQLSchemas.scala
```diff
@@ -64,4 +66,21 @@ object GraphQLSchemas {
...
+      implicit final val jsonEncoderWidthInsertInput: io.circe.Encoder[WidthInsertInput] =
+        ???
+      implicit final val jsonEncoderWidthInsertInputL: io.circe.Encoder[Array[WidthInsertInput]] =
+        ???
```
rpiaggio, line 83: `clue` macro should have defined `Encoder[ExploreResizableWidthInsertInput]` (in the `Implicits` object), so this may not be necessary.

common/src/main/scala/explore/GraphQLSchemas.scala
cquiroz: Just testing, either Array or List produce an exception during macro expansion

rpiaggio (approved):
LGTM!

common/src/main/scala/explore/GraphQLSchemas.scala
rpiaggio: Might be a bug, what's the exception?

cquiroz: I get a very long exception

```
[error] /Users/cquiroz/Projects/explore/common/src/main/scala/explore/common/UserPreferencesQueries.scala:31:4: java.util.NoSuchElementException: None.get
[error]  at scala.None$.get(Option.scala:627)
[error]  atscala.None$.get(Option.scala:626)
[error]  atclue.macros.GraphQLImpl.resolveVariables(GraphQL.scala:236)
[error]  atclue.macros.GraphQLImpl.$anonfun$addVars$1(GraphQL.scala:273)
[error]  atscala.Function$.$anonfun$chain$2(Function.scala:23)
[error]  atscala.collection.LinearSeqOps.foldLeft(LinearSeq.scala:169)
[error]  atscala.collection.LinearSeqOps.foldLeft$(LinearSeq.scala:165)
[error]  atscala.collection.immutable.List.foldLeft(List.scala:79)
[error]  atscala.Function$.$anonfun$chain$1(Function.scala:23)
[error]  atclue.macros.GraphQLImpl.$anonfun$expand$6(GraphQL.scala:462)
[error]  atcats.effect.internals.IORunLoop$.step(IORunLoop.scala:241)
[error]  atcats.effect.IO.unsafeRunTimed(IO.scala:338)
[error]  atcats.effect.IO.unsafeRunSync(IO.scala:256)
[error]  atclue.macros.Macro.macroResolve(Macro.scala:19)
[error]  atclue.macros.Macro.macroResolve$(Macro.scala:17)
[error]  atclue.macros.GraphQLImpl.macroResolve(GraphQL.scala:27)
[error]  atclue.macros.GraphQLImpl.resolve(GraphQL.scala:31)
```
and more

common/src/main/scala/explore/common/UserPreferencesQueries.scala
cquiroz: Does it mean the document is sent when the macro is executed?

rpiaggio: When the query is executed.

common/src/main/scala/explore/GraphQLSchemas.scala
rpiaggio: Thanks, I'll take a look

rpiaggio: I see the exception pops up during `@GraphQL` expansion, when resolving query variables. This is possible due to the fact that only types defined in the schema (or base types) can be used as query variables. The query is compiled and validated against the schema at compile time by literally looking for type definitions, and so type aliases can't be used.

There is also a bug in `clue` when reporting this kind of issues, which I will address.

cquiroz: Thanks for the explanation. I managed to make it work.

cquiroz:
Ok this is ready for a review. I want to summarize what this PR does.

The target and obs trees are resizables but the position is lost once you switch tabs or reload.
This PR will store the width and reuse it on startup.
This is a quite simple usage and it is mostly about testing the waters on storing user preferences

Note that we assume this is discardable data, if we fail to save we ignore it and failures to read will use a default value.

The values are per user and not synchronized in real time, if you have two instances open and resize on one you will not see the changes in the other
If you were to reload you'd see this.

toddburnside (approved):
LGTM

PR #575, opened 2021-01-26
Use scalajs-react-table for MagnitudesForm.

toddburnside:
[ch568]
Magnitudes table should look exactly the same as the previous version with the SemanticUI table.

The code for adding controls to columns is MUCH simpler than for ag-grid (ReactTableHelpers).

image

### DIFF
--- a/common/src/main/scala/explore/utils/ReactTableHelpers.scala
+++ b/common/src/main/scala/explore/utils/ReactTableHelpers.scala
@@ -1,3 +1,6 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
 package explore.utils
 
 import cats._
@@ -6,7 +9,7 @@ import eu.timepit.refined.types.string._
 import explore._
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
-import lucuma.core.util.{Display, Enumerated}
+import lucuma.core.util.{ Display, Enumerated }
 import lucuma.ui.forms._
 import lucuma.ui.optics._
 import monocle.Lens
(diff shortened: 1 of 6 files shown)


cquiroz:
Do you have a screenshot?

common/src/main/scala/explore/utils/ReactTableHelpers.scala
```diff
@@ -0,0 +1,120 @@
...
+import monocle.Lens
+import react.semanticui.elements.button.Button
+import reactST.reactTable.mod._
+
+import scalajs.js
+
+object ReactTableHelpers {
+
+  /**
+   * Create a AgGridColumn Builder for a column which edits the value in a FormInputEv.
```
cquiroz, line 24: `AgGridColumn`?

common/src/main/scala/explore/utils/ReactTableHelpers.scala
```diff
@@ -0,0 +1,120 @@
...
+        FormInputEV(id = newId,
+                    value = va.zoom(lens),
+                    validFormat = validFormat,
+                    changeAuditor = changeAuditor,
+                    disabled = disabled,
+                    modifiers = modifiers
+        )
+      }
+      .build
+      .cmapCtorProps[(CellProps[View[A], _]) with js.Object](_.cell.row.original)
```
cquiroz, line 53: `neat`

cquiroz (approved):
LGTM

common/src/main/scala/explore/utils/ReactTableHelpers.scala
toddburnside: Oops.

toddburnside:
> Do you have a screenshot?

I can post one, but it looks exactly the same as what is in staging/production.

common/src/main/scala/explore/utils/ReactTableHelpers.scala
toddburnside: Looks like I did too much copy/paste of documentation from the ag-grid implementation. I'll go through that.

PR #576, opened 2021-01-29
Use safari hack only on safari or it triggers a chrome bug

cquiroz:
To get 100vh to work in Safari iOS we used a custom safari property but it produced an issue in chrome

This fixes by making the css conditional to iOS


toddburnside (approved):
LGTM

PR #584, opened 2021-02-02
Update loglevel and highchart facade versions.

rpiaggio:


build.sbt
```diff
@@ -118,9 +118,6 @@ lazy val common = project
   .settings(commonSettings: _*)
   .settings(commonJsLibSettings: _*)
   .settings(
-    npmDependencies in Compile ++= Seq(
-      "loglevel" -> "1.6.8"
```
rpiaggio, line 122: This is imported transitively by the facade lib.

targeteditor/src/main/scala/explore/targeteditor/SkyPlot.scala
```diff
@@ -106,7 +106,7 @@ object SkyPlot {
           val millisSinceEpoch = instant.toEpochMilli.toDouble
 
           def point(value: Double): Chart.Data =
-            PointOptionsObject()
+            PointOptionsObject(js.undefined)
```
rpiaggio, line 109: For some reason, ST doesn't define a default value here.

cquiroz (approved):
LGTM

PR #585, opened 2021-02-03
Responsive target tile part one

cquiroz:
This is the first part of the work to get the observation tab to be more complete

It includes work to change the styles of the tiles as they move following:
https://philipwalton.com/articles/responsive-components-a-solution-to-the-container-queries-problem/

It also adds storing and reading the user layout on the db

[skip-ch]

### DIFF
--- a/common/src/main/scala/explore/common/UserPreferencesQueries.scala
+++ b/common/src/main/scala/explore/common/UserPreferencesQueries.scala
@@ -11,11 +11,17 @@ import clue.macros.GraphQL
 import clue.{ GraphQLClient, GraphQLOperation }
 import crystal.react.implicits._
 import explore.GraphQLSchemas.UserPreferencesDB
-import explore.GraphQLSchemas.UserPreferencesDB.Types._
 import explore.model.ResizableSection
+import explore.model.GridLayoutSection
+import explore.model.layout._
 import japgolly.scalajs.react.Callback
+import japgolly.scalajs.react.raw.JsNumber
 import lucuma.core.model.User
 import lucuma.core.util.Gid
+import clue.data.Input
+import react.common._
+import react.gridlayout.{ BreakpointName => _, _ }
+import scala.collection.immutable.SortedMap
 
 object UserPreferencesQueries {
 
@@ -60,6 +66,23 @@ object UserPreferencesQueries {
           }
         }
    """
+
+    def storeWidthPreference[F[_]: Effect](
+      userId:  Option[User.Id],
+      section: ResizableSection,
+      width:   Int
+    )(implicit
+      cl:      GraphQLClient[F, UserPreferencesDB]
+    ): Callback =
+      userId.map { i =>
+        UserWidthsCreation
+          .execute[F](
+            WidthUpsertInput(i, section, width)
+          )
+          .attempt
+          .runAsyncAndForgetCB
+      }.getOrEmpty
+
   }
 
   /**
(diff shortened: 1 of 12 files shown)


common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -60,6 +66,23 @@ object UserPreferencesQueries {
...
         }
    """
+
+    def storeWidthPreference[F[_]: Effect](
+      userId:  Option[User.Id],
+      section: ResizableSection,
+      width:   Int
+    )(implicit
+      cl:      GraphQLClient[F, UserPreferencesDB]
+    ): Callback =
```
rpiaggio, line 76: Just a matter of having a consistent style: in other `*Queries` objects, there's nothing react specific, with the methods returning `IO`s, and the conversion to `Callback`s being done on the components that call the queries.

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -100,19 +123,105 @@ object UserPreferencesQueries {
...
+      userId:       Option[User.Id],
+      area:         GridLayoutSection,
+      defaultValue: LayoutsMap
+    )(implicit cl:  GraphQLClient[F, UserPreferencesDB]): F[LayoutsMap] =
+      (for {
+        uid <- OptionT.fromOption[F](userId)
+        c   <-
+          OptionT.pure(
+            GridLayoutPositionsBoolExp(user_id =
+              Input(StringComparisonExp(Input(Gid[User.Id].show(uid))))
```
rpiaggio, line 169: `uid.show` ?

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -100,19 +123,105 @@ object UserPreferencesQueries {
...
+
+    // Gets the layout of a section.
+    // This is coded to return a default in case
+    // there is no data or errors
+    def queryWithDefault[F[_]: MonadError[*[_], Throwable]](
+      userId:       Option[User.Id],
+      area:         GridLayoutSection,
+      defaultValue: LayoutsMap
+    )(implicit cl:  GraphQLClient[F, UserPreferencesDB]): F[LayoutsMap] =
+      (for {
```
rpiaggio, line 164: I'm not sure the monad transformer is really needed here.

Isn't the same achieved with `userId.fold(F.pure(defaultValue){uid => ...}` ?

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -100,19 +123,105 @@ object UserPreferencesQueries {
...
+    )(implicit
+      cl:      GraphQLClient[F, UserPreferencesDB]
+    ): Callback =
+      userId.map { uid =>
+        UserGridLayoutUpsert
+          .execute[F](
+            layouts.layouts.flatMap { bl =>
+              bl.layout.l.map(i =>
+                GridLayoutPositionsInsertInput(
+                  user_id = Input(Gid[User.Id].show(uid)),
```
rpiaggio, line 210: If you `import clue.data.syntax._` you can use `value.assign` instead of `Input(value)` (similar to `.some` in cats).

rpiaggio (approved):
Nice. LGTM, with some comments.

toddburnside (approved):
LGTM. Is there CSS for the new styles you are applying?

common/src/main/scala/explore/common/UserPreferencesQueries.scala
cquiroz: Ok, that sounds a good practice, these should return `IO`

common/src/main/scala/explore/common/UserPreferencesQueries.scala
cquiroz: Not completely, I want that errors also are handled and go to the default value

common/src/main/scala/explore/common/UserPreferencesQueries.scala
cquiroz: That's great let me fix it

cquiroz:
I'm not yet applying styles, I plan to concentrate on that on the next PR

PR #588, opened 2021-02-04
Some improvements to the user preferences queries

cquiroz:
Some improvements to the queries suggested by @rpiaggio and it should fix a bug when the db is empty for a user


toddburnside (approved):
LGTM

rpiaggio (approved):
LGTM

PR #589, opened 2021-02-05
Asterisms in target view tree

rpiaggio:
Allows:
 * Dragging observations between targets and asterisms.
 * Dragging targets into asterisms.
 * Removing targets from asterisms.
 * Create/delete asterisms.
 * Undo/redo of all of the above.
 * URLs for asterisms.
 * Auto expand target or observation if URL with observation assigned to them is invoked.

Pending (future PR):
 * Unassigned observations list.
 * Collapse target while dragging it into an asterism (we don't care about the target's observations in that case, and it may be confusing).


rpiaggio:
I'm deploying a review app for us to play with.

cquiroz (approved):
I skimmed the code and all seems right. I checked the test app and it looks great.

PR #590, opened 2021-02-05
Store fov on user preferences

cquiroz:
I've wanted for a while to make the fov on targets persistent and this PR does it.
The fov is stored on a per user/target basis. The changes are not reflected immediately as it maybe confusing if you have more than one window open

As usual this is treated as non critical data, any errors are ignored and a default is used:

screencast 2021-02-05 14-49-59

### DIFF
--- a/common/src/main/scala/explore/common/UserPreferencesQueries.scala
+++ b/common/src/main/scala/explore/common/UserPreferencesQueries.scala
@@ -243,21 +243,19 @@ object UserPreferencesQueries {
     )(implicit cl: GraphQLClient[F, UserPreferencesDB]): F[Angle] =
       (for {
         r <-
-          OptionT
-            .liftF[F, Option[Long]] {
-              query[F](uid.show, tid.show).map { r =>
-                r.lucuma_target_preferences_by_pk.map(_.fov)
-              }
+          query[F](uid.show, tid.show)
+            .map { r =>
+              r.lucuma_target_preferences_by_pk.map(_.fov)
             }
-            .handleErrorWith(_ => OptionT.none)
-      } yield r.map(Angle.fromMicroarcseconds)).value.map(_.flatten.getOrElse(defaultFov))
+            .handleError(_ => none)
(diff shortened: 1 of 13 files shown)


toddburnside (approved):
Nice.

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -221,4 +223,73 @@ object UserPreferencesQueries {
...
+      (for {
+        r <-
+          OptionT
+            .liftF[F, Option[Long]] {
+              query[F](uid.show, tid.show).map { r =>
+                r.lucuma_target_preferences_by_pk.map(_.fov)
+              }
+            }
+            .handleErrorWith(_ => OptionT.none)
+      } yield r.map(Angle.fromMicroarcseconds)).value.map(_.flatten.getOrElse(defaultFov))
```
rpiaggio, line 253: Maybe `r.flatMap` and then `getOrElse` directly on the `OptionT`?

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -221,4 +223,73 @@ object UserPreferencesQueries {
...
+  }
+
+  @GraphQL(debug = false)
+  object UserTargetPreferencesUpsert extends GraphQLOperation[UserPreferencesDB] {
+    val document =
+      """mutation target_preferences_upsert($objects: [lucuma_target_insert_input!]! = {}) {
+        insert_lucuma_target(objects: $objects, on_conflict: {constraint: lucuma_target_pkey, update_columns: target_id}) {
+          affected_rows
+        }
+      }"""
```
rpiaggio, line 263: Just FYI, for cases like this, where the mutation expects an `Array` but we are always going to call it with a single object, we can have the GraphQL mutation build the `Array`.

For example, here the mutation could be declared:
```
mutation target_preferences_upsert($object: lucuma_target_insert_input!) {
  insert_lucuma_target(objects: [$object], ...) {
    ...
  } 
}
```

rpiaggio (approved):
LGTM

PR #593, opened 2021-02-08
Mag table

cquiroz:
This PR is about making the magnitudes table more responsive.
It uses a technique described [here](https://css-tricks.com/responsive-data-tables/) to move the headers to the side to make it easier to read on narrow screens.

Some screenshots:

screencast 2021-02-08 13-16-49

image

### DIFF
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -70,6 +70,10 @@ body.light-theme {
   color: var(--explore-accent-color);
 }
 
+tfoot {
+  background-color: var(--table-footer-background);
+}
+
 //----------
 // Main grid
 //----------
@@ -841,6 +845,26 @@ img.partner-split-flag {
   font-size: larger;
 }
 
+// For very small tables hide the headers and add fake "header" to the left of the row
+.tile-xs-width .explore-magnitudes-container {
+  .ui.table,
+  thead,
+  tbody,
+  th,
+  td,
+  tr {
+    // SUI already has an !ímportant
+    display: block !important;
+  }
+
+  // Hide table headers (but not display: none;, for accessibility)
+  thead tr {
+    position: absolute;
+    top: -9999px;
+    left: -9999px;
+  }
+}
+
 .ui.table.partner-splits-editor-table {
   th:last-child,
   td:last-child,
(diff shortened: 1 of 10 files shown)


targeteditor/src/main/scala/explore/targeteditor/MagnitudeForm.scala
```diff
@@ -174,16 +202,21 @@ object MagnitudeForm {
...
+            <.div(
+              <.label("Magnitudes"),
+              Segment(attached = SegmentAttached.Attached,
+                      compact = true,
+                      clazz = ExploreStyles.MagnitudesTableContainer
+              )(
+                tableMaker.makeTable(
+                  options = options,
+                  data = props.magnitudes.toListOfViews(_.band).toJSArray,
+                  headerCellFn = Some(c => <.th(Column.all.get(c.id.toString).foldMap(_.label))),
```
toddburnside, line 214: If you use `.setHeader` on the column definitions above you can should be able to just use `TableMaker.basicHeaderCellFn()`.  That will also apply any properties necessary for react-table to the header cells. In this case it is only 'colspan` and `role`, but in other cases they can be significant.

I think this would also eliminate the need for the Column list, too.

common/src/main/resources/less/style.less
```diff
@@ -1027,3 +1051,74 @@ html {
...
+    padding-left: 0.25em;
+    white-space: nowrap;
+  }
+
+  td:first-child::before {
+    font-weight: normal;
+  }
+
+  // It is a bit unfortunate that we have to hardcode the column names here
+  td:nth-of-type(1)::before {
```
toddburnside, line 1090: Yeah, this is a bit of a bummer. But, it's amazing you could make the tables responsive at all. :)

toddburnside (approved):
Very cool. LGTM.

rpiaggio (approved):
Nice. LGTM

targeteditor/src/main/scala/explore/targeteditor/MagnitudeForm.scala
```diff
@@ -174,16 +180,24 @@ object MagnitudeForm {
...
+                      compact = true,
+                      clazz = ExploreStyles.MagnitudesTableContainer
+              )(
+                tableMaker.makeTable(
+                  options = options,
+                  data = props.magnitudes.toListOfViews(_.band).toJSArray,
+                  headerCellFn = Some(c =>
+                    TableMaker
+                      .basicHeaderCellFn(Css.Empty)(c)
+                  ),
```
toddburnside, line 195: `headerCellFn = Some(TableMaker.basicHeaderCellFn())` should also work. 

It looks like `scalajs-react-table` should use `Css.Empty` instead of `Css("")` for it's function default values. Not sure how I missed that.

toddburnside (approved):

PR #594, opened 2021-02-08
Visual improvements in Target View tree

rpiaggio:
A few visual improvements to the target view tree:

* Only show asterisms panel if there are asterisms.
* Only make targets draggable if there are asterisms.
* Show number of targets in asterisms.
* When dragging a target to an asterism:
  * Don't "move" it but instead "copy" it. In other words, the original target is still rendered in its place while dragging.
  * Show only a card with the target name. Don't show target's observations anymore.
  * Don't paint other targets while hovering over them.

**BEFORE:**

**AFTER:**

[skip-ch]

### DIFF
--- a/common/src/main/resources/less/style.less
+++ b/common/src/main/resources/less/style.less
@@ -424,6 +424,7 @@ body.light-theme {
   align-items: stretch;
   min-height: 0;
   height: 100%;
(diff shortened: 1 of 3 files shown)


cquiroz (approved):
LGTM

PR #600, opened 2021-02-11
Switch from SizeMe to ResizeDetector.

rpiaggio:
Also, resizer logic was moved to a `div` in `Routing`.

For the moment the facade is here but it will eventually be moved a standalone project.

(And cats update).

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -6,16 +6,16 @@ package explore.components
 import cats.syntax.all._
 import explore.Icons
 import explore.components.ui.ExploreStyles
-import explore.model.enum.TileSizeState
 import explore.components.ui.ExploreStyles._
+import explore.model.Constants
+import explore.model.enum.TileSizeState
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common._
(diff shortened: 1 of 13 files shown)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -48,9 +48,9 @@ object Tile {
       .builder[Props]
       .stateless
       .render_PC { (p, c) =>
-        SizeMe(monitorHeight = true) { s =>
+        ResizeDetector() { s =>
```
rpiaggio, line 51: ResizeDetector tracks height by default.

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -339,7 +330,10 @@ object ObsTabContents {
         }
       )
       .renderBackend[Backend]
-      .componentDidMount($ => $.backend.readWidthPreference *> $.backend.readLayoutPreference)
+      .componentDidMount($ =>
+        $.backend.readWidthPreference($.props.userId.get) *>
```
cquiroz, line 334: This is unrelated. Would it be possible to combine these into one call?

cquiroz (approved):
Looks great, and great work hunting this bug

PR #601, opened 2021-02-11
Avoid remount when switching between 1 or 2 panels

rpiaggio:
When a resize forced a switch between 1 and 2 panel mode, the right side was being remounted. This forced requeries of the DB and resetting the event subscriptions.

Setting a `key` + preserving the DOM structure seems to do the trick.


cquiroz (approved):
Oh this is interesting to know, great discovery

PR #603, opened 2021-02-13
Keep subscriptions upon SSO token refresh.

rpiaggio:
Also:
* `Terminating` connection status no longer exists in `clue`.
* Name clients for better logging in `clue`.
* Allocate more memory for Github actions build.
* Update Scala.js to `1.5.0`.

### DIFF
--- a/common/src/main/webpack/dev.webpack.config.js
+++ b/common/src/main/webpack/dev.webpack.config.js
@@ -47,7 +47,8 @@ const Web = Merge(
         rewrites: [
           { from: /^\/conf.json$/, to: '/conf/development.conf.json' },
         ]
-      }
+      },
+      https: true
     },
     plugins: [
       // Needed to enable HMR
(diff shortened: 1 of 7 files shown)


common/src/main/webpack/dev.webpack.config.js
```diff
@@ -47,7 +47,8 @@ const Web = Merge(
         rewrites: [
           { from: /^\/conf.json$/, to: '/conf/development.conf.json' },
         ]
-      }
+      },
+      https: true
```
cquiroz, line 51: I guess you get a cert warning on the browser but that's ok, we could consider making an explore CA to issue cents for local development

cquiroz (approved):
Great, tHanks

common/src/main/webpack/dev.webpack.config.js
rpiaggio: It bothers the first time only. Unfortunately, there's not way to configure WDS to serve both http and https. I guess we could issue a cert for `local.lucuma.xyz`, yes.

PR #605, opened 2021-02-15
Combine the calls to get the obs tab width and layout

cquiroz:
We are doing two calls to hasura to get the tab width and layout positions
This PR combines them into a single call

### DIFF
--- a/common/src/main/scala/explore/common/UserPreferencesQueries.scala
+++ b/common/src/main/scala/explore/common/UserPreferencesQueries.scala
@@ -175,17 +175,13 @@ object UserPreferencesQueries {
           OptionT
             .liftF[F, (Int, SortedMap[react.gridlayout.BreakpointName, (Int, Int, Layout)])] {
               query[F](uid.show, c, resizableArea.value).map { r =>
-                val w = r.explore_resizable_width_by_pk.map(_.width).getOrElse(defaultValue._1)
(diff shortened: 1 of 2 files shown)


common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -165,20 +173,24 @@ object UserPreferencesQueries {
           )
         r   <-
           OptionT
-            .liftF[F, SortedMap[react.gridlayout.BreakpointName, (Int, Int, Layout)]] {
-              query[F](c).map { r =>
+            .liftF[F, (Int, SortedMap[react.gridlayout.BreakpointName, (Int, Int, Layout)])] {
+              query[F](uid.show, c, resizableArea.value).map { r =>
+                val w = r.explore_resizable_width_by_pk.map(_.width).getOrElse(defaultValue._1)
                 if (r.grid_layout_positions.isEmpty) defaultValue
```
rpiaggio, line 179: When there's a defined `width` but not a `layout`, this seems to override the defined `witdh` with the default. Is this the intended behavior?

cquiroz: Thank you for pointing this out, I added a fix for this

rpiaggio (approved):
LGTM

PR #606, opened 2021-02-16
Fix for non int dimensions

cquiroz:
`ResizeDetector` can produce non int values and this was throwing an exception


toddburnside (approved):
LGTM

PR #614, opened 2021-02-17
Setup the target tile on the observations tab

cquiroz:
This PR aims to set the target tile on the observation tab. It will now read the target (asterisms not supported) of an obs and display part of the target tab inside the tile

It also contains some explorations on how to make the target tab more responsive.

### DIFF
--- a/common/src/main/scala/explore/components/ResponsiveComponent.scala
+++ b/common/src/main/scala/explore/components/ResponsiveComponent.scala
@@ -1,0 +1,41 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.components
+
+import cats.syntax.all._
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import react.common._
+import react.common.implicits._
+import react.resizeDetector.ResizeDetector
+
+final case class ResponsiveComponent(
+  widthBreakpoints:  List[(Int, Css)],
+  heightBreakpoints: List[(Int, Css)] = Nil,
+  clazz:             Css = Css.Empty
+) extends ReactPropsWithChildren[ResponsiveComponent](ResponsiveComponent.component)
+
+object ResponsiveComponent {
+  type Props = ResponsiveComponent
+
+  // Explicitly never reuse as we are not considering the content
+  implicit val propsReuse: Reusability[ResponsiveComponent] = Reusability.never
+
+  val component =
+    ScalaComponent
+      .builder[Props]
+      .stateless
+      .render_PC { (p, c) =>
+        ResizeDetector() { s =>
+          val heightClass =
+            p.heightBreakpoints.findLast(_._1 < s.height.orEmpty).foldMap(_._2)
+          val widthClass  =
+            p.widthBreakpoints.findLast(_._1 < s.width.orEmpty).foldMap(_._2)
+          <.div(p.clazz |+| widthClass |+| heightClass, s.targetRef, c)
+        }
+      }
+      .configure(Reusability.shouldComponentUpdate)
+      .build
+
+}
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -7,7 +7,6 @@ import cats.syntax.all._
 import explore.Icons
 import explore.components.ui.ExploreStyles
 import explore.components.ui.ExploreStyles._
-import explore.model.Constants
 import explore.model.enum.TileSizeState
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
@@ -16,6 +15,7 @@ import react.common.implicits._
 import react.resizeDetector.ResizeDetector
 import react.semanticui.collections.menu._
 import react.semanticui.elements.button.Button
+import explore.model.Constants
 
 final case class TileButton(body: VdomNode)
 
@@ -36,7 +36,10 @@ object Tile {
 
   // Explicitly never reuse as we are not considering the content
   implicit val propsReuse: Reusability[Tile] = Reusability.never
-  val defaultBreakpoints                     =
+  val heightBreakpoints                      =
+    List((200, TileXSH), (700 -> TileSMH), (1024 -> TileMDH))
+
+  val widthBreakpoints                       =
     List((300                            -> TileXSW),
          (Constants.TwoPanelCutoff.toInt -> TileSMW),
          (768                            -> TileMDW),
@@ -49,8 +52,6 @@ object Tile {
       .stateless
       .render_PC { (p, c) =>
         ResizeDetector() { s =>
-          val widthClass     =
-            defaultBreakpoints.findLast(_._1 < s.width.getOrElse(0)).map(_._2).getOrElse(TileSMW)
           val maximizeButton =
             Button(
               as = <.a,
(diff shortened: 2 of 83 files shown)


common/src/main/scala/explore/components/ResponsiveComponent.scala
```diff
@@ -0,0 +1,41 @@
...
+package explore.components
+
+import cats.syntax.all._
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import react.common._
+import react.common.implicits._
+import react.resizeDetector.ResizeDetector
+
+final case class ResponsiveComponent(
```
rpiaggio, line 13: Nice!

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -49,8 +52,6 @@ object Tile {
       .stateless
       .render_PC { (p, c) =>
         ResizeDetector() { s =>
```
rpiaggio, line 54: Should we remove this line, since `ResizeDetector` is now invoked within `ResponsiveComponent` below?

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -255,24 +273,27 @@ object ObsTabContents {
...
-                  NonEmptyList.of(TargetEditSubscription.subscribe[IO](targetId))
-                ) { targetOpt =>
-                  <.div(
-                    <.div(
-                      (props.userId.get, targetOpt.get).mapN { case (uid, _) =>
-                        val stateView = ViewF.fromState[IO]($).zoom(State.options)
-                        TargetBody(uid, targetId, targetOpt.zoom(_.get)(f => _.map(f)), stateView)
+              <.div(
+                ^.key := s"target-$targetId",
+                Tile("Target")(
```
rpiaggio, line 278: You can do `Tile("Target").withKey(s"target-$targetId")(` to avoid the extra `div`. Same applies to the "Notes" `Tile` above.

Also, can we remove the outer `div`? (the one with `key := "target"`)

common/src/main/scala/explore/components/Tile.scala
cquiroz: Indeed I removed it but I think it came back on the rebase

explore/src/main/scala/explore/tabs/ObsTabContents.scala
cquiroz: Nice tip, we can't remove the outer one as react grid layout needs a fixed id

cquiroz: Maybe LiveRenderMod should take a key?

model/src/main/scala/explore/model/ObsSummary.scala
```diff
@@ -0,0 +1,71 @@
...
+  duration:          Duration = Duration.of(93, ChronoUnit.MINUTES),
+  observationTarget: Option[Either[Asterism.Id, Target.Id]]
+)
+
+object ObsSummary {
+  implicit val eqObsSummary: Eq[ObsSummary] = Eq.by(x => (x.id, x.name, x.observationTarget))
+
+  implicit val obsSummaryEncoder: Encoder[ObsSummary] = new Encoder[ObsSummary] {
+    final def apply(c: ObsSummary): Json = {
+      val common = Json.obj(("id", c.id.asJson), ("name", c.name.asJson))
```
rpiaggio, line 37: I think we can unify the logic a bit (and avoid intermediate allocations and the `deepMerge`) with something like:

``` scala
  val common = JsonObject(...)
  (c.observationTarget match {
    case None => common
    case Some(t) => common.add("observationTarget", t match {
      case Right(tid) => Json.obj("target_id" -> tid.asJson)
      case Left(aid) => Json.obj("asterism_id" -> aid.asJson)
    })
  }).asJson
```

Would it work if we add to `common` something like: `"observationTarget" ->

model/src/main/scala/explore/model/ObsSummary.scala
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  implicit val obsTargetDecoder: Decoder[Either[Asterism.Id, Target.Id]] =
+    new Decoder[Either[Asterism.Id, Target.Id]] {
+      final def apply(c: HCursor): Decoder.Result[Either[Asterism.Id, Target.Id]] =
+        for {
+          aid <- c.downField("asterism_id").as[Option[Asterism.Id]]
+          tid <- c.downField("target_id").as[Option[Target.Id]]
+        } yield (aid, tid) match {
+          case (Some(aid), _)    => Left(aid)
```
rpiaggio, line 56: Do we want to make `None` explicit in the 2nd position in this case? Otherwise, if both are defined it will match here instead of producing an error.

model/src/main/scala/explore/model/ObsSummary.scala
```diff
@@ -0,0 +1,71 @@
...
+          case (None, Some(tid)) => Right(tid)
+          case _                 => sys.error("graphql schema doesn't allow this")
```
rpiaggio, line 58: We can return a `DecodingFailure` instead of calling `sys.error`.

rpiaggio: Actually, if we remove this method from here, won't the `eitherDecoder` you added to `decoders` be picked up if we add the `import`?

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -410,7 +410,10 @@ object TargetObsList {
 
       def renderObsBadge(obs: ObsAttached): TagMod =
         ObsBadge(
-          ObsSummary(obs.id, obs.name.orEmpty),
+          ObsSummary(id = obs.id,
+                     name = obs.name,
+                     observationTarget = None
+          ), // FIXME Add the target id
```
rpiaggio, line 416: Regarding the comment: I don't think we want the target id here since the badge is already nested within the target (or asterism). (According to the spec, we have to render the observation badge differently in different views)

rpiaggio (changes requested):
A couple of things commented inline.

rpiaggio (approved):

explore/src/main/scala/explore/tabs/ObsTabContents.scala
rpiaggio: It does, since it uses the same mechanism in `ReactProps`. However, the `.withKey` has to go after the render function. Eg:

``` scala
  LiveQueryRenderMod[ObservationDB,
                                     TargetEditQuery.Data,
                                     Option[TargetEditQuery.Data.Target]
                  ](
                    TargetEditQuery.query(targetId),
                    _.target,
                    NonEmptyList.of(TargetEditSubscription.subscribe[IO](targetId))
                  ) { targetOpt =>
                       ...
                  }.withKey("key")
```

cquiroz: This works fine on the `LiveQueryRenderMod` element but I'm having troubles with the one on `Tile`. I'll address that in another PR

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
cquiroz: 👍

cquiroz:
I added the suggested fixes and also ran OrganizeImports but in a separate commit for easier review

rpiaggio:
Looks great. Thank you and thank you for organizing the imports.

PR #615, opened 2021-02-18
Golden tests

cquiroz:
I have been thinking to check if golden tests could be useful and the query with obs seems a good candidate. I wish more samples were tested tested though


toddburnside:
This seems like a good idea. I'm not sure if there are any real downsides.

model/jvm/src/test/resources/explore/model/ObsSummary-hA4vBbb7oekKaTc747Dj3SDfXQglCjFai3vhoR_a9FC=.json
```diff
@@ -0,0 +1,7 @@
+{
+  "id" : "o-1",
+  "name" : "噾䑩雌涭坻啰཈⭨璞꒺欝濓년델퀻痤▥썧誖됧᫼蒞⩂벶ᚻ伺Ȇ쬟쌕畤믑䑄っẾ잦㣤،塅热蹯묍殐୩픃즨㌩ꤽ廙蹽ꓣ첡ﾟආ⤾垓亂뷸ᤪⱵ椼鳩䣇뉁⻵",
+  "observationTarget" : {
+    "target_id" : "t-160d2501da90d1db"
```
cquiroz, line 5: So there is only one sample so no asterism. I don't know if you can run it somehow to have multiple samples

rpiaggio:
This seems like a good idea. It will give us an extra safety net.

model/jvm/src/test/resources/explore/model/ObsSummary-hA4vBbb7oekKaTc747Dj3SDfXQglCjFai3vhoR_a9FC=.json
rpiaggio: We can pass a `count: Int` to `GoldenCodecTests` indicating the number of samples we want. If we add enough, we should have both kinds of observations generated.

cquiroz: Great find, let me try it

PR #619, opened 2021-02-19
Add verticalMultiline.atDefnSite to .scalafmt.config

toddburnside:
This resulted in a lot of minor formatting changes.

There was only one similar to what I had in tmp-api (Restorer.scala), which I think is a big improvement. 

Other than that, I think some of the changes are a little better, some are a little worse, and some are just different. I have no strong opinion about whether it is an overall improvement or if we should make the change.

### DIFF
--- a/common/src/main/scala/explore/common/UserPreferencesQueries.scala
+++ b/common/src/main/scala/explore/common/UserPreferencesQueries.scala
@@ -106,7 +106,8 @@ object UserPreferencesQueries {
       userId:       Option[User.Id],
       area:         ResizableSection,
       defaultValue: Int
-    )(implicit cl:  GraphQLClient[F, UserPreferencesDB]
+    )(implicit
+      cl:           GraphQLClient[F, UserPreferencesDB]
     ): F[Int] =
       (for {
         uid <- OptionT.fromOption[F](userId)
@@ -167,7 +168,8 @@ object UserPreferencesQueries {
       layoutSection: GridLayoutSection,
       resizableArea: ResizableSection,
       defaultValue:  (Int, LayoutsMap)
-    )(implicit cl:   GraphQLClient[F, UserPreferencesDB]
+    )(implicit
+      cl:            GraphQLClient[F, UserPreferencesDB]
     ): F[(Int, LayoutsMap)] =
       (for {
         uid <- OptionT.fromOption[F](userId)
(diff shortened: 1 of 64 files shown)


cquiroz (approved):
I think I like it

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -106,7 +106,8 @@ object UserPreferencesQueries {
       userId:       Option[User.Id],
       area:         ResizableSection,
       defaultValue: Int
-    )(implicit cl:  GraphQLClient[F, UserPreferencesDB]): F[Int] =
+    )(implicit cl:  GraphQLClient[F, UserPreferencesDB]
+    ): F[Int] =
```
rpiaggio, line 110: Ouch

toddburnside: Yeah, this falls into the "some are a little worse" category. I do like the other places where it puts the final case class closing paren on the same line as the last parameter, though. 🤷

rpiaggio (commented):
I do like:
* That all parameters in all parameter lists are aligned.

I don't like:
* That the closing parenthesis is moved to the line of the last parameter. When it was in the next line, it nicely delimited visually the end of the parameter block.

I definitely don't like:
* That single-parameter lists are not closed in the same line.

In other words, it seems the closing parenthesis at the end of single- and multi-parameter lists behave exactly the reverse as I would expect (which would be same line for single- and next line for multi-).

toddburnside:
>     * That the closing parenthesis is moved to the line of the last parameter. When it was in the next line, it nicely delimited visually the end of the parameter block.

I agree that it is a little less clear with `continuationIndet.defnSite = 2`. I think that is why the default for that parameter is 4 - to visually delimit the end of the parameter block. It would also align the parameters up with `extends`. But it's all a matter of taste.

Maybe we should just stick with what we have and use `// format: off` for egregious cases?

cquiroz:
I really wish dotty comes with a std conf for scalafmt that everybody uses

toddburnside:
> I really wish dotty comes with a std conf for scalafmt that everybody uses

😆 That is what the default configuration for scalafmt was supposed to be - but nobody could agree.

rpiaggio:
Sorry for the long silence. Maybe we can play a bit more with the other `verticalMultiline` options? Let me see if I can get to a configuration that addresses all our woes and if that happens I'll send a PR to this PR.

toddburnside:
> Sorry for the long silence. Maybe we can play a bit more with the other `verticalMultiline` options? Let me see if I can get to a configuration that addresses all our woes and if that happens I'll send a PR to this PR.

No worries. As I said above, I'm not strongly attached to this. I'm just glad we have something.

toddburnside:
Will be replaced with a different PR.

PR #621, opened 2021-02-23
Creating new observation

cquiroz:
This PR enables creating observations on the UI entering the name

screencast 2021-02-23 12-13-31

### DIFF
--- a/common/src/main/scala/explore/components/InputModal.scala
+++ b/common/src/main/scala/explore/components/InputModal.scala
@@ -16,12 +16,13 @@ import japgolly.scalajs.react.vdom.html_<^._
 import lucuma.ui.forms.FormInputEV
 import monocle.macros.Lenses
 import react.common._
-import react.semanticui.collections.form.Form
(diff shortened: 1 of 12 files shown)


common/src/main/scala/explore/components/InputModal.scala
```diff
@@ -0,0 +1,96 @@
...
+) extends ReactProps[InputModal](InputModal.component)
+
+object InputModal {
+  type Props = InputModal
+
+  @Lenses
+  final case class State(inputValue: String)
+
+  object State {
+    val Zero = State("")
```
toddburnside, line 43: Is this used?

toddburnside (approved):
LGTM

common/src/main/scala/explore/components/InputModal.scala
cquiroz: Good catch, at some moment state had more fields and unused Zero to initialize but I could remove it now

PR #624, opened 2021-02-24
New clue. Wait for connections before render.

rpiaggio:
Update to new clue, which changes logic a bit and renames things.

Part of the change is that now we wait for web socket clients to be connected and initialized before rendering things that use them.

### DIFF
--- a/common/src/main/scala/explore/components/state/ConnectionManager.scala
+++ b/common/src/main/scala/explore/components/state/ConnectionManager.scala
@@ -5,8 +5,6 @@ package explore.components.state
 
 import cats.effect.IO
 import cats.syntax.all._
-import clue.TerminateOptions
-import clue.WebSocketCloseParams
 import crystal.react.implicits._
 import eu.timepit.refined.cats._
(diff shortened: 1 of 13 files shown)


common/src/main/scala/explore/components/state/ConnectionManager.scala
```diff
@@ -17,37 +15,48 @@ import io.circe.syntax._
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common.ReactProps
+import react.semanticui.elements.icon.Icon
+import react.semanticui.sizes._
 
-final case class ConnectionManager(ssoToken: NonEmptyString)
-    extends ReactProps[ConnectionManager](ConnectionManager.component)
+final case class ConnectionManager(ssoToken: NonEmptyString, onConnect: IO[Unit])(
```
cquiroz, line 21: Great that we move the onINit here and make it more generic

cquiroz (approved):
LGTM

toddburnside (approved):
👍

PR #629, opened 2021-02-25
Some more format options.

rpiaggio:
I haven't found a way to keep `(parameter)` in one line when it's a secondary single-parameter group, and I'm not thrilled about them being split.

However, I find this formatting easier to read with the dangling parens.

What do you think?


toddburnside (approved):
Agree that having the line breaks for a single secondary parameter is not ideal, but I'm fine with this. It seems like the second parameter set is tough for scalafmt.

PR #631, opened 2021-02-26
Support deletion of observations

cquiroz:
This PR adds the ability to delete observations. It also fixes a few UI details when the obs list is
empty and handles updating the local model with the remote

https://user-images.githubusercontent.com/3615303/109335051-407b6200-7840-11eb-859a-980465324660.mp4


common/src/main/scala/explore/Icons.scala
```diff
@@ -39,4 +41,5 @@ object Icons {
   val Maximize            = Icon(className = "expand alt")
   val Checkmark           = Icon("checkmark")
   val Remove              = Icon("remove")
+  val Trash               = Delete.size(Small).fitted(true).clazz(ExploreStyles.TrashIcon)
```
rpiaggio, line 44: Good idea

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
```diff
@@ -43,16 +50,29 @@ object ObsBadge {
...
-                  case NameAndConf        => obs.name.orEmpty
-                  case ConfAndConstraints => obs.conf
-                }
+                <.div(
+                  ExploreStyles.ObservationCardHeader,
+                  props.layout match {
+                    case NameAndConf        => obs.name.orEmpty
+                    case ConfAndConstraints => obs.conf
+                  },
+                  deleteButton
```
rpiaggio, line 74: Is the delete button showing in observations badges in the target view? Do we want that?

observationtree/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -41,19 +45,50 @@ object ObsList {
...
   protected val component =
     ScalaComponent
       .builder[Props]
       .render_P { props =>
+        def deleteLocal(id: Observation.Id, focus: Option[ObsSummary]) =
+          props.observations.mod(l => l.filterNot(_.id === id)) *>
+            (focus match {
+              case Some(s) => props.focused.set(FocusedObs(s.id).some)
+              case None    => props.focused.set(none)
+            })
```
rpiaggio, line 71: Unify both cases in `props.focused.set(focus.map(s => FocusedObs(s.id)))` ?

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
cquiroz: At the moment yes, it is a good question, I'd say in general yes but it isn't wired at the moment

cquiroz: I added a change that will hide the delete button on the targets tree

observationtree/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -41,19 +45,50 @@ object ObsList {
...
+
+        def createLocal(name: String): IO[Observation.Id] =
+          IO(Observation.Id(PosLong.unsafeFrom(Random.nextInt().abs.toLong + 1))).flatTap { oid =>
+            props.observations.mod(l =>
+              (ObsSummary(oid, name.some, observationTarget = none) :: l).reverse
```
rpiaggio, line 73: This will reverse the existing list. I would expect this to cause some flicker between the local change and the new list to come from the server. Maybe we can just append the new element? Both operations are `O(n)` anyway. (Or do 2 reversals)

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
```diff
@@ -1,23 +1,30 @@
...
+import react.semanticui.elements.button.Button
+import react.semanticui.sizes._
 import react.semanticui.views.card._
 
 final case class ObsBadge(
   obs:      ObsSummary,
   layout:   ObsBadge.Layout,
-  selected: Boolean = false
+  selected: Boolean = false,
+  deleteCB: Observation.Id => Callback
```
rpiaggio: Would it be OK to expose `IO` instead of `Callback`? Just to use the same style throughout.

rpiaggio: Would it be OK to expose `IO` instead of `Callback`? Just to use the same style throughout.

observationtree/src/main/scala/explore/observationtree/ObsList.scala
toddburnside: Won't the position of the new element depend on the sorting order? Is it always in chronological order of when they were added at the moment?

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
cquiroz: I thought about that and I was tempting to do `id => F[Unit]` rather than `IO` but as we know type params on `Props` require some hacks downstream.
I guess fixing to IO would be fine

observationtree/src/main/scala/explore/observationtree/ObsList.scala
cquiroz: Right we need to reverse the inner list.
@toddburnside This mimics what the server will do which is to append to the end

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
cquiroz: Slightly related, I left out `deleteCB` from `Reusability` but it should be possible using the `Reuse` mechanism.
Would that work on `IO` as well?

rpiaggio: If that's `scalajs-react`'s `Reusable`, then I would expect it to work, since it's generic. However, I experimented with this approach a bit and found it added complexity at the call site, since we have to build the `Reusable`. In this case at least, since the delete function won't really change, I think it's enough to leave out the function out of the reusability.

rpiaggio (approved):
LGTM

PR #632, opened 2021-02-26
Retry whoami and handle app init errors.

rpiaggio:
Using `cats-retry` for this. Maybe we want to use it in other cases too. For the moment I only added it to `whoami` calls, I'm not sure which of the other ones are idempotent.


cquiroz (approved):
LGTM

PR #633, opened 2021-03-02
Modify queries for addition of paging to API

toddburnside:


rpiaggio (approved):
LGTM

swalker2m:
At some point (perhaps when we switch to a real production version), the server should probably limit the page size.  In that case of course, you may only receive a subset of the requested data.

PR #634, opened 2021-03-02
[ch-718] Add undo to the observation tree.

rpiaggio:


observationtree/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -44,48 +53,118 @@ object ObsList {
...
+      c:                                         TransactionalClient[F, ObservationDB]
+    ): F[Unit] =
+      ObsQueries.ProgramDeleteObservation.execute[F](id).void
+
+    private def setObsWithIndex(
+      observations:       View[ObservationList],
+      focused:            View[Option[Focused]],
+      obsId:              Observation.Id,
+      obsWithIndexSetter: Adjuster[ObservationList, obsListMod.ElemWithIndex],
+      nextToFoucs:        Option[ObsSummary]
```
toddburnside, line 89: Typo. Unless "Foucs" is a thing is Spanish.

toddburnside (approved):
LGTM

observationtree/src/main/scala/explore/observationtree/ObsList.scala
rpiaggio: Good catch, thank you

PR #635, opened 2021-03-02
[Ch-726] Fix concurrency issues in undoer.

rpiaggio:
We were waiting for the action to complete before updating the stacks. We now update the stacks before invoking the action, which preserves the order even if the actions complete in an order different than the invocation one.


toddburnside (approved):
LGTM

PR #636, opened 2021-03-02
Unify styles between observation and target trees

rpiaggio:
This is just a refactor. It brings the query executions from the `*Queries` object into the component, and passes the GraphQL client implicitly to the invoker methods instead of calling `AppCtx` in each of them. Also, handler methods now return `IO[Unit]` instead of `Callback` and `randomInt` is now called within `IO`.

### DIFF
--- a/observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
+++ b/observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
@@ -246,9 +246,9 @@ object TargetObsList {
           .mod(targetWithIndexSetter.set(targetWithIndex)) >>
           // 2) Send mutation & adjust focus
           targetWithIndex.fold(
-            focused.set(nextToFoucs.map(f => Focused.FocusedTarget(f.id))) *> removeTarget(targetId)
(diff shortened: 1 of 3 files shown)


observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -543,7 +626,7 @@ object TargetObsList {
                             clazz = ExploreStyles.DeleteButton |+| ExploreStyles.JustifyRight,
                             onClickE = (e: ReactMouseEvent, _: ButtonProps) =>
                               e.stopPropagationCB *>
-                                deleteTarget(targetId, undoCtx.setter, focusOnDelete)
+                                deleteTarget(targetId, undoCtx.setter, focusOnDelete).runAsyncCB
```
toddburnside, line 629: The above is the last remaining `*>` if you want to change  it to `>>` for consistency.

toddburnside (approved):
LGTM

observationtree/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: :+1: thank you

PR #649, opened 2021-03-11
Request names when creating targets and asterisms.

rpiaggio:
When a target is created, a catalog search is triggered.

I had to move some code from the `targeteditor` module into `common`, so that it's callable from the target tree.

Also, there's some (WIP) code to attempt to have reusable functions, similar to what `Reusable` provides but with different ergonomics (reusable values can be specified manually).


toddburnside (approved):
LGTM

PR #650, opened 2021-03-12
Set name/coords/magnitudes in single operation.

rpiaggio:
When renaming a target, two `set` operations were triggered: once by `InputEV`'s  `onBlur`, which changed the name, and another one setting the coords and magnitudes returned by the catalog search. This meant that 2 undo operations were required to return to the previous state.

This PR makes sure the search form's `InputEV` operates on a copy of the name, and then all 3 fields (name, coords, magnitudes) are set together atomically after the catalog search.


toddburnside (approved):
LGTM

PR #651, opened 2021-03-12
Constraint tree

toddburnside:
[ch-730] Does not include add/delete.


model/shared/src/main/scala/explore/model/ObsSummary.scala
```diff
@@ -63,12 +70,24 @@ object ObsSummary {
         }
     }
 
+  private case class CS(id: ConstraintSet.Id)
+
```
toddburnside, line 74: Maybe there's a better way to parse the nested constraint set -> id?
`"constraintSet": { "id": "c-2"}`
But, this is the best I could come up with.

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
```diff
@@ -0,0 +1,135 @@
...
+import scala.collection.immutable.SortedSet
+import scala.concurrent.duration._
+
+final case class ConstraintSetTabContents(
+  userId:      ViewOpt[User.Id],
+  focused:     View[Option[Focused]],
+  expandedIds: View[SortedSet[ConstraintSet.Id]],
+  size:        ResizeDetector.Dimensions
+) extends ReactProps[ConstraintSetTabContents](ConstraintSetTabContents.component) {
+  def isCsSelected: Boolean = focused.get.isDefined
```
rpiaggio, line 39: We might want to check if the focused element is a constraint set. Otherwise, if navigating here from another view while focused in another object type, this will be `true`.

model/jvm/src/test/resources/explore/model/ObsSummary-u9pv5ogAVP04lI9PYNyoP9w_EECItUaAtJl4c534TuK=.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id" : "o-361857f48a6a4d36",
+  "name" : "돚ꑰ쯁Ღ숅Ⱆ◖䎨ꃧ卙嵐蚨ƨ웥筅ᑤ헒ᖖဩ镾뎻蔳꺀콸뺗搌࠲濹痘྄ 圿অ菜ႇ疮⢓빀ﯱﷁ匊୴늑凇싒興谉X慮⨩㵿杻ꨁᬝ堊雀砑꣫㵃䄖賙ꈭ➂Ō敩ۜ",
+  "observationTarget" : {
+    "type" : "Target",
+    "target_id" : "t-1"
+  },
```
rpiaggio, line 7: Let's make sure that in the golden test cases we have all 3 cases for `observationTarget`: at leas one with an asterism, one with a target, and one with `null`. 

It has taken me a few tries to get this, maybe we want to increase the number of cases.

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
toddburnside: Hmmm. Yeah. You're right. I guess I just copied this from `TargetTabContents.scala` when I created the file. I guess we probably need to update `TargetTabContents.scala`, too?

model/shared/src/main/scala/explore/model/RootModel.scala
```diff
@@ -18,16 +20,17 @@ import monocle.Lens
...
-  focused:               Option[Focused] = none,
-  targetViewExpandedIds: TargetViewExpandedIds = TargetViewExpandedIds(),
-  searchingTarget:       Set[Target.Id] = HashSet.empty,
-  userSelectionMessage:  Option[NonEmptyString] = none
+  vault:                        Option[UserVault],
+  tabs:                         EnumZipper[AppTab],
+  focused:                      Option[Focused] = none,
+  targetViewExpandedIds:        TargetViewExpandedIds = TargetViewExpandedIds(),
+  searchingTarget:              Set[Target.Id] = HashSet.empty,
+  constraintSetViewExpandedIds: SortedSet[ConstraintSet.Id] = SortedSet.empty,
```
rpiaggio, line 32: What do you think of putting all the expanded ids in one class?

model/shared/src/test/scala/explore/model/arb/ArbFocused.scala
```diff
@@ -39,12 +48,16 @@ trait ArbFocused {
   implicit val focusedAsterismCogen: Cogen[Focused.FocusedAsterism] =
     Cogen[Asterism.Id].contramap(_.asterismId)
 
+  implicit val focusedConstraintSetCoget: Cogen[Focused.FocusedConstraintSet] =
```
rpiaggio, line 51: Typo

model/jvm/src/test/resources/explore/model/ObsSummary-u9pv5ogAVP04lI9PYNyoP9w_EECItUaAtJl4c534TuK=.json
toddburnside: Yeah. I see what you mean. now that we have constraint set, too, we'll probably want to increase the number of cases. I'll do that and make sure we have a mix.

model/shared/src/test/scala/explore/model/arb/ArbFocused.scala
```diff
@@ -30,6 +33,12 @@ trait ArbFocused {
   val focusedTargetGen: Gen[Focused.FocusedTarget] =
     arbitrary[Target.Id].map(FocusedTarget.apply)
 
+  val focusedAsterismGen: Gen[Focused.FocusedAsterism] =
+    arbitrary[Asterism.Id].map(FocusedAsterism.apply)
```
rpiaggio, line 37: Good catch. This was missing.

model/shared/src/main/scala/explore/model/RootModel.scala
toddburnside: I guess there's no reason not, too. And just call it expandedIds. I can do that.

model/shared/src/test/scala/explore/model/arb/ArbFocused.scala
rpiaggio: We are still missing `focusedAsterismGen` and `focusedConstraintSetGen` in line 27. Could you add those please?

toddburnside: Sure.

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsQueries.scala
```diff
@@ -0,0 +1,204 @@
...
+  @GraphQL
+  object ObservationEditSubscription extends GraphQLOperation[ObservationDB] {
+    val document = """
+      subscription {
+        observationEdit(programId: "p-2") {
+          id
+        }
+      }    
+    """
+  }
```
rpiaggio, line 131: This is the same as `ObsQueries.ProgramObservationsEditSubscription`. We should reuse that one. It's also duplicate in `TargetObsQueries`, let's remove it from there too.

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
```diff
@@ -0,0 +1,383 @@
...
+
+      val constraintSets        = props.constraintSetsWithObs.get.constraintSets
+      val constraintSetsWithIdx = constraintSets.toList.zipWithIndex
+
+      val unassignedObs = obsByConstraintSet.get(none).orEmpty
+
+      def renderObsBadge(obs: ObsSummary): TagMod =
+        ObsBadge(
+          obs,
+          ObsBadge.Layout.ConfAndConstraints,
```
rpiaggio, line 170: I think we want to show the observation's pointing name instead of the conf in this case.

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
```diff
@@ -0,0 +1,383 @@
...
+    def getDraggedStyle(style:   TagMod, snapshot: Draggable.StateSnapshot): TagMod =
+      if (!snapshot.isDragging)
+        TagMod.empty
+      else if (!snapshot.isDropAnimating)
+        style
+      else
+        TagMod(style, ^.transitionDuration := "0.001s")
+
+    def getListStyle(isDragging: Boolean): TagMod =
+      ExploreStyles.DraggingOver.when(isDragging)
```
rpiaggio, line 156: At some point we probably want to abstract away common logic with other views. We can do that later, but can we do it with these 2 methods (and maybe `renderObsBadge`) where the code is the same as in `TargetObsList`?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
rpiaggio: Yes, please.

model/shared/src/main/scala/explore/model/ObsSummary.scala
rpiaggio: I think it's OK.

model/shared/src/main/scala/explore/model/RootModel.scala
rpiaggio: :+1:

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
toddburnside: I've changed all of these to check the type, but we may want to think about this behavior for another PR. It seems to me like we might want Focused to be a product instead of a sum type.

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
```diff
@@ -0,0 +1,153 @@
...
+        }
+      )
+      .renderPS { ($, props, state) =>
+        AppCtx.withCtx { implicit ctx =>
+          val treeResize =
+            (_: ReactEvent, d: ResizeCallbackData) =>
+              ($.setStateLIn[IO](TwoPanelState.treeWidth)(d.size.width) *>
+                UserWidthsCreation
+                  .storeWidthPreference[IO](props.userId.get,
+                                            ResizableSection.ConstraintSetsTree,
```
cquiroz, line 81: Nice

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -40,13 +40,15 @@ import react.semanticui.sizes._
...
-  targetViewExpandedIds: View[TargetViewExpandedIds],
-  size:                  ResizeDetector.Dimensions
+  userId:      ViewOpt[User.Id],
+  focused:     View[Option[Focused]],
+  searching:   View[Set[Target.Id]],
+  expandedIds: View[ExpandedIds],
+  size:        ResizeDetector.Dimensions
 ) extends ReactProps[TargetTabContents](TargetTabContents.component) {
-  def isTargetSelected: Boolean = focused.get.isDefined
+  def isTargetSelected: Boolean = focused.get.collect { case Focused.FocusedTarget(_) =>
```
cquiroz, line 49: Can this be done with `exist`?

model/jvm/src/test/scala/explore/model/CirceGoldenSuite.scala
```diff
@@ -9,5 +9,5 @@ import io.circe.testing.ArbitraryInstances
 import munit.DisciplineSuite
 
 class CirceGoldenSuite extends DisciplineSuite with ArbitraryInstances {
-  checkAll("GoldenCodec[ObsSummary]", GoldenCodecTests[ObsSummary](5).goldenCodec)
+  checkAll("GoldenCodec[ObsSummary]", GoldenCodecTests[ObsSummary](10).goldenCodec)
```
cquiroz, line 12: Great to see this getting traction. Why did you increase the amount of samples? were there conditions not included?

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
cquiroz: 👍

cquiroz (approved):
Looking great, it looks very extensive

model/jvm/src/test/scala/explore/model/CirceGoldenSuite.scala
toddburnside: I was trying to get a mix of `pointings` and contraint sets. It was hard to get all of the permutations. I guess another option would be to set the number high when generating samples, then delete some of the generated cases with overlap and set the number to match how many files are left.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
toddburnside: Everything I could think of that used `exist` was a lot less concise because you need total function. It can be done, though, of course.

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
toddburnside: Good point. Do you think Observation name in the header and pointing in the description? Or, `conf` in the header?

We can also use actual contraints when ObsBadge is used in TargetObsList and elsewhere. But, we'll have to pull the necessary information from the API. The constraints is currently displayed as `<0.7" <0.3 mag Bright`, This would require pulling most of the constraint set from the API, or we could pull the name and use that instead. I'll look at the design doc to see if it specifies what should be displayed.

toddburnside: The design shows NameAndConf. I'll go with that unless we get told otherwise.

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
rpiaggio: How would a product work? The idea is that in trees we can model whether, say, a target is selected or one of its observations.

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
rpiaggio: :+1:

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
toddburnside: Yeah. I was looking at that code today, and the sum is needed. But, `Focused` is also being used in the `*TabContents` to determine whether to display the tree or right panel on narrow screens. This results in what seems to be odd behavior to me. For example, if you select a target in the target tab, then you'll also be displaying an empty right hand panel in constraints whether or not a constraint is selected. Alternatively, if you close a constraint and go back to the tree, you'll also go back to the tree in targets and observations whether or not you had one selected previously.

rpiaggio: > For example, if you select a target in the target tab, then you'll also be displaying an empty right hand panel in constraints whether or not a constraint is selected.

I might be missing something, but I think this shouldn't happen now that we are checking the type. The only case I think this can happen is when having an observation selected and switching views. I'm not sure whether this is desirable or we want to avoid it.

toddburnside: You're right. I was distracted when I wrote that. That was the old behavior. The new behavior is that if you "open" a target on the target tab, you "close" any constraint that was "open" in the constraints tab, etc. Which seems like it would make switching between tabs less smooth. But, maybe it's OK.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
cquiroz: Ah right, I thought exist could take a partial

PR #652, opened 2021-03-16
Update Explore for API changes.

toddburnside:
This update Explore to match changes from 2 PRs in the ODB API:

- Changes to Observation/Pointing sharing in https://github.com/gemini-hlsw/lucuma-tmp-api/pull/186
- Addition of NonEmptyString GraphQL scalar value and use for names in https://github.com/gemini-hlsw/lucuma-tmp-api/pull/194


cquiroz (approved):
LGTM

common/src/main/scala/explore/GraphQLSchemas.scala
```diff
@@ -6,9 +6,11 @@ package explore
 import clue.data._
 import clue.data.syntax._
 import clue.macros.GraphQLSchema
+import eu.timepit.refined.types.string.NonEmptyString
```
rpiaggio, line 9: Can we make the type mapping explicit in the `Scalars` section?

common/src/main/scala/explore/common/TargetQueries.scala
```diff
@@ -121,7 +114,7 @@ object TargetQueries {
        t.magnitudes
       )
     )(s =>
-      TargetResult.name.set(s._1.value) >>>
+      TargetResult.name.set(s._1) >>>
```
rpiaggio, line 117: Does it work if we also change the getter too in line 111 to `t.name`? Maybe we can now use `explore.optics.disjointZip`.

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
```diff
@@ -72,7 +72,7 @@ object ObsBadge {
                 <.div(
                   ExploreStyles.ObservationCardHeader,
                   props.layout match {
-                    case NameAndConf        => obs.name.orEmpty
+                    case NameAndConf        => obs.name.fold("")(_.value)
```
rpiaggio, line 75: Shall we use some placeholder here like `<UNNAMED>` or `-----` ?

cquiroz: I recall a report of some people being annoyed at `<UNAMED>` 🤷

observationtree/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -129,7 +130,7 @@ object ObsList {
       // Temporary measure until we have id pools.
       val newObs = IO(Random.nextInt()).map(int =>
         ObsSummary(Observation.Id(PosLong.unsafeFrom(int.abs.toLong + 1)),
-                   name.some,
+                   NonEmptyString.from(name).toOption,
```
rpiaggio, line 133: We can move the conversion to `InputModal` by having its `onComplete` callback take a `NonEmptyString` parameter.  (Maybe its `initialValue` can be an `Option[NES]` too).

rpiaggio (approved):
LGTM.

Probably we can also simplify the decoders in `TargetObsQueries.TargetIdName` and `.AsterismIdName`. I think they can be automatically derived now.

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
rpiaggio: Ah, let's avoid that one then (which we are not using anymore for new objects). How about dashes? The idea is to make explicit that there is no name. This situation shouldn't happen when using only the UI, but it can happen when using the API.

common/src/main/scala/explore/GraphQLSchemas.scala
toddburnside: Yep.

common/src/main/scala/explore/common/TargetQueries.scala
toddburnside: Good catch. I wasn't aware of `disjointZip` but it works. :)

observationtree/src/main/scala/explore/observationtree/ObsBadge.scala
toddburnside: Switched to dashes.

observationtree/src/main/scala/explore/observationtree/ObsList.scala
toddburnside: Good idea.

toddburnside:
> Probably we can also simplify the decoders in `TargetObsQueries.TargetIdName` and `.AsterismIdName`. I think they can be automatically derived now.

Good catch. `TargetIdName` can be derived, but `AsterismIdName` cannot because of the extra `nodes` layer for the targets.

rpiaggio:
> Good catch. `TargetIdName` can be derived, but `AsterismIdName` cannot because of the extra `nodes` layer for the targets.

Ah, right. We can still remove a few lines from it by using `.as[Option[NonEmptyString]]`

toddburnside:
> Ah, right. We can still remove a few lines from it by using `.as[Option[NonEmptyString]]`

Yep. I did that part of it. :)

PR #654, opened 2021-03-17
Fix issue when un-assigning observations

toddburnside:
On the targets and constraints tabs, it was not possible to drag an observation to the "Unassigned" area if there were no other unassigned observations.


cquiroz (approved):
LGTM

rpiaggio:
Good catch :+1:

PR #655, opened 2021-03-17
Error-handling crystal.

rpiaggio:
The new version of `crystal` would force us to handle errors (or log them) in effects started from Callbacks resulting from `.runAsync*CB` methods.

~Opening as draft for the moment until the new `crystal` is released, so we can see how code would look like, in case anyone has ideas to improve this.~


common/src/main/scala/explore/components/ConnectionsStatus.scala
```diff
@@ -54,7 +54,7 @@ object ConnectionsStatus {
   val component = ScalaComponent
     .builder[Props]
     .render(_ =>
-      AppCtx.withCtx { ctx =>
+      AppCtx.runWithCtx { ctx =>
```
cquiroz, line 57: I recall we discussed changing this name to make it more explicit 👍

common/src/main/scala/explore/components/UserSelectionForm.scala
```diff
@@ -62,50 +68,52 @@ object UserSelectionForm {
...
-                onClick = p.login >> p.message.set(none).runAsyncCB
-              ).when(s.showButtons),
-              Button(content = "Continue as Guest",
-                     size = Big,
-                     clazz = ExploreStyles.LoginBoxButton,
-                     onClick = p.guest >> p.message.set(none).runAsyncCB,
-                     icon = Icons.UserAstronaut
-              ).when(s.showButtons),
-              p.message.get.whenDefined(message =>
+        AppCtx.runWithCtx { implicit ctx =>
```
cquiroz, line 71: Why do we need the ctx now if we didn't before?

cquiroz (approved):
Looks good, some minor comments inline

common/src/main/scala/explore/components/UserSelectionForm.scala
toddburnside: For the implicit Logger, I believe.

toddburnside (approved):
LGTM

common/src/main/scala/explore/components/UserSelectionForm.scala
rpiaggio: > For the implicit Logger, I believe.

Yes

PR #662, opened 2021-03-18
Update constraint summary in obs tab obs badge when constraint changes

toddburnside:
Summary was not updating when the constraint was edited via the API.


rpiaggio (approved):
LGTM

PR #665, opened 2021-03-18
Pass regular children to ReflexContainer.

rpiaggio:
`ReflexContainer` expects a regular list of react children, so we can pass that instead of an array to avoid having to provide keys for the elements.

### DIFF
--- a/observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
+++ b/observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
@@ -336,72 +336,70 @@ object ConstraintSetObsList {
                       val obsSelected = props.focused.get
                         .exists(f => csObs.map(obs => FocusedObs(obs.id)).exists(f === _))
 
-                      <.span(^.key := csId.toString)(
-                        Droppable(csId.toString, renderClone = renderClone) {
-                          case (provided, snapshot) =>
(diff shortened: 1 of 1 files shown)


cquiroz (approved):
LGTM

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
```diff
@@ -304,39 +304,39 @@ object ConstraintSetObsList {
...
+                        ),
+                        Icons.ChevronRight
+                      )
 
-                        val obsSelected = props.focused.get
-                          .exists(f => csObs.map(obs => FocusedObs(obs.id)).exists(f === _))
+                      val obsSelected = props.focused.get
+                        .exists(f => csObs.map(obs => FocusedObs(obs.id)).exists(f === _))
 
+                      <.span(^.key := csId.toString)(
```
toddburnside, line 339: Do we need this `span` along with the other changes you made?

toddburnside (approved):
LGTM

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
rpiaggio: Good catch. I was just trying things out here. I'll remove it, it's not necessary.

PR #666, opened 2021-03-21
Several UI improvements.

rpiaggio:
* `react-reflex` facade now allows passing `scalajs-react` style attributes for styling, using the mechanism provided by `react-common`.
* `react-reflex` facade allows defining `ReactHandler` that are not direct children of a `ReactElement`.
* "Unassigned observations" header is now painted when dragging over it. Useful to drag onto it when the section is collapsed (Targets or Constraints View).
* Selected target/asterism/constraints are now painted when dragging over them. Before this, the selected color took precedence.
* Clicking on unassigned observations no longer selects them, since there is nothing to show in this case when in Targets or Constraints View.

### DIFF
--- a/common/src/main/scala/react/reflex/ReflexWithHandle.scala
+++ b/common/src/main/scala/react/reflex/ReflexWithHandle.scala
@@ -1,0 +1,59 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package react.reflex
+
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+
+import scalajs.js
+
+trait ReflexEvents extends js.Object
+
+object ReflexWithHandle {
+  // Yes, this is a hack. We use it because react-reflex expects a ReactHandle to be the
+  // direct first child of a ReactElement. This makes it impossible to make the ReactHandle
+  // part of, say, a react-beautiful-dnd Droppable, which would have to go between the
+  // ReactElement and the ReactHandle. Since ReactElement injects properties into its
+  // ReactHandle child, things break down if there's anything in between.
+  //
+  // We are therefore using:
+  // 1) An undocumented feature of react-reflex. If a boolean property "withHandle" is passed
+  // to ReactElement, it will inject into its children the properties that ReactHandle needs.
+  // (These properties are modeled in ReflexHandle.HandleProps).
+  // 2) We create a Scala component that will take ReflexHandle.HandleProps + a render function,
+  // to which it will provide the ReflexHandle.HandleProps. The render function can then pass this
+  // to a ReactHandle within.
+  // 3) Since the Scala component needs to capture the unboxed properties that ReactElement will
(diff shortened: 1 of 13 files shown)


common/src/main/scala/react/reflex/ReflexWithHandle.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+
+import scalajs.js
+
+trait ReflexEvents extends js.Object
+
+object ReflexWithHandle {
+  // Yes, this is a hack. We use it because react-reflex expects a ReactHandle to be the
```
cquiroz, line 14: This sounds good, do you think there is a risk that react-reflex may deprecate this in the future?

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
```diff
@@ -407,36 +407,44 @@ object ConstraintSetObsList {
...
-                    provided.innerRef,
-                    provided.droppableProps,
-                    props.getListStyle(snapshot.isDraggingOver)
-                  )(
-                    <.div(ExploreStyles.ObsTree)(
-                      <.div(ExploreStyles.ObsScrollTree) {
-                        Segment(
-                          vertical = true,
-                          clazz = ExploreStyles.ObsTreeGroup
+              ReflexElement(size = 36,
```
cquiroz, line 410: Eventually we may want to persist this size, right?

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
```diff
@@ -407,36 +407,44 @@ object ConstraintSetObsList {
...
+                      ReflexHandle(provided = reflexProvided)(
+                        Header(block = true,
+                               clazz =
+                                 ExploreStyles.ObsTreeHeader |+| ExploreStyles.ObsTreeGroupHeader
                         )(
-                          unassignedObs.zipWithIndex.toTagMod(
-                            (props.renderObsBadgeItem _).tupled
-                          ),
-                          provided.placeholder
+                          <.span(ExploreStyles.TargetLabelTitle)("UNASSIGNED OBSERVATIONS"),
```
cquiroz, line 428: I may rather not use capitals here and if we need them we set the style via css

cquiroz (approved):
LGTM

common/src/main/scala/react/reflex/ReflexWithHandle.scala
rpiaggio: There's a chance. It doesn't have a lot of stars and forks but it has 15 contributors and it seems actively maintained for the moment.

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
rpiaggio: We can persist the size of each element. Reflex elements can be size aware, and there's an `onResize` callback.

observationtree/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
rpiaggio: Good idea. I will do that.

PR #668, opened 2021-03-24
Use vite2 instead of webpack

cquiroz:
This is fairly big change on how we build the app. Before we used scala.js `fullOptS::webpack` command to bulid a big single js file which contains the scala.js code and the dependencies.

This process is fairly resource intensive and at some moment we were not able to build on heroku, thus terminating our ability to do review apps

Also it ties us to `scalajs-bundler` which is a good project but it is hard to adapt to the constantly evolving world of js bundlers.

The approach in **this** PR is radically different. Instead of taking all the code as a monolithic piece of code we can now produce ES modules and deliver them directly to the browser without bundling.

We can take advantage of a new breed of tools that forgo bundling and instead deliver ES modules only with a strong focus on performance.
Two were tested `snowpack` and `vite2`. While both are similar `vite2` is more appropriate for scala.js which in dev mode generates thousands of modules and need strong caching support.

This new style has some benefits and drawbacks that I will try to describe:

# Pros
* More efficient final bundling. While it takes approximately the same time, doing the final build in two stages (first on scala, then on node) is more efficient in memory usage, enough to get heroku to do it.
* For the same reason above we can get review apps again.
* Near instantaneous browser updates when changing assets or js code.
* We can now use the `SmallestModules` strategy on scala.js. With this scala.js will produce thousands of modules (one per class approximately) and updates will only need to rewrite those changes making development faster (in theory)
* More natural to work with the node ecosystem as we remove the odd step of `scalajs-bundler`
* It will allow us to transition to a world where we load modules dynamically reducing the impact of the large scala.js outputted code

# Cons
* On development the initial load is slower as it needs to read many thousand modules
* At the moment the output js is substantially larger. This is due to limitations on `scala.js` with ES modules rather than the js side tools.
* Needs to run two tools in parallel during development.
* It can only run on modern browsers that support ES modules.
* We loose `packtracker` support

It seems this is a good option moving forward. I'll leave this PR as draft as I want to keep testing how this works in practice. I'll try to do a few features using it before committing.

Some other notes:
* resource files are now in `common/src/main/webapp`. We want them outside the sbt set of resources to make it more efficient
* For now this is only enabled for the main `explore` app

# TODO
- [x] Fix to get the `conf` working on `development`
- [x] Fix jquery import
- [x] more testing
- [x] Disable testing on Heroku


toddburnside:
Wow. A lot of changes and a nice write-up. It's looking good. :)

build.sbt
```diff
@@ -1,54 +1,11 @@
...
-
-addCommandAlias(
-  "quickTest",
-  "modelJVM/test"
-)
-
-addCommandAlias(
-  "fixImports",
-  "; scalafix OrganizeImports"
-)
```
rpiaggio, line 51: Can we keep `quickTest` and `fixImports`?

common/src/main/public/development.conf.json
```diff
@@ -0,0 +1,11 @@
+{
```
rpiaggio, line 1: I'm guessing we need to have a `conf.json` now since there's no instance of copying `development.conf.json`. Would a symbolic link work, so we don't have duplicate things?

cquiroz: This is in the TODO. I need to somehow proxy this but I'm not there yet, this is a workaround.

build.sbt
cquiroz: Right, I deleted too much

explore/package.json
```diff
@@ -0,0 +1,51 @@
+{
```
rpiaggio, line 1: Will this allow us to use dependabot to keep JS dependencies updated?

cquiroz: I'd expect so, I'm still a bit undecided on whether to keep two package.json files or only one

rpiaggio:
Impressive work, thank you! I'm looking forward to this setup. Thank you for the detailed explanation too.

> On development the initial load is slower as it needs to read many thousand modules

How long does this take?

cquiroz:
image

cquiroz:
So it is in my host 280secs app for 5.2k files on the initial load

rpiaggio:
> So it is in my host 280secs app for 5.2k files on the initial load

Ok, but if I understand correctly, then we can just leave that process running "forever" in the background. If we change the JS dependencies, do we have to reload everything?

cquiroz:
The initial load is slow really because we are using the SmallestModule philosophy in scala.js

I hardly see a regular js project having 5k modules

I hope/expect that SmallestModules will make development with scala.js faster though 

If you change deps I think you need to reload though I haven't tried

rpiaggio:
I'm running on this branch and it's working great, thank you and congratulations!

Changes are reflected much quicker than before.

I had an issue with the self signed certificate. It was solved by commenting them in `vite.conig.js` and running with `npx vite --https`.

cquiroz:
Ok this seems to be in a usable state. @rpiaggio tested it and deployment works.
Let me know if we should merge and I'll disable doing CI in heroku to get this on green

rpiaggio:
Ship it!

PR #671, opened 2021-03-25
Merge the subprojects into a single `explore` subproject

cquiroz:
As discussed on the `ui` channel we may want to remove the subprojects and have a single `explore` sub projcet.
I'm keeping `model` as it is cross built and I also kept `commons` as I think these should go to `lucuma-ui` eventually but I could be convinced to merge it with `explore`


toddburnside (commented):
Can the various `Test.scala` files be deleted?

cquiroz:
Good point, I'll do that

toddburnside (approved):
Looks fine to me. I've made quite a few changes to files that have been moved - I'll just make copies of them and copy them to the correct places after I merge in these changes.

cquiroz:
@toddburnside We can do this later on, moving them is very easy

toddburnside:
> @toddburnside We can do this later on, moving them is very easy

It's not a big deal for me, since I haven't made a commit yet. But, it'd probably be a good idea to see what state Raúl is in.

rpiaggio (approved):
LGTM. I'm working only in `common` at the moment, so I won't be affected. Agreed to keep `common` separate for now. Most things there will probably go to `lucuma-ui` or their own facade libraries.

PR #672, opened 2021-03-26
GraphQL types generated with new clue using scalafix

rpiaggio:
As you know, we are dropping `clue` macros in favor of generating code with `scalafix`.

* Same as before, an `object` annotated with `@GraphQLSchema` gets the types generated. For the moment, the template object name must end with `*GQL`. The trait and object containing the generated types will have the suffix stripped. Also, the GraphQL schema must be in a `.graphql` file named without the `*GQL` suffix. In the future there will probably be a more flexible/configurable way of naming things.
* To run the generator, run in sbt `scalafix GraphQLGen` (You can give it a try).
* The generated code is now part of the project and checked in.
* ~For the moment, only schema generation is implemented, queries are coming next.~ Queries are migrated too. Small change is that the template for a query/mutation/subscription must now be a `trait` instead of an `object`. Same naming conventions (ie: `*GQL` suffix) is followed as for schemas.
* I split the `GraphQLSchemas` object into a file for each schema, to better organize code now that it contains a huge bunch of code.
* ~The code I'm checking in had `scalafmt` run after generation. As discussed, we may not want this in the future if we want to `--check` in CI.~ The generated code now has `format: off` and `format: on` around it so we can `--check`.

~BTW: We probably want `ConstraintsPanel` and `ConstraintsQueries`. They are superseded by Todd's ongoing work. I needed to comment them out since they referred to the obsolete `ExploreDB`, which I now removed. I left them there for the moment just in case, please let me know if I can obliterate.~ New constraints editor was merged and adapted.

~This is a draft PR to showcase how things will look. I'll mark as done after we migrate the queries too.~


build.sbt
```diff
@@ -113,6 +111,30 @@ lazy val common = project
...
+    //       .in(Compile)
+    //       .toTask(s" GraphQLGen --out-from=$outFrom --out-to=$outTo") // --settings=$settings")
+    //       .value
+    //     (to ** "*.scala").get
+    //   }
+    // }.taskValue
+    // scalafix GraphQLGen --settings=GraphQLGen.schemaDirs=["/Users/rpiaggio/gemini/explore/common/src/main/resources/graphql/schemas"]
+    // scalafix GraphQLGen --settings={"GraphQLGen":{"schemaDirs":["gen/input/src/main/resources/graphql/schemas"]},"rules":["GraphQLGen"]}
+    // THIS ALMOST WORKS (IT WOULD IF IT WAS JUST A STRING): scalafix GraphQLGen --settings.schemaDirs=["gen/input/src/main/resources/graphql/schemas"]
+    // scalafix GraphQLGen --out-from=(.+).scala --out-to=$1.gql.scala
```
rpiaggio, line 137: Please disregard all this. It's an ongoing experiment towards an sbt plugin (which I'm not sure will be necessary).

common/src/main/scala/explore/schemas/implicits.scala
```diff
@@ -0,0 +1,26 @@
+package explore.schemas
+
+import clue.data.syntax._
+import lucuma.core.model.Magnitude
+import ObservationDB.Types.MagnitudeInput
+import java.math.MathContext
+import UserPreferencesDB.Types.WidthUpsertInput
+import UserPreferencesDB.Types.ExploreResizableWidthInsertInput
+
+object implicits {
```
rpiaggio, line 10: Implicits have to be defined outside of the template objects, since they refer to generated types that don't exist before the first generation. This prevents the code from compiling, and scalafix won't run until the code compiles.

toddburnside:
> BTW: We probably want `ConstraintsPanel` and `ConstraintsQueries`. They are superseded by Todd's ongoing work. I needed to comment them out since they referred to the obsolete `ExploreDB`, which I now removed. I left them there for the moment just in case, please let me know if I can obliterate.

Keeping them but commented out is fine. I have replaced the content of both files in my current work but they will still exist. I hadn't got to the point of deleting `ExploreDB` but it was on my list. I will be cleaning up the temporary `Constraints` model and related classes/code.

build.sbt
```diff
@@ -34,11 +34,13 @@ inThisBuild(
...
     addCompilerPlugin(
-      ("org.typelevel" % "kind-projector" % kindProjectorVersion).cross(CrossVersion.full)
+      ("org.typelevel"                    % "kind-projector" % kindProjectorVersion).cross(CrossVersion.full)
     ),
     description := "Explore",
     scalacOptions += "-Ymacro-annotations",
-    Global / onChangedBuildSource := ReloadOnSourceChanges
+    Global / onChangedBuildSource := ReloadOnSourceChanges,
+    scalafixDependencies += "edu.gemini" %% "clue-generator" % Settings.LibraryVersions.clue,
+    scalafixScalaBinaryVersion := "2.13"
```
cquiroz, line 43: Will this change in dotty?

cquiroz:
I'm not sure how much can I review but I support the concept.
It is kind of a pity not to make it with macros but it is hard to battle scala 3 compatibility

Also reductions on compile size are welcome

Did you test it with the hasura scheme?

rpiaggio:
> > BTW: We probably want `ConstraintsPanel` and `ConstraintsQueries`. They are superseded by Todd's ongoing work. I needed to comment them out since they referred to the obsolete `ExploreDB`, which I now removed. I left them there for the moment just in case, please let me know if I can obliterate.
> 
> Keeping them but commented out is fine. I have replaced the content of both files in my current work but they will still exist. I hadn't got to the point of deleting `ExploreDB` but it was on my list. I will be cleaning up the temporary `Constraints` model and related classes/code.

Ok, It sounds they are not needed as reference anymore. I will delete tem and then you can add back the new versions.

build.sbt
rpiaggio: Yes, that's the scala version Scalafix runs in, which may or may not be the the same as the project on which it's being run. It defaults to 2.12 so we have to set it if it takes another value. In our case, we are only publishing the `GraphQLGen` rule for 2.13, so we have to change it.

rpiaggio:
> Did you test it with the hasura scheme?

Yes, everything's working, including the `UserPreferencesDB`.

rpiaggio:
This is pretty much done. This is how code will look like with scalafix.

There are a couple of rough edges on the clue side that I'd like to fix before releasing it and merging this, so that we have `--check` working.

I feel theres A LOT of room for improvement here. We can do that in the next steps, but for now we are unblocked to move to Scala 3 (and probably we have faster compilation times now).

Suggestions are more than welcome!

rpiaggio:
`--check` is now working, added it to the tests. @cquiroz you may want to add this to the vite branch when you merge.

Marking the PR as ready.

rpiaggio:
Oh wait, we need a scalafix release...

rpiaggio:
Ok, it's out...

toddburnside (approved):
Great work!

Sometime I really need to look at how all of this code get generated. Looks good.

PR #674, opened 2021-03-29
Small theme improvements

cquiroz:
As part of another PR I did a few theme fixes but I moved them to a separate PR. It is mostly about the separator and a few borders that were too contrasty for my taste


toddburnside (approved):
👍

PR #682, opened 2021-03-30
Ch732 edit constraints

toddburnside:
There are a few little improvements that might be desired for the editing of the elevation range values, but I'd like to get this merged first. Everything currently works, and it'll be easier to talk about potential enhancements without the noise of all the other changes.

image


common/src/main/resources/less/style.less
```diff
@@ -951,6 +951,18 @@ tfoot {
   }
 }
 
+.constraints-grid {
+  padding: 0.5em;
+
+  .elevation-range-picker {
+    width: 18ex;
```
cquiroz, line 958: Any reason to prefer ex over em?

common/src/main/scala/explore/model/reusability.scala
```diff
@@ -27,9 +26,11 @@ object reusability {
   implicit def focusedReuse: Reusability[Focused]                                    = Reusability.derive
   implicit def idListReuse[Id, A: Reusability]: Reusability[KeyedIndexedList[Id, A]] =
     Reusability.by(_.toList)
-  implicit def constraintsSummaryReuse: Reusability[ConstraintsSummary]              = Reusability.derive
-  implicit def obsSummaryReuse: Reusability[ObsSummary]                              =
-    Reusability.by(x => (x.id, x.name, x.status, x.conf, x.constraints, x.duration, x.pointingId))
+  implicit def AirMassRangeReuse: Reusability[AirMassRange]                          = Reusability.byEq
```
cquiroz, line 29: I suppose some of these will go to `lucuma-ui`

explore/src/main/scala/explore/constraints/ConstraintsQueries.scala
```diff
@@ -3,110 +3,135 @@
...
-          cloud_cover
-          image_quality
-          sky_background
-          water_vapor
+          cloudExtinction
+          imageQuality
+          skyBackground
+          waterVapor
+          elevationRange {
+            type: __typename
```
cquiroz, line 40: What's the type used for?

cquiroz (approved):
Looking great, I added a few minor questions
Do you have a screenshot?

common/src/main/resources/less/style.less
toddburnside: Hmmm. No. I think this was a neural meltdown. I'll change to em.

toddburnside:
> Do you have a screenshot?

Added to description.

explore/src/main/scala/explore/constraints/ConstraintsQueries.scala
toddburnside: It is needed in the decoder to determine whether to decode an AirMassRange or an HourAngleRange.

swalker2m:
In the "Observations" view, when the "Constraints" panel is collapsed down to the bare minimum, the sky background and water vapor are summarized with icons.  At least that's the case in the mockups (e.g., Figure 2.23 A partially executed sequence).  Just a possible thought for the future, if we end up using them it might help to reinforce those icons by adding them to the options in the editor.

toddburnside:
> Just a possible thought for the future, if we end up using them it might help to reinforce those icons by adding them to the options in the editor.

That's a good idea. If/when we get icons we can look into that. I think we'd need to enhance the select controls to support that.

PR #684, opened 2021-03-30
Observations always have a name

cquiroz:
This improves the signature on `ObsSummary` and assumes there is always a name. Note the UI doesn't let you create observations with an empty name.


toddburnside (approved):
LGTM

cquiroz:
The design has change thus this is obsolete

PR #690, opened 2021-04-03
New crystal with contexts.

rpiaggio:
Crystal now uses React's context for storing, well, context. It provides a component to store context (`ContextProvider`) and a mechanism where consumers can use just the value of a context, or get a `View` which allows modifying said context. Also, it supports multiple contexts.

This PR includes a little demo on how to use it which is not meant to go into production. After we have implemented a proper help system, we can remove this:

Kapture 2021-04-03 at 12 00 41

One disadvantage of this approach is that we can't access contexts from lifecycle methods. Although React provides a way to do this, support in `scalajs-react` is pending (https://github.com/japgolly/scalajs-react/issues/890). To work around this temporarily, context is passed implicitly in props from parents to child components who need it in lifecycle methods.

The old `AppRoot` has been renamed to `StateProvider`.


common/src/main/scala/explore/common/ConstraintSetObsQueries.scala
```diff
@@ -197,14 +197,9 @@ object ConstraintSetObsQueries {
...
-    (
-      View[ConstraintSetsWithObs] => VdomNode
-    ) => LiveQueryRenderMod[ObservationDB, ConstraintSetsObsQuery.Data, ConstraintSetsWithObs]
-
-  val ConstraintSetObsLiveQuery: LiveQueryRenderer =
-    render =>
-      AppCtx.runWithCtx { implicit appCtx =>
+  val ConstraintSetObsLiveQuery =
+    ScalaFnComponent[View[ConstraintSetsWithObs] => VdomNode](render =>
+      AppCtx.using { implicit appCtx =>
```
cquiroz, line 202: Is it so that we execute this on `IO` when calling `using`? Or is it different than `runWitCtx` in that sense

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -75,14 +75,27 @@ object TopBar {
...
                 borderless = true,
                 tabular = MenuTabular.Right
               )(
                 MenuItem(
                   <.span(
                     ExploreStyles.MainTitle,
-                    "Explore"
+                    "Explore",
+                    HelpCtx.usingView { help =>
+                      val helpMsg = help.zoom(HelpContext.msg)
```
cquiroz, line 86: Nice, this will be useful

cquiroz (approved):
Thanks a lot, this was a lot of work but will be useful in a few important scenarios

common/src/main/scala/explore/common/ConstraintSetObsQueries.scala
rpiaggio: The inner workings are different than before. `Ctx.using` (or `Ctx.usingView`) return a `VdomElement`, so they can only be used within `render`. In other words, they act as a component; and since it's declarative, a name with `run` in it didn't seem fitting anymore. Also, the function passed to them must be a render function, returning a `VdomNode`.

This is a functionality provided by React, and is basically a way of "teleporting" props further down the component tree, without the need of explicitly passing them in all intermediate components. If the context changes, it is automatically propagated to all `.using(View)` blocks in the currently rendered component tree.

toddburnside (approved):
LGTM

PR #693, opened 2021-04-04
Update sbt

cquiroz:
Note we should downgrade log4cats, version 2.x is meant to be used with `cats-effect` 3


toddburnside (approved):
LGTM

cquiroz:
I'll wait a bit before merging. I noted some performance hit

PR #698, opened 2021-04-05
Remove user defined name

cquiroz:
This PR removes the user defined name of observations and instead shows the target/asterism name and the obs id

It will be discussed if we should replace the obs id with a different type of index

image


cquiroz:
This needs some rework

PR #701, opened 2021-04-07
Use sttp cats fetch backend

cquiroz:
Just noted there is a cats-effect fetch backend recently introduced in sttp. This lets us have nicer code


rpiaggio (approved):
Nice LGTM.

BTW, there are other uses in `SSOClient`.

Maybe we should create a single `backend` and keep it in the context?

toddburnside (approved):
LGTM

cquiroz:
Good point, I'll update the other bits as well

rpiaggio:
Nice, you unified the backend too in SSO. Thank you!

PR #705, opened 2021-04-10
Type rework.

rpiaggio:
This PR attempts to establish a criterion to reconcile the various types which represent ad-hoc query results with model types we need in order to pass said results to common components (like `ObsBadge`).

The idea is to decode always to ad-hoc types (in most of the cases the generated ones) and convert to model types at the last moment when needed.

**Pros:** We don't need to force queries to our model types.
**Cons:** We need to do manual conversions. This may be automatable to some extent, using for example `chimney`. I have not explored this route yet since it's not clear whether chimney will have a Scala 3 release.

Also, I've introduced optics based on generated types, which are therefore are only available in `common` and not in `model`. In order to test these optics, I've created `model-testkit` project (and a `model-tests` to which the tests where moved). The tests in `common` can only be run in a JS environment.

`ObsSummary` is now a trait, with a bunch of other traits that can be mixed in to represent possible fields. `ObsBadge` will decide the layout based on the mixins. Also, it shows the obs id, with code copied from #698. Observations do not have a name anymore.


common/src/main/scala/explore/common/ConstraintSetObsQueries.scala
```diff
@@ -101,7 +103,9 @@ object ConstraintSetObsQueries {
       }
 
       object Observations {
-        type Nodes = ObsSummary
+        object Nodes {
+          type ConstraintSet = ConstraintsSummary
```
cquiroz, line 107: Wherre is `ConstraintSummary` defined?

common/src/main/scala/explore/common/ConstraintSetObsQueries.scala
```diff
@@ -202,11 +255,29 @@ object ConstraintSetObsQueries {
   // format: on
   /* END: Generated by clue. Will be replaced when regenerating. */
 
+  private def convertPointing(
+    pointing: ConstraintSetsObsQuery.Data.Observations.Nodes.ObservationTarget
+  ): Pointing =
+    pointing match {
```
cquiroz, line 261: Nice

common/src/main/scala/explore/common/ObsQueries.scala
```diff
@@ -133,9 +189,27 @@ object ObsQueries {
   // format: on
   /* END: Generated by clue. Will be replaced when regenerating. */
 
+  private def convertPointing(
```
cquiroz, line 192: It seems we need to define this for each query that will use our `model` classes, right? Seems a good trade off

common/src/main/scala/explore/common/ConstraintSetObsQueries.scala
rpiaggio: It's in `explore.model`. It was already there, that part of the logic didn't change.

In this case, we are still decoding directly to our model type, since we are always querying the same set of fields.

common/src/main/scala/explore/common/TargetObsQueries.scala
```diff
@@ -163,14 +165,13 @@ object TargetObsQueries {
           observations(programId: "p-2", first: 2147483647) {
             nodes {
               id
-              name
-              observationTarget {
+              pointing: observationTarget {
                 type: __typename
                 ... on Target {
-                  target_id: id
+                  targetId: id
```
cquiroz, line 171: Are you using `targetId` rather than `target_id` to get nicer field names?

common/src/main/scala/explore/common/ObsQueries.scala
rpiaggio: Yes, that's the idea (the "con" I mentioned in the PR description).

common/src/main/scala/explore/common/TargetObsQueries.scala
rpiaggio: Exactly. The aliased names will be used to generate the case classes to store results, so we might as well go with Scala style.

explore/src/main/scala/explore/observationtree/ConstraintSetObsList.scala
```diff
@@ -105,7 +105,7 @@ object ConstraintSetObsList {
       ConstraintSetsWithObs.obs.composeGetter(
         obsWithIndexGetter
           .composeOptionLens(first)
-          .composeOptionLens(ObsSummary.constraints)
+          .composeOptionLens(ObsSummaryWithPointingAndConstraints.constraints)
```
cquiroz, line 108: 👍

explore/src/main/scala/explore/observationtree/TargetObsList.scala
```diff
@@ -760,7 +779,7 @@ object TargetObsList {
                               <.span(ExploreStyles.ObsTreeGroupHeader)(
                                 <.span(
                                   opIcon,
-                                  asterism.name.value,
+                                  asterism.name.map(_.value),
```
cquiroz, line 782: Would this show `Some(name)` on the UI?

model/shared/src/main/scala/explore/model/Pointing.scala
```diff
@@ -0,0 +1,37 @@
...
+package explore.model
+
+import cats.Eq
+import cats.syntax.all._
+import eu.timepit.refined.cats._
+import eu.timepit.refined.types.string.NonEmptyString
+import lucuma.core.model.Asterism
+import lucuma.core.model.Target
+
+sealed trait Pointing
```
cquiroz, line 13: Cool

project/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.4.4
```
cquiroz, line 1: 1.5.0?

cquiroz:
Looks great, thanks for this

cquiroz (approved):
LGTM

project/project/build.properties
rpiaggio: I'm not sure how this file got here, it's in `project/project`. I'll try deleting it.

explore/src/main/scala/explore/observationtree/TargetObsList.scala
rpiaggio: I think it will just show `name` and skip it if it's `None`, as per `scalajs-react`'s conversion of an `Option[String]` into `VdomNode`.

PR #708, opened 2021-04-10
Ch761 prototype help system

cquiroz:
Prototype of the help system based on the demoed version. There will likely be changes but this is a good start


common/src/main/scala/explore/components/HelpIcon.scala
```diff
@@ -0,0 +1,45 @@
...
+import explore.HelpCtx
+import explore.Icons
+import explore.components.ui.ExploreStyles
+import explore.implicits._
+import explore.model.Help
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import react.common._
+
+final case class HelpIcon(id: Help.Id) extends ReactProps[HelpIcon](HelpIcon.component) {}
```
rpiaggio, line 20: Brackets can be removed?

common/src/main/scala/explore/contexts.scala
```diff
@@ -4,16 +4,32 @@
...
+import sttp.model.Uri
 
 object AppCtx extends Ctx[IO, AppContextIO]
 
 @Lenses
-case class HelpContext(msg: Option[String])
+case class HelpContext(
+  rawUrl:        Uri,
+  editUrl:       Uri,
+  user:          String Refined MatchesRegex["[\\w-_]*"],
```
rpiaggio, line 23: If the regexes are `+` instead of `*` we can prevent empty strings.

explore/src/main/scala/explore/HelpBody.scala
```diff
@@ -0,0 +1,132 @@
...
+                  ExploreStyles.HelpBody,
+                  s.content match {
+                    case Ready(a)         =>
+                      ReactMarkdown(content = a,
+                                    clazz = ExploreStyles.HelpMarkdownBody,
+                                    linkTarget = "_blank",
+                                    imageConv
+                      ): VdomNode
+                    case Pending(_)       => <.div(ExploreStyles.HelpMarkdownBody, "Loading...")
+                    case crystal.Error(_) =>
```
rpiaggio, line 108: Maybe we want to discriminate on the error.

It seems that if there's a timeout for example, then the user will get "Not found" and the option to create the page.

rpiaggio (approved):
Looks awesome to me. 👍

common/src/main/scala/explore/contexts.scala
cquiroz: You're right. good idea

explore/src/main/scala/explore/HelpBody.scala
cquiroz: I'm not sure what to offer in case of timeout, just an error message, or maybe a retry button?

cquiroz:
A few notes:

* I changed the source of the md files from the wiki to reglar files, This reduces the options for easy editing online, but lets us accept third party PRs.
* This PR includes a facade for `react-markdwon`

explore/src/main/scala/explore/HelpBody.scala
rpiaggio: Maybe we can just discriminate the "not found" error and in all other cases show a generic error?

The retry button sounds good too, although if the user closes and opens the help panel it will work as a retry right?

cquiroz:
Addressed PR comments

toddburnside (approved):
Looks great!

PR #723, opened 2021-04-14
Break GraphQL generated sources dependency cycle.

rpiaggio:
The previous GraphQL type generator required the project that used the generated types to compile so that the generator could run. This required us a lot of times to break the cycle by commenting/generating/uncommenting, which was... inconvenient at best. (Although I'm afraid in time this would become plainly painful).

This PR proposes a different approach, in which the GraphQL schemas and queries (and their templates) are defined in a separate project `common-graphql` (which depends on `model` but NOT on `common`). Clue's generator is run as an sbt source generator in `common`, providing it with all the types to run the queries.

Although I don't like having the queries in a separate project, this is probably less frustrating than having to manually comment and uncomment code when running the generator.

A side effect of this approach is that the generated code is not checked in anymore, which means we don't have to check if it's up to date. Also, the generator is run automatically whenever the queries or schema change.

Furthermore, since the template files and the code generated from them are kept separately, I've changed the clue generator (https://github.com/gemini-hlsw/clue/pull/158) so that now the generated code replaces the template, which simplifies the generator quite a bit.

As a final note, I would have liked to keep the generated types in `trait`s, for example `trait ObsQueriesGQL` and have `object ObsQueries extends ObsQueriesGQL`. However, this confused the `Reusability.derive` macro. I've therefore opted for 2 separate `object`s, which unfortunately means we have to import both in usage sites.

(Unrelatedly, some JS dependencies and facades were also updated in this PR).

BTW, CI will fail until the new clue is released.


common/src/main/scala/explore/common/ConstraintSetObsQueries.scala
```diff
@@ -28,10 +25,23 @@ import lucuma.ui.reusability._
...
+  sealed trait A
+  object A {
+    case class B(b: Int) extends A
+    case class C(c: String) extends A
+  }
+  implicit val reuseA: japgolly.scalajs.react.Reusability[A] = {
+    import japgolly.scalajs.react._
+    japgolly.scalajs.react.Reusability.derive[A]
+  }
+
```
toddburnside, line 44: Is this needed?

common/src/main/scala/explore/common/package.scala
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore
+
+package object common {
+  // These are needed for Reusability derivation macro to work when defined within traits.
```
toddburnside, line 7: Will these be needed? If so, a better comment might be useful.

common/src/main/scala/explore/common/ConstraintSetObsQueries.scala
rpiaggio: Good catch, thank you. No, that's not needed, I was just testing the `Reusability.derive` under different circumstances. (If `ConstraintSetObsQueries` is a trait instead of an object, it doesn't work).

common/src/main/scala/explore/common/package.scala
rpiaggio: These aren't needed either, and are leftovers from different things I tried.

build.sbt
```diff
@@ -202,7 +201,11 @@ lazy val commonJsLibSettings = lucumaScalaJsSettings ++ commonLibSettings ++ Seq
   dependencyOverrides ++= ScalaJSReact.value
 )
 
-lazy val commonES = Seq(
+lazy val commonModule = Seq(
+  scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
```
cquiroz, line 205: I guess we should use ESModule kind

cquiroz (approved):
I.m not sure if I follow well the change but I'm certainly willing to test it

build.sbt
rpiaggio: MUnit in ScalaJS requires CommonJSModules (https://scalameta.org/munit/docs/getting-started.html#scalajs-setup). I will apply the setting only for `Test` though.

cquiroz: Ok, this is not very important but I think it should work with ESModules too

PR #725, opened 2021-04-15
Reusable functions

rpiaggio:
~This is a base PR showing how we would have reusable functions using `scalajs-react` directly.~

~In a 2nd PR we can compare usage with rolling out our own `Reuse`.~

Reusable functions passed to components via `Reusable.fn` and added convenience syntax. (`View` keeps assuming update function is stable and keeps reusability only by value).


rpiaggio:
Pushed a new commit. It turns out we can get the more convenient syntax by using a few implicits.

rpiaggio:
I've therefore discarded using our own `Reuse`. I will convert the rest of the functions passed as props before making this PR ready for review.

rpiaggio:
Furthermore, I considered making a version of `ViewF` where the mod function is actually a `Reusable` function, but I can't think of a real-world example where we use a mod function that depends on other values. For all practical cases, I think it's pretty safe to assume the mod function is stable and reusable, and save ourselves the additional complexity.

WDYT @cquiroz @toddburnside ?

cquiroz:
I'm not sure I have a strong (on any really) opinion, I'll happily defer to you

toddburnside:
> For all practical cases, I think it's pretty safe to assume the mod function is stable and reusable, and save ourselves the additional complexity.

This seems reasonable to me. I can't think of any examples off hand, either.

explore/src/main/scala/explore/observationtree/TreeComp.scala
```diff
@@ -1,172 +1,172 @@
 // Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
```
rpiaggio, line 1: We are not using this component, but we will probably use it in the future, that's why I'm leaving it commented out for the moment. Migration to reusable functions is not trivial in this case.

cquiroz (approved):
Looks great!

toddburnside (approved):
LGTM. Let's give it a try!

PR #727, opened 2021-04-16
Fix react-markdown in production

cquiroz:
`react-markdown` is being used for the help system but it fails in production.
After chasing down this issue, my conclusion is that `vite/rollup` has some troubles with modules using commonjs style rather than es style

https://github.com/vitejs/vite/issues/2862

This is still odd because it works properly if I use it on the plain js side, thus there is something odd on the way scala.js interacts with commonjs imports

The solution I found is kind of convoluted but seems to work:

Remove the `react-markdown` dependency and internalize the `.js` files from `react-markdown` but convert them to es via [cjstoesem](https://github.com/wessberg/cjstoesm/)

This feels very hacky but the only alternative I can think of is to reimplement in scala.
I'd like to at least test this in prod.


rpiaggio (approved):
:+1: hope this fixes the issue in prod

toddburnside (approved):
🤞 👍

PR #734, opened 2021-04-19
ConstraintsQueries move to comon-graphql.

rpiaggio:
When moving GraphQL queries to `common-graphql` project I missed the fact that `ConstraintsQueries` was in the `explore` project and thus was not migrated to the new "managed sources" mechanism.

This PR fixes that.


cquiroz (approved):
Looks good

toddburnside (approved):
👍

PR #742, opened 2021-04-21
Update ConstraintsSummary model

cquiroz:
Add totalCount of observations using a constraint set

### DIFF
--- a/model/shared/src/main/scala/explore/model/ConstraintsSummary.scala
+++ b/model/shared/src/main/scala/explore/model/ConstraintsSummary.scala
@@ -4,12 +4,16 @@
 package explore.model
 
 import cats._
+import eu.timepit.refined.auto._
 import eu.timepit.refined.cats._
+import eu.timepit.refined.types.numeric.NonNegInt
 import eu.timepit.refined.types.string.NonEmptyString
 import io.circe.Decoder
 import io.circe.Encoder
-import io.circe.generic.semiauto._
+import io.circe.HCursor
+import io.circe.Json
 import io.circe.refined._
+import io.circe.syntax._
 import lucuma.core.enum.CloudExtinction
 import lucuma.core.enum.ImageQuality
 import lucuma.core.enum.SkyBackground
@@ -24,7 +28,8 @@ final case class ConstraintsSummary(
   imageQuality:    ImageQuality,
   cloudExtinction: CloudExtinction,
   skyBackground:   SkyBackground,
-  waterVapor:      WaterVapor
+  waterVapor:      WaterVapor,
+  obsCount:        NonNegInt
 ) {
   def summaryString: String =
     s"${imageQuality.label} ${cloudExtinction.label} ${skyBackground.label} ${waterVapor.label}"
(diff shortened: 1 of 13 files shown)


model/shared/src/main/scala/explore/model/ConstraintsSummary.scala
```diff
@@ -24,7 +28,8 @@ final case class ConstraintsSummary(
   imageQuality:    ImageQuality,
   cloudExtinction: CloudExtinction,
   skyBackground:   SkyBackground,
-  waterVapor:      WaterVapor
+  waterVapor:      WaterVapor,
+  obsCount:        NonNegInt
```
cquiroz, line 32: I wonder if this should be an `Option` and only read it in the obs view

toddburnside: 🤔 It definitely isn't needed everywhere. On the other hand, do we want to deal with the optionality in the places it is required?

cquiroz: Yeah that's the question, and definitely I don't want to have different classes for different queries

toddburnside: I guess we could do `getOrElse(0)` where it is needed. It's a little loss of safety, but in practice it isn't likely to be a problem.

rpiaggio: Why do we need the count in `ConstraintsSummary`? I thought `ConstraintsSummary` was only being used to build the badges. I would expect it in `ConstraintSetModel` and maybe the logic to display warning within `ConstraintsPanel`, which will be shared in both the Constraints and Observation views.

rpiaggio: Answering my own question: we probably need it in the table to show when no item is selected in the constraints view. In that case, we do need it in `ConstraintsSummary` if that's what we are using there. In that case, we should probably use the same live query to feed data to both the tree and the table.

cquiroz: I want to reuse the constraint editor on the obs view (like we do for targets ) and that takes a ConstraintsSummary, I just want to get the obs count there to implement the custom logic for shared/not shared constraint sets

rpiaggio: I must be missing something. Doesn't the constraint editor (`ConstraintsPanel`) take a `ConstraintSetModel`?

cquiroz: You're right. we only pass the constraints to query a `ConstrantSetModel` and use it in `ConstraintsPanel`

I'm thinking of showing if you the constraint is shared on the tile, thus I need it on the Obs query.
u

rpiaggio: I'm going to try making `ConstraintsSummary` a trait with the 4 enums and the `summayString` method and let clue generate types to store results but force them to extend such trait. This should be in line with the approach of not decoding into our model and instead converting (or in this case extending), which would in turn remove the need for our own decoder.

cquiroz: Ok let's close this

PR #743, opened 2021-04-23
Render in tiles' titles.

rpiaggio:
By using React's portals, tiles contents can now render things into the tile's title.

`Tile` now accepts a `render` function `Tile.RenderInTitle ~=> VdomNode` instead of children. The contents renderer therefore now receive a `Tile.RenderInTitle` which is just a `VdomNode ~=> VdomNode` that when called, will create a portal into a `<.span` in the title.

Undo buttons have been migrated to use this mechanism.

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -7,10 +7,13 @@ import cats.syntax.all._
 import explore.Icons
 import explore.components.ui.ExploreStyles
 import explore.components.ui.ExploreStyles._
+import explore.implicits._
 import explore.model.Constants
 import explore.model.enum.TileSizeState
 import japgolly.scalajs.react._
+import japgolly.scalajs.react.internal.JsUtil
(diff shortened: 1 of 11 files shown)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -45,11 +53,16 @@ object Tile {
          (1024                           -> TileLGW)
     )
 
+  def renderPortal(mountNode: Option[raw.ReactDOM.Container], info: VdomNode): VdomNode =
+    mountNode.map(node => ReactPortal(info, node))
```
cquiroz, line 57: Is ReactPortak defined in scalajs-react?

cquiroz (approved):
Extraordinary, this is going to be very useful

toddburnside (approved):
Very cool!

common/src/main/scala/explore/components/Tile.scala
rpiaggio: Yes

PR #747, opened 2021-04-24
ConstraintsSummary as trait.

rpiaggio:
This PR aligns the use of `ConstraintSummary` with `ObsSummary`, in which we don't decode results directly into our model classes, converting or extending instead. This gives us more flexibility in building queries. We also avoid the need to provide decoders.


cquiroz (approved):
Great

PR #755, opened 2021-04-26
Move the info-ref to the backend

cquiroz:

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -39,8 +39,6 @@ object Tile {
 
   type RenderInTitle = VdomNode ~=> VdomNode
 
-  val infoRef = Ref.toVdom[html.Element]
-
   // Explicitly never reuse as we are not considering the content
   implicit val propsReuse: Reusability[Tile] = Reusability.never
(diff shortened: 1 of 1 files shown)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -58,58 +56,69 @@ object Tile {
...
+            secondary = true,
+            clazz = ExploreStyles.TileTitleMenu,
+            tabular = MenuTabular.Right
+          )(
+            MenuItem(as = <.a)(p.title)
+          ),
+          <.span(ExploreStyles.TileTitleInfo,
+                 ^.untypedRef := infoRef,
+                 ^.key := s"tile-info-${p.title}",
+                 <.span
```
rpiaggio, line 99: We can probably remove this inner `span` too.

rpiaggio (approved):
LGTM. Thank you! This was an oversight on my part.

common/src/main/scala/explore/components/Tile.scala
cquiroz: I wasn't sure what was that used for, I'll remove it in the next PR

PR #756, opened 2021-04-27
Ch780 support GitHub style markdown

cquiroz:
This PR adds support for github flavored markdown, things like strike-through and tables.
I tried a new technique to do the build that lets me remove the internalized `react-markdown` and depend on the npm package. Remains to be tested in staging.


toddburnside (approved):
I don't fully understand the build change, but let's give it a try in staging.

cquiroz:
Basically we need to use different type of imports for production and development, this is because the dependency uses common js but in dev, vite will convert it to es

common/src/main/scala/react/markdown/ReactMarkdown.scala
```diff
@@ -9,24 +9,53 @@ import react.common._
...
 import scala.scalajs.js
+import scala.scalajs.js.JSConverters._
 import scala.scalajs.js.annotation._
 
+sealed trait ReactMarkdownPlugin extends js.Object
+
+object ReactMarkdownPlugin {
+
+  val RemarkGFM =
+    if (scala.scalajs.runtime.linkingInfo.productionMode) RemarkGFMProd else RemarkGFMDev
```
rpiaggio, line 20: Nice

PR #761, opened 2021-04-29
Ch785 add a constraints panel to show the obs constraints

cquiroz:
This PR ads a panel to edit constraints on the observation view. It mostly reuses the UI done for the constraints tab.

It also contains a combo box to quickly assign a constraint set to an obs

NOTE: Editions in this panel will affect directly the set, i.e. all observations using that set will be affected.
This may not be the preferred way but I'll leave that as a task for later.

image

### DIFF
--- a/common/src/main/scala/explore/common/ObsQueries.scala
+++ b/common/src/main/scala/explore/common/ObsQueries.scala
@@ -9,7 +9,6 @@ import explore.AppCtx
 import explore.components.graphql.LiveQueryRenderMod
 import explore.data.KeyedIndexedList
 import explore.implicits._
-import explore.model.ConstraintsSummary
 import explore.model.ObsSummaryWithPointingAndConstraints
 import explore.model.Pointing
 import explore.model.reusability._
@@ -24,6 +23,7 @@ import ObsQueriesGQL._
 object ObsQueries {
 
   type ObservationList = KeyedIndexedList[Observation.Id, ObsSummaryWithPointingAndConstraints]
+  type ConstraintsInfo = List[ProgramObservationsQuery.Data.ConstraintSets.Nodes]
 
   private def convertPointing(
     pointing: ProgramObservationsQuery.Data.Observations.Nodes.ObservationTarget
@@ -35,23 +35,9 @@ object ObsQueries {
         Pointing.PointingAsterism(id, name, Nil)
     }
 
-  implicit class ConstraintSetsObsQueryDataOps(
-    val self: ProgramObservationsQuery.Data.ConstraintSets.Nodes
-  ) extends AnyVal {
-    def asConstraintSummary: ConstraintsSummary =
-      new ConstraintsSummary {
-        val id              = self.id
-        val name            = self.name
-        val imageQuality    = self.imageQuality
-        val cloudExtinction = self.cloudExtinction
-        val skyBackground   = self.skyBackground
-        val waterVapor      = self.waterVapor
-      }
-  }
-
   private val programObservationsQueryoObservationListGetter
-    : Getter[ProgramObservationsQuery.Data, (List[ConstraintsSummary], ObservationList)] = data => {
-    val cs  = data.constraintSets.nodes.map(_.asConstraintSummary)
+    : Getter[ProgramObservationsQuery.Data, (ConstraintsInfo, ObservationList)] = data => {
+    val cs  = data.constraintSets.nodes
     val obs = KeyedIndexedList.fromList(
       data.observations.nodes.map(node =>
         ObsSummaryWithPointingAndConstraints(node.id,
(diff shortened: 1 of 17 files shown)


common/src/main/scala/explore/common/ObsQueries.scala
```diff
@@ -34,9 +35,24 @@ object ObsQueries {
...
+      new ConstraintsSummary {
+        val id              = self.id
+        val name            = self.name
+        val imageQuality    = self.imageQuality
+        val cloudExtinction = self.cloudExtinction
+        val skyBackground   = self.skyBackground
+        val waterVapor      = self.waterVapor
+      }
+  }
+
```
rpiaggio, line 51: This can be avoided by making `ProgramObservationsQuery.Data.ConstraintSets.Nodes` extend `ConstraintsSummary`. Clue's generator can be instructed to do so by declaring `trait Nodes extends ConstraintsSummary` in the object hierarchy.

For example: https://github.com/gemini-hlsw/explore/blob/master/common-graphql/src/main/scala/explore/common/ObsQueriesGQL.scala#L48

model-tests/shared/src/test/scala/explore/model/ConstraintsSummarySuite.scala
```diff
@@ -0,0 +1,12 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats.kernel.laws.discipline.EqTests
+import explore.model.arb.all._
+import munit.DisciplineSuite
+
+class ConstraintsSummarySuite extends DisciplineSuite {
+  checkAll("Eq[ConstraintsSummary]", EqTests[ConstraintsSummary].eqv)
```
rpiaggio, line 11: It seems I unintentionally removed this. Thank you for reinstating!

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -171,120 +174,122 @@ object ConstraintsPanel {
...
-                validFormat = ValidFormatInput
-                  .forRefinedTruncatedBigDecimal[AirMassRange.Value, 1](airMassErrorMsg)
-                  .composeValidFormat(
-                    ValidFormatNec.gte(
-                      TruncatedRefinedBigDecimal[AirMassRange.Value, 1](state.get.airMass.min).get,
-                      "Must be >= Min"
+              ReactFragment(
+                FormInputEV(
+                  id = "minam",
+                  label = Label("Min", HelpIcon("constraints/main/minam.md")),
```
toddburnside, line 192: Aren't these the buttons that Andy just removed?

common/src/main/scala/explore/common/ObsQueries.scala
cquiroz: Ok I'll try that

toddburnside (approved):
👍

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
cquiroz: This is rebased on top of Andy's Pr and I tried to keep those changes but I'll review again

common/src/main/scala/explore/common/ObsQueries.scala
```diff
@@ -45,15 +47,21 @@ object ObsQueries {
...
   implicit class ProgramObservationsQueryDataOps(val self: ProgramObservationsQuery.Data.type)
       extends AnyVal {
     def asObservationList = programObservationsQueryoObservationListGetter
   }
 
   val ObsLiveQuery =
-    ScalaFnComponent[View[ObservationList] ~=> VdomNode](render =>
+    ScalaFnComponent[View[(ConstraintsInfo, ObservationList)] ~=> VdomNode](render =>
```
cquiroz, line 59: Probably next version will be worth putting this in a case class

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -54,7 +57,7 @@ object Tile {
   def renderPortal(mountNode: Option[raw.ReactDOM.Container], info: VdomNode): VdomNode =
     mountNode.map(node => ReactPortal(info, node))
 
-  implicit val rawContainerReuse: Reusability[raw.ReactDOM.Container] = Reusability.always
+  implicit val rawContainerReuse: Reusability[raw.ReactDOM.Container] = Reusability.never
```
cquiroz, line 60: @rpiaggio, why was this `always`?

rpiaggio: The DOM `Element` (`<span>`)  where the portal is mounted will always be the same during the lifetime of the `Tile`. Therefore, by setting this to `always` we ensure that the same `ReactPortal` is used if the `Tile` rerenders for any reason, otherwise it will be recreated.

Was this preventing necessary rerenders?

cquiroz: Actually I didn't see any difference, However the fulll Tile is set to always re render because we don't really know what's inside the body, though perhaps it can be done via `Reuse`

rpiaggio: BTW, now that we have reusable functions maybe we want to derive `propsReuse`? Or use a `.by(...)` so we can take the `render` function into account too.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/Tile.scala
rpiaggio: If we change reuse to be smarter, I believe we should still keep `rawContainerReuse` as `never`.

cquiroz: Ok, I'll do that on another PR, to learn a bit about `Reuse`

PR #762, opened 2021-04-30
Close the help when clicking on the main content

cquiroz:
This is a small improvement to the help system. Now you can click outside the help to close it

### DIFF
--- a/explore/src/main/scala/explore/ExploreLayout.scala
+++ b/explore/src/main/scala/explore/ExploreLayout.scala
@@ -12,6 +12,7 @@ import explore.implicits._
 import explore.model._
 import explore.model.reusability._
 import japgolly.scalajs.react._
+import japgolly.scalajs.react.component.builder.Lifecycle.RenderScope
 import japgolly.scalajs.react.extra.router._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common._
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -68,6 +75,9 @@ object ExploreLayout {
                 props.r.renderP(props.view)
               )
             )
+          )(
+            ^.onClick -->
+              helpView.set(none).runAsyncAndForgetCB.when_($.state.helpVisible)
```
rpiaggio, line 80: What happens if we always set `none` here without tracking the status in the state?

cquiroz: Actually we should be able to simplify. I got a bit confused by this one getting called twice but it was avoided by preventing propagation on the help icon

explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -76,10 +86,11 @@ object ExploreLayout {
   private val component =
     ScalaComponent
       .builder[Props]
-      .stateless
-      .render_P { p =>
-        IfLogged(p.view)((renderFn _).reusable(p))
+      .initialState(State(false))
+      .renderP { ($, p) =>
+        IfLogged(p.view)((renderFn($) _).reusable(p))
```
rpiaggio, line 91: I'm not sure I would recommend this pattern of using `$` to build a reusable function. In this case the state is used in the function, but the state's reusability is not taken into account when computing the reusability of the function.

In other places I've added a `View[State]` as a regular parameter to `renderFn`, and then you can pass a `ViewF.fromState[IO]($)`.

rpiaggio (approved):
LGTM

cquiroz:
This was simplified a lot taking the State out of `ExploreLayout`

PR #763, opened 2021-04-30
Make tile reusable

cquiroz:
As discussed in a previous PR we could enhance tiles to support `Reusability` via `Reusable`. I maybe even able to understand what's going on

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -39,7 +39,7 @@ object Tile {
 
   type RenderInTitle = VdomNode ~=> VdomNode
 
-  implicit val propsReuse: Reusability[Tile] = Reusability.derive
+  implicit val propsReuse: Reusability[Tile] = Reusability.derive && Reusability.by(_.render)
 
   val heightBreakpoints =
     List((200, TileXSH), (700 -> TileSMH), (1024 -> TileMDH))
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -13,19 +13,17 @@ import explore.model.enum.TileSizeState
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.internal.JsUtil
 import japgolly.scalajs.react.vdom.html_<^._
+import lucuma.ui.reusability._
 import org.scalajs.dom.html
 import react.common._
 import react.common.implicits._
 import react.semanticui.collections.menu._
 import react.semanticui.elements.button.Button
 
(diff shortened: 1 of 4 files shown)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -41,8 +39,7 @@ object Tile {
 
   type RenderInTitle = VdomNode ~=> VdomNode
 
-  // Explicitly never reuse as we are not considering the content
-  implicit val propsReuse: Reusability[Tile] = Reusability.never
+  implicit val propsReuse: Reusability[Tile] = Reusability.derive
```
rpiaggio, line 42: This is probably not taking the `render` parameter into account. I think we can add ` && Reusability.by(_.render)` at the end.

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -212,6 +210,25 @@ object ObsTabContents {
         }.getOrEmpty
       }
 
+    def constraintsSelectorFn(
+      constraintsSetId: Option[ConstraintSet.Id],
+      obsSummaryOpt:    Option[ObsSummary],
+      observations:     ConstraintsInfo
+    )(implicit
+      client:           clue.TransactionalClient[IO, explore.schemas.ObservationDB]
```
rpiaggio, line 218: This is great. I hadn't realized that converting a method to a function with `fn _` resolves the implicits. I was jumping through hoops all over to pass implicits to these render functions. I can simplify all that now.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/Tile.scala
rpiaggio: I wonder if this is still the case in Scala 3, which has context functions. I'll give it a try before changing anything.

explore/src/main/scala/explore/tabs/ObsTabContents.scala
rpiaggio: I wonder if this is still the case in Scala 3, which has context functions. I'll check that before changing anything.

toddburnside (approved):
👍

common/src/main/scala/explore/components/Tile.scala
cquiroz: Shouldn't `derive` take into account all fields?

rpiaggio: I'm not sure how the macro is implemented. There is a possibility it only takes a look at the first parameter group.

PR #765, opened 2021-04-30
Experiment with dynamic modules

cquiroz:
I've been tempted to test dynamic module loading for a while and this is my first attempt.
The help system is not used right away and in many cases won't be used. It would be better to load the code and dependencies on demand.

With the changes in this PR we'll create 2 js files, the main containing explore and another with the help windown and its dependencies (ReactMarkdown, et.al.)

Only when you open the help the files are loaded from then net and executed

I noted there maybe some duplication on code we could use to further reduce size.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ common/src/main/publicdev/
 # vite
 server.cert
 server.key
+stats.html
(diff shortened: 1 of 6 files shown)


rpiaggio (approved):
This is totally awesome and it seems quite straightforward to implement. Great job!

LVGTM

toddburnside (approved):
Nice!

cquiroz:
I'll merge and test that works correctly in staging

PR #770, opened 2021-05-03
Reduce read timeout

cquiroz:
Logout isn't working properly. Reducing the timeout seems to fix it though it isn't clear the exact mechanism


rpiaggio (approved):
LGTM

PR #774, opened 2021-05-05
Migrate to Cats Effect 3

rpiaggio:
Explore is working, but we need to wait for an official cats-retry release.

Also, I'm thinking of alternatives to having implicit `Dispatcher`.


common/src/main/scala/explore/components/graphql/LiveQueryRenderMod.scala
```diff
@@ -40,25 +38,21 @@ final case class LiveQueryRenderMod[S, D, A](
...
-    implicit val timer: Timer[F]
-    implicit val cs: ContextShift[F]
-  }
+  trait Props[F[_], S, D, A] extends Render.LiveQuery.Props[F, ViewF[F, *], S, D, A]
 
   final case class State[F[_], S, D, A](
     queue:                   Queue[F, A],
     subscriptions:           NonEmptyList[GraphQLSubscription[F, _]],
-    cancelConnectionTracker: CancelToken[F],
+    cancelConnectionTracker: F[Unit],
```
cquiroz, line 55: I thought the cancel token would be `() => Future[Unit]`

rpiaggio: That's for things that are running from outside of the effect system, started with `unsafe*Run`. In this case, everything is still happening within `F` since it's started with `f.start`. So the `CancelToken` is `fiber.cancel` (which is a `IO[Unit]`).

cquiroz (approved):
Very nice, thanks

common/src/main/scala/explore/components/graphql/LiveQueryRenderMod.scala
rpiaggio: Just to be clear: we could have done things this way before, and it would have been simpler from the start.

toddburnside (approved):
LGTM

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -30,7 +30,7 @@ object UserPreferencesQueries {
   implicit class UserWidthsCreationOps(val self: UserWidthsCreation.type) extends AnyVal {
     import self._
 
-    def storeWidthPreference[F[_]: Effect](
+    def storeWidthPreference[F[_]: ApplicativeError[*[_], Throwable]](
```
cquiroz, line 33: This is nice, a lot less power than `Effect`

common/src/main/scala/explore/utils/package.scala
```diff
@@ -28,17 +28,17 @@ package object utils {
       }
     }
 
+  val gitHash = BuildInfo.gitHeadCommit
```
cquiroz, line 31: 👍

cquiroz (approved):
Looks great, thanks

common/src/main/scala/explore/common/UserPreferencesQueries.scala
rpiaggio: Right, and you need a `Dispatcher` to run it.

explore/src/main/scala/explore/Explore.scala
```diff
@@ -4,36 +4,188 @@
...
+  override final def run: IO[Unit] = {
+    japgolly.scalajs.react.extra.ReusabilityOverlay.overrideGloballyInDev()
+
+    def initialModel(vault: Option[UserVault]) = RootModel(
+      vault = vault,
+      tabs = EnumZipper.of[AppTab]
+    )
+
+    val fetchConfig: IO[AppConfig] = {
+      // We want to avoid caching the static server redirect and the config files (they are not fingerprinted by webpack).
```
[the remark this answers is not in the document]
rpiaggio: I guess the comment is obsolete now. But yes, even with fingerprinting we have the issue of not knowing the resulting name.

PR #775, opened 2021-05-05
Support latex on the help tab

cquiroz:
This is an experiment to support latex expressions on the help window.

Thanks to dynamic load it doesn't increase the size of the core app a lot

image


toddburnside (approved):
Nice addition.

PR #776, opened 2021-05-06
Git hash for version in Heroku.

rpiaggio:


build.sbt
```diff
@@ -63,8 +63,6 @@ lazy val model = crossProject(JVMPlatform, JSPlatform)
   .jsSettings(commonModuleTest: _*)
   .jvmSettings(commonJVMSettings)
 
-val curTime = System.currentTimeMillis()
```
rpiaggio, line 66: This was replaced by `buildInfoOptions += BuildInfoOption.BuildTime`

cquiroz: We moved it out or resource changes would trigger a recompilation, this shouldn't be an issue with vite anymore

cquiroz (approved):

build.sbt
rpiaggio: You're right, I had forgotten about this.

PR #781, opened 2021-05-07
Improvements on tile size management

cquiroz:
The way making tiles minimize and restore wasn't very good. This PR improves the situation by moving the grid to a separate component in charge of managing the tiles

This also makes the constraints tile minimizable

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -4,6 +4,7 @@
 package explore.components
 
 import cats.syntax.all._
+import eu.timepit.refined.types.string.NonEmptyString
 import explore.Icons
 import explore.components.ui.ExploreStyles
 import explore.components.ui.ExploreStyles._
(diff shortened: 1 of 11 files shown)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -21,6 +22,7 @@ import react.semanticui.collections.menu._
 import react.semanticui.elements.button.Button
 
 final case class Tile(
+  id:                Tile.TileId,
```
rpiaggio, line 25: What is the `id` used for?

rpiaggio (approved):
LGTM I look forward to seeing how this works.

common/src/main/scala/explore/components/Tile.scala
cquiroz: It is the same id used in the grid, it makes some bits of the code simpler

One issue I gave is there is a certain duplication of states, this PR removes one by making the state of the tile dependent on the height

PR #782, opened 2021-05-08
Generate BuildInfo only once

cquiroz:
This may seem trivial but when doing SmallestModules vite detects an extra changed module on each compilation

```
/Users/cquiroz/Projects/explore/explore/target/scala-2.13/explore-fastopt/explore.BuildInfo$.js
```


rpiaggio (approved):
LGTM

PR #790, opened 2021-05-11
Fix HMR.

rpiaggio:
Fix HMR by skipping IOApp and running the app directly through the global IORuntime.

Also, release dispatcher on reload.


toddburnside (approved):
I think this really makes more sense than trying to pretend we are an IOApp.

PR #791, opened 2021-05-14
Basic config form

cquiroz:
This PR builds the first form on the configurations tile. This form should have an imaging alternate but at this level it only has spectroscopy

Note there is no persistence or any real use of this data but this PR will let us get feedback on the UI

image

### DIFF
--- a/model/shared/src/main/scala/explore/model/SpectroscopyConfigurationOptions.scala
+++ b/model/shared/src/main/scala/explore/model/SpectroscopyConfigurationOptions.scala
@@ -1,0 +1,50 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats.Eq
+import eu.timepit.refined.types.numeric.PosBigDecimal
+import eu.timepit.refined.types.numeric.PosInt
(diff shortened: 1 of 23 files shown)


model/shared/src/main/scala/explore/model/SpectroscopyConfigurationOptions.scala
```diff
@@ -0,0 +1,50 @@
...
+import eu.timepit.refined.types.numeric.PosInt
+import eu.timepit.refined.cats._
+import explore.model.enum.FocalPlaneOptions
+import explore.model.enum.SpectroscopyCapabilities
+import lucuma.core.math.Angle
+import lucuma.core.math.Wavelength
+import monocle.macros.Lenses
+
+@Lenses
+final case class SpectroscopyConfigurationOptions(
```
rpiaggio, line 17: We will probably switch to using generated classes once there is a model to generate classes from.  But in the meanwhile we do need this.

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/model/SpectroscopyConfigurationOptions.scala
cquiroz: Indeed, and it wouldn't be stored in the state but rather come from a query

toddburnside (approved):
LGTM

PR #799, opened 2021-05-17
Support ESC to close the help

cquiroz:
We've talked for a while to support shortcuts and these should probably be added sooner rather than later.
I'm experimenting with react-hotkey with a simple shortcut to close the help window pressing ESC.

### DIFF
--- a/explore/src/main/scala/explore/ExploreLayout.scala
+++ b/explore/src/main/scala/explore/ExploreLayout.scala
@@ -15,6 +15,7 @@ import japgolly.scalajs.react._
 import japgolly.scalajs.react.extra.router._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common._
+import react.hotkeys._
 import react.semanticui.modules.sidebar.Sidebar
 import react.semanticui.modules.sidebar.SidebarAnimation
 import react.semanticui.modules.sidebar.SidebarDirection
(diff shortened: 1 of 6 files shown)


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -44,39 +45,43 @@ object ExploreLayout {
...
-            direction = SidebarDirection.Right,
-            animation = SidebarAnimation.Overlay,
-            visible = helpView.get.isDefined
-          )(
-            helpView.get
-              .map { h =>
-                // Lazy load the React component for help
-                val prom = js.dynamicImport {
-                  new HelpLoader().loadHelp(helpCtx.get, h)
+        GlobalHotKeys(keyMap = KeyMap("CLOSE_HELP" -> "ESC"),
```
rpiaggio, line 48: We should probably activate this only when the help is open, so that we can use the ESC key for other purposes in other parts of the application.

I think this can be achieved by just moving this `GlobalHotKeys` to the help component.

Also, it is recommended to use `GlobalHotKeys` as a stand alone component (the exception is when there's another `GlobalHotKeys` further down the tree which should override this one): https://github.com/greena13/react-hotkeys#globalhotkeys-component

rpiaggio (approved):
LGTM with some comments

explore/src/main/scala/explore/ExploreLayout.scala
cquiroz: Thanks, I implemented it as you suggested and it seems a lot better

PR #804, opened 2021-05-20
Unselect button in views

rpiaggio:


cquiroz:
How will this work on small screens?

rpiaggio:
> How will this work on small screens?

That's a good question. Now it's shown at the top of the list, the same as in large screens, but maybe it makes more sense to show it in the "right" panel, probably in a more concise form?

On further thought, maybe we should rethink the whole navigation in small screens? If there's a table with all elements when navigating into a view maybe it makes sense to navigate directly there, instead of the tree. Also the left/right navigation may be confusing now that we have tree and table. Maybe the "left" side could be toggled between a table and a tree view?

cquiroz:
On small screens we show only one side, either the tree or the "body"

cquiroz (approved):
LGTM

toddburnside (approved):
👍

PR #809, opened 2021-05-21
Update vite and plugin-commonjs

cquiroz:
The last vite update broke the build but it seems this can be solved by making the version of `plugin-commonjs` explicit


toddburnside (approved):
LGTM

PR #815, opened 2021-05-24
Add extra fields to the imaging options ui

cquiroz:
Adds 3 more fields to the imaging config params ui

### DIFF
--- a/explore/src/main/scala/explore/config/ImagingConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/ImagingConfigurationPanel.scala
@@ -144,9 +144,11 @@ object ImagingConfigurationPanel {
                     HelpIcon("configuration/signal_to_noise.md"),
                     ExploreStyles.SkipToNext
             ),
-            FormInputEV(id = "signal-to-noise",
-                        value = signalToNoise,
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/config/ImagingConfigurationPanel.scala
```diff
@@ -114,6 +128,37 @@ object ImagingConfigurationPanel {
...
+              changeAuditor = ChangeAuditor.fromFormat(formatArcsec).optional,
+              disabled = false
+            ),
+            <.label("S / N",
+                    HelpIcon("configuration/signal_to_noise.md"),
+                    ExploreStyles.SkipToNext
+            ),
+            FormInputEV(id = "signal-to-noise",
+                        value = signalToNoise,
+                        validFormat = ValidFormatInput.fromFormatOptional(formatPosBigDecimal)
```
toddburnside, line 149: No change auditor desired?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/config/ImagingConfigurationPanel.scala
cquiroz: You're right, we should have a `ChangeAuditor` defined here

PR #816, opened 2021-05-25
Alternate way to treat shared constraints. Add labels to all buttons.

rpiaggio:
image


cquiroz:
Starting to look quite busy. Why are we adding labels to the buttons?

rpiaggio:
> Starting to look quite busy. Why are we adding labels to the buttons?

Agreed. Andy requested the labels in the last meeting, if i understood correctly.

toddburnside:
> Agreed. Andy requested the labels in the last meeting, if i understood correctly.

I think what Andy mostly wanted was labels for the `+` buttons at the top of the tree views so it was clear what they do. I don't think the copy/paste/undo/redo, etc. need labels.

rpiaggio:
> > Agreed. Andy requested the labels in the last meeting, if i understood correctly.
> 
> I think what Andy mostly wanted was labels for the `+` buttons at the top of the tree views so it was clear what they do. I don't think the copy/paste/undo/redo, etc. need labels. It's also more consistent with the target view where the add target and add asterism buttons have labels.

He did mention the Undo/Redo buttons. Maybe I didn't understand correctly and he mentioned those as cases where it was clear what they did from the icons, but not the "New" buttons? I'll ask him.

rpiaggio:
BTW, the buttons with icon + label are supposed to look closer to this, but the current style doesn't make both sections very noticeable:

image

rpiaggio:
There was a story requesting labels in buttons which I had missed: https://app.clubhouse.io/lucuma/story/799/add-labels-to-buttons

common/src/main/webapp/less/style.less
```diff
@@ -1504,10 +1502,21 @@ html {
...
-
 .explore-button-summary {
   width: 100%;
 }
+
+.explore-tile-title-button-new {
+  margin-left: 1em !important; // important to override SUI's style.
+}
+
+// This fixes what seems to be a bug in SUI (React?), where compact button groups are still very wide.
```
cquiroz, line 1513: Is it mentioned as a bug somewhere? Perhaps in FomanticUI has been fixed?

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -194,76 +191,49 @@ object ConstraintsPanel {
...
+                s"These constraints cannot be modified because they are shared by $obsCount observations.",
+                <.br,
+                "Please edit in Constraints View, create a New one or a Copy."
+              )
+            ),
+            wide = PopupWide.Very,
+            position = PopupPosition.TopCenter
           )
-          .when(props.multiEditWarnings)
+        ).when(!props.allowMultiEdit)
```
cquiroz, line 210: `.unless(props....)`? It is just taste feel free to ignore

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -194,76 +191,49 @@ object ConstraintsPanel {
...
           FormInputEV(
             id = "name",
             label = Label("Name", HelpIcon("constraints/main/name.md")),
             clazz = ExploreStyles.ConstraintsNameField,
             value = nameView,
             validFormat = ValidFormatInput.nonEmptyValidFormat,
             errorClazz = ExploreStyles.InputErrorTooltip,
-            errorPointing = LabelPointing.Below
+            errorPointing = LabelPointing.Below,
+            disabled = readonly
```
cquiroz, line 236: I wonder if you can disable the whole form in one go, I suspect no

explore/src/main/scala/explore/tabs/ConstraintsTile.scala
```diff
@@ -18,39 +23,77 @@ import explore.components.ui.ExploreStyles
...
-            AssignConstraintSetToObs
-              .execute(csId, obsId.id)
-              .runAsyncAndForgetCB *> Callback.log(s"Set to $csId")
-          }.getOrEmpty,
-        options = observations.map(s => new SelectItem(value = s.id.show, text = s.name.value))
+      <.span(
+        Select(
+          value = constraintSetId.map(_.show).orEmpty, // Set to empty string to clear
+          placeholder = "Select a constraint set",
+          // size = sizes.Small,
```
cquiroz, line 77: Commented code?

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -172,7 +169,7 @@ object ConstraintsPanel {
                 constraintSetId = id.assign,
                 // programId = constraintSet.get.programId,
                 programId = "p-2",
-                name = NonEmptyString.unsafeFrom(constraintSet.get.name + " - Copy"),
+                name = NonEmptyString.unsafeFrom(s"${constraintSet.get.name} [$id]"),
```
cquiroz, line 172: Do we want to show the `id` in the combobox?

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
rpiaggio: Good catch! Will change.

cquiroz (commented):
The approach around making it explicit that you need to copy before edíitng is worth testing.
I'm a bit worried on how busy it looks, I'd like to keep the tiles relatively clean, in fact I'd prefer not have undo buttons and use shortcuts instead tough that brings another set of issues.

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
rpiaggio: I investigated that initially, but I couldn't find a way.

explore/src/main/scala/explore/tabs/ConstraintsTile.scala
rpiaggio: Ah, good catch. The commented code doesn't work; I left it commented out while investigating adding it to the facade, but React SUI doesn't really support it.

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
rpiaggio: If we show the id the combobox we can keep the same name as the original when copying. Is that what you are suggesting?

rpiaggio:
> The approach around making it explicit that you need to copy before edíitng is worth testing.
> I'm a bit worried on how busy it looks, I'd like to keep the tiles relatively clean, in fact I'd prefer not have undo buttons and use shortcuts instead tough that brings another set of issues.

We could definitely hide the undo/redo buttons when it's in read-only mode (I'll do that). So, the "Shared" warning and the undo/redo buttons will never be there at the same time. This will unclutter a bit.

common/src/main/webapp/less/style.less
rpiaggio: I dag a bit deeper and it's not actually a bug. The thing is, labeled+`icon` buttons should actually have a `labelPosition` or they are not rendered as expected (they are rendered much smaller, which is what I want in this case).

So, I'm fixing this to introduce a `verycompact` style that we can apply to buttons to render them as we want.

rpiaggio:
Ok, now things look either like this:
image

or like this:
image

cquiroz (approved):
Let's user test this

PR #818, opened 2021-05-25
Sort selected frequencies by central wavelength

cquiroz:

### DIFF
--- a/model/shared/src/main/scala/explore/model/ImagingConfigurationOptions.scala
+++ b/model/shared/src/main/scala/explore/model/ImagingConfigurationOptions.scala
@@ -3,6 +3,7 @@
 
 package explore.model
 
+import cats.Order
 import cats.syntax.all._
 import coulomb.Quantity
 import eu.timepit.refined.types.numeric.PosBigDecimal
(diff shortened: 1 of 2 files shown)


model/shared/src/main/scala/explore/model/ImagingConfigurationOptions.scala
```diff
@@ -24,16 +27,21 @@ sealed abstract class AvailableFilter {
   val range: Option[Quantity[Int, Nanometer]]
 }
 
+object AvailableFilter {
+  implicit val order: Order[AvailableFilter]       = Order.by(x => (x.centralWavelength, x.tag))
+  implicit val ordering: Ordering[AvailableFilter] = order.toOrdering
```
rpiaggio, line 32: I'm almost sure that cats provides an implicit conversion from `Order` to `Ordering` if you import `cats.syntax.all._` (probably?) at the use site.

rpiaggio (approved):
LGTM

toddburnside (approved):
LGTM

model/shared/src/main/scala/explore/model/ImagingConfigurationOptions.scala
cquiroz: That's correct but you need to import `cats.implicits._` for that to work, Strange

PR #819, opened 2021-05-26
Targets summary table

rpiaggio:
There's a bunch of interactions between the summary table and the tree on the left that we have to take care of.

image

Coming in next PR: attempting to make columns selectable.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
+++ b/explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
@@ -9,6 +9,7 @@ import japgolly.scalajs.react.vdom.html_<^._
 import explore.common.TargetObsQueries._
 import reactST.reactTable._
 import explore.model.Focused
+import explore.model.ExpandedIds
 import crystal.react.implicits._
 import react.semanticui.collections.table._
 import lucuma.core.model.Target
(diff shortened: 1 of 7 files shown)


explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
```diff
@@ -0,0 +1,146 @@
...
+          )
+          .setHeader("Observations")
+      ).toJSArray
+
+      TargetTableComponent(
+        Table(celled = true, selectable = true, striped = true, compact = TableCompact.Very),
+        header = true
+      )(TargetTable.Options(columns, props.pointingsWithObs.targets.toList.toJSArray))
+    }
+  // .builder[Props]
```
cquiroz, line 136: Commented code?

cquiroz (approved):
Looks good, I guess it will have more iterations like the idea of filtering Columns, etc

explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
```diff
@@ -0,0 +1,146 @@
...
+import reactST.reactTable._
+
+import scalajs.js.JSConverters._
+
+final case class TargetSummaryTable(
+  pointingsWithObs: PointingsWithObs,
+  focused:          View[Option[Focused]],
+  expandedIds:      View[ExpandedIds]
+)(implicit val ctx: AppContextIO)
+// extends ReactProps[TargetSummaryTable](TargetSummaryTable.component)
```
toddburnside, line 32: Comment here, too.

toddburnside (approved):
LGTM

PR #820, opened 2021-05-26
Unfocus when navigating to deleted element.

rpiaggio:
I'm not entirely sure if this logic goes in the tree. It seems like the best place given the circumstances: namely, all the information is there at the right time (after the tree data is retrieved).

Otherwise, we have to create an intermediate wrapper for each view.


cquiroz (approved):
Great, thanks

PR #830, opened 2021-06-02
Ch932 Add columns not in spec + remember hidden columns + revive Reuse

rpiaggio:
In this PR:
* More columns are added to the target summary table.
* Hidden columns are stored globally.
* `SUITable` now detects sortBy and *Layout plugins, and default renderers react accordingly.
* Fix rendering into tile titles via portals, without needing to render the tile contents twice.
* `Reuse` makes comeback (in `lucuma-ui`).

It turns out that `Reusable` didn't do what I though it did. Or, more correctly, it did too much of it: when comparing functions for reusability it checked reference equality, besides the `Reusability` of the instantiated parameters. This forces us to define all reusable functions as `val`s, which IMHO interferes with readability and mainteinability (and nothing prevents passing a non-stable reference, which silently nullifies reusability).

I revived our old `Reuse` in `lucuma-ui`, https://github.com/gemini-hlsw/lucuma-ui/pull/270, which Explore now uses. `Reuse` can be now converted into a `Reusable`, in case it's needed.

Defining reusable functions inline allows us to use different idioms, depending on the usage. I converted some of them, but many other existing uses could be converted too.


cquiroz (approved):
LGTM

PR #834, opened 2021-06-03
Fix sorting in target summary table. Decouple table values from render function.

rpiaggio:
- Decouple values and rendering in table columns.
- Correctly sort columns in target summary table.

### DIFF
--- a/explore/src/main/scala/explore/config/SequenceTable.scala
+++ b/explore/src/main/scala/explore/config/SequenceTable.scala
@@ -92,6 +92,8 @@ object SequenceTable {
 
   private val StepTableComponent = new SUITable(StepTable)
 
+  import StepTable.syntax._
+
   val component =
     ScalaFnComponent[Props] { props =>
       val bracketDef =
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/config/SequenceTable.scala
```diff
@@ -116,60 +118,63 @@ object SequenceTable {
...
         StepTable
-          .Column(
-            "guide",
-            s => if (s.guided) Icons.Crosshairs.copy(clazz = ExploreStyles.StepGuided).raw else null
-          )
-          .setHeader(""),
+          .Column("guide", _.guided)
+          .setHeader("")
+          .setCell(cell =>
+            if (cell.value) Icons.Crosshairs.copy(clazz = ExploreStyles.StepGuided) else EmptyVdom
```
cquiroz, line 140: Will `Icons.Cros...).when(cell.value)` work?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/config/SequenceTable.scala
rpiaggio: Unfortunately, it doesn't, since `when` returns a `TagMod` and we need a `VdomNode` here.

toddburnside (approved):
LGTM

PR #836, opened 2021-06-04
Switch target form to have labels next to controls and a few small fixes

cquiroz:
This PR converts the target forms to have the labels next to the field

image

### DIFF
--- a/common/src/main/scala/explore/components/HelpIcon.scala
+++ b/common/src/main/scala/explore/components/HelpIcon.scala
@@ -17,31 +17,29 @@ import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common._
 
-final case class HelpIcon(id: Help.Id) extends ReactProps[HelpIcon](HelpIcon.component)
+final case class HelpIcon(id: Help.Id)
 
 object HelpIcon {
   type Props = HelpIcon
 
   type HelpId = NonEmptyFiniteString[20]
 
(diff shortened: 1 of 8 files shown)


common/src/main/scala/explore/components/HelpIcon.scala
```diff
@@ -17,31 +17,29 @@ import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common._
 
-final case class HelpIcon(id: Help.Id) extends ReactProps[HelpIcon](HelpIcon.component)
+final case class HelpIcon(id: Help.Id)
```
cquiroz, line 20: @rpiaggio I converted this to a functional component for "fun", Does it seem ok?

toddburnside (approved):
LGTM

common/src/main/scala/explore/components/HelpIcon.scala
rpiaggio: I wonder what your definition of fun is.

It looks great. I'm planning to generalize the implicit conversion into `react-common` by having `ReactProps` accept a functional component as parameter, but until we have that we have to manually provide the conversion.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/HelpIcon.scala
toddburnside: It's a `fun`ctional component.

PR #837, opened 2021-06-04
Sticky columns.

rpiaggio:
Kapture 2021-06-04 at 18 15 57

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
+++ b/explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
@@ -103,7 +103,6 @@ object TargetSummaryTable {
           column("type", _ => ())
             .setCell(_ => Icon("star"))
             .setWidth(30),
-          // .setClassName("sticky"),
           column("name", TargetResult.name.get)
             .setCell(cell =>
               <.a(^.onClick ==> (_ =>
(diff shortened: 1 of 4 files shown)


cquiroz:
What is sticky in this context? That the first column always shows up?

explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
```diff
@@ -93,7 +100,10 @@ object TargetSummaryTable {
 
       val columns =
         (List(
-          column("type", _ => ()).setCell(_ => Icon("star")),
+          column("type", _ => ())
+            .setCell(_ => Icon("star"))
+            .setWidth(30),
+          // .setClassName("sticky"),
```
toddburnside, line 106: Leftover comment? Another one below, too.

toddburnside (approved):
Very cool.

toddburnside:
> What is sticky in this context? That the first column always shows up?

**Two** columns! 😄

explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
rpiaggio: yup, thx. will remove

rpiaggio:
> What is sticky in this context? That the first column always shows up?

Exactly, sticky columns cannot be scrolled out. We may want sticky headers too.

PR #841, opened 2021-06-08
New schema without sharing ConstraintSets.

rpiaggio:
* Constraint Sets are now part of Observations.
* Everything should be working in Observation View.
* Constraints View is "Under Construction" for now. I want to rethink it with multiple selection and bulk editing.
* Live Query/Subscription components now use Reuse functions throughout and have the `Pot` rendering logic factored out, so we can have intermediate components between them and where the `Pot` is actually used.


cquiroz (approved):
Looks great, it is a big change :D

toddburnside (approved):
LGTM.

rpiaggio:
Included in #853

PR #842, opened 2021-06-11
Spectroscopy mode algorithm

cquiroz:
This is the first pass to port the mode selection algorithm. There are some parts missing like scoring but it gives the basis to start working on the UI

We may want to move this to a separate lib if we want to build a graphql backend


rpiaggio:
Nice.

Is the CSV only needed for testing or does it contain data needed by the algorithm?

rpiaggio (approved):
LGTM

cquiroz:
Right now only for testing but I want to load it during execution as well

PR #852, opened 2021-06-16
Attempts to run CI when getting PRs from scala-steward

cquiroz:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,12 @@
 name: ci
 
 on:
-  push
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -1,7 +1,12 @@
 name: ci
 
 on:
-  push
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
rpiaggio, line 9: Won't this disable CI if we attempt a PR on another branch? (I don't know if omitting this makes the default all of them or any of them)

cquiroz: I think so. Do we ever do that, maybe for intermediate branches?

cquiroz: Maybe this is a better config:
```
on: [push, pull_request]
```

.mergify.yml
```diff
@@ -2,7 +2,7 @@
 pull_request_rules:
   - name: automatically merge scala-steward's PRs
     conditions:
-      - author=geminihlswadmin
+      - author=scala-steward
```
rpiaggio, line 5: Good catch. I missed this when switching back to Scala Steward.

.github/workflows/ci.yml
rpiaggio: We have done "PR on top of PR".

PR #859, opened 2021-06-17
Table to show the available configuration modes

cquiroz:
This is the first pass of a table to show the configurations.
At the moment it only shows the list of all configurations as we need to refactor the way the selection is made as well as the list of columns displayed

image


toddburnside (approved):
LGTM

PR #860, opened 2021-06-18
Use coulomb refined

cquiroz:
I'm getting a spurious error in heroku on the line below. This change seems to fix it


toddburnside (approved):
Seems odd that the original would cause any errors. But, this is a little nicer, anyway. 😄

PR #873, opened 2021-06-22
Setup Font awesome pro

cquiroz:
We bought a license for Font Awesome Pro and this PR sets it up. It took longer than I expected and there are still some bits missing but at least now we can remove all SUI icons

There is a bit of mismatch between SUI and the new font awesome pro and it required reviewing each icon

Note that Font Awesome Pro can be used as a webfont but I opted to use the React component which uses svg instead. This should allow us to reduce the size of the css doing tree-shaking but that's still pending

Now we'll need to setup a license on each of our dev machines and it is also setup on CI. There maybe some issues for PRs from forks that may need some extra work

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,8 @@ val reactJS              = "16.13.1"
 val FUILess              = "2.8.7"
 val kindProjectorVersion = "0.13.0"
 
+ThisBuild / Test / bspEnabled := false
+
 addCommandAlias(
   "quickTest",
   "modelTestsJVM/test"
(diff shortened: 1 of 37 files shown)


common/src/main/scala/react/fa/FontAwesomeIcon.scala
```diff
@@ -0,0 +1,266 @@
...
+import scala.scalajs.js
+import scala.scalajs.js.JSConverters._
+import scala.scalajs.js.annotation._
+import scala.scalajs.js.|
+
+/**
+ * Facade for the react component for FontAwesomeIcons
+ * See: https://fontawesome.com/v5.15/how-to-use/on-the-web/using-with/react
+ */
+final case class FontAwesomeIcon(
```
cquiroz, line 21: This should go to lucuma-ui

build.sbt
```diff
@@ -7,6 +7,8 @@ val reactJS              = "16.13.1"
 val FUILess              = "2.8.7"
 val kindProjectorVersion = "0.13.0"
 
+ThisBuild / Test / bspEnabled := false
```
toddburnside, line 10: What prompted this?

cquiroz: I'm trying to get metals to use bsp as it seems more reliable than bloop but it chokes on tests that use node modules

toddburnside (approved):

rpiaggio (approved):
LGTM 🥳

Thank you for doing this tedious work.

PR #879, opened 2021-06-23
Update README and cache headers

cquiroz:
Checked that no cache headers are set in production. This PR should fix it
I also added some bits of documentation to the README


rpiaggio (approved):
LGTM

PR #881, opened 2021-06-24
Rename Display to Visible

cquiroz:
We already have an enum `Display` in lucuma core, we should rather rename this one


rpiaggio (approved):
LGTM

PR #882, opened 2021-06-24
Status dropdown in observation badges.

rpiaggio:
* Now showing status and duration from the ODB in badges.
* Also in this PR: `Action` which should allow better organization of code when using undoer (will refactor existing usages in future PR).
* Remove Logger/AppContext from all the places where it's no longer needed because we use sync effects now.

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -654,6 +654,7 @@ tfoot {
   margin: 0 0;
   background-color: var(--obs-badge-background);
   border: solid 1px var(--site-border-color);
+  font-size: inherit;
 }
 
 .card.obs-badge:last-child {
@@ -1370,7 +1371,12 @@ html {
 }
 
 .ui.button.delete-button.obs-delete-button {
-  margin-left: 0;
+  margin: 0;
+  padding: 0 !important;
+
+  svg {
+    margin: 0 !important;
+  }
 }
 
 .crash-message {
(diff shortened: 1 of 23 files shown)


cquiroz:
Do you have a screenshot?

common/src/main/webapp/less/style.less
```diff
@@ -1370,7 +1371,12 @@ html {
 }
 
 .ui.button.delete-button.obs-delete-button {
-  margin-left: 0;
+  margin: 0;
+  padding: 0 !important;
+
+  svg {
+    margin: 0 !important;
```
cquiroz, line 1378: Weird that you need an `important` here

explore/src/main/scala/explore/observationtree/TreeComp.scala
```diff
@@ -113,7 +113,7 @@
 //             setter
 //               .set(tree,
 //                    getSet.getter.get,
-//                    { value: props.treeMod.ElemWithIndex =>
+//                    { value: props.treeMod.ElemWithIndexOpt =>
```
cquiroz, line 116: Change in commented code?

rpiaggio:
> Do you have a screenshot?

image

explore/src/main/scala/explore/observationtree/TreeComp.scala
rpiaggio: If we ever uncomment it, we want that change ;)

common/src/main/webapp/less/style.less
rpiaggio: There's another style from SUI that seems to take precedence otherwise... Maybe there's another way around this?

image

toddburnside (approved):
LGTM

rpiaggio:
I'm still looking for a decent progress bar with discrete steps. Maybe we can reuse the N&S one from seqexec, although it seems a bit overkill.

toddburnside:
> it seems a bit overkill.

If it's worth doing, it's worth overdoing.

PR #892, opened 2021-06-25
Actions in ObsList

rpiaggio:
Refactor `ObsList` to use undoer's `Action` in the hopes of making code more legible/maintainable.

An `Action` is just a wrapper for:
* An `access` (which is just a `getter` and a `setter`, and can be provided in the form of separate functions, a `Lens`, `GetAdjust`, or other optics in the future).
* An `onSet` effect.
* An optional `onRestore` effect (if omitted, `onSet` will be used upon restore).

This is all the information an `UndoContext` needs to perform an undoable action (and undo/redo it). Maybe there's a better name than `Action` for this. Suggestions welcome.

I'd like to know if you find the resulting code easier to read/write than the original one.

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -654,7 +654,7 @@ tfoot {
   margin: 0 0;
   background-color: var(--obs-badge-background);
   border: solid 1px var(--site-border-color);
-  font-size: inherit;
+  font-size: smaller;
 }
 
 .card.obs-badge:last-child {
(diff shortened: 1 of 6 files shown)


common/src/main/webapp/less/style.less
```diff
@@ -654,7 +654,7 @@ tfoot {
   margin: 0 0;
   background-color: var(--obs-badge-background);
   border: solid 1px var(--site-border-color);
-  font-size: inherit;
+  font-size: smaller;
```
rpiaggio, line 657: This is unrelated. It's a fix to a previous PR which resulting in showing text too big in badges.

toddburnside (approved):
LLGTM (Looks like gibberish to me). I don't fully understand it, but it looks OK as far as I can tell.

rpiaggio:
Well, the idea is that code is now more understandable. Should I keep refactoring? ;)

toddburnside:
> Well, the idea is that code is now more understandable. Should I keep refactoring? ;)

Do you think you can get it down to my level? 😮 

One of these days I need to sit down and go through the undo stuff again and renew my understanding of it. But, I don't think you should hold your PR up until then.

cquiroz (approved):
LGTM

PR #898, opened 2021-06-30
Show spectroscopy modes in the table

cquiroz:
This shows the modes for spectroscopy in the config tables. Some notes:

* There is no  filtering nor sorting yet
* The UI is not finished and feedback is welcome
* I'm trying an encoding with Dependent types to avoid existentials and seems to be working fine, again feedback is welcome

### DIFF
--- a/model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
+++ b/model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
@@ -13,17 +13,20 @@ import eu.timepit.refined.types.numeric._
 import eu.timepit.refined.types.string._
 import explore.model.enum.SpectroscopyCapabilities
 import fs2.data.csv._
+import lucuma.core.enum.F2Disperser
+import lucuma.core.enum.GmosSouthDisperser
+import lucuma.core.enum.GnirsDisperser
+import lucuma.core.enum.GpiDisperser
 import lucuma.core.enum.ImageQuality
 import lucuma.core.enum.Instrument
 import lucuma.core.math.Angle
 import lucuma.core.math.Wavelength
 import lucuma.core.math.units._
 import lucuma.core.util.Enumerated
-import monocle.macros.GenLens
-import spire.math.Rational
 import monocle.Getter
 import monocle.Lens
-import lucuma.core.enum.GmosSouthFpu
+import monocle.macros.GenLens
+import spire.math.Rational
 
 sealed trait FocalPlane extends Product with Serializable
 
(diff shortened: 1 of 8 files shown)


model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
```diff
@@ -35,10 +46,141 @@ object FocalPlane {
     Enumerated.of(SingleSlit, MultipleSlit, IFU)
 }
 
+trait InstrumentRow {
+  def instrument: Instrument
+
+  type Disperser
+  val disperser: Disperser
```
swalker2m, line 53: I'm not familiar with doing this kind of thing.  What does having abstract types like this do for us over having instrument-specific `InstrumentRow` types in an ADT?   Instead of matching on the `InstrumentRow` you match on the different `Disperser` types but in that case there are no guarantees about what type possibilities may exist.

Nothing prevents you from having a GMOS `Filter` and an F2 `Disperser` for example?

If I have an `InstrumentRow` containing a GMOS longslit definition, how do I turn it into a GMOS configuration?

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
```diff
@@ -35,10 +46,141 @@ object FocalPlane {
...
+          }
+        }
+      case i @ Instrument.GmosSouth  =>
+        (decodeGmosSouthDisperser(disperser0), decodeGmosSouthFilter(filter0)).mapN { case (d, f) =>
+          new InstrumentRow {
+            val instrument = i
+            type Disperser = GmosSouthDisperser
+            val disperser = d
+            type Filter = Option[GmosSouthFilter]
+            val filter = f
```
swalker2m, line 127: These `InstrumentRow`s are spectroscopy-specific?  I think imaging will allow, for example, multiple filters.

swalker2m (approved):
My ignorance generated a couple of questions but I don't want to stand in the way.  Ignoring the UI aspects which I really cannot pretend to comment upon then .... :+1:

cquiroz:
This is a bit of an experiment but it maybe easier as you said to have a simple ADT rather than type variables. Good point that I may need to implement an equivalent version for imaging

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
cquiroz: I was hoping you had an answer for how to turn this into a config. 😀

PR #904, opened 2021-07-02
Fixes

cquiroz:
We are entering wavelengths in microns so values like 0.3 are common but when you start typing with a 0 it fails as it is not possible to have a wavelength 0.
This PR fixes it by storing the values in the interim as a Quantity where a value of 0 is legal
This also fixes a deployment issue with the csv table


cquiroz:
I'll merge it anyway for testing

model/shared/src/main/scala/explore/model/SpectroscopyConfigurationOptions.scala
```diff
@@ -4,36 +4,52 @@
...
 import explore.model.enum.SpectroscopyCapabilities
 import explore.modes.FocalPlane
 import lucuma.core.math.Angle
 import lucuma.core.math.Wavelength
+import lucuma.core.math.units.Micrometer
 import monocle.macros.Lenses
 
+/**
+ * We want to store wavelengths in units to allow 0 in the UI
+ * but later we convert these units for calculations were Wavelength 0 is not allowed
```
swalker2m, line 21: We need to allow 0 in the UI as a temporary state?

swalker2m (commented):
Belated, but looks good. :+1:

model/shared/src/main/scala/explore/model/SpectroscopyConfigurationOptions.scala
cquiroz: Yes, when you start typing e.g. 0.2, the input tries to convert the first 0 into a 0 wavelength and it fails thus forbidding the 0 to be entered which is not very good
In this model 0 is allowed but ignored until a valid wavelength is entered
This is noticeable here because we are using microns. With nanometers you most likely wouldn't see it

PR #915, opened 2021-07-06
Fix Reuse being too agressive

cquiroz:
This makes the table not refresh if the wavelength param changes


swalker2m (approved):
Looks great, understanding that I don't know what I'm looking at.  Pending CI :+1:

cquiroz:
`Reusability` is an optimization to avoid refreshes on the screen unless strictly needed. In this case we only would do refreshes if the table rows would change but this missed that we now had columns that are calculated thus they change if e.g. the wavelength changes

PR #929, opened 2021-07-11
Monocle3

cquiroz:
Switch to Monocle 3, note this is not ready to be merged as we need to also upgrade scalajs-react


toddburnside:
Upgrade done in #954

PR #935, opened 2021-07-13
Magnitude list editor

toddburnside:
This accommodates the changes in https://github.com/gemini-hlsw/lucuma-tmp-api/pull/318. However, it only makes use of `replaceList` to replace the entire magnitudes list. This is equivalent to the previous editing method. A different PR may be created to utilize the more fine-grained editing afforded by `editList`.


swalker2m (commented):
Makes sense to do it this way to me as a first step, and the code looks good to me. :+1:  I'll let Carlos review since it's in the UI.

cquiroz (approved):
LGTM, thanks

PR #938, opened 2021-07-18
Store the spectroscopy parameters in the backend

cquiroz:
This PR updates the UI to read ans save the parameters for the spectorscopy mode in the backend

### DIFF
--- a/common-graphql/src/main/scala/explore/schemas/ObservationDB.scala
+++ b/common-graphql/src/main/scala/explore/schemas/ObservationDB.scala
@@ -35,46 +35,52 @@ trait ObservationDB {
     type DatasetFilename  = String
     // Refined
     type NonEmptyString   = eu.timepit.refined.types.string.NonEmptyString
+    type PosInt           = eu.timepit.refined.types.numeric.PosInt
+    type PosBigDecimal    = eu.timepit.refined.types.numeric.PosBigDecimal
     // Time
     type Instant          = java.time.Instant
   }
 
   object Enums {
-    type CatalogName         = enum.CatalogName
-    type CloudExtinction     = enum.CloudExtinction
-    type EphemerisKeyType    = enum.EphemerisKeyType
-    type GcalArc             = enum.GcalArc
-    type GcalContinuum       = enum.GcalContinuum
-    type GcalDiffuser        = enum.GcalDiffuser
-    type GcalFilter          = enum.GcalFilter
-    type GcalShutter         = enum.GcalShutter
-    type GmosAmpCount        = enum.GmosAmpCount
-    type GmosAmpReadMode     = enum.GmosAmpReadMode
-    type GmosCustomSlitWidth = enum.GmosCustomSlitWidth
-    type GmosDetector        = enum.GmosDetector
-    type GmosDtax            = enum.GmosDtax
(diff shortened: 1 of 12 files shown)


common-graphql/src/main/scala/explore/schemas/ObservationDB.scala
```diff
@@ -35,46 +35,52 @@ trait ObservationDB {
...
+    type InstrumentType           = enum.Instrument
+    type MagnitudeBand            = enum.MagnitudeBand
+    type MagnitudeSystem          = enum.MagnitudeSystem
+    type MosPreImaging            = enum.MosPreImaging
+    type ObsStatus                = enum.ObsStatus
+    type ObsActiveStatus          = enum.ObsActiveStatus
+    type SkyBackground            = enum.SkyBackground
+    type StepType                 = enum.StepType
+    type WaterVapor               = enum.WaterVapor
+    // temporal
```
rpiaggio, line 80: temporal because they will make it to `lucuma-core`?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -46,62 +51,89 @@ import sttp.model._
...
-        _ => IO.unit
-      )
-
-    def apply[A](
-      lens: Lens[State, A]
-    ): View[A] =
-      apply(lens.get, lens.modify)
-
-  }
+  val dataIso: Iso[SpectroscopyRequirementsData, SpectroscopyConfigurationOptions] =
```
rpiaggio, line 81: Why are there 2 classes?

explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -46,62 +51,89 @@ import sttp.model._
...
+    Iso[SpectroscopyRequirementsData, SpectroscopyConfigurationOptions] { s =>
+      def wavelengthToMicro(w: Wavelength) = w.micrometer.toValue[BigDecimal]
```
rpiaggio, line 83: Can we use an `Iso[Wavelength, Quantity[BigDecimal, Micrometer]]` ?

common-graphql/src/main/scala/explore/schemas/ObservationDB.scala
cquiroz: Yes, once we do the big merge with the monocle update

explore/src/main/scala/explore/config/ConfigurationPanel.scala
cquiroz: I should have documented this. On the UI we need to support a plain 0 which is an illegal `Wavelength` value. Thus on the UI model we use a `Quantity` which is legal with value `0`

explore/src/main/scala/explore/config/ConfigurationPanel.scala
cquiroz: You can't do that. `Wavelength` of 0 is illegal while `Quantity` of 0 is

rpiaggio: I see, I missed that. It would have to be a `Prism` then. Or we could leave it as it is.

PR #954, opened 2021-07-22
Bump explore and all dependencies to versions supporting monocle 3 and scalajs-react 2.

rpiaggio:


explore/src/main/scala/explore/observationtree/obsTree.scala
```diff
@@ -5,26 +5,35 @@ package explore.observationtree
 
 import explore.model.ObsSummary
 import monocle.Lens
-import monocle.macros.Lenses
+import monocle.macros.GenLens
```
cquiroz, line 8: Maybe we should use `Focus` everywhere

cquiroz (approved):
LGTM

explore/src/main/scala/explore/observationtree/obsTree.scala
rpiaggio: Oh, I missed we were using `GenLens`. Is there a difference with `Focus`, or just the new name?

rpiaggio:
I should mention that this PR attempts to be a minimal set of changes to allow migration (and unblock the new features).

However, we still have some overlap between crystal and scalajs-react 2, albeit with slightly different approaches and with the latter probably enables simpler ways of doing things.

I'm going to look at these next and try to coalesce things into scalajs-react, removing from crystal what's no longer necessary.

PR #958, opened 2021-07-26
Update to CE3 and revert fix to reset the IO runtime

cquiroz:
Cats effect 3.2.0 provides a way to reset the js global runtime. We can use this in the context of HMR

### DIFF
--- a/explore/src/main/scala/explore/Explore.scala
+++ b/explore/src/main/scala/explore/Explore.scala
@@ -4,9 +4,9 @@
 package explore
 
 import cats.effect.IO
+import cats.effect.IOApp
 import cats.effect.SyncIO
 import cats.effect.std.Dispatcher
-import cats.effect.unsafe.IORuntime
 import cats.syntax.all._
 import cats.~>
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/Explore.scala
```diff
@@ -201,13 +200,14 @@ object ExploreMain {
...
       }
-      .onError { t =>
+      .void
+      .handleErrorWith { t =>
         logger.error(t)("Error initializing") >>
           crash(s"There was an error initializing Explore:<br/>${t.getMessage}")
       }
   }
+
+  // TODO Release dispatcher on stop.
```
rpiaggio, line 212: Given the nature of web apps, I think we can get away with not implementing this.

rpiaggio (approved):
Nice LGTM

explore/src/main/scala/explore/Explore.scala
cquiroz: Right, I wasn't sure about that

toddburnside (approved):
LGTM. 😄

PR #959, opened 2021-07-26
Attempt increase of node memory to 6G.

rpiaggio:


toddburnside (approved):
Worth a try.

cquiroz:
I guess this should be changed in heroku for this to take effect. Heroku doesn't look at GHA config

rpiaggio:
> I guess this should be changed in heroku for this to take effect. Heroku doesn't look at GHA config

Ah ok, for some reason this comment had not made it into my open PR tab when I merged, so I read it after merging. I'll change in Heroku too.

PR #970, opened 2021-08-02
Updates a few libs and try to reduce the amount of work in production…

cquiroz:
This updates a few libs including vite but still cannot find the solution to go beyond 2.3.0
It will also skip running the visualization plugin in production thus saving a few cycles

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -18,10 +18,12 @@ const fontImport = ViteFonts({
 // https://vitejs.dev/config/
 export default ({ command, mode }) => {
   const scalaClassesDir = path.resolve(__dirname, "explore/target/scala-2.13");
+  const isProduction = mode == "production";
   const sjs =
-    mode == "production"
+    isProduction
       ? path.resolve(scalaClassesDir, "explore-opt")
       : path.resolve(scalaClassesDir, "explore-fastopt");
+  const rollupPlugins = isProduction ? [] : [visualizer()];
   const common = path.resolve(__dirname, "common/");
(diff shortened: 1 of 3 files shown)


vite.config.js
```diff
@@ -110,7 +112,7 @@ export default ({ command, mode }) => {
     },
     build: {
       rollupOptions: {
-        plugins: [visualizer()],
+        plugins: rollupPlugins
```
rpiaggio, line 115: `rollupPlugins` is defined in `rollup` and is where the logic is to run different things in each environment?

rpiaggio (approved):
LGTM

vite.config.js
```diff
@@ -18,10 +18,12 @@ const fontImport = ViteFonts({
 // https://vitejs.dev/config/
 export default ({ command, mode }) => {
   const scalaClassesDir = path.resolve(__dirname, "explore/target/scala-2.13");
+  const isProduction = mode == "production";
   const sjs =
-    mode == "production"
+    isProduction
       ? path.resolve(scalaClassesDir, "explore-opt")
       : path.resolve(scalaClassesDir, "explore-fastopt");
+  const rollupPlugins = isProduction ? [] : [visualizer()];
```
cquiroz, line 26: @rpiaggio Here we select that in production we don't run the visualizer

toddburnside (approved):
LGTM

vite.config.js
rpiaggio: Missed that, thank you! 👍

PR #985, opened 2021-08-04
Update aladin-lite to work in production

cquiroz:
Aladin was updated to use es-modules but we had babel transpiling aladin to emulate es for older browsers. This is in contradiction to how vite assembles the modules. 
The solution was to set babel to emit non emulated es-modules and this seems to work now in production


arturog8m (approved):
Excellent! Thanks for tracking down this problem.

PR #991, opened 2021-08-06
Adapt ElevationRange editor to new View logic

rpiaggio:
We discovered that the interaction of `getDerivedStateFromProps` and the new logic of `ViewF.withOnMod` could yield surprising results.

This is an attempt to rewrite the ElevationRange editor by making the derived state read-only and always modifying the original data directly. This removes the need to use `withOnMod`.

### DIFF
--- a/explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
+++ b/explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
@@ -236,7 +236,7 @@ object ConstraintsPanel {
                 changeAuditor = ChangeAuditor.accept.decimal(1),
                 clazz = ExploreStyles.ElevationRangeEntry
               ),
-              <.label("Min"),
+              <.label("Max"),
               FormInputEV(
                 id = "maxam",
                 value = airMassView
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -136,19 +132,48 @@ object ConstraintsPanel {
         )
       }
 
-      def updateElevationRange(ert: ElevationRangeType) =
-        ert match {
-          case AirMass   => erView.set(state.get.airMass)
-          case HourAngle => erView.set(state.get.hourAngle)
-        }
+      val erTypeView: View[ElevationRangeType] =
```
toddburnside, line 135: Nice work. Creating these views didn't really occur to me.

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -218,47 +239,43 @@ object ConstraintsPanel {
               <.label("Min"),
```
toddburnside, line 239: You might as well change the label to "Max" here.

toddburnside (approved):
LGTM

PR #1000, opened 2021-08-09
Don't run linter on dependabot PRs.

rpiaggio:
Dependabot PRs run on `push` and `pull_request` no longer have access to `secrets` (see https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-797125425).

This makes the lint action fail with a 403 on dependabot PRs.

We could switch to running on `pull_request_target`, but I think we don't actually need the linter on dependabot PRs, which is what this PR is about.


cquiroz (approved):
Let's try this

PR #1007, opened 2021-08-09
Improve target grid layout

rpiaggio:
When the plot was introduced to the target panel it broke the grid. This PR should fix it.


common/src/main/webapp/less/style.less
```diff
@@ -39,6 +39,11 @@ a {
   color: var(--site-link-color);
 }
 
+// Override SUI's default, which prevented 
+.ui.form .fields .field .ui.input input, .ui.form .field .ui.input input {
+  width: 100%;
```
rpiaggio, line 45: SUI's default of `auto` was preventing input fields from shrinking in small screens.

common/src/main/webapp/less/style.less
```diff
@@ -932,15 +937,9 @@ tfoot {
...
-.target-aladin-cell {
-  height: 100px;
-  min-height: 300px; // don't collapse in single column mode.
-}
-
 .tile-xl-width,
 .tile-lg-width,
-.tile-md-width,
-.tile-sm-width {
+.tile-md-width {
```
rpiaggio, line 943: Intentionally increased the cutoff.

cquiroz:
I haven't looked at the details but I think we need to somewhat calculate the height of the area allocated to aladin and pass that as a height value unlike now that (I think) it is fixed

rpiaggio:
> I haven't looked at the details but I think we need to somewhat calculate the height of the area allocated to aladin and pass that as a height value unlike now that (I think) it is fixed

We can do that for aladin and the plot when we move them to their own panels.

package.json
```diff
@@ -59,10 +59,10 @@
     "vite-plugin-fonts": "^0.2.2"
   },
   "scripts": {
-    "lint-dark": "sed '/^[\\.\\}\\@]/d;/^[[:blank:]]*\\..*/d;/\\/\\/.*/d' common/src/main/webapp/less/variables-dark.less > vars.css;npx stylelint common/src/main/webapp/less; rm vars.css",
-    "fix-dark": "sed '/^[\\.\\}\\@]/d;/^[[:blank:]]*\\..*/d;/\\/\\/.*/d' common/src/main/webapp/less/variables-dark.less > vars.css;npx stylelint --fix common/src/main/webapp/less; rm vars.css",
-    "lint-light": "sed '/^[\\.\\}\\@]/d;/^[[:blank:]]*\\..*/d;/\\/\\/.*/d' common/src/main/webapp/less/variables-light.less > vars.css;npx stylelint common/src/main/webapp/less; rm vars.css",
-    "fix-light": "sed '/^[\\.\\}\\@]/d;/^[[:blank:]]*\\..*/d;/\\/\\/.*/d' common/src/main/webapp/less/variables-light.less > vars.css;npx stylelint --fix common/src/main/webapp/less; rm vars.css",
+    "lint-dark": "sed '/^[\\.\\}\\@]/d;/^[[:blank:]]*\\..*/d;/\\/\\/.*/d' common/src/main/webapp/less/variables-dark.less > vars.css;trap \"rm vars.css\" EXIT;npx stylelint common/src/main/webapp/less",
```
cquiroz, line 62: `trap`?

cquiroz:
Do you have a screenshot?

package.json
rpiaggio: Think of it as a `finally`. The `rm` will be executed at the end, but this way the exit code of the script will be the one returned by `stylelint`.

rpiaggio:
> Do you have a screenshot?

Large screen:
image

Medium screen:
image

Small screen:
image
image


I see the magnifying glass is truncated in the small screen. Will fix.

cquiroz (approved):
LGTM

PR #1019, opened 2021-08-13
Select observation configuration

rpiaggio:
* Shows selected configuration from the DB and allows changing it.
* For XT, only GMOS longslit (single FPU) modes are enabled.
* New GMOS B480 modes added (thanks Carlos).
* Modes with multiple FPU options are now flattened into several lines, each with only one FPU.
* Selected configuration is sticky.
* Undo/redo supported.
* For the moment, if there is a selected configuration that falls outside the current requirements, it is not shown. In a future PR it will be shown anyway.
* UI look and feel is, of course, subject to discussion.
* Undo logic is a bit more streamlined. Also `zoom` on `UndoSetter`. This is useful for encompassing 2 different parts of the model (in this case: science requirements and configuration) under the same undo stack.

Kapture 2021-08-13 at 12 34 21


toddburnside (approved):
LGTM

cquiroz:
Looks great. We may want to have a way to unselect a mode?

rpiaggio:
> Looks great. We may want to have a way to unselect a mode?

Good idea. I'll implement it for now so that if you click on the select row it will be unselected.

PR #1020, opened 2021-08-13
Update react-markdown and vite to its latest versions

cquiroz:
For quite a while `react-markdown` has been a problem in the build because it is partially CJS and ES and has forbidden us to update vite
`react-markdown` 7.0.0 is full ES thus we can update it and simplify some bits. It tooks some effort to properly pollyfil some dependencies but seems to work in dev and prod


explore/src/main/webapp/index.html
```diff
@@ -14,6 +14,11 @@
       var global = window;
     }
   </script>
+  <script type="module">
+    import process from "process";
+
+    window.process = process;
```
rpiaggio, line 20: Where is `process` used?

rpiaggio (approved):
LGTM

explore/src/main/webapp/index.html
cquiroz: That's the secret sauce. react-markdown needs it but it is not provided in the dev enviroment in vite so it needs a polyfill

PR #1025, opened 2021-08-17
Improvements in config selection

rpiaggio:
* Use virtualized table for faster rendering. NOTE: selected row is no longer sticky. This will be addressed in future PR.
* Filtered mode list is memoized via hooks.
* Enabled modes are at the top of the list.

There's some commented code left on purpose which might help how we deal with row selection.

### DIFF
--- a/common/src/main/scala/reactST/reactTable/SUITable.scala
+++ b/common/src/main/scala/reactST/reactTable/SUITable.scala
@@ -115,12 +115,13 @@ class SUITable[
       val tableInstance = props.instance
 
       val tableRender: TableRender[D, TableInstanceD] = (props.table: Any) match {
-        case table: Table => tableInstance => table(tableInstance.getTableProps())
+        case table: Table =>
+          tableInstance => table.copy(as = layout.tag)(tableInstance.getTableProps())
         case other        => other.asInstanceOf[TableRender[D, TableInstanceD]]
       }
 
       val rowRender: RowRender[D] = (props.row: Any) match {
-        case row: TableRow => rowData => row(rowData.getRowProps())
+        case row: TableRow => rowData => row.copy(as = layout.tag)(rowData.getRowProps())
         case other         => other.asInstanceOf[RowRender[D]]
       }
 
@@ -132,17 +133,17 @@ class SUITable[
 
       val headerCellRender: HeaderCellRender[D, ColumnInstanceD] =
         (props.headerCell: Any) match {
-          case headerCell: TableHeaderCell => _ => headerCell
+          case headerCell: TableHeaderCell => _ => headerCell.copy(as = layout.tag)
           case fn                          => fn.asInstanceOf[HeaderCellRender[D, ColumnInstanceD]]
         }
 
       val bodyCellRender: BodyCellRender[D] = (props.cell: Any) match {
-        case cell: TableCell => _ => cell
+        case cell: TableCell => _ => cell.copy(as = layout.tag)
         case fn              => fn.asInstanceOf[BodyCellRender[D]]
       }
 
       val footerCellRender: HeaderCellRender[D, ColumnInstanceD] = (props.footerCell: Any) match {
-        case footerCell: TableHeaderCell => _ => footerCell
+        case footerCell: TableHeaderCell => _ => footerCell.copy(as = layout.tag)
         case fn                          => fn.asInstanceOf[HeaderCellRender[D, ColumnInstanceD]]
       }
 
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -85,7 +85,9 @@ a {
   input[type="week"],
   input[type="number"],
   input[type="email"],
-  input[type="url"] { font-size: 16px; }
+  input[type="url"] {
+    font-size: 16px;
+  }
 }
 
 @media screen and (device-aspect-ratio: 40/71) {
@@ -101,7 +103,9 @@ a {
   input[type="week"],
   input[type="number"],
   input[type="email"],
-  input[type="url"] { font-size: 16px; }
+  input[type="url"] {
+    font-size: 16px;
+  }
 }
 
 @media screen and (device-aspect-ratio: 375/667) {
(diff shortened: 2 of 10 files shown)


common/src/main/scala/reactST/reactTable/SUITable.scala
```diff
@@ -132,17 +133,17 @@ class SUITable[
...
       val headerCellRender: HeaderCellRender[D, ColumnInstanceD] =
         (props.headerCell: Any) match {
-          case headerCell: TableHeaderCell => _ => headerCell
+          case headerCell: TableHeaderCell => _ => headerCell.copy(as = layout.tag)
           case fn                          => fn.asInstanceOf[HeaderCellRender[D, ColumnInstanceD]]
         }
 
       val bodyCellRender: BodyCellRender[D] = (props.cell: Any) match {
-        case cell: TableCell => _ => cell
+        case cell: TableCell => _ => cell.copy(as = layout.tag)
```
cquiroz, line 141: Does this convert the table tags to something else?

common/src/main/webapp/less/style.less
```diff
@@ -1198,7 +1212,9 @@ img.partner-split-flag {
   margin-bottom: 0;
 }
 
-i.icon.user-astronaut::before { content: "\f4fb"; }
+i.icon.user-astronaut::before {
```
cquiroz, line 1215: Unrelated but I suspect we don't need this anymore with FontAwesome icons in react

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -218,35 +211,77 @@ object SpectroscopyModesTable {
...
-            TableHeaderCell(clazz = ExploreStyles.Sticky |+| ExploreStyles.ModesHeader)(
-              ^.textTransform.capitalize.when(c.id.toString =!= ResolutionColumnId.value),
-              ^.textTransform.none.when(c.id.toString === ResolutionColumnId.value)
-            ),
-          row = (rowData: Row[SpectroscopyModeRow]) =>
-            TableRow(
-              disabled = !enabledRow(rowData.original),
-              clazz = ExploreStyles.ModeSelected.when_(
-                props.scienceConfiguration.get.exists(conf => equalsConf(rowData.original, conf))
+  // protected def ensureContains(matrix: List[SpectroscopyModeRow], conf: Option[ScienceConfigurationData]): List[SpectroscopyModeRow] =
```
cquiroz, line 214: Commented code

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -218,35 +211,77 @@ object SpectroscopyModesTable {
...
+  //     else
+
+  protected def enabledRow(row: SpectroscopyModeRow): Boolean =
+    List(Instrument.GmosNorth, Instrument.GmosSouth).contains_(row.instrument.instrument) &&
+      row.focalPlane === FocalPlane.SingleSlit
+
+  val component =
+    ScalaFnComponent
+      .withHooks[Props]
+      .useMemoBy(_.spectroscopyRequirements)(props =>
```
cquiroz, line 227: This looks super interesting, could you expand a bit on what it does?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -218,35 +211,77 @@ object SpectroscopyModesTable {
...
-              ^.onClick --> props.scienceConfiguration.set(toggleRow(rowData.original)),
-              props2Attrs(rowData.getRowProps())
-            )
-        )(tableInstance)
+          val (enabled, disabled) = rows.partition(enabledRow)
+          (enabled ++ disabled).toJSArray
+        }
       )
-    }
+      .useMemoBy($ =>
```
cquiroz, line 244: You can have multiple memo?

cquiroz (approved):
Amazing work

common/src/main/scala/reactST/reactTable/SUITable.scala
rpiaggio: There's a typleclass mechanism in place in `SUITable` to render HTML table elements or div blocks depending whereas the react-table builder has a layout plugin.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
rpiaggio: Yup, we will probably use it to show the selected config when it's not in the list.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
rpiaggio: Yes, you can memoize different values with different dependencies.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
rpiaggio: In this case, the filtered table (the 2nd parameter) will only be recomputed if the requirements (the 1st parameter) are different than in the last render. Difference in this case is according to `Reusability`. In other words, as long as `spectroscopyRequirements` can be reused, then the memoized list of filtered configurations is also reused instead of computing a fresh one.

The filtered table (whether it is a freshly computed one or a memoized one) is passed as an additional parameter to the `render` method.

cquiroz: Coul we use this to get the spectroscpy matrix? Right now it is being loaded on every mount I think

rpiaggio: Maybe. The use case is a bit different since we want the fetch to happen asynchronously. I'll check.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: I thought about using `IO.memoize` for that but didn't follow it through

PR #1031, opened 2021-08-18
Style the scrollbars

cquiroz:
The scrollbars look a bit too dark are are difficult to spot. Styling of scrollbars in css is not very well defined but some improvements are possible

Explore 2021-08-18 12-59-29


rpiaggio (approved):
LGTM

PR #1033, opened 2021-08-19
Mangler options for scala.js

cquiroz:
This is an experimental terser config. It doesn't work proprely with dynamically loaded components but it maybe fixable via exclusions


rpiaggio:
Superseded by #1098

PR #1034, opened 2021-08-19
Several fixes [ch-1017] [ch-1015]

rpiaggio:
* In Observation View, when no modes match the requirements Explore no longer breaks.
* In Target View, date in night elevation plot now matches the selected date (no more 1 month offset).
* In Target View, semester plots now show visibility time above 30 degrees (instead of 0).
* In Target View, semester plots now render a data point every 3 days (instead of 1), which increases threefold the rendering speed.


cquiroz (approved):
Excellent, thanks

PR #1036, opened 2021-08-20
[ch-1001] Implement plot suggestions.

rpiaggio:
* "Morning 12deg Twilight" was mislabeled as "Evening"
* The graph grid was on top of the Twilight labels which made them difficult to read.
* When the target is not visible display the message "Target is below horizon."
* Enable the Elevation, Parallactic Angle, and Sky Brightness curves by default.


cquiroz (approved):
LGTM

PR #1041, opened 2021-08-23
Modify the resizable property of tiles depending on the state

cquiroz:

### DIFF
--- a/common/src/main/scala/explore/components/TileController.scala
+++ b/common/src/main/scala/explore/components/TileController.scala
@@ -25,6 +25,7 @@ import react.common._
 import react.gridlayout._
 
 import scala.concurrent.duration._
+import scala.scalajs.js.|
 
 final case class TileController(
   userId:           Option[User.Id],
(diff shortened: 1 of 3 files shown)


common/src/main/scala/explore/components/TileController.scala
```diff
@@ -72,6 +73,19 @@ object TileController {
       .filter(_.i.forall(_ === id.value))
       .andThen(layoutItemHeight)
 
+  def tileResizable(id: Tile.TileId): Traversal[LayoutsMap, Boolean | Unit] =
```
rpiaggio, line 76: Why `Boolean | Unit`?

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/TileController.scala
cquiroz: Yeah that isn't nice, it comes form the facade so I should fix there

PR #1043, opened 2021-08-23
Improvements to target-search

cquiroz:
This fixes a bug where sometimes the call to do simbad search fails. It was solved by adding a retry as in other cases

Some UI fixes too:
* Properly square the go to target button in aladin
* Better positioning of the search icon
* Animate when search is in progress
* Better position of the search error tooltip

### DIFF
--- a/common/src/main/scala/explore/common/RetryHelpers.scala
+++ b/common/src/main/scala/explore/common/RetryHelpers.scala
@@ -1,0 +1,35 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.common
+
+import cats.Applicative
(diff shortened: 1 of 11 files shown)


common/src/main/scala/explore/common/RetryHelpers.scala
```diff
@@ -0,0 +1,35 @@
...
+import retry.RetryDetails._
+import retry.RetryPolicies._
+import retry._
+
+import java.{ util => ju }
+import scala.concurrent.duration._
+
+import ju.concurrent.TimeUnit
+
+trait RetryHelpers {
```
rpiaggio, line 17: Nice!

rpiaggio (approved):
LGTM

rpiaggio:
I wonder if we want to use the retry logic in all other places. Eg: fetching the spectroscopy matrix.

cquiroz:
Probably we should

PR #1044, opened 2021-08-23
Add help icon to the sky plot section

cquiroz:
Explore 2021-08-23 16-55-57

### DIFF
--- a/common/src/main/scala/explore/components/HelpIcon.scala
+++ b/common/src/main/scala/explore/components/HelpIcon.scala
@@ -15,7 +15,9 @@ import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import react.common._
 
-final case class HelpIcon(id: Help.Id)
+import scala.scalajs.js
+
+final case class HelpIcon(id: Help.Id, clazz: js.UndefOr[Css] = js.undefined)
 
 object HelpIcon {
   type Props = HelpIcon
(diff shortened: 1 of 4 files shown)


common/src/main/scala/explore/components/HelpIcon.scala
```diff
@@ -24,17 +26,23 @@ object HelpIcon {
...
-        Icons.Info
-          .fixedWidth()
-          .clazz(ExploreStyles.HelpIcon)
-      )
+  implicit val helpReusability: Reusability[HelpIcon] =
+    Reusability.by(x => (x.id.value, x.clazz.map(_.htmlClass)))
+
+  val component = ScalaFnComponent
+    .withHooks[Props]
+    .renderWithReuse { p =>
```
rpiaggio, line 34: Since there are no hooks defined, I'm guessing it's not possible to use `renderWithReuse` if you just do `ScalaFnComponent[Props]` ?

common/src/main/webapp/less/style.less
```diff
@@ -1506,6 +1506,12 @@ html {
   color: var(--site-text-color);
 }
 
+.explore-help-icon-floating {
```
rpiaggio, line 1509: This seems specific to the plot. Maybe the name should reflect that?

rpiaggio (approved):
LGTM

common/src/main/webapp/less/style.less
cquiroz: Well I think there maybe more floating help icons in the future

common/src/main/scala/explore/components/HelpIcon.scala
cquiroz: That's right, you can't get `renderWithReuse` unless you have hooks

common/src/main/webapp/less/style.less
rpiaggio: Agreed. But with the same positioning?

common/src/main/scala/explore/components/HelpIcon.scala
rpiaggio: We should probably suggest supporting this to japgolly.

cquiroz: Yes, but don't know how to say. The only think I know about `hooks` is they exist in react

PR #1061, opened 2021-09-01
useTable hook

rpiaggio:
Fully embrace hooks when using react-table.

I'm opening this as draft to enable discussion. There are probably improvements that can be made.

In particular, I'm not sure about:
* Requiring `Reusability` for columns and rows. Although technically makes sense, it is awkward. Also it is actually required by `useMemo`.
* `useTable_` to pass the additional options manipulation function. This is only required when used as a secondary hook, which makes the API a bit inconsistent.


explore/src/main/scala/explore/config/SequenceTable.scala
```diff
@@ -89,142 +89,144 @@ object SequenceTable {
...
+      .setHeader("ROI")
+      .setCell(_.value.shortName),
+    StepTable
+      .Column("sn", _ => "")
+      .setHeader("S/N")
+  )
+
+  implicit private val colReuse: Reusability[List[ColumnInterface[StepLine[_]]]] =
+    Reusability.always
+  implicit private val dataReuse: Reusability[List[StepLine[_]]]                 = Reusability.byRefOr_==
```
cquiroz, line 167: Does this mean we run reusability for each row? I guess with a virtualized list it is fine but maybe slow for large non virtualized tables?

rpiaggio: I agree we don't want to check all the elements in the list. That's why it's implemented as `byRefOr_==`.

explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
```diff
@@ -174,94 +182,76 @@ object TargetSummaryTable {
...
-            .setHiddenColumns(
-              props.hiddenColumns.get.toList
-                .map(col => col: IdType[TargetResult])
-                .toJSArray
-            )
-        )
+          )
+      }
+      .useMemoBy((props, _) => props.pointingsWithObs)((_, _) => _.targets.toList) // Memo rows
+      .useTableBy_(TargetTable)(
```
toddburnside, line 188: What is the difference between `useTableBy()` and `useTableBy_()`?

toddburnside: I see. With or without `modOpts`(now I just need to understand where that is coming from...). I wonder if we should make that clearer in the names?

toddburnside: Ah-ha. I grok it now. 🕺

rpiaggio: >I wonder if we should make that clearer in the names?

Agreed, I'll do that.

toddburnside (approved):
Very nice. I like the new syntax. - Oops. This was meant for the scalajs-react-table PR. But, I like the syntax here, too. :)

common/src/main/scala/reactST/reactTable/SUITable.scala
```diff
@@ -198,18 +208,18 @@ class SUITable[
...
 
-    def apply(options: TableOptsD): Component = builder(tableMaker.use(options))
+  //   // def apply(options: TableOptsD): Component = builder(tableDef.use(options))
 
-    def apply(columns: js.Array[_ <: UseTableColumnOptions[D]], data: js.Array[D]): Component =
-      apply(tableMaker.Options(columns, data))
-  }
+  //   // def apply(columns: js.Array[_ <: UseTableColumnOptions[D]], data: js.Array[D]): Component =
+  //   //   apply(tableDef.Options(columns, data))
+  // }
```
toddburnside, line 222: Delete these comments?

toddburnside (approved):
👍

PR #1073, opened 2021-09-06
Don't highlight date when navigating months.

rpiaggio:
The changes in `styles.less` is just auto-formatting.

### DIFF
--- a/common/src/main/webapp/sass/datepicker.scss
+++ b/common/src/main/webapp/sass/datepicker.scss
@@ -1,13 +1,11 @@
 @use "react-datepicker/src/stylesheets/datepicker.scss";
 
 .react-datepicker {
-  font-family: "Lucida Grande", "Lucida Sans Unicode", Arial, Helvetica,
-    sans-serif;
+  font-family: inherit;
 }
 
 .react-datepicker-wrapper {
-  font-family: "Lucida Grande", "Lucida Sans Unicode", Arial, Helvetica,
-    sans-serif;
+  font-family: inherit;
 }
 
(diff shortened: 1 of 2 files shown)


common/src/main/webapp/sass/datepicker.scss
```diff
@@ -1,14 +1,22 @@
 @use "react-datepicker/src/stylesheets/datepicker.scss";
 
 .react-datepicker {
-  font-family: "Lucida Grande", "Lucida Sans Unicode", Arial, Helvetica, sans-serif;
+  font-family: "Lucida Grande", "Lucida Sans Unicode", Arial, Helvetica,
```
cquiroz, line 4: Should we use the same font as the system?

rpiaggio: Good idea. I actually changed to `inherit` so that it's consistent with the rest of the UI.

common/src/main/webapp/sass/datepicker.scss
```diff
@@ -1,14 +1,20 @@
...
+  font-family: inherit;
 }
 
-.react-datepicker__input-container input{
+.react-datepicker__input-container input {
   background: var(--form-input-background);
   color: var(--form-input-color);
 }
+
+// https://stackoverflow.com/questions/63284785/react-datepicker-remove-coloring-of-same-day-in-other-months
```
toddburnside, line 16: Good stackoverflow-ing.

toddburnside (approved):
LGTM

PR #1075, opened 2021-09-07
Add RV, redshift columns to target summary table.

rpiaggio:


explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
```diff
@@ -72,6 +73,9 @@ object TargetSummaryTable {
     "epoch"        -> "Epoch",
     "pmra"         -> "µ RA",
     "pmdec"        -> "µ Dec",
+    "rv"           -> "RV",
```
cquiroz, line 76: Do we want to display the three versions in the table?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/TargetSummaryTable.scala
rpiaggio: As per Andy's request, yes. We only show `RV` initially.

cquiroz (approved):
LGTM, I'd like at a later date review alternatives to support conversions in a single column but this will do for now

PR #1077, opened 2021-09-07
Auto scroll to selected mode in configuration table.

rpiaggio:
We still have to find a way to show the selected mode when it's not among the current filtered list, or when it's scrolled away. But for the moment this is an improvement.

Also, fixes stripes, which would change as the user scrolled, and were not shown when scrolling horizontally in Chrome.


cquiroz (approved):
Cool, thanks

toddburnside (approved):
LGTM

PR #1080, opened 2021-09-08
Fix: constraints selector dropdown appears behind lower tiles

toddburnside:


common/src/main/scala/explore/model/reusability.scala
```diff
@@ -59,4 +60,5 @@ object reusability {
   implicit def modelUndoStacksReuse[F[_]]: Reusability[ModelUndoStacks[F]]           = Reusability.derive
   // Move to lucuma-ui
   implicit val semesterReuse: Reusability[Semester]                                  = Reusability.derive
+  implicit val cssReuse: Reusability[Css]                                            = Reusability.by(_.htmlClass)
```
toddburnside, line 63: I was surprised I couldn't find a reuseability for Css anywhere. I can move this and the above to lucuma-ui and cut a release of it. (Unless I missed an instance somewhere)

cquiroz: Maybe it should be in scalajs-react-common

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -413,14 +413,19 @@ object ObsTabContents {
                   props.searching,
                   state.zoom(State.options)
                 ),
+                // The ExploreStyles.ConstraintsTile css adds a z-index to the constraints tile react-grid wrapper
+                // so that the constraints selector dropdown always appears in front of any other tiles. If more
+                // than one tile ends up having dropdowns in the tile header, we'll need something more complex such
+                // as changing the css classes on the various tiles when the dropdown is clicked to control z-index.
```
cquiroz, line 419: Neat

common/src/main/scala/explore/model/reusability.scala
toddburnside: Actually, I did find an instance, but only in seqexec

cquiroz (approved):
Great investigative work to solve this one, thanks

common/src/main/scala/explore/model/reusability.scala
toddburnside: > Maybe it should be in scalajs-react-common

Oh, good point.

rpiaggio (approved):
LGTM thanks for tackling this tricky issue

PR #1084, opened 2021-09-09
Fix moon so it appears behind dropdown

toddburnside:
In narrow windows, the moon would appear in front of the magnitude dropdown.
image

By placing the moon after the sky plot in the DOM, I was able to remove the z-index on the moon that was causing the problem.
image


cquiroz (approved):
LGTM

PR #1090, opened 2021-09-14
Sidereal time. Button controls in plot.

rpiaggio:
I'm changing the plot controls to buttons because:
- They provide a consistent experience when there are more than 2 options (which now happens with the time display).
- Takes up less screen space.

Of course, we can always go back to sliders and I presume radio buttons for time display.

image


toddburnside (approved):
I'm not the final arbiter of the UI, of course, but I like the new look.

cquiroz:
Looks nice. Did you try with the accent color (a type of green) which maybe not very readable but we could try
We could also check the font. I think the plot's font is not the same as the one for explore

rpiaggio:
Font: I made it inherit from the page now, so it's uniform.

Colors: I'm trying the accent color and I'm not sure it looks nice as a background color. The color I'm using is the one react-datepicker uses, so it's consistent at least with the selected date.

cquiroz:
Right it doesn't look that nice nor easy to read, let's use the blue one

PR #1093, opened 2021-09-16
Remove keywords from proposals

toddburnside:
This also fixes a tab order issue when changing to a single column grid (see comment in style.less).


cquiroz (approved):
LGTM, thanks

PR #1098, opened 2021-09-20
Run terser after build

rpiaggio:
This PR switches to run `vite` programatically. This allows us to run things on `vite`'s output. In this case, we are running `terser` with our own handling of the `nameCache` for property mangling, since `vite` doesn't seem to handle it properly.

Strangely, the best results are obtained when running `terser` both within `vite` (without property mangling, just compressing) and afterwards (with property mangling).

For example, Explore's main chunk (`index.js`) results with:
- No terser: ~32.5 Mb.
- Terser only after vite: ~19.5 Mb.
- Terser only in vite: ~16.3 Mb.
- Terser both in and after vite: ~10.5MB.

Supersedes #1033

### DIFF
--- a/.github/workflows/heroku.yml
+++ b/.github/workflows/heroku.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Vite - Build
         run: |
           npm install
-          npx vite build
+          node build
         env:
           NODE_OPTIONS: --max_old_space_size=6144
       - name: Heroku - Deploy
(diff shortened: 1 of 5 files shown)


cquiroz (approved):
Wow this is amazing, I guess it won't affect the dev server

rpiaggio:
> Wow this is amazing, I guess it won't affect the dev server

It shouldn't. This only runs in the production build.

PR #1101, opened 2021-09-21
Fix full.yml for new build.

rpiaggio:
full.yml was failing due to npm arguments present in the `build` command.


toddburnside (approved):
LGTM

PR #1102, opened 2021-09-21
Report brotli improvement.

rpiaggio:
The brotli improvement is not so impressive `1259.69 KiB --> 1030.15 KiB (81.78%)` but still...

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,7 +1,6 @@
 const path = require("path")
 const fs = require("fs")
 const process = require("process")
-const size = require("human-format")
 const { build } = require("vite")
 const { minify } = require("terser")
 const humanFormat = require("human-format")
--- a/build.js
+++ b/build.js
@@ -1,11 +1,12 @@
 const path = require("path")
 const fs = require("fs")
 const process = require("process")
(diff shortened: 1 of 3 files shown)


build.js
```diff
@@ -1,11 +1,12 @@
 const path = require("path")
 const fs = require("fs")
 const process = require("process")
-const size = require('human-format');
+const size = require("human-format")
```
toddburnside, line 4: Is this `require` needed? It seems like the one three lines below is what is being used.

toddburnside (approved):
LGTM

cquiroz (approved):
Looks good, should build.js be done using ES imports?

build.js
rpiaggio: You're right, I brought it in twice. Thanks!

rpiaggio:
> Looks good, should build.js be done using ES imports?

If I try using `import` I get `SyntaxError: Cannot use import statement outside a module`. I admit I'm a bit new to this. Is there something I'm missing?

cquiroz:
Ah don't worry, just use `require`

PR #1104, opened 2021-09-21
Upgrade to clue v0.18.0

rpiaggio:
Also minor fixes in new build.

NOTE: This PR now brings `http4s-core` into the build since `clue` now uses `http4s`'s `Uri` instead of `sttp`'s. In Explore we still use the `sttp` client, so we now have both as dependencies.. However, the build output size does not increase noticeably (~200 Kb).


cquiroz (approved):
Cool, LGTM

PR #1105, opened 2021-09-21
Migrate to http4s-client

rpiaggio:


.github/workflows/full.yml
```diff
@@ -22,6 +22,6 @@ jobs:
           NODE_OPTIONS: '--max-old-space-size=6144'
           FONTAWESOME_NPM_AUTH_TOKEN: ${{ secrets.FONTAWESOME_NPM_AUTH_TOKEN }}
         run: |
-          sbt stage
+          sbt -v -J-Xmx6g stage
```
rpiaggio, line 25: Unrelated change, but CI builds were failing because of OOM.

common/src/main/scala/explore/common/SSOClient.scala
```diff
@@ -3,29 +3,33 @@
...
 import retry._
-import sttp.client3._
-import sttp.client3.impl.cats.FetchCatsBackend
 
 import java.time.Instant
 import java.time.temporal.ChronoUnit
 import java.{ util => ju }
 import scala.concurrent.duration._
 
+// import ju.concurrent.TimeUnit
```
toddburnside, line 31: remove?

.github/workflows/full.yml
cquiroz: Maybe because of the addition of http4s?

rpiaggio: It was already happening. See for example https://github.com/gemini-hlsw/explore/pull/1060. We actually increased the memory a while ago for the Heroku build but forgot to do it here too.

toddburnside (approved):
Nice.

.github/workflows/full.yml
rpiaggio: Ah #1060 already has http4s-core (but not http4s-client), so maybe?

PR #1118, opened 2021-09-22
Use firebase for review apps and staging.

rpiaggio:
Since we actually run our builds in GitHub, there's nothing really tying us to Heroku. Therefore, I'm exploring other hosting options. I played a bit with Netlify and Firebase, and I decided to go for the latter. (We could also explore other options, like AWS's Amplify).

Reasons for these are:
 - Options for serving static assets from Heroku are limited.
 - The buildpack for serving static assets hasn't been maintained in a year.
 - It doesn't support brotli compression.
 - Cold-serving Explore from Heroku takes between ~3s (rarely) to ~6s (and sometimes more). Serving from Firebase consistently takes ~1.8s. (From Netlify takes ~3s).
 - We can have review apps for all PRs in Firebase (which this PR does).
 
Things we lose by moving away from Heroku:
- Nice workflow UI.
- Promoting from staging to production without rebuilding. There may be ways to achieve this. We could investigate implementing a cloud function that copies resource from one app to the other. 

I don't consider these big losses compared to the gains.

### DIFF
--- a/.github/workflows/full.yml
+++ b/.github/workflows/full.yml
@@ -48,3 +48,22 @@ jobs:
         run: yarn bundlemon
         env:
           BUNDLEMON_PROJECT_ID: 614a60c0e759f5000834cacc
+
+      - name: Deploy review app to Firebase
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
+        uses: FirebaseExtended/action-hosting-deploy@v0
+        with:
+          repoToken: '${{ secrets.GITHUB_TOKEN }}'
+          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_EXPLORE_GEMINI }}'
+          projectId: explore-gemini
+          target: 'staging'
+
+      - name: Deploy staging app to Firebase
+        if: github.event_name == 'push'
+        uses: FirebaseExtended/action-hosting-deploy@v0
+        with:
+          repoToken: '${{ secrets.GITHUB_TOKEN }}'
+          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_EXPLORE_GEMINI }}'
+          projectId: explore-gemini
+          target: 'staging'
+          channelId: live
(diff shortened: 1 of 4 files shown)


cquiroz:
What about the domain names? to get sso working?

.github/workflows/full.yml
```diff
@@ -48,3 +48,22 @@ jobs:
...
+          target: 'staging'
+
+      - name: Deploy staging app to Firebase
+        if: github.event_name == 'push'
+        uses: FirebaseExtended/action-hosting-deploy@v0
+        with:
+          repoToken: '${{ secrets.GITHUB_TOKEN }}'
+          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_EXPLORE_GEMINI }}'
+          projectId: explore-gemini
+          target: 'staging'
```
toddburnside, line 68: target is `staging` in both places. Is that correct?

rpiaggio:
> What about the domain names? to get sso working?

They are not set up yet. I'll ask Rob to assign `lucuma.xyz` one. (Or do you have access to the DNS?) Until then, SSO is not working.

rpiaggio:
I don't think the GH action supports attaching and reporting a `lucuma.xyz` URL when creating review apps. Seems unlikely. Anyway, we can still check them out as guest for the moment.

cquiroz:
I'd say let's givi it a try

toddburnside:
> I'd say let's givi it a try

Yeah. Testing it in parallel with Heroku seems like a good plan.

cquiroz (approved):
LGTM

rpiaggio:
Just in case we explore other options in the future, I'm documenting my experience with Netlify:

- It doesn't support deploying the same artifacts to different environments. For example, it doesn't support runtime configuration via environment variables. Different `conf.json` would have to be deployed manually to staging and production.
- It doesn't  support deploying to different environments from the same branch. It requires a separate `production` branch, for example.
- As already mentioned, it's a bit slower than Firebase.

cquiroz:
Slower on what sense?

cquiroz:
And heroku?

toddburnside:
>     * Cold-serving Explore from Heroku takes between ~3s (rarely) to ~6s (and sometimes more). Serving from Firebase consistently takes ~1.8s. (From Netlify takes ~3s).

PR #1119, opened 2021-09-23
Change constraints undo/redo so only keeps stack for most recent "thing"

toddburnside:
This eliminates the separate undo stacks for ConstraintSets, ConstraintGroups and ConstraintGroupLists in favor of maintaining only one stack at a time. This eliminate most or all strange interactions between undo at the different "levels".

So, if you are editing a constraint group, the undo stack for that will be kept until you edit a different constraint group or drag and drop to change the constraint group list. Note that a editing a constraint set on the observation tab is the same as editing a single element constraint group on the constraints tab - so the same undo stack would be used both places.

### DIFF
--- a/common/src/main/scala/explore/model/ConstraintsUndoStacks.scala
+++ b/common/src/main/scala/explore/model/ConstraintsUndoStacks.scala
@@ -1,0 +1,45 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats.Eq
+import cats.syntax.all._
+import explore.common.ConstraintGroupQueries.ConstraintGroupList
+import explore.undo.UndoStacks
+import lucuma.core.model.Observation
+
+import scala.collection.immutable.SortedSet
+
+sealed trait ConstraintsUndoStacks[F[_]] {
+  import ConstraintsUndoStacks._
+
+  def listStacks: UndoStacks[F, ConstraintGroupList] = this match {
+    case ConstraintGroupUndoStacks(_, _) => UndoStacks.empty[F, ConstraintGroupList]
+    case ConstraintListUndoStacks(us)    => us
+  }
+
+  def groupStacks(obsIds: SortedSet[Observation.Id]): UndoStacks[F, ConstraintSet] = this match {
+    case ConstraintGroupUndoStacks(ids, undoStacks) if ids === obsIds => undoStacks
+    case _                                                            => UndoStacks.empty[F, ConstraintSet]
+  }
+}
+
+object ConstraintsUndoStacks {
+  case class ConstraintGroupUndoStacks[F[_]](
+    ids:        SortedSet[Observation.Id],
+    undoStacks: UndoStacks[F, ConstraintSet]
+  ) extends ConstraintsUndoStacks[F]
+
+  case class ConstraintListUndoStacks[F[_]](undoStacks: UndoStacks[F, ConstraintGroupList])
+      extends ConstraintsUndoStacks[F]
+
+  def empty[F[_]] = ConstraintListUndoStacks(UndoStacks.empty[F, ConstraintGroupList])
+
+  implicit def eqConstraintsUndoStacks[F[_1]]: Eq[ConstraintsUndoStacks[F]] = Eq.instance {
+    case (ConstraintListUndoStacks(l1), ConstraintListUndoStacks(l2))           => l1 === l2
+    case (ConstraintGroupUndoStacks(i1, s1), ConstraintGroupUndoStacks(i2, s2)) =>
+      i1 === i2 && s1 === s2
+    case _                                                                      => false
+  }
+}
--- a/explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
+++ b/explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
@@ -20,6 +20,8 @@ import explore.components.undo.UndoButtons
 import explore.implicits._
 import explore.model.AirMassRange
 import explore.model.ConstraintSet
+import explore.model.ConstraintsUndoStacks
+import explore.model.ConstraintsUndoStacks._
 import explore.model.ElevationRange
 import explore.model.Help
 import explore.model.HourAngleRange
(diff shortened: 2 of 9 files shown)


cquiroz (approved):
Looks great. We need to find ways to cope with undo complexities even if some functionality is limited

toddburnside:
> Looks great. We need to find ways to cope with undo complexities even if some functionality is limited

I think this approach will also work for the targets. There will just be more than 2 cases.

rpiaggio:
>if you are editing a constraint group, the undo stack for that will be kept until you edit a different constraint group 

Why do we need to do this? (Invalidating other groups' stacks)

common/src/main/scala/explore/model/ConstraintsUndoStacks.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  def groupStacks(obsIds: SortedSet[Observation.Id]): UndoStacks[F, ConstraintSet] = this match {
+    case ConstraintGroupUndoStacks(ids, undoStacks) if ids === obsIds => undoStacks
+    case _                                                            => UndoStacks.empty[F, ConstraintSet]
+  }
+}
+
+object ConstraintsUndoStacks {
+  case class ConstraintGroupUndoStacks[F[_]](
+    ids:        SortedSet[Observation.Id],
```
rpiaggio, line 30: This is a question that goes beyond the scope of this PR but I've been curious for a while: why is this a `SortedSet` and not just a `Set`?

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -300,7 +304,12 @@ object ConstraintsPanel {
     }
 
     def render(props: Props, state: State) = AppCtx.using { implicit appCtx =>
-      renderFn(props, state, UndoContext(props.undoStacks, props.constraintSet))
+      val csUndoView: View[UndoStacks[IO, ConstraintSet]] =
+        props.undoStacks.zoom(_.groupStacks(props.obsIds))(mod =>
+          csus => ConstraintGroupUndoStacks(props.obsIds, mod(csus.groupStacks(props.obsIds)))
+        )
```
rpiaggio, line 310: Can we encapsulate the mod operations in `ConstraintsUndoStacks`? Maybe they pass enough laws to build an `Adjuster`? (And then we can zoom using a `GetAdjust`)

rpiaggio:
I'm not sure I follow the logic completely. If I understand correctly, now the model has only one `ConstraintUndoStacks` which can alternate between being `ConstraintGroupUndoStacks` or a `ConstraintListUndoStacks`, depending on what was edited last.

Can you explain a bit how this works, maybe in a comment in `ConstraintUndoStacks.scala`, and why we decided to go this way? We discussed in Slack the use cases that generated problems otherwise, but I can't recall them.

I'm not 100% comfortable with what seem to be unrelated stacks being lost when I edit a CS, but I can be convinced otherwise if we find out that it's too cumbersome to support this soundly.

common/src/main/scala/explore/model/ConstraintsUndoStacks.scala
cquiroz: At least in the past I often used `SortedSet/Map` in places because monocle would provide some lenses only for sorted collections. This was relaxed on the latest versions of monocle

toddburnside: I'm sorting the constraint group list by lowest observation id, and this seemed the easiest way to accomplish that. (Although it turned out that the `Order` defined by cats for SortedSets didn't work for that purpose, so I had to define a new one. So, maybe there is a better way to do it.)

I chose that sort order because it seemed the most stable ordering for the list. There isn't a constraint id, and an ordering by name would potentially be reordering the list with every edit. It seems a bit arbitrary, but I figured I'd do it that way until I got feedback requesting otherwise.

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
toddburnside: I started doing that originally, but didn't know the right optic and was unsure about the legality - esp. with the group stack one requiring the observation ids. So, I decided to do it "in place" since each was used only once. But, I can check out an `Adjuster` and see if an implementation would be lawful.

common/src/main/scala/explore/model/ConstraintsUndoStacks.scala
rpiaggio: Ah I see now. Having the `Observation.Id`s sorted within each group is an optimization when sorting the groups.

toddburnside:
> > if you are editing a constraint group, the undo stack for that will be kept until you edit a different constraint group
> 
> Why do we need to do this? (Invalidating other groups' stacks)

We don't strictly need to do that for constraints. But, if we edit a group, we need to deal with the stacks for any individual constraint set in that group (the easiest way to `deal with it` being to clear them), and when editing an individual constraint set we need to deal with the stacks for any group it is in. And, if we retain stacks for other groups, we also need to clean up stacks that are orphaned when group memberships change - not a big deal, but one more thing to do. The interaction with ConstraintGroupList undo stacks is a different issue.

> I'm not 100% comfortable with what seem to be unrelated stacks being lost when I edit a CS, but I can be convinced otherwise if we find out that it's too cumbersome to support this soundly.

Right. This is the more extreme option that I mentioned in the gist. I may have misinterpreted, but I thought we had to decided to go this route in our slack conversation, in part because of the upcoming target changes and wanting some consistency with whatever we do for them. With targets, if we edit a target group, we not only need to worry about the stacks for other target groups like we do for constraint sets, but also related asterism groups and environment groups. (I am considering the editing of an individual item as a group of one.)

> I'm not sure I follow the logic completely. If I understand correctly, now the model has only one ConstraintUndoStacks which can alternate between being ConstraintGroupUndoStacks or a ConstraintListUndoStacks, depending on what was edited last.

Right. But nothing happens to ConstraintUndoStacks until an edit is actually made somewhere. So, just switching to view another constraint set doesn't change it.

> Can you explain a bit how this works, maybe in a comment in ConstraintUndoStacks.scala, and why we decided to go this way? We discussed in Slack the use cases that generated problems otherwise, but I can't recall them

If we decide to go this route, I can definitely do that.

toddburnside:
> If we decide to go this route, I can definitely do that.

I might also change some names. 😆 I couldn't come up with good ones at the time so just made some verbose ones.

toddburnside:
Terribly out of date, and we don't know what, if anything, we're going to with potential odd behaviors of undo/redo.

PR #1127, opened 2021-09-28
Proposal category groups

toddburnside:
image
Some additional styling may be desired for the header.

This is not a new issue, but we probably want to keep the width from resizing when longer items are selected, too.

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -282,11 +282,15 @@ tfoot {
 }
 
 // Used as part of creating headers in a FormSelect.
+// Selector is very specific to avoid affecting other controls
 .ui.dropdown.selection .menu .item.header.disabled {
-  // This needs !important because SUIR sets it to 'all' via inline style.
-  // Selector is very specific to avoid affecting other controls
-  pointer-events: none !important;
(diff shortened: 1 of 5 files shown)


cquiroz (approved):
The list looks great

rpiaggio:
Looks very nice.

A very minor thing: if I click on a header, the selection doesn't change (which is the expected behavior) but the dropdown still collapses. Is there a way to avoid this?

rpiaggio (approved):
LGTM

toddburnside:
> A very minor thing: if I click on a header, the selection doesn't change (which is the expected behavior) but the dropdown still collapses. Is there a way to avoid this?

I've got a couple of ideas.

toddburnside:
> > A very minor thing: if I click on a header, the selection doesn't change (which is the expected behavior) but the dropdown still collapses. Is there a way to avoid this?
> 
> I've got a couple of ideas.

If we don't use the `pointing-events: none` it fixes this issue. However, hovering over the header did have a highlighting effect. I didn't like that, but I guess that's really a matter of personal preference. I removed `pointing-events` and set some CSS to make hovering do nothing. We can change that as desired.

PR #1133, opened 2021-09-28
Change scalafmt.conf

toddburnside:
Use file from lucuma-ui

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,10 +1,8 @@
 version = "3.0.5"
-style = default
+
+style = defaultWithAlign
 
 maxColumn = 100
-
-// Vertical alignment is pretty, but leads to bigger diffs
-align.preset = most
 
 rewrite.rules = [
   AvoidInfix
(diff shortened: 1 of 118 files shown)


rpiaggio:
I missed the comment in lucuma-ui and I see now that we lost aligning `=>`s in `match/case`s and `<-` in `for`s. Or there seems to be a low threshold to keep them aligned. It'd be cool to just get those back.

toddburnside:
> I missed the comment in lucuma-ui and I see now that we lost aligning `=>`s in `match/case`s and `<-` in `for`s. Or there seems to be a low threshold to keep them aligned. It'd be cool to just get those back.

Yeah. It's a bit odd. For example, in ConnectionsStatus.scala some, but not all of the `=>` are aligned.

cquiroz (approved):
LGTM

toddburnside:
Interestingly, it only seems to be possible to get the alignment of the `=>` and `<-` with `align.preset = most`, not with any of the individual settings. Of course, that also brings back the aligned `=`, even for definitions separated by new lines. 

Also, I like the `align.arrowEnumeratorGenerator = true` setting, which gives 

image

instead of 

image

This is something we had previously with the `most` preset, but can also be set independently.

I wonder if we should wait for the fix to https://github.com/scalameta/scalafmt/issues/2775 in https://github.com/scalameta/scalafmt/pull/2753 and go back to the `most` preset?

cquiroz:
I also prefer the alignment inside the for loops

We can certainly wait for upstream fixes

toddburnside:
> I also prefer the alignment inside the for loops
> 
> We can certainly wait for upstream fixes

The failure to sometimes not align the `=>` and `<-` had to do with multiline statements. `align.multiline = true` needed to be set, but that also caused the extreme alignment of the `=` even across spaces. There have been 2 PRs merged to fix the multiline issues. 

Unless someone objects, I'll close this PR and see what happens when a new version of scalafmt-core is released. I would like to compare what happens with `align.preset = most` vs what is in this PR with the addition of `align.multiline = true` and `align.arrowEnumeratorGenerator = true`.

toddburnside:
Just a note for future work. If we use `align.preset = most` I think we can get rid of all (or at least most) of the `align.tokens` settings as they are included in `most`.

toddburnside:
Waiting on new version of scalafmt.

PR #1137, opened 2021-09-30
Save sort order of constraints summary for duration of session

toddburnside:
Not yet being saved to the user preferences service.

### DIFF
--- a/explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
+++ b/explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
@@ -34,6 +34,8 @@ import reactST.reactTable._
 import reactST.reactTable.mod.Cell
 import reactST.reactTable.mod.DefaultSortTypes
 import reactST.reactTable.mod.IdType
+import reactST.reactTable.mod.SortingRule
+import reactST.reactTable.mod.TableState
 
 import scala.collection.immutable.SortedSet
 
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
```diff
@@ -73,10 +76,20 @@ object ConstraintsSummaryTable {
...
+  private val columnClasses: Map[String, Css]                                             = Map(
     "edit" -> (ExploreStyles.Sticky |+| ExploreStyles.ConstraintsSummaryEdit)
   )
 
+  private def toSortingRules(tuples: List[(String, Boolean)])                             = tuples.map { case (id, b) =>
+    SortingRule[ConstraintGroup](id).setDesc(b)
+  }.toJSArray
+
+  private def fromTableState(state: TableState[ConstraintGroup]): List[(String, Boolean)] = state
+    .asInstanceOf[ConstraintsTable.StateType]
```
rpiaggio, line 88: I would expect things to work if we define `state: ConstraintsTable.StateType` and avoid the cast.

rpiaggio (approved):
LGTM.

explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
toddburnside: Except that `tableInstance.state` returns a `TableState` (no sorting state mixed in). I don't know of another way to obtain the state, but I may be missing something.

rpiaggio: Ah, right `TableInstance` facade is generated by ScalablyTyped and is not type-parametrized beyond `D`. Sorry, got mixed up.

cquiroz:
Should we store this in the persistent user preferences?

toddburnside:
> Should we store this in the persistent user preferences?

That's the next step.

PR #1147, opened 2021-10-05
Show sky brightness percentiles in plot.

rpiaggio:
image

Also in this PR:
- Add `scalajs-react-virtuoso` as direct dependency, since we are using it directly.
-


cquiroz (approved):
LGTM

PR #1150, opened 2021-10-05
Edit single observation constraint set on constraints tab.

toddburnside:
This also optimistically updates the constraint group list in memory. Waiting for a round trip from the server became too painful with all of the changes that can happen when editing the constraints for a single observation.

### DIFF
--- a/explore/src/main/scala/explore/Routing.scala
+++ b/explore/src/main/scala/explore/Routing.scala
@@ -76,7 +76,7 @@ object Routing {
             RootModel.expandedIds.andThen(ExpandedIds.constraintSetObsIds)
           ),
           model.zoom(RootModel.undoStacks).zoom(ModelUndoStacks.forConstraintList),
-          model.zoom(RootModel.undoStacks).zoom(ModelUndoStacks.forBulkConstraintSet),
+          model.zoom(RootModel.undoStacks).zoom(ModelUndoStacks.forConstraintGroup),
           model.zoom(RootModel.constraintSummaryHiddenColumns),
           model.zoom(RootModel.constraintSummarySorting),
           size
@@ -116,7 +116,10 @@ object Routing {
             targetTab
           )
           | staticRoute("/configurations", ConfigurationsPage) ~> render(SequenceEditor())
-          | staticRoute("/constraints", ConstraintsBasePage) ~> renderP(constraintSetTab))
+          | staticRoute("/constraints", ConstraintsBasePage) ~> renderP(constraintSetTab)
+          | dynamicRouteCT(
+            ("/constraints/obs" / id[Observation.Id]).xmapL(ConstraintsObsPage.obsId)
+          ) ~> renderP(constraintSetTab))
 
       val configuration =
         rules
--- a/explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
+++ b/explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
@@ -105,6 +105,9 @@ object ConstraintGroupObsList {
 
       val constraintGroups = props.constraintsWithObs.get.constraintGroups
 
+      // if a single observation is selected
+      val singleObsSelected = props.focused.get.collect { case FocusedObs(_) => true }.nonEmpty
+
       val state   = ViewF.fromStateSyncIO($)
       val undoCtx = UndoContext(
         props.undoStacks,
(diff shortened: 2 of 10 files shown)


explore/src/main/scala/explore/Routing.scala
```diff
@@ -116,7 +116,10 @@ object Routing {
             targetTab
           )
           | staticRoute("/configurations", ConfigurationsPage) ~> render(SequenceEditor())
-          | staticRoute("/constraints", ConstraintsBasePage) ~> renderP(constraintSetTab))
+          | staticRoute("/constraints", ConstraintsBasePage) ~> renderP(constraintSetTab)
+          | dynamicRouteCT(
+            ("/constraints/obs" / id[Observation.Id]).xmapL(ConstraintsObsPage.obsId)
```
cquiroz, line 121: Will there be a url for a given obs in constraints? Isn't that kind of the inverse of the hirerachy? as in obs have constraints?

explore/src/main/scala/explore/tabs/ConstraintSetTabContents.scala
```diff
@@ -140,26 +185,68 @@ object ConstraintSetTabContents {
...
+
+                mergeWithIds.fold(splitList) { idsToMerge =>
+                  val combined = idsToMerge ++ idsToEdit
+                  splitList
+                    .removed(idsToMerge)
+                    .removed(idsToEdit)
+                    .updated(combined, ConstraintGroup(newCg.constraintSet, combined))
+                }
+              }
+              .getOrElse(cgl) // shouldn't happen
```
cquiroz, line 231: Famous last words

explore/src/main/scala/explore/Routing.scala
toddburnside: Well, isn't editing the constraint for an observation in the constraints tab kind of the inverse of the hierarchy? 😆 Anyway, it seems consistent to me, but I can remove it if desired.

cquiroz: Right, it matches what we see on screen. I was thinking it didn't make sense to do that but looking at it in practice it doesn't seem too bad

explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
```diff
@@ -189,16 +198,21 @@ object ConstraintGroupObsList {
...
                         TagMod.when(props.expandedIds.get.contains(obsIds))(
                           cgObs.zipWithIndex.toTagMod { case (obs, idx) =>
-                            props.renderObsBadgeItem(selectable = false,
-                                                     highlightSelected = false,
-                                                     linkToObsTab = true
+                            props.renderObsBadgeItem(
+                              selectable = true,
+                              highlightSelected = true,
+                              forceHighlight = groupSelected && !singleObsSelected,
+                              linkToObsTab = false,
```
rpiaggio, line 213: We could keep `linkToObsTab`. That triggers on double click, right?

explore/src/main/scala/explore/observationtree/ViewCommon.scala
```diff
@@ -10,24 +10,33 @@ import explore.components.ui.ExploreStyles
...
 import react.beautifuldnd._
 
 trait ViewCommon {
   def focused: View[Option[Focused]]
 
-  def renderObsBadge(obs: ObsSummary, highlightSelected: Boolean = true): TagMod =
+  def renderObsBadge(
+    obs:               ObsSummary,
+    highlightSelected: Boolean = true,
+    forceHighlight:    Boolean = false // if true, overrides highlightSelected
```
rpiaggio, line 26: Why do we need `forceHighlight`?

rpiaggio: Ah, is it to highlight all observations in a group?

rpiaggio: Ah, is it to highlight all observations in a group?

explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
toddburnside: We could, but I was worried about inadvertent double clicks sending the user off somewhere. It doesn't seem as necessary if they can edit the constraint set for the observation here. If we want to keep a link to the observation, I think it might be nicer to make the observation id in the upper right of the badge be a link. That would be more obvious. With the double click, there isn't any indication that is going to happen.

explore/src/main/scala/explore/observationtree/ViewCommon.scala
toddburnside: Exactly.

PR #1160, opened 2021-10-07
SiderealTargetEditor

rpiaggio:
The target editor is back online.

There are still tons of commented code.

The `SiderealTargetEditor` is pretty much done, but it's glued to the rest of explore by tenuous code (in `TargetTile`) that, for the moment, allows editing only the first of the science targets (and crashes if there are none).

BONUS: Publicly released versions of all dependencies :)


toddburnside (approved):
LGTM

PR #1168, opened 2021-10-12
Target editor with table

rpiaggio:
This is far from being done, but I changed our model, so I'd like to merge before going on so that we can sync.


model/shared/src/main/scala/explore/model/ModelOptics.scala
```diff
@@ -3,11 +3,11 @@
 
 package explore.model
 
-import lucuma.core.math.Coordinates
-import lucuma.core.math.Declination
-import lucuma.core.math.RightAscension
-import lucuma.core.model.SiderealTracking
-import monocle.Lens
+// import lucuma.core.math.Coordinates
```
rpiaggio, line 6: I think we can remove this file, but I'm not sure yet.

rpiaggio:
For the moment `TargetTile.scala` is glue code where I'm not removing commented code yet. All the rest is reviewable.

cquiroz (approved):
LGTM

PR #1172, opened 2021-10-13
Multi-target ids.

rpiaggio:
Target editor can now edit a single or multiple targets. Depending on the case, we may receive from the API a single `Target.Id` or multiple ones.

This PR introduces `ScienceTarget.Id` which is a sum type containing cases to wrap a single `Target.Id` or a `NonEmptySet[Target.Id]`.

All relevant mentions of `Target.Id` (undo stacks, searching, etc) are switched to `ScienceTarget.Id`.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinCell.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinCell.scala
@@ -15,6 +15,7 @@ import explore.common.UserPreferencesQueriesGQL._
 import explore.components.ui.ExploreStyles
 import explore.implicits._
 import explore.model.ModelOptics
+import explore.model.ScienceTarget
 import explore.model.TargetVisualOptions
 import explore.model.reusability._
 import japgolly.scalajs.react.ReactMonocle._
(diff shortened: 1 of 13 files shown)


cquiroz (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -84,7 +84,8 @@ object AladinCell extends ModelOptics {
         implicit val ctx = props.ctx
         $.setStateL(State.fov)(fov) >>
           UserTargetPreferencesUpsert
-            .updateFov[IO](props.uid, props.tid, fov.x)
+            // TODO Accept multiple Target ids.
```
toddburnside, line 87: This is another area that is a bit problematic for the UI. It will be difficult to not "reset" the FOV when updating the observations in a target * group.

toddburnside: Except that names are not unique. Even if the API is changed to require uniqueness within a target environment, I don't think they'll need to be unique within a program.

PR #1173, opened 2021-10-13
Implicit conversions from model types to clue inputs

toddburnside:


cquiroz:
I'll Probably Move my conversions to another package

PR #1174, opened 2021-10-14
Restore targets tab with list of target lists.

toddburnside:
No summary table or editor, yet. Lots of work left to do in TargetTabContents.scala, so I left a bunch of commented code.

As mentioned in the meeting, I added an icon to the group heading if there is an "unmoored" version of the target list. It looks like this: 
image
If we go this route, we'll probably want to add more styling. It should be pretty easy to make it a button to create an unmoored instance of it if we're asked to do so.

### DIFF
--- a/model/shared/src/main/scala/explore/model/ObsSummary.scala
+++ b/model/shared/src/main/scala/explore/model/ObsSummary.scala
@@ -35,7 +35,6 @@ object ObsSummary {
 
 trait ObsWithConstraints extends ObsSummary {
   val constraints: ConstraintsSummary
-  // val targetEnvId: TargetEnvironment.Id
 
   lazy val constraintsSummary = constraints.summaryString
 }


model/shared/src/main/scala/explore/model/TargetListGroup.scala
```diff
@@ -0,0 +1,50 @@
...
+import lucuma.core.model.Target
+import lucuma.core.model.TargetEnvironment
+import monocle.Focus
+import monocle.Lens
+
+import scala.collection.immutable.SortedSet
+
+case class TargetListGroup(
+  obsIds:       SortedSet[Observation.Id],
+  targetEnvIds: SortedSet[TargetEnvironment.Id],
```
rpiaggio, line 20: Should this be a `NonEmptySet` ? Or is it used only for unmoored environments?

model/shared/src/main/scala/explore/model/ObsSummary.scala
```diff
@@ -34,6 +35,7 @@ object ObsSummary {
 
 trait ObsWithConstraints extends ObsSummary {
   val constraints: ConstraintsSummary
+  // val targetEnvId: TargetEnvironment.Id
```
rpiaggio, line 38: Commented code

model/shared/src/main/scala/explore/model/TargetListGroup.scala
toddburnside: It could be a NonEmptySet, but since it is going to be inside another SortedSet (for the expanded ids) it has to have an ordering. I don't think a NonEmptySet does and I didn't see another NonEmptySortedSet with a brief look. Of course, the next question is: do the expanded ids need to be sorted sets?

toddburnside: Perfect. I didn't realize that. I'll make it a NonEmptySet in the next PR.

PR #1176, opened 2021-10-15
Add/delete targets in groups

rpiaggio:
No undo/redo for these actions yet.

Kapture 2021-10-14 at 21 00 50


explore/src/main/scala/explore/targeteditor/TargetEnvEditor.scala
```diff
@@ -44,30 +69,135 @@ object TargetEnvEditor {
...
+                ): VdomNode
+              )
+            )
+          ),
           TargetTable(
-            props.targetEnv.get.scienceTargets.toList.map(_._2),
+            props.targetEnv.zoom(TargetEnv.scienceTargets),
             props.hiddenColumns,
+            selectedTargetId,
+            // selectedTargetId,
```
toddburnside, line 191: Duplicate commented code.

model/shared/src/main/scala/explore/model/ScienceTarget.scala
```diff
@@ -119,6 +125,13 @@ object ScienceTarget {
   }
 
   object Id {
+    def fromTargetIdList(targetIds: List[Target.Id]): Option[ScienceTarget.Id] =
```
toddburnside, line 128: Nice. I was thinking I would need something like this.

explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
```diff
@@ -45,6 +45,7 @@ import lucuma.ui.reusability._
 import monocle.Iso
 import react.common._
 import react.semanticui.collections.form.Form
+import react.semanticui.elements.divider._
```
toddburnside, line 48: CI Build says `import explore.components.Tile` above is unused.

explore/src/main/scala/explore/targeteditor/TargetEnvEditor.scala
```diff
@@ -44,30 +69,135 @@ object TargetEnvEditor {
...
+    name:           NonEmptyString,
+    searching:      View[Set[ScienceTarget.Id]],
+    selectedTarget: View[Option[ScienceTarget.Id]]
+  )(implicit ctx:   AppContextIO): IO[Unit] =
+    TargetEnvQueriesGQL.AddSiderealTarget
+      .execute(
+        List(targetEnv.get.id),
+        newTarget(name).toCreateInput
+      ) >>= { response =>
+      val targetIds = response.updateScienceTargetList.flatMap(_.edits.map(_.target.id))
```
toddburnside, line 97: I guess this currently only handles a single environment? I guess when you support multiple you can pull the target environment id back from `updateScienceTargetList` for each new target id. I need to think through how the right information gets back to the target group lists. I realize now that, for editing, I'll need the target ids for each observation, too. Not just the target environment id. But, updating the individual observations when a group is being edited and a target is added isn't clear to me right now. 🤔

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/TargetEnvEditor.scala
rpiaggio: Yes, we can request the `targetEnvironmentId` for each added `target.id`.

Let's see what issues we have to deal with when we attempt to glue everything together. I'll start by adding support for multi-id target environments.

PR #1223, opened 2021-10-29
Give some ids their own types

toddburnside:
Makes types for TargetEnvIdObsId and TargetEnvIdObsIdSet. When we switch to scala 3, these can be made opaque types. This is a test PR before doing the same for all ids in model/package.scala.

### DIFF
--- a/common/src/main/scala/explore/common/TargetListGroupQueries.scala
+++ b/common/src/main/scala/explore/common/TargetListGroupQueries.scala
@@ -15,6 +15,7 @@ import explore.components.graphql.LiveQueryRenderMod
 import explore.implicits._
 import explore.model.ObsSummaryWithConstraints
 import explore.model.TargetEnv
+import explore.model.TargetEnvIdObsId
 import explore.model.TargetEnvIdObsIdSet
 import explore.schemas.implicits._
 import explore.utils._
@@ -38,7 +39,7 @@ object TargetListGroupQueries {
   // This is used for sorting the TargetListGroupObsList. If we change to sort by name or something
   // else, we can remove this.
   implicit val orderSortedSet: Order[TargetEnvIdObsIdSet] =
-    Order.by(_.toList.map(t => (t._2, t._1)))
+    TargetEnvIdObsIdSet.orderByObsThenTargetEnv
 
   type ObservationResult = TargetListGroupObsQuery.Data.Observations.Nodes
   val ObservationResult = TargetListGroupObsQuery.Data.Observations.Nodes
@@ -58,7 +59,7 @@ object TargetListGroupQueries {
       targetListGroups
         .get(targetEnvObsIds)
         .fold(observations.mapFilter { obsSumm =>
-          if (targetEnvObsIds.contains((obsSumm.targetEnvId, obsSumm.id.some)))
+          if (targetEnvObsIds.contains(TargetEnvIdObsId((obsSumm.targetEnvId, obsSumm.id.some))))
             obsSumm.scienceTargetIds.some
           else none
         }.combineAll)(_.targetIds)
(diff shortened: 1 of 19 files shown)


model/shared/src/main/scala/explore/model/NonEmptySetWrapper.scala
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
```
toddburnside, line 4: I wonder if this file should go in `explore.model.util` to match the pattern in `lucuma.core`? The current `utils.scala` could be made a package object in the new package and references changed (to remove the `s`).

common/src/main/scala/explore/common/TargetListGroupQueries.scala
```diff
@@ -38,7 +39,7 @@ object TargetListGroupQueries {
   // This is used for sorting the TargetListGroupObsList. If we change to sort by name or something
   // else, we can remove this.
   implicit val orderSortedSet: Order[TargetEnvIdObsIdSet] =
```
rpiaggio, line 41: Although I came up with the name `TargetEnvIdObsId`, I'm not too happy with it. If anyone can think of a better one  it would be welcome.

explore/src/main/scala/explore/observationtree/TargetListGroupObsList.scala
```diff
@@ -231,10 +205,10 @@ object TargetListGroupObsList {
         val cgObs           = obsIds.toList.map(id => observations.get(id)).flatten
         // if this group or something in it is selected
         val groupSelected   =
-          props.selected.get.optValue.exists(_.intersect(targetEnvObsIds).nonEmpty)
+          props.selected.get.optValue.exists(_.hasIntersect(targetEnvObsIds))
```
rpiaggio, line 208: `intersects` ?

model/shared/src/main/scala/explore/model/NonEmptySetWrapper.scala
rpiaggio: Sounds good to me.

model/shared/src/main/scala/explore/model/NonEmptySetWrapper.scala
```diff
@@ -0,0 +1,59 @@
...
+import cats.data.NonEmptySet
+import cats.syntax.all._
+import monocle.Iso
+
+import scala.collection.immutable.SortedSet
+
+final class NonEmptySetWrapper[IdSet, Id](self: IdSet, iso: Iso[IdSet, NonEmptySet[Id]]) {
+  private val selfSet = iso.get(self)
+
+  // expose the NonEmptySet methods
```
rpiaggio, line 15: Maybe we can `@inline` these?

rpiaggio (approved):
This is totally awesome. It improves the code much more than I expected and reduces the cognitive load when dealing with these types. Thank you!

LGTM!

rpiaggio:
Can't wait for the other types 😁

common/src/main/scala/explore/common/TargetListGroupQueries.scala
toddburnside: Well, it is a `TargetEnv.id`. So we could call it a `TargetEnvId`. I think on the UI we just need to be aware that a `TargetEnv` is not a `TargetEnvironment`. We already have that distinction and I don't think it is an issue in practice.

explore/src/main/scala/explore/observationtree/TargetListGroupObsList.scala
toddburnside: Much better.

model/shared/src/main/scala/explore/model/NonEmptySetWrapper.scala
toddburnside: Certainly. :)

common/src/main/scala/explore/common/TargetListGroupQueries.scala
toddburnside: Except we're already using `TargetEnvIdSet` for a `NonEmptySet[TargetEnvironment]`

toddburnside: Will do in separate PR if we do it.

PR #1248, opened 2021-11-03
Back to Callback

rpiaggio:
Use `Callback` again instead of `SyncIO`, now that we have a `scalajs-react` bundle that uses `CallbackTo` for sync effects and `IO` for async.

### DIFF
--- a/common/src/main/scala/explore/model/AppContext.scala
+++ b/common/src/main/scala/explore/model/AppContext.scala
@@ -21,6 +21,7 @@ import explore.schemas._
 import explore.utils
 import io.circe.Json
 import japgolly.scalajs.react.Callback
+import japgolly.scalajs.react.util.DefaultEffects.{ Sync => DefaultS }
 import lucuma.schemas._
 import org.http4s._
 import org.http4s.dom.FetchClientBuilder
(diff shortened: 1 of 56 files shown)


common/src/main/scala/explore/model/AppContext.scala
```diff
@@ -106,38 +107,38 @@ case class AppContext[F[_]](
...
-        f(logger.warn(message))
-      def info(message: => String): SyncIO[Unit]                =
-        f(logger.info(message))
-      def debug(message: => String): SyncIO[Unit]               =
-        f(logger.debug(message))
-      def trace(message: => String): SyncIO[Unit]               =
-        f(logger.trace(message))
-    }
-  }
+  // val syncLogger: Logger[DefaultS] = {
```
toddburnside, line 117: Commented code?

explore/src/main/scala/explore/HelpBody.scala
```diff
@@ -136,11 +136,11 @@ object HelpBody {
         }
       }
       .componentDidMount { $ =>
-        implicit val ctx = $.props.ctx
+        // implicit val ctx = $.props.ctx
```
toddburnside, line 139: More commented code.

common/src/main/scala/explore/model/AppContext.scala
rpiaggio: Thx for catching these :+1:

toddburnside (approved):
LGTM - but I don't think tree.txt is needed.

rpiaggio:
> LGTM - but I don't think tree.txt is needed.

That one slipped in :D Removed in latest commit.

PR #1289, opened 2021-11-16
Calculate all rows on the configuration list and support sorting

cquiroz:
At the moment we only query the visible itc rows, however to support sorting we need to query each row
This PR does it prioritizing the visible rows

Note this leads to quite many requests and though in many cases the http cache will be hit it still places some extra load on the browser. We need to add an extra layer of persistent caching to improve it

### DIFF
--- a/explore/src/main/scala/explore/config/ConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/ConfigurationPanel.scala
@@ -16,6 +16,7 @@ import explore.components.Tile
 import explore.components.ui.ExploreStyles
 import explore.components.undo.UndoButtons
 import explore.implicits._
+import explore.model.AvailableFilter
 import explore.model.ImagingConfigurationOptions
 import explore.model.SpectroscopyConfigurationOptions
 import explore.model.display._
@@ -36,6 +37,8 @@ import react.common._
 import react.semanticui.collections.form.Form
 import react.semanticui.sizes._
 
+import scala.collection.SortedSet
+
 final case class ConfigurationPanel(
   obsId:            Observation.Id,
   scienceDataUndo:  UndoCtx[ScienceData],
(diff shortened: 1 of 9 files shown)


explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -46,13 +49,17 @@ final case class ConfigurationPanel(
 object ConfigurationPanel {
   type Props = ConfigurationPanel
 
-  implicit val propsReuse: Reusability[Props] = Reusability.derive
-  implicit val stateReuse: Reusability[State] = Reusability.never
+  implicit val propsReuse: Reusability[Props]                          = Reusability.derive
+  implicit val filterReuse: Reusability[AvailableFilter]               = Reusability.byEq
+  implicit val filterSetReuse: Reusability[SortedSet[AvailableFilter]] = Reusability.by(_.toSet)
+  implicit val optionsReuse: Reusability[ImagingConfigurationOptions]  = Reusability.derive
```
rpiaggio, line 55: Maybe this should go in `explore.model.reusability`.  Also, `lucuma.ui.reusability` can derive a `Reusability[SortedSet[A]]` from a `Reusability[A]`.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -168,33 +223,11 @@ object SpectroscopyModesTable extends ItcColumn {
...
-              val secs = t.toMillis / 1000.0 * c
-              f"$secs%.0f s"
-            case Right(ItcResult.Pending)         =>
-              Icons.Spinner.spin(true)
-            case Right(ItcResult.SourceTooBright) =>
-              Popup(content = "Source too bright", trigger = Icons.SunBright.color("yellow"))
-          }
-          <.div(ExploreStyles.ITCCell, content)
-        }
+        .setCell(c => Reusable.implicitly(c.value).map(p => itcCell(p)))
```
rpiaggio, line 226: After a 2nd look at this, I don't think `itcCell` is being reused here.

The render function passed to `setCell` goes straight to `react-table` in JS land. So, it seems the `Reusable` is being recreated in each render here, without it getting compared to the previous instance.

Of course, there's a chance I may be wrong. Do you notice an improvement over just having `.setCell(c => itcCell(c.value))` ?

cquiroz: I'm not sure either. I guess there will be nobody to handle the reusabillity result?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -155,6 +159,57 @@ object SpectroscopyModesTable extends ItcColumn {
...
+  }
+
+  implicit val reuseQueryResponse: Reusability[EitherNec[ItcQueryProblems, ItcResult]] =
+    Reusability.byEq
+
+  def sortItcFun(
+    itc: ItcResultsCache,
+    cw:  Option[Wavelength],
+    sn:  Option[PosBigDecimal]
+  ): SortByFn[SpectroscopyModeRow] =
```
rpiaggio, line 193: I wonder if we can use an `Ordering[(SpectroscopyModeRow, ItcResultsCache, Option[Wavelength], Option[PosBigDecimal])]` (or `cats.Order`, which we can test); and then use `setSortByFn` on the column.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
rpiaggio: Exactly, so the reusable value is just being resolved locally. In that case, we could be degrading performance instead of improving it.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: I'll investigate this in another PR

PR #1295, opened 2021-11-19
FOV from catalog in target search popup.

rpiaggio:
Automatically set FOV of Aladin in the target search popup if the angular size is present in the catalog.

For the moment this only applies to Aladin within the target search popup, future PR will address initializing the instance of Aladin in the target editor.


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -387,7 +387,7 @@ object SpectroscopyModesTable extends ItcColumn {
         )
       )((props, _, itcResults, _, _, _, _, _) => { case (wv, sn, rows, range) =>
         implicit val ctx = props.ctx
-        range.value.map(r => updateITCOnScroll(wv, sn, r, rows, itcResults.withEffect)).orEmpty
+        range.value.map(r => updateITCOnScroll(wv, sn, r, rows, itcResults)).orEmpty
```
cquiroz, line 390: ???

cquiroz (approved):
Great improvement, thanks

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -29,7 +29,7 @@ import react.common._
 import scala.annotation.nowarn
 import scala.concurrent.duration._
 
-final case class AladinContainer(
+final case class AladinContainer( // TODO Use target's FOV if not set in user prefs.
```
toddburnside, line 32: Depending on the target model we choose, saving the target FOV in user prefs will be more or less difficult. Currently in staging, I don't think it really works with the unshared TargetEnv (although I'd have to look). Maybe we could use target name? With an unshared model, using target id has limitations.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
rpiaggio: Ooops, sorry. I was trying things out and this slipped through.

This is unrelated, but it seems the `withEffect` is not required.

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
rpiaggio: Understood.

PR #1299, opened 2021-11-22
Unstackable target tables.

rpiaggio:
Don't stack columns in target tables in small screens.

Changes in `SUITable` are just general improvements, which are necessary when using `useFlexLayout` (which I ended up not using, but I'm leaving the changes there in case we ever do).


common/src/main/scala/reactST/reactTable/SUITable.scala
```diff
@@ -145,7 +160,7 @@ class SUITable[D, Plugins, Layout](
             )
           )
         )
-      }))
+      })(^.key := "header"))
```
cquiroz, line 163: Would this mean that if we have multiple tables, we'd have repeated keys? Maybe it is not a problem but what about nested tables?

cquiroz (approved):
LGTM

PR #1302, opened 2021-11-22
Convert ConfigurationPanel to a functional component

cquiroz:
This contains a bit of refactoring and the conversion of `ConfigurationPanel` to a functional component, this is mostly to get some practice using hooks

### DIFF
--- a/explore/src/main/scala/explore/config/ConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/ConfigurationPanel.scala
@@ -86,63 +86,55 @@ object ConfigurationPanel {
       op.runS(SpectroscopyRequirementsData()).value
     }
 
-  private def renderFn(
-    props:           Props,
-    scienceDataUndo: UndoContext[ScienceData],
-    mode:            StateSnapshot[ScienceMode],
-    imaging:         StateSnapshot[ImagingConfigurationOptions]
-  )(implicit ctx:    AppContextIO): VdomNode = {
-    val requirementsCtx = scienceDataUndo.zoom(ScienceData.requirements)
-
-    val requirementsViewSet = UndoView(props.obsId, requirementsCtx)
-
-    val isSpectroscopy = mode.value === ScienceMode.Spectroscopy
-
-    val spectroscopy = requirementsViewSet(
-      ScienceRequirementsData.spectroscopyRequirements,
-      UpdateScienceRequirements.spectroscopyRequirements
-    )
-
-    // val imaging = state.zoomStateL(State.imagingOptions)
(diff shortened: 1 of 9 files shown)


explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -100,70 +101,70 @@ object ConfigurationPanel {
...
-      .renderBackend[Backend]
-      .configure(Reusability.shouldComponentUpdate)
-      .build
+    ScalaFnComponent
+      .withHooks[Props]
+      .useStateSnapshotWithReuse(
+        State(ScienceMode.Spectroscopy, ImagingConfigurationOptions.Default)
+      )
+      .render { (props, state) =>
+        AppCtx.using { implicit appCtx =>
```
rpiaggio, line 162: We already have the context in `props.ctx`, no need to go get it from the `AppCtx` object. We can either `implicit val ctx = props.ctx` or pass it explicitly to `renderFn`.

(This could have been optimized like in the class component version too).

cquiroz: an nice

explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -100,70 +101,70 @@ object ConfigurationPanel {
...
-    ScalaComponent
-      .builder[Props]
-      .initialState(State(ScienceMode.Spectroscopy, ImagingConfigurationOptions.Default))
-      .renderBackend[Backend]
-      .configure(Reusability.shouldComponentUpdate)
-      .build
+    ScalaFnComponent
+      .withHooks[Props]
+      .useStateSnapshotWithReuse(
+        State(ScienceMode.Spectroscopy, ImagingConfigurationOptions.Default)
```
rpiaggio, line 159: With hooks we can call `useState*` multiple times. We could avoid defining the `State` case class and have a `mode` state and a `imagingOptions` state. But it works fine either way.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/ConfigurationPanel.scala
cquiroz: That's interesting, I'll try that a bit

rpiaggio: You can even use a StateSnapshot for one of them and a regular state for the other one.

toddburnside (approved):
LGTM

PR #1303, opened 2021-11-23
Use StateSnapshot

cquiroz:
I'm not sure why this works, it seems the cache is not being updated unlike when using `Hooks.UseState`


cquiroz:
@rpiaggio has confirmed this fails due to an upstream bug on `StateSnapshot`. I'm discarding this change

PR #1305, opened 2021-11-23
Sticky columns.

rpiaggio:
This was all a bit trickier than I expected but it seems to be working now.

Applies to all target tables.

Kapture 2021-11-23 at 14 39 56


toddburnside (approved):
LGTM

PR #1308, opened 2021-11-25
ITC Requests queue

cquiroz:
The modes tables requires one ITC request per row, To support sorting we need to hit each possible row. The current naive implementation can thus send dozens or hundreds of requests and we can see it starts choking the browser. In particular the websocket downstream channel gets blocked with all those requests

This PR is maybe overkill but address the situation creating a queue of requests that are being sent at a certain intervals, thus filling the modes table a bit more slowly but on the other hand making the rest of the app more responsive.

There are a few details for future work:
* We should be able to stop the queue cleanly on unmount
* We may want the ability to cancel o de prioritize some requests
* Maybe parallelism can be increased without negatively affecting responsiveness

I have to say it is quite impressive that we can use the regular cats-effect and fs2 constructs to do this relatively complex engine on the browser

### DIFF
--- a/explore/src/main/scala/explore/config/ConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/ConfigurationPanel.scala
@@ -94,7 +94,9 @@ object ConfigurationPanel {
       .withHooks[Props]
       .useStateSnapshotWithReuse(none[ITCRequestsQueue[IO]])
       .useEffectOnMountBy { (props, st) =>
+        // Create the requests queue and put it on the state
         implicit val ctx: AppContextIO = props.ctx
+        // TODO proper cleanup
         ITCRequestsQueue.build[IO]((r: Option[ITCRequestsQueue[IO]]) => st.setState(r).to[IO])
       }
       .useStateSnapshotWithReuse[ScienceMode](ScienceMode.Spectroscopy)
--- a/explore/src/main/scala/explore/config/ConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/ConfigurationPanel.scala
@@ -3,6 +3,7 @@
 
 package explore.config
 
+import cats.effect.IO
 import cats.syntax.all._
 import coulomb.Quantity
 import crystal.react.implicits._
@@ -14,6 +15,7 @@ import explore.components.Tile
 import explore.components.ui.ExploreStyles
 import explore.components.undo.UndoButtons
 import explore.implicits._
+import explore.itc._
 import explore.model.ImagingConfigurationOptions
 import explore.model.SpectroscopyConfigurationOptions
 import explore.model.display._
@@ -21,6 +23,7 @@ import explore.model.reusability._
 import explore.undo.UndoContext
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.extra._
+import japgolly.scalajs.react.util.syntax._
 import japgolly.scalajs.react.vdom.html_<^._
 import lucuma.core.enum.ScienceMode
 import lucuma.core.math.Wavelength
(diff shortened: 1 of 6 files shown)


vite.config.js
```diff
@@ -95,7 +95,6 @@ module.exports = ({ command, mode }) => {
               cert: fs.readFileSync("server.cert"),
             },
       watch: {
-        usePolling: true,
```
rpiaggio, line 98: I think I added this at some point to fix that vite was not detecting changes in my filesystem due to reasons beyond my comprehension. Is it affecting performance?

cquiroz: Yes, I see `node` using 50% of cpu even when not doing any changes. Removing polling gets it back to normal

swalker2m:
I have a naive question.  Could/should we put more of this on the ITC server itself?  Instead of 100 requests one large request for all the rows and then stream responses as they are calculated.

cquiroz:
Not naive at all. The ITC server supports sending multiple modes in a single requests but doing so affects caching performance. It would be harder to benefit from HTTP cache doing them in batches as they would rarely match the url.
Another downside is that you'd get a set of responses instead of one by one.
Anyway the option is there if it can be used, especially for CLI clients

cquiroz:
Now that I think about, the ITC server is not supporting websockets but maybe it could

rpiaggio:
> Now that I think about, the ITC server is not supporting websockets but maybe it could

Just elaborating on the idea: So the query would actually be a GQL subscription where we pass all the queries toghether and receive a stream or responses. We could remove from the request the results that are cached.

cquiroz:
The HTTP cache also makes the requests not hit ITC and it reduces its load. I suppose we could cache the results in memory on ITC?

explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -89,52 +92,58 @@ object ConfigurationPanel {
   protected val component =
     ScalaFnComponent
       .withHooks[Props]
+      .useStateSnapshotWithReuse(none[ITCRequestsQueue[IO]])
+      .useEffectOnMountBy { (props, st) =>
+        // Create the requests queue and put it on the state
+        implicit val ctx: AppContextIO = props.ctx
+        // TODO proper cleanup
+        ITCRequestsQueue.build[IO]((r: Option[ITCRequestsQueue[IO]]) => st.setState(r).to[IO])
```
rpiaggio, line 100: Maybe we can separate concerns by having `build` return a `F[ITCRequestsQueue]` and then flatMap to `setState` locally ?

vite.config.js
rpiaggio: Thanks for tracking this down, I'm having the CPU issue as well.

explore/src/main/scala/explore/config/ITCRequestsQueue.scala
```diff
@@ -0,0 +1,166 @@
...
+    E: Temporal[F],
+    L: Logger[F],
+    T: TransactionalClient[F, ITC]
+  ): F[Unit] =
+    Stream
+      .fromQueueNoneTerminated(f)
+      .meteredStartImmediately(500.milliseconds) // Limit how fasts requests are created
+      .evalMap(f => ITCRequestsQueue.doRequest(f))
+      .compile
+      .drain
```
rpiaggio, line 56: I think that limiting concurrency might be a more fitting solution than using a Stream. We could send N requests simultaneously initially, and make sure a new one isn't started until another is done.

Looking around on how to achieve this I found this which uses a semaphore: https://gist.github.com/gvolpe/44e2263f9068efe298a1f30390de6d22.

But we can try that in another PR.

rpiaggio (approved):
LGTM

PR #1310, opened 2021-11-26
Use the constraints set in the program

cquiroz:
This PR links the conditions on the UI to the ones used on ITC, changes on conditions will trigger a re calcualtion of ITC times

https://user-images.githubusercontent.com/3615303/143581071-0a017297-c2c8-4a9e-8463-dc4fdb2cec21.mp4

### DIFF
--- a/common/src/main/scala/explore/common/ObsQueries.scala
+++ b/common/src/main/scala/explore/common/ObsQueries.scala
@@ -57,18 +57,24 @@ object ObsQueries {
 
   case class ScienceData(
     requirements:  ScienceRequirementsData,
-    configuration: Option[ScienceConfigurationData]
+    configuration: Option[ScienceConfigurationData],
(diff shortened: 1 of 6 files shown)


common/src/main/scala/explore/common/ObsQueries.scala
```diff
@@ -56,18 +56,24 @@ object ObsQueries {
 
   case class ScienceData(
     requirements:  ScienceRequirementsData,
-    configuration: Option[ScienceConfigurationData]
+    configuration: Option[ScienceConfigurationData],
+    constraints:   ConstraintSet
```
rpiaggio, line 60: Why do we need to add the constraints to `ScienceData`? I can't find where it's used.

rpiaggio (approved):
LGTM

PR #1311, opened 2021-11-26
New crystal, hooks. ITC queries changes.

rpiaggio:
This PR includes several changes:

- The `View[A]` type alias as a `ViewF[DefaultS, A]` is now moved into `crystal`. This is the cause of so many files in the PR, since the `import` was changed.
- A couple of methods on `View` were moved into `crystal`, where they belong.
- Usage of new hooks, such as `useStateView`, `useSingleEffect` and `useSerialStateView`.
- The mechanism to avoid overwhelming the browser when we send a batch of ITC request was changed: instead of a stream, we now use a `parTraverse` with limited concurrency (through a semaphore).
- ITC queries are no longed triggered when scrolling on the config modes table (all rows are loaded when any input changes).
- ITC queries are canceled when a new batch of queries is started and when the component is unmounted (via `useSingleEffect` hook).


explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -92,60 +90,49 @@ object ConfigurationPanel {
   protected val component =
     ScalaFnComponent
       .withHooks[Props]
-      .useStateSnapshotWithReuse(none[ITCRequestsQueue[IO]])
-      .useEffectOnMountBy { (props, st) =>
-        // Create the requests queue and put it on the state
+      .useStateView[ScienceMode](ScienceMode.Spectroscopy)
```
cquiroz, line 93: Nice

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -351,18 +323,20 @@ object SpectroscopyModesTable {
         }
       )
       // itc results cache
-      .useState(
-        ItcResultsCache(Map.empty[ItcResultsCache.CacheKey, EitherNec[ItcQueryProblems, ItcResult]])
+      .useSerialStateView(
```
cquiroz, line 326: Is it serial, because the counter is now handled in the View, right?

cquiroz (approved):
This is great, thanks

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
rpiaggio: Yes

PR #1313, opened 2021-11-26
Fix charset warning

cquiroz:
A full build generates a set of extra @charset entries which are not needed. This PR fixes it:
See
https://github.com/vitejs/vite/issues/5519


rpiaggio (approved):
LGTM

PR #1322, opened 2021-11-29
Make the browserslist explicit

cquiroz:


package.json
```diff
@@ -72,5 +72,10 @@
     "fix-light": "sed '/^[\\.\\}\\@]/d;/^[[:blank:]]*\\..*/d;/\\/\\/.*/d' common/src/main/webapp/less/variables-light.less > vars.css;trap \"rm vars.css\" EXIT;npx stylelint --fix common/src/main/webapp/less",
     "build": "node --unhandled-rejections=strict build.js",
     "serve": "vite preview"
-  }
+  },
+  "browserslist": [
```
rpiaggio, line 76: When is this used? I'm suspecting stylelint?

rpiaggio (approved):
LGTM

package.json
cquiroz: Yes, but more important it is used to do auto prefixing, anyway I saw a warning that the list should rather be explicit

toddburnside: I had to do a search for what this was about. Came up with https://stackoverflow.com/questions/55510405/what-is-the-significance-of-browserslist-in-package-json-created-by-create-react

PR #1328, opened 2021-12-01
Pass the targets to be used on ITC calculations

cquiroz:
We need some parameters coming from the observation target in ITC. Previously it was hardcoded but now it will be taken from teh values entered on the UI

This PR highlighted that ITC expects one target but there maybe N defined on the asterism. In that case we'll calculate ITC for each science target and take the maxiumum time indicated for the set of targets

### DIFF
--- a/common/src/main/scala/explore/schemas/implicits.scala
+++ b/common/src/main/scala/explore/schemas/implicits.scala
@@ -157,10 +157,7 @@ object itcschema {
     implicit class ScienceDataOps(val s: ObsQueries.ScienceData) extends AnyVal {
       // From the list of targets selects the ones relevant for ITC
       def itcTargets: List[ITCTarget] = s.targets.scienceTargets
-        .collect { case (_, s: SiderealTarget) =>
-          s.tracking.radialVelocity
-        }
-        .collect { case Some(rv) =>
+        .collect { case (_, SiderealTarget(_, SiderealTracking(_, _, _, _, Some(rv), _), _, _)) =>
           // We can only process targets with a radial velocity
           ITCTarget(rv)
         }
--- a/common/src/main/scala/explore/schemas/implicits.scala
+++ b/common/src/main/scala/explore/schemas/implicits.scala
@@ -3,9 +3,12 @@
 
 package explore.schemas
 
+import cats.syntax.all._
 import clue.data.Input
 import clue.data.syntax._
 import explore.common.ITCQueriesGQL
+import explore.common.ObsQueries
+import explore.model.ITCTarget
 import explore.modes.GmosNorthSpectroscopyRow
 import lucuma.core.math._
 import lucuma.core.model._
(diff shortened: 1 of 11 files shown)


explore/src/main/scala/explore/config/ITCRequests.scala
```diff
@@ -69,64 +79,68 @@ object ITCRequests {
         InstrumentModes(m.toGmosNITCInput)
       }
       // Discard values in the cache
-      .filterNot { case m =>
-        cache.get.cache.contains((wavelength, signalToNoise, constraints, m))
+      .filterNot { case mode =>
+        val params = ITCRequestParams(wavelength, signalToNoise, constraints, targets, mode)
```
rpiaggio, line 83: We may want to map into `ITCRequestParams` here and make `itcRows` a `List[ITCRequestParams]`. Otherwise we are creating it twice (once here and once in line 92).

common/src/main/scala/explore/schemas/implicits.scala
```diff
@@ -149,5 +153,19 @@ object itcschema {
       def toITCInput: RadialVelocityInput =
         RadialVelocityInput(metersPerSecond = r.rv.value.assign)
     }
+
+    implicit class ScienceDataOps(val s: ObsQueries.ScienceData) extends AnyVal {
+      // From the list of targets selects the ones relevant for ITC
+      def itcTargets: List[ITCTarget] = s.targets.scienceTargets
+        .collect { case (_, s: SiderealTarget) =>
```
rpiaggio, line 160: IIRC, class matches are discouraged, being equivalent to `.isInstaceOf`. I think we can pattern match directly to `SiderealTarget(_, SiderealTracking(_, _, _, _, Some(rv), _), _, _)`.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -66,7 +68,9 @@ object SpectroscopyModesTable {
   type ColId = NonEmptyString
 
   implicit val reuseProps: Reusability[Props] =
-    Reusability.by(x => (x.scienceConfiguration, x.spectroscopyRequirements, x.constraints))
+    Reusability.by(x =>
+      (x.scienceConfiguration, x.spectroscopyRequirements, x.constraints, x.targets)
+    )
```
rpiaggio, line 73: Can we turn this into a `Reusability.derive` and use `Reusability.always` for `SpectroscopyModesMatrix`? Or receive a `Reusable[SpectroscopyModesMatrix]`.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -56,6 +57,7 @@ final case class SpectroscopyModesTable(
   scienceConfiguration:     View[Option[ScienceConfigurationData]],
   spectroscopyRequirements: SpectroscopyRequirementsData,
   constraints:              ConstraintSet,
+  targets:                  Option[List[ITCTarget]],
```
rpiaggio, line 60: What's the semantic difference between `None` and `Some(Nil)` ?

rpiaggio (approved):
LGTM. Made a couple of suggestions.

common/src/main/scala/explore/schemas/implicits.scala
cquiroz: Are you sure? I thought this was kosher

toddburnside (approved):
As usual, Raul beat me to it and did a great review. LGTM

common/src/main/scala/explore/schemas/implicits.scala
cquiroz: A benefit is to combine both `collect` in only one

PR #1332, opened 2021-12-02
Select the magnitude from the one closest to the requested wavelength

cquiroz:
Take the magnitudes from the model's target to use with ITC rather than the harcoded value
This PR also simplifies a bit how the configuratoin modes are handled

### DIFF
--- a/explore/src/main/scala/explore/itc/ITCResultsCache.scala
+++ b/explore/src/main/scala/explore/itc/ITCResultsCache.scala
@@ -10,26 +10,24 @@ import eu.timepit.refined.types.numeric.PosBigDecimal
 import explore.model.ConstraintSet
 import explore.model.ITCTarget
 import explore.modes._
-import explore.schemas.itcschema.implicits._
 import japgolly.scalajs.react._
 import lucuma.core.enum._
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/itc/ITCResultsCache.scala
```diff
@@ -57,7 +55,7 @@ final case class ItcResultsCache(
 object ItcResultsCache {
 
   def enabledRow(row: SpectroscopyModeRow): Boolean =
-    List(Instrument.GmosNorth, Instrument.GmosSouth).contains_(
+    List[Instrument](Instrument.GmosNorth).contains_(
```
toddburnside, line 58: Why only GmosNorth now?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/itc/ITCResultsCache.scala
cquiroz: Currently ITC only supports GN and it was being filtered twice, This one is not needed anymore but would need to be restored as we enable more instruments

PR #1342, opened 2021-12-08
Add support for Gmos South ITC

cquiroz:
Now it will do ITC requests for Gmos South modes and it also includes some improvements trying to display the visible rows first

### DIFF
--- a/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
+++ b/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
@@ -328,6 +328,13 @@ object SpectroscopyModesTable {
       .map(selected => rows.indexWhere(row => equalsConf(row, selected)))
       .filterNot(_ == -1)
 
+  protected def visibleRows(visibleRange: ListRange, rows: List[SpectroscopyModeRow]) = {
+    val s = visibleRange.startIndex.toInt
+    val e = visibleRange.endIndex.toInt
+
+    (for { i <- s to e } yield rows.lift(i)).collect { case Some(m) => m }.toList
+  }
+
   val component =
     ScalaFnComponent
       .withHooks[Props]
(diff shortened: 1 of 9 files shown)


toddburnside (approved):
LGTM

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -418,8 +424,16 @@ object SpectroscopyModesTable {
                                     itcResults.async,
                                     itcProgress.async
                       )
-                }.orEmpty
+                  }
+                  .getOrElse(IO.unit)
```
rpiaggio, line 428: Is the issue with multiple invocations solved this way? Does it have to do with using `.getOrElse(IO.unit)` instead of `.orEmpty`?

rpiaggio: Upon closer look, it probably has to do with removing `rows` from the dependencies.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -418,8 +424,16 @@ object SpectroscopyModesTable {
...
               )
+
+          val isCacheEmpty = itcResults.value.get.isEmpty
+
+          // Only send the requests if the cache is empty, send the visible rows first
+          submitRows(range.value.foldMap(visibleRows(_, sortedRows)) ++ sortedRows)
+            .whenA(isCacheEmpty)
```
rpiaggio, line 435: What happens if any of the ITC parameters change but the cache is not emptied?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: I think it is about `rows`

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: You're right, this is failing

PR #1345, opened 2021-12-10
Client logging

cquiroz:
It was requested to see more ITC logging. This PR adds a control to change the log level (Only in dev and staging)

image

The log level is stored in the browser local storage.

Given how  logging is setup it cannot be changed at runtime. Instead the value is changed and  we do a page reload

### DIFF
--- a/common/src/main/scala/explore/model/ExploreLocalPreferences.scala
+++ b/common/src/main/scala/explore/model/ExploreLocalPreferences.scala
@@ -13,6 +13,7 @@ import io.circe.Json
 import io.circe.parser.decode
 import io.circe.syntax._
 import log4cats.loglevel.LogLevelLogger
+import lucuma.core.util.Enumerated
 import monocle.Focus
 import org.scalajs.dom.window
 import typings.loglevel.mod.LogLevelDesc
@@ -21,6 +22,25 @@ import typings.loglevel.mod.LogLevelDesc
 case class ExploreLocalPreferences(level: LogLevelDesc)
 
 object ExploreLocalPreferences {
+  implicit val levelEnumerated: Enumerated[LogLevelDesc] =
+    new Enumerated[LogLevelDesc] {
+      val all = List(LogLevelDesc.TRACE,
+                     LogLevelDesc.DEBUG,
+                     LogLevelDesc.INFO,
+                     LogLevelDesc.WARN,
+                     LogLevelDesc.ERROR,
+                     LogLevelDesc.SILENT
+      )
+
+      def tag(l: LogLevelDesc): String =
+        if (l == LogLevelDesc.TRACE) "TRACE"
+        else if (l == LogLevelDesc.DEBUG) "DEBUG"
+        else if (l == LogLevelDesc.INFO) "INFO"
+        else if (l == LogLevelDesc.WARN) "WARN"
+        else if (l == LogLevelDesc.ERROR) "ERROR"
+        else "SILENT"
+    }
+
   val Default = ExploreLocalPreferences(LogLevelLogger.Level.INFO)
 
   val StorageKey = "ExplorePreferences"
@@ -28,7 +48,7 @@ object ExploreLocalPreferences {
 
   val level = Focus[ExploreLocalPreferences](_.level)
 
-  implicit val eqExploreLocalpreferences: Eq[ExploreLocalPreferences] = Eq.by(_.level.toString)
+  implicit val eqExploreLocalpreferences: Eq[ExploreLocalPreferences] = Eq.by(_.level)
 
   implicit class LevelOps(val l: LogLevelDesc) extends AnyVal {
     def value: String =
@@ -55,7 +75,7 @@ object ExploreLocalPreferences {
     )
   }
 
-  implicit val decodeFoo: Decoder[ExploreLocalPreferences] = new Decoder[ExploreLocalPreferences] {
+  implicit val decoder: Decoder[ExploreLocalPreferences] = new Decoder[ExploreLocalPreferences] {
     final def apply(c: HCursor): Decoder.Result[ExploreLocalPreferences] =
       for {
         l <- c.downField(LevelKey).as[Option[String]]
--- a/common/src/main/scala/explore/model/ExploreLocalPreferences.scala
+++ b/common/src/main/scala/explore/model/ExploreLocalPreferences.scala
@@ -1,0 +1,82 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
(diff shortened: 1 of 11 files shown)


common/src/main/scala/explore/model/ExploreLocalPreferences.scala
```diff
@@ -0,0 +1,86 @@
...
+    case _        => LogLevelDesc.INFO
+  }
+
+  implicit val encoder: Encoder[ExploreLocalPreferences] = new Encoder[ExploreLocalPreferences] {
+    final def apply(a: ExploreLocalPreferences): Json = Json.obj(
+      (LevelKey, Json.fromString(a.level.value))
+    )
+  }
+
+  implicit val decodeFoo: Decoder[ExploreLocalPreferences] = new Decoder[ExploreLocalPreferences] {
```
rpiaggio, line 58: Foo?

common/src/main/scala/explore/model/ExploreLocalPreferences.scala
```diff
@@ -0,0 +1,86 @@
...
+  def levelFromString(s: String): LogLevelDesc = s match {
+    case "TRACE"  => LogLevelDesc.TRACE
+    case "DEBUG"  => LogLevelDesc.DEBUG
+    case "WARN"   => LogLevelDesc.WARN
+    case "ERROR"  => LogLevelDesc.ERROR
+    case "SILENT" => LogLevelDesc.SILENT
+    case _        => LogLevelDesc.INFO
+  }
+
+  implicit val encoder: Encoder[ExploreLocalPreferences] = new Encoder[ExploreLocalPreferences] {
```
rpiaggio, line 52: We could abstract away `Encoder[LogLevelDesc]` and `Decoder[LogLevelDesc]`. I think then the `Encoder[ExploreLocalPreferences]` at least could be derived. Maybe the decoder too.

rpiaggio: Actually, maybe we can have an `Enumerated[LogLevelDesc]` and use all the functionality from there?Including getting an arbitrary.

common/src/test/scala/explore/model/arb/ArbExploreLocalPreferences.scala
```diff
@@ -0,0 +1,25 @@
...
+import org.scalacheck.Cogen
+import org.scalacheck.Cogen._
+import org.scalacheck.Gen
+import typings.loglevel.mod.LogLevelDesc
+
+trait ArbExploreLocalPreferences {
+
+  implicit val localPreferencesArb = Arbitrary[ExploreLocalPreferences] {
+    for {
+      level <- Gen.const(LogLevelDesc.INFO)
```
rpiaggio, line 17: Do we want to generate the other values? Actually if we make an `Enumerated` we can use its functionality.

explore/src/main/scala/explore/Explore.scala
```diff
@@ -121,7 +124,15 @@ object ExploreMain extends IOApp.Simple {
...
-      .flatMap { implicit dispatcher =>
+      .flatMap { d =>
+        for {
+          p <- ExploreLocalPreferences.loadPreferences[IO]
+          l <- setupLogger(p)
+        } yield (d, l, p)
+      }
+      .flatMap { param =>
+        implicit val (dispatcher, logger, _)                     = param
+        implicit val localPreferences                            = param._3
```
rpiaggio, line 135: Can `localPreferences` be obtained from the unapply in the previous line?

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -140,11 +144,23 @@ object TopBar {
...
                       ),
+                      DropdownItem()(
+                        Icons.BarCodeRead.fixedWidth(),
+                        "Log Level",
+                        DropdownMenu(
+                          DropdownItem(onClick = setLogLevel(LogLevelDesc.INFO))(
+                            Checkbox(label = "Info", checked = level != LogLevelDesc.DEBUG)
+                          ),
+                          DropdownItem(onClick = setLogLevel(LogLevelDesc.DEBUG))(
+                            Checkbox(label = "Debug", checked = level == LogLevelLogger.Level.DEBUG)
```
rpiaggio, line 155: Should this be `checked level == LogLevelDesc.DEBUG`?

rpiaggio: Maybe we can use `===`? (Sorry to insist :D but the `Eq` would come for free with `Enumerated` )

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -35,46 +36,50 @@ import react.semanticui.modules.dropdown.DropdownItem
...
-      .builder[TopBar]
-      .initialState(State(false, currentTheme))
-      .render { $ =>
-        val p            = $.props
-        val currentTheme = $.state.theme
+  def flipTheme(theme: Theme): Theme =
+    if (theme === Theme.Dark) Theme.Light else Theme.Dark
 
+  private val component =
+    ScalaFnComponent
```
rpiaggio, line 62: Nice

explore/src/main/scala/explore/Explore.scala
cquiroz: It can but I don't need it to be implicit, though now that I look at it again it is `implicit` anyway

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -35,46 +36,50 @@ import react.semanticui.modules.dropdown.DropdownItem
...
+            appCtx.sso.logout >> props.onLogout
+
+          val level = props.preferences.level
+
+          def setLogLevel(l: LogLevelDesc): Callback =
+            ExploreLocalPreferences
+              .storePreferences[IO](
+                props.preferences.copy(level = l)
+              )
+              .runAsync *> Callback(window.location.reload(false))
```
rpiaggio, line 82: I would suggest doing `(....storePreferences[IO](...) *> IO(window.location.reload(false))).runAsync`.

That way we make sure `storePreferences` completes before reloading.

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -108,20 +113,19 @@ object TopBar {
                         Reuse.always(
                           DropdownItem(text = "About Explore", icon = Icons.Info.fixedWidth())
                         ),
-                        Reuse.always(
+                        Reuse.never(
```
rpiaggio, line 116: Nice catch

rpiaggio (approved):
LGTM. It's awesome to have this functionality, especially in staging. Thanks!

explore/src/main/scala/explore/TopBar.scala
cquiroz: Good suggestion. I'll add the `Enumerated`

PR #1346, opened 2021-12-10
Display each catalog query result independently.

rpiaggio:
For Simbad, we make multiple wildcard queries for the same name. Previously, we waited for all of these queries to finish before displaying the results.

With this PR each query adds its results to the displayed ones as they complete. This results in a more snappy experience, especially in the presence of wildcard queries that would take too much time (like `Gaia DR2 1896909625251586176`). Also, a non-wildcard query is added to the batch.

Besides all this, I noticed a significant performance degradation in development WRT to recent versions. This seems to be caused by `cats-effect` tracing (see https://github.com/typelevel/cats-effect/issues/2634). I wasn't able to configure `cats-effect` with tracing disabled. (See also https://github.com/typelevel/cats-effect/blob/series/3.x/docs/tracing.md#configuration).

I'm therefore downgrading `cats-effect` (and `fs2` which brings it in transitively) until we find a solution. I'm leaving here the things I tried in `vite.conf.js` and `main.jsx` so we can work on top of that.

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -7,7 +7,7 @@ object Settings {
 
   object LibraryVersions {
     val cats              = "2.7.0"
-    val catsEffect        = "3.3.0"
+    val catsEffect        = "3.2.9"
     val catsRetry         = "3.1.0"
     val circe             = "0.14.1"
     val circeGolden       = "0.3.0"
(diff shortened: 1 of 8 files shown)


project/Settings.scala
```diff
@@ -7,15 +7,15 @@ object Settings {
 
   object LibraryVersions {
     val cats              = "2.7.0"
-    val catsEffect        = "3.3.0"
+    val catsEffect        = "3.2.9"
```
cquiroz, line 10: 👍

cquiroz (approved):
LGTM

toddburnside (approved):
LGTM

PR #1353, opened 2021-12-14
Format the times according to the result

cquiroz:

### DIFF
--- a/explore/src/main/scala/explore/itc/ItcColumn.scala
+++ b/explore/src/main/scala/explore/itc/ItcColumn.scala
@@ -31,7 +31,9 @@ sealed trait ItcResult extends Product with Serializable
 object ItcResult {
   case object SourceTooBright                                     extends ItcResult
   case object Pending                                             extends ItcResult
-  case class Result(exposureTime: FiniteDuration, exposures: Int) extends ItcResult
+  case class Result(exposureTime: FiniteDuration, exposures: Int) extends ItcResult {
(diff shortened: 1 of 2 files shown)


cquiroz:
image

explore/src/main/scala/explore/itc/ItcColumn.scala
```diff
@@ -31,7 +31,9 @@ sealed trait ItcResult extends Product with Serializable
 object ItcResult {
   case object SourceTooBright                                     extends ItcResult
   case object Pending                                             extends ItcResult
-  case class Result(exposureTime: FiniteDuration, exposures: Int) extends ItcResult
+  case class Result(exposureTime: FiniteDuration, exposures: Int) extends ItcResult {
```
rpiaggio, line 34: When modeling a general representation of a duration of time, we should probably use `java.time.Duration`. It provides arithmetic features and a standarized `toString`, which can be made human presentable with a one-liner (https://stackoverflow.com/a/40487511).

`scala.concurrent.Duration` is meant to be used with `scala.concurrent`.

rpiaggio (approved):
LGTM. We can switch to `java.time` at another instance.

explore/src/main/scala/explore/itc/ItcColumn.scala
cquiroz: I thought about using `Duration` but I feared increasing the js size a lot. But I could try

PR #1357, opened 2021-12-14
Update cats-effect and set process.env variable

cquiroz:

### DIFF
--- a/explore/src/main/webapp/main.jsx
+++ b/explore/src/main/webapp/main.jsx
@@ -12,7 +12,13 @@ import { Explore } from "@sjs/main.js";
 
 // Setting this here shouldn't be necessary if we get `vite-plugin-environment` to work.
 // but for now we can survive setting this only on dev
-if (import.meta.env.DEV && process && process.env) {
+if (!process) {
+  process = {
+    env: {}
+  };
+}
+
+if (import.meta.env.DEV) {
   process.env = {CATS_EFFECT_TRACING_MODE: "none"};
 }
 
--- a/explore/src/main/webapp/main.jsx
+++ b/explore/src/main/webapp/main.jsx
@@ -11,9 +11,11 @@ import "react-circular-progressbar/dist/styles.css";
 import { Explore } from "@sjs/main.js";
 
 // Setting this here shouldn't be necessary if we get `vite-plugin-environment` to work.
-// See vite.config.js. Although configured there, the value is not getting to runtime for some reason.
-// Anyway, for the moment this is not having the desired effect either, and we are downgrading cats-effect.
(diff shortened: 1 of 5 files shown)


vite.config.js
```diff
@@ -53,11 +46,9 @@ module.exports = ({ command, mode }) => {
   const publicDir = mode == "production" ? publicDirProd : publicDirDev;
   return {
     // TODO Remove this if we get EnvironmentPlugin to work.
```
toddburnside, line 48: Should the comment be removed, too?

explore/src/main/webapp/main.jsx
```diff
@@ -11,9 +11,11 @@ import "react-circular-progressbar/dist/styles.css";
...
 
 // Setting this here shouldn't be necessary if we get `vite-plugin-environment` to work.
-// See vite.config.js. Although configured there, the value is not getting to runtime for some reason.
-// Anyway, for the moment this is not having the desired effect either, and we are downgrading cats-effect.
-// process.env.CATS_EFFECT_TRACING_MODE="none";
+// but for now we can survive setting this only on dev
+if (import.meta.env.DEV && process && process.env) {
+  process.env = {CATS_EFFECT_TRACING_MODE: "none"};
+}
+
```
toddburnside, line 18: Why does this work in dev but not the other environments? Or is that one of the `mysteries`?

toddburnside (approved):
LGTM

explore/src/main/webapp/main.jsx
cquiroz: I'm not sure, `process env`  should not be present on the browser  being a `node` thing but I guess the dev server injects while the build mode does not

toddburnside: Fun!

rpiaggio: Do we want to define `process` and `process.env` if they are not present? Do we want to preserve other potential properties defined in `process.env` ?

vite.config.js
```diff
@@ -53,11 +46,9 @@ module.exports = ({ command, mode }) => {
...
-    define: {
-      'process.env.CATS_EFFECT_TRACING_MODE': "none"
-    },
     root: "explore/src/main/webapp",
     publicDir: publicDir,
+    envPrefix: ["VITE_", "CATS_EFFECT_"],
```
rpiaggio, line 51: This is a good idea, it should carry over environment variables into the app. I think it's probably redundant with lines 15-17 (in case vite honors it).

rpiaggio (approved):
LGTM. Thank you for taking care of this!

PR #1358, opened 2021-12-15
ITC error

cquiroz:
It is not always clear that we are missing some data to calculate ITC. This PR adds some indications to the forms to make it clearer:

image

image


toddburnside (approved):
Looks great!

PR #1360, opened 2021-12-18
Fix a bug with the urls for the help system

cquiroz:
Also updates `github-markdown-css`

### DIFF
--- a/explore/src/main/scala/explore/HelpBody.scala
+++ b/explore/src/main/scala/explore/HelpBody.scala
@@ -36,7 +36,8 @@ final case class HelpBody(base: HelpContext, helpId: Help.Id)(implicit val ctx: 
     path.segments.init.foldLeft(base.rawUrl / base.user.value / base.project.value / "main")(
       (uri, segment) => uri / segment.encoded
     )
-  private val url            = (rootUrl / "main").withPath(path)
+  private val mainUrl        = rootUrl / "main"
+  private val url            = (rootUrl / "main").withPath(mainUrl.path.concat(path))
   private val rootEditUrl    = base.editUrl / base.user.value / base.project.value
   private val newPage        = (rootEditUrl / "new" / "main")
     .withQueryParam("filename", path.segments.mkString("/"))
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/HelpBody.scala
```diff
@@ -75,7 +76,7 @@ object HelpBody {
         load(props.url).flatMap(v => state.set(Pot.fromTry(v)).to[IO])
       }
       .render { (props, state) =>
-        val imageConv = (s: Uri) => props.baseUrl.withPath(s.path)
+        val imageConv = (s: Uri) => props.baseUrl.withPath(props.baseUrl.path.concat(s.path))
```
toddburnside, line 79: Seems weird that you need to do it this way. But, if it works, it works. :)

toddburnside (approved):
LGTM

explore/src/main/scala/explore/HelpBody.scala
```diff
@@ -36,7 +36,8 @@ final case class HelpBody(base: HelpContext, helpId: Help.Id)(implicit val ctx:
     path.segments.init.foldLeft(base.rawUrl / base.user.value / base.project.value / "main")(
       (uri, segment) => uri / segment.encoded
     )
-  private val url            = (rootUrl / "main").withPath(path)
+  private val mainUrl        = rootUrl / "main"
+  private val url            = (rootUrl / "main").withPath(mainUrl.path.concat(path))
```
rpiaggio, line 40: It took me a bit to understand what's going on here (and in line 79). Maybe we can extract a tiny method , say `concatToUriPath(uri: Uri, path: Uri.Path): Uri = uri.withPath(uri.path.concat(path))` ?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/HelpBody.scala
cquiroz: yeah, it is a bit weird

PR #1391, opened 2021-12-30
Targets unshared, again

toddburnside:
Initial changes for the unshared targets. Much remains to be done, for example:

- Fix and re-add the target summary table.
- For new targets, create the id locally so we don't have to wait for the round trip.
- Show the target list on the targets tab, and allow dragging to asterisms - and potentially editing the targets directly.

I will do some more renaming of classes/values in later PRs. There seemed to be plenty of changes for one PR...

There are still some issues with the updating the tree locally after adding/removing targets from asterisms. I will continue looking into that in the morning. Everything gets resolved after the server roundtrip, so we can merge before that is fixed if we'd like.


common-graphql/src/main/scala/explore/common/TargetListGroupQueriesGQL.scala
```diff
@@ -20,52 +21,58 @@ object TargetListGroupQueriesGQL {
   trait TargetListGroupObsQuery extends GraphQLOperation[ObservationDB] {
```
rpiaggio, line 21: Should we rename this to `AsterismGroupObsQuery` ?

common-graphql/src/main/scala/explore/common/TargetListGroupQueriesGQL.scala
```diff
@@ -113,9 +124,20 @@ object TargetListGroupQueriesGQL {
   @GraphQL
   trait ReplaceScienceTargetListMutation extends GraphQLOperation[ObservationDB] {
```
rpiaggio, line 125: `UpdateTargetEnvironmentMutation`?

common/src/main/scala/explore/common/TargetListGroupQueries.scala
```diff
@@ -14,64 +14,52 @@ import crystal.react.reuse._
...
 import lucuma.ui.reusability._
 import monocle.Focus
 import monocle.Getter
 
 import scala.collection.immutable.SortedMap
+import scala.collection.immutable.SortedSet
 
 import TargetListGroupQueriesGQL._
 
 object TargetListGroupQueries {
```
rpiaggio, line 37: More pontential `TargetList` -> `Asterism` renames.

common/src/main/scala/explore/common/TargetQueries.scala
```diff
@@ -33,61 +37,72 @@ import TargetQueriesGQL._
...
         )
 
     def apply[A](
       modelLens: Lens[SiderealTarget, A],
-      remoteSet: A => EditSiderealInput => EditSiderealInput
+      remoteSet: A => EditTargetInput => EditTargetInput
     ): View[A] = apply(modelLens.get, modelLens.modify, remoteSet)
   }
 
+  def toTargetEndo(setSidereal: Endo[EditSiderealInput]): Endo[EditTargetInput] = eti =>
```
rpiaggio, line 67: Nice 👍

common/src/main/scala/explore/common/TargetListGroupQueries.scala
toddburnside: Yeah. As I noted in the description, I'm going to do more renaming. But, it seemed like the diff was going to be hard enough to follow - file renaming doesn't get tracked well in git. 
 
I will make another commit (in this PR) with a bunch of renaming.

explore/src/main/scala/explore/targeteditor/TargetEnvEditor.scala
```diff
@@ -27,17 +27,18 @@ import lucuma.core.model.SiderealTarget
...
 import lucuma.core.model.User
 import lucuma.ui.reusability._
-import monocle.function.Index._
+import monocle.Optional
 import react.common.ReactFnProps
 import react.semanticui.elements.button._
 import react.semanticui.shorthand._
 import react.semanticui.sizes._
 
 final case class TargetEnvEditor(
```
rpiaggio, line 36: Maybe we should rename `TargetEnv` to something else? Or do we still handle that concept?

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
```diff
@@ -44,7 +44,7 @@ import scala.concurrent.duration._
 
 final case class TargetSelectionPopup(
   trigger:          Reuse[Button],
-  onSelected:       Target ==> Callback
+  onSelected:       (Option[Target.Id], Target) ==> Callback
```
rpiaggio, line 47: Nice

model/shared/src/main/scala/explore/model/util/NonEmptySetWrapper.scala
```diff
@@ -74,6 +77,9 @@ final class NonEmptySetWrapper[IdSet, Id](self: IdSet, iso: Iso[IdSet, NonEmptyS
   @inline
   def intersects(other: IdSet): Boolean = intersect(other).nonEmpty
 
+  @inline
+  def firstAndOnly: Option[Id] = if (selfSet.length === 1) selfSet.head.some else none
```
rpiaggio, line 81: `single` ?

model-testkit/shared/src/main/scala/explore/model/arb/ArbObsSummary.scala
```diff
@@ -32,14 +31,13 @@ trait ArbObsSummary {
...
       duration     <- arbitrary[Duration]
-      targetEnvId  <- arbitrary[TargetEnvironment.Id]
       targets      <- arbitrary[Set[Target.Id]]
     } yield ObsSummaryWithConstraints(id,
                                       constraints,
                                       status,
                                       activeStatus,
                                       duration,
-                                      targetEnvId,
+                                      // targetEnvId,
```
rpiaggio, line 40: Commented code

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
rpiaggio: Should we have case class for this? Seems to be used in a few places.

rpiaggio (approved):
LGTM

Looks great, thank you for doing this! 

I like the feeling of having a lighter mental load.

common/src/main/scala/explore/common/TargetListGroupQueries.scala
rpiaggio: :+1: I missed that line in the description

explore/src/main/scala/explore/targeteditor/TargetEnvEditor.scala
toddburnside: I think it should probably be the AsterismEditor.

model/shared/src/main/scala/explore/model/util/NonEmptySetWrapper.scala
toddburnside: Sounds good. I couldn't think of a good name at the moment I created it....

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
toddburnside: Yeah, might be more clear. `TargetWithId` is actually used more places and is just an alias for a similar tuple. I've been wondering if that, along with `SiderealTargetWithId` and `NonsiderealTargetWithId`, should be bumped up to a case class.

PR #1393, opened 2021-12-30
Some lib upgrades and rewrite of LogoutTracker

cquiroz:

### DIFF
--- a/common/src/main/scala/explore/components/state/LogoutTracker.scala
+++ b/common/src/main/scala/explore/components/state/LogoutTracker.scala
@@ -15,59 +15,48 @@ import explore.utils.ExploreEvent
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import lucuma.broadcastchannel._
-import monocle.Focus
-import react.common.ReactProps
-// import japgolly.scalajs.react.callback.CallbackCats._
+import react.common.ReactFnProps
 
 final case class LogoutTracker(
   setVault:   Option[UserVault] ==> Callback,
   setMessage: NonEmptyString ==> Callback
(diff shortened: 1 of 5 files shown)


common/src/main/scala/explore/components/state/LogoutTracker.scala
```diff
@@ -15,59 +15,48 @@ import explore.utils.ExploreEvent
...
-      }
-      .componentWillUnmount(
-        _.state.bc.map(bc => IO(bc.close()).attempt.void).orEmpty
+  protected implicit val stateReuse: Reusability[BroadcastChannel[ExploreEvent]] =
+    Reusability.always
+
+  val component = ScalaFnComponent
+    .withHooks[Props]
+    // Create a nonce
+    .useState(System.currentTimeMillis)
```
rpiaggio, line 38: Can we use `useMemo` here since it never changes afterwards?

common/src/main/scala/explore/components/state/LogoutTracker.scala
```diff
@@ -15,59 +15,48 @@ import explore.utils.ExploreEvent
...
+          case ExploreEvent.Logout.event =>
+            (props.setVault(none) >> props.setMessage(
+              "You logged out in another instance"
+            )).to[IO].whenA(x.value =!= nonce.value.toString)
+          case _                         => IO.unit
+        })
+
+      state
+        .setState(bc.some)
+        .flatMap(_ => CallbackTo(Callback(bc.close()).attempt))
```
rpiaggio, line 54: `>>` or `>*` instead of `.flatMap(_ =>` ?

rpiaggio (approved):
LGTM

PR #1398, opened 2021-12-31
Rewrite Tile as Fn compoanent

cquiroz:
Another rewrite, Took me a bit to understand how to properly use a Ref in a hook

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -13,7 +13,6 @@ import explore.model.Constants
 import explore.model.enum.TileSizeState
 import japgolly.scalajs.react.Key
 import japgolly.scalajs.react._
-import japgolly.scalajs.react.util.JsUtil
 import japgolly.scalajs.react.vdom.html_<^._
 import japgolly.scalajs.react.{ facade => Raw }
 import lucuma.ui.reusability._
@@ -38,13 +37,14 @@ final case class Tile(
   key:               js.UndefOr[Key] = js.undefined,
   controllerClass:   Option[Css] = None // applied to wrapping div when in a TileController.
 )(val render:        Tile.RenderInTitle ==> VdomNode)
-    extends ReactProps[Tile](Tile.component) {
+    extends ReactFnProps[Tile](Tile.component) {
   def showMaximize: Boolean                                                                =
     state === TileSizeState.Minimized || (canMaximize && state === TileSizeState.Normal)
   def showMinimize: Boolean                                                                =
     state === TileSizeState.Maximized || (canMinimize && state === TileSizeState.Normal)
   def withState(state: TileSizeState, sizeStateCallback: TileSizeState ==> Callback): Tile =
     copy(state = state, sizeStateCallback = sizeStateCallback)(render)
+  protected val infoRef                                                                    = Ref.toVdom[html.Element]
 }
 
 object Tile {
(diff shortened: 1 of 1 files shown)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -38,13 +37,14 @@ final case class Tile(
...
 )(val render:        Tile.RenderInTitle ==> VdomNode)
-    extends ReactProps[Tile](Tile.component) {
+    extends ReactFnProps[Tile](Tile.component) {
   def showMaximize: Boolean                                                                =
     state === TileSizeState.Minimized || (canMaximize && state === TileSizeState.Normal)
   def showMinimize: Boolean                                                                =
     state === TileSizeState.Maximized || (canMinimize && state === TileSizeState.Normal)
   def withState(state: TileSizeState, sizeStateCallback: TileSizeState ==> Callback): Tile =
     copy(state = state, sizeStateCallback = sizeStateCallback)(render)
+  protected val infoRef                                                                    = Ref.toVdom[html.Element]
```
rpiaggio, line 47: Is this used?

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -71,89 +67,72 @@ object Tile {
...
-      <.div(ExploreStyles.Tile |+| ExploreStyles.FadeIn, p.key.whenDefined(^.key := _))(
-        <.div(
-          ExploreStyles.TileTitle,
-          p.back.map(b => <.div(ExploreStyles.TileButton, b)),
-          Menu(
+  val component =
+    ScalaFnComponent
+      .withHooks[Props]
+      // info ref
+      .useRefToVdom[html.Element]
```
rpiaggio, line 74: Cool, I didn't know this existed.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/Tile.scala
cquiroz: Nope, I'll remove it

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -71,89 +67,72 @@ object Tile {
...
+      .render { (p, infoRef) =>
```
cquiroz, line 75: I tried to do `renderWithReuse` and I'm facing troubles to define `Reusability` for the ref

rpiaggio: A ref is a mutable holder (like cats-effect `Ref`). The ref is always the same, even though its `current` value may change. So I think it's safe to use `Reusability.always`.

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -53,13 +56,11 @@ object Tile {
...
-  protected case class State(portalNode: Option[Raw.ReactDOM.Container] = None)
-
   implicit val propsReuse: Reusability[Tile] = Reusability.derive && Reusability.by(_.render)
 
-  implicit val rawContainerReuse: Reusability[Raw.ReactDOM.Container] = Reusability.always
-
-  implicit val stateReuse: Reusability[State] = Reusability.derive
+  implicit val rawContainerReuse: Reusability[html.Element]                  = Reusability.always
+  implicit val rawDomContainerReuse: Reusability[Raw.ReactDOM.Container]     = Reusability.always
+  implicit def ref[F[_]]: Reusability[Ref.FullF[F, Node, Node, HTMLElement]] = Reusability.always
```
cquiroz, line 63: Kind of ugly

PR #1404, opened 2022-01-04
Create target id locally and other improvements

toddburnside:
Improves the responsiveness of the api by creating the target id locally and a few other changes. Also fixes a bug in the mod function for the asterism view.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
+++ b/explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
@@ -113,7 +113,7 @@ object AsterismEditor {
                 )
               ),
               onSelected = Reuse
-                .by((props.obsIds, props.asterism))(_ match {
+                .by((props.obsIds, props.asterism, selectedTargetId))(_ match {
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -106,14 +113,9 @@ object AsterismEditor {
                 )
               ),
               onSelected = Reuse
-                .always(_ match {
+                .by((props.obsIds, props.asterism))(_ match {
```
rpiaggio, line 116: Should `selectedTargetId` be part of the reusability conditions?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
toddburnside: Yeah. It probably should. It wasn't being used for reuse before, and I think it works because `selectedTargetId` is only set, not modified. But it is probably a recipe for weird behavior in the future. I'll add it to the reusability.

PR #1412, opened 2022-01-05
Port TargetTabContent

cquiroz:
I started this before the target model change and maybe stepping into that work. We could merge this later if you prefer

### DIFF
--- a/common/src/main/scala/explore/syntax/ui/package.scala
+++ b/common/src/main/scala/explore/syntax/ui/package.scala
@@ -1,3 +1,6 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
 package explore.syntax
 
 import org.scalajs.dom.Window
--- a/common/src/main/scala/explore/syntax/ui/package.scala
+++ b/common/src/main/scala/explore/syntax/ui/package.scala
@@ -1,0 +1,11 @@
+package explore.syntax
+
+import org.scalajs.dom.Window
+import explore.model.Constants
+
+package object ui {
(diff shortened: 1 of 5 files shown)


common/src/main/scala/explore/syntax/ui/package.scala
```diff
@@ -0,0 +1,11 @@
+package explore.syntax
+
+import org.scalajs.dom.Window
+import explore.model.Constants
+
+package object ui {
+  implicit class WindowOps(val self: Window) extends AnyVal {
```
rpiaggio, line 7: Nice

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -329,26 +306,35 @@ object TargetTabContents {
...
-      .initialState(
-        State(TwoPanelState.initial[ObsIdSet](SelectedPanel.Uninitialized),
-              TargetVisualOptions.Default
+    ScalaFnComponent
+      .withHooks[Props]
+      .useStateView(TwoPanelState.initial[ObsIdSet](SelectedPanel.Uninitialized))
+      .useStateView(TargetVisualOptions.Default)
+      .useEffectOnMountBy { (props, panels, _) =>
+        implicit val ctx = props.ctx
+        Callback.log("On mount") *>
```
rpiaggio, line 316: A bit of nitpicking: since we already have an implicit `AppContextIO` in context, this could be for example `Logger[IO].debug` or `.trace`, so that we can control the log level.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/tabs/TargetTabContents.scala
cquiroz: I'll remove this, it is a left over while I was developing

PR #1422, opened 2022-01-08
Use resize-detector as a hook

cquiroz:
I found out that `resizeDetector` can be used as a hook. This PR adds a facade and ports `TargetTabContents` to use it. We may want to port the other components and delete the component based facade

It also fixes a bug on setting the width of the resizable part of the UI

### DIFF
--- a/common/src/main/scala/react/resizeDetector/hooks.scala
+++ b/common/src/main/scala/react/resizeDetector/hooks.scala
@@ -1,0 +1,150 @@
+package react.resizeDetector
+
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.{ facade => Raw }
+import org.scalajs.dom.html
+import react.common._
+import react.resizeDetector.ResizeDetector._
+
+import scala.annotation.nowarn
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+import scala.scalajs.js.|
+
+@js.native
+protected trait ReactResizeDetectorDimensions extends js.Object {
+  val height: js.UndefOr[Raw.JsNumber]
+  val width: js.UndefOr[Raw.JsNumber]
+}
+
+@js.native
+trait UseResizeDetectorReturnJS extends ReactResizeDetectorDimensions {
+  val ref: Raw.React.RefFn[html.Element]
+}
+
+sealed trait UseResizeDetectorReturn {
+  val height: Option[Int]
+  val width: Option[Int]
+  val ref: Ref.Simple[html.Element]
+}
+
+object UseResizeDetectorReturn {
+  def fromJS(r: UseResizeDetectorReturnJS): UseResizeDetectorReturn = new UseResizeDetectorReturn {
+    val height = r.height.toOption.map(_.toInt)
+    val width  = r.width.toOption.map(_.toInt)
+    val ref    =
+      Ref.fromJs(r.ref.asInstanceOf[facade.React.RefHandle[html.Element | Null]])
+  }
+
+  implicit val reusabilityUseResizeDetectorReturn: Reusability[UseResizeDetectorReturn] =
+    Reusability.by(x => (x.height, x.width))
+}
+
+@js.native
+protected trait UseResizeDetectorProps extends Props {
+  var targetRef: Raw.React.RefFn[html.Element]
+}
+
+object UseResizeDetectorProps {
+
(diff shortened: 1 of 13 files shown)


toddburnside (approved):
I'm not the hooks expert, but it LGTM. 😄

common/src/main/scala/react/resizeDetector/hooks.scala
```diff
@@ -0,0 +1,153 @@
...
+  def fromJS(r: UseResizeDetectorReturnJS): UseResizeDetectorReturn = new UseResizeDetectorReturn {
+    val height = r.height.toOption.map(_.toInt)
+    val width  = r.width.toOption.map(_.toInt)
+    val ref    =
+      Ref.fromJs(r.ref.asInstanceOf[facade.React.RefHandle[html.Element | Null]])
+  }
+
+  implicit val reusabilityUseResizeDetectorReturn: Reusability[UseResizeDetectorReturn] =
+    Reusability.by(x => (x.height, x.width))
+}
```
rpiaggio, line 44: Can we extract the traits defined in `ResizeDetector` and reuse them in both places?

common/src/main/scala/react/resizeDetector/hooks.scala
```diff
@@ -0,0 +1,153 @@
...
+    p
+  }
+}
+
+object HooksApiExt {
+
+  object mod {
+    @JSImport("react-resize-detector", JSImport.Namespace)
+    @js.native
+    @nowarn
```
rpiaggio, line 89: Why is `@nowarn` needed?

rpiaggio: If `base` is not to be referenced from outside maybe we can make it `private`? And define it directly as `js.Dynamic`.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -319,20 +320,21 @@ object TargetTabContents {
                                 Constants.InitialTreeWidth.toInt
           )
           .attempt
-          .map {
+          .flatMap {
             case Right(w) =>
```
rpiaggio, line 324: `_.toOption.foldMap` ?

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -319,20 +320,21 @@ object TargetTabContents {
...
+                .zoom(TwoPanelState.treeWidth[ObsIdSet])
+                .set(w)
+                .to[IO]
+            case Left(_)  => IO.unit
           }
-          .runAsyncAndForget
+          .runAsync
       }
-      .renderWithReuse { (props, tps, opts) =>
+      .useResizeDetector(UseResizeDetectorProps())
```
rpiaggio, line 333: Can we call `.useResizeDetector()` ?

rpiaggio (approved):
LGTM

common/src/main/scala/react/resizeDetector/hooks.scala
cquiroz: Somehow it detects that is not used. If I make it private it should do fine I thin

common/src/main/scala/react/resizeDetector/hooks.scala
cquiroz: They are not exactly the same. Instead of `ref` it has a `targetRef`. I was hoping to delete the component based one later

explore/src/main/scala/explore/tabs/TargetTabContents.scala
cquiroz: I guess we should be able to add a default. let me check it

common/src/main/scala/react/resizeDetector/hooks.scala
rpiaggio: I missed that. We can still reuse `DimensionsJS`. We can still remove the class component version later. `Props` is being reused.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
rpiaggio: Ah ok, I saw a `()` version defined, but it's only at the low level.

PR #1430, opened 2022-01-12
Add target library to target tab

toddburnside:
Adds the target library to the target tab and allows drag-n-drop of targets onto asterisms. This feature is a little limited because it only allows dropping onto asterisms, not individual observations. 

I'm pretty happy with the behavior of the expanded ids during drag and drop, but I don't change the selected (edited) item at all. At some point, we'll probably need to revisit that for both this and constraint set drag and drop. That will cause some additional work in the undo/redo `actions`.

Undo/redo works pretty well, but intermixing drag and drop with modifying target lists in the editor can sometimes cause odd behavior.

https://user-images.githubusercontent.com/6035943/149044333-98480715-9b58-4d80-8671-3ca1430c89b7.mov


explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
```diff
@@ -77,28 +81,25 @@ object AsterismGroupObsList {
...
-     * NOT in the selection - in which case we just drag the individual item.
+     * When we're dragging an observation, we can either be dragging a single observation or a group
+     * of them. If we have a selection, and the drag id is part of the selection, we drag all the
+     * items in the selection. However, the user may have something selected, but be dragging
+     * something that is NOT in the selection - in which case we just drag the individual item.
      */
-    def getDraggedIds(dragId: String, props: Props): Option[ObsIdSet] =
-      parseDragId(dragId).map { obsId =>
+    def getDraggedIds(dragId: String, props: Props): Option[Either[Target.Id, ObsIdSet]] =
+      parseDragId(dragId).map(_.map { obsId =>
```
cquiroz, line 97: Would it make sense to use `Nested` here?

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
```diff
@@ -292,10 +327,37 @@ object AsterismGroupObsList {
               "Asterism Summary"
             )
           ),
-          <.div(ExploreStyles.ObsTree)(
-            <.div(ExploreStyles.ObsScrollTree)(
-              asterismGroups.toTagMod(renderGroup)
-            )
+          ReflexContainer()(
```
cquiroz, line 330: What is `Reflex`?

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
```diff
@@ -292,10 +327,37 @@ object AsterismGroupObsList {
...
+              ReflexElement(minSize = 36, clazz = ExploreStyles.ObsTreeSection)(
+                Header(block = true, clazz = ExploreStyles.ObsTreeHeader)("Asterisms"),
+                <.div(ExploreStyles.ObsTree)(
+                  <.div(ExploreStyles.ObsScrollTree)(
+                    asterismGroups.toTagMod(renderAsterismGroup)
+                  )
+                )
+              ),
+              ReflexSplitter(propagate = true),
+              ReflexElement(minSize = 36, clazz = ExploreStyles.ObsTreeSection, withHandle = true)(
```
cquiroz, line 341: Is `minSize` something that can be in css or is it a param used by the component

cquiroz (approved):
Looks nice, LGTM

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
toddburnside: It provides the resizable containers.

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
toddburnside: Hmmm. Perhaps. I don't have a strong feeling for how Nested works, but I'll look into it.

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
toddburnside: Heh. I don't know. I just resurrected it from the old targets tab, but I'll take a look.

toddburnside: It is a parameter to the JS component.

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
toddburnside: I don't think `Nested` is a good abstraction for here. Nested with an Either provides an easy way to map over the right side of the `Option[Either[]]`, at the expense of making the left more difficult. In this case, the `Either` is not being used in a `right biased` way. The fact that the observations ids are on the `Right` and I can use `map` to modify them is just a coincidence. Which makes me wonder if `Either` is even really the correct abstraction since it is now right-biased. A scala 3 union type would be better, I think. I'll change this to a match to better reflect the intentions.

rpiaggio: Maybe we can use a little ADT?

rpiaggio (approved):
I must admit I'm a bit lost in the logic but LGTM. Thanks!

toddburnside:
> I must admit I'm a bit lost in the logic but LGTM. Thanks!

In the undo/redo logic? Yeah, that was the hardest part of this by far. It took me embarrassingly long to get working...

PR #1433, opened 2022-01-12
Port ObsTabContent

cquiroz:
Port the class to a functional component and to use the `useResizeDetector` hook

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -77,7 +77,7 @@ object Tile {
       .withHooks[Props]
       // info ref
       .useRefToVdom[html.Element]
-      .renderWithReuse { (p, infoRef) =>
+      .renderWithReuse { (p, contentRef) =>
         val maximizeButton =
           Button(
             as = <.a,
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -192,315 +190,293 @@ object ObsTabContents {
...
-            .flatMap(v => $.modStateIn[IO](State.fovAngle.replace(v)))
-            .runAsyncAndForget
-        }.getOrEmpty
+            )
+            .toList
+        )
       }
+    }(obsView)
+
+  // TODO Use this method
```
toddburnside, line 230: Why aren't we using this now?

toddburnside (approved):
LGTM

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -77,7 +77,7 @@ object Tile {
       .withHooks[Props]
       // info ref
       .useRefToVdom[html.Element]
-      .renderWithReuse { (p, infoRef) =>
```
rpiaggio, line 80: Well, the ref to the title area was called `info` to differentiate from the actual tile contents. Maybe `titleRef`?

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/Tile.scala
cquiroz: Ah ok. I thought it was the actual content. I'll revert

explore/src/main/scala/explore/tabs/ObsTabContents.scala
cquiroz: Because aladin has problems setting the fov on startup

PR #1438, opened 2022-01-13
New brightness model

rpiaggio:
This PR is the first step in supporting the brightness model in Explore. I think it is a good time to merge in order to avoid painful rebases if we diverge further.

For the moment, very little changes visually. Source profile and spectral definition are not displayed yet. Also, changes to brightnesses are not persisted since the API doesn't support it yet.

The only visual change is that the band cannot be changed anymore for existing brightnesses. That is, when a band is added to the table it has that band value forever. Supporting editing existing bands is not straightforward since bands are just keys in a Map now (I made an attempt but had trouble with undo and I'm not sure it's worth pursuing further).

Other changes:
- Reorganized optics so that they are all together.
- Some changes to `ReactTableHelpers` so that they don't require a `Lens` from the row type, and act directly on the cell type. Now the types of the rows are not `View`s (though they contain one), so using `zoom` is not possible.


common/src/main/public/development.conf.json
```diff
@@ -1,6 +1,6 @@
 {
   "environment": "DEVELOPMENT",
-  "odbURI": "wss://lucuma-odb-development.herokuapp.com/ws",
+  "odbURI": "wss://lucuma-odb-master.herokuapp.com/ws",
```
cquiroz, line 3: Should this be `staging`?

common/src/main/scala/explore/schemas/implicits.scala
```diff
@@ -66,36 +72,168 @@ object implicits {
       ParallaxModelInput(microarcseconds = p.μas.value.value.assign)
   }
 
-  implicit class SiderealTargetOps(sidereal: SiderealTarget) {
+  implicit class UnnormalizedSedOps(u: UnnormalizedSED) {
```
cquiroz, line 75: Should these be in `lucuma-schemas`?

explore/src/main/scala/explore/targeteditor/BrightnessesEditor.scala
```diff
@@ -0,0 +1,258 @@
...
+      State(usedBands, Band.all.diff(usedBands.toList).headOption)
+    }
+  }
+
+  protected def defaultBandUnits: Band => Units Of Brightness[T] // Abstract
+
+  implicit protected def propsReuse: Reusability[Props] // Abstract
+  implicit protected val stateReuse: Reusability[State] = Reusability.derive
+
+  implicit protected val displayTaggedUnits: Display[Units Of Brightness[T]] = Display[Units].narrow
```
cquiroz, line 69: What is `narrow`?

cquiroz (approved):
Looks great, I agree we should merge soon to avoid more painful merges

common/src/main/public/development.conf.json
rpiaggio: It should actually be `development`, but that's still in the old API. After we merge this, we can publish the latest API to `lucuma-odb-development` and remove `lucuma-odb-master`.

explore/src/main/scala/explore/targeteditor/BrightnessesEditor.scala
rpiaggio: It adapts the instance of `Display[A]` as a `Display[B]`, where `B <: A`. I'll look into moving this to lucuma-core.

PR #1441, opened 2022-01-14
Make case classes for targets with ids.

toddburnside:


model/shared/src/main/scala/explore/model/TargetWithId.scala
```diff
@@ -0,0 +1,73 @@
...
+    for {
+      id     <- c.get[Target.Id]("id")
+      target <- c.as[Target]
+    } yield TargetWithId(id, target)
+  )
+
+  implicit val eqTargetWithId: Eq[TargetWithId] = Eq.by(x => (x.id, x.target))
+}
+
+case class TargetWithOptId(optId: Option[Target.Id], target: Target)
```
toddburnside, line 61: Would it be preferable to put all the case classes at the top of the file and the objects further down?

cquiroz (approved):
Great, looks better this way

model/shared/src/main/scala/explore/model/TargetWithId.scala
rpiaggio: I think I'd prefer that yes

rpiaggio (approved):
LGTM. It improves readability a lot. Thank you!

PR #1455, opened 2022-01-16
AngularSize now defined in lucuma-catalog, instead of being in target model.

rpiaggio:
This PR goes on top of https://github.com/gemini-hlsw/explore/pull/1446


explore/src/main/scala/explore/targets/TargetSearchResult.scala
```diff
@@ -0,0 +1,30 @@
...
+import japgolly.scalajs.react.Reusability
+import lucuma.catalog.AngularSize
+import lucuma.catalog.CatalogTargetResult
+import lucuma.core.model.Target
+
+final case class TargetSearchResult(
+  targetWithOptId: TargetWithOptId,
+  angularSize:     Option[AngularSize]
+) {
+  lazy val optId: Option[Target.Id] = targetWithOptId.optId
```
cquiroz, line 18: Why lazy?

cquiroz (approved):
Looks good, I guess we should look into storing the angular size on the user preferences or is it going to be on the odb?

explore/src/main/scala/explore/targets/TargetSearchResult.scala
rpiaggio: So as not to allocate it unless it's needed.

rpiaggio:
> Looks good, I guess we should look into storing the angular size on the user preferences or is it going to be on the odb?

This PR is to actually remove it from the ODB. I agree we should initialize visualization preferences with this and store in the users preferences DB. I'll create a story for this.

PR #1457, opened 2022-01-18
Fix title ref rendering in Tile.

rpiaggio:
`Tile` needs to rerender after the `ref` is set to the title strip. The rerender didn't happen if the tile contents were static.

This PR fixes the issue by saving the ref in state, which will force a rerender when set.

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -4,6 +4,8 @@
 package explore.components
 
 import cats.syntax.all._
+import crystal.react.hooks._
+import crystal.react.implicits._
 import crystal.react.reuse._
 import eu.timepit.refined.types.string.NonEmptyString
 import explore.Icons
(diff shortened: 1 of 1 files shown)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -75,8 +73,8 @@ object Tile {
   val component =
     ScalaFnComponent
       .withHooks[Props]
-      // info ref
-      .useRefToVdom[html.Element]
+      // infoRef - We use this mechanism instead of a regular Ref in order to force a rerender when it's set.
+      .useStateView(none[html.Element])
```
cquiroz, line 77: Would you recommend in genera not to use `useRefXXX`?

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -117,27 +118,28 @@ object Tile {
               MenuItem(as = <.a)(p.title)
             ),
             p.control.map(b => <.div(ExploreStyles.TileControl, b)),
-            <.span(ExploreStyles.TileTitleStrip,
-                   ExploreStyles.FixedSizeTileTitle.when(!p.canMinimize && !p.canMaximize)
-            ).withRef(infoRef),
+            <.span(^.untypedRef(setInfoRef))(
```
cquiroz, line 121: I don't recall this on the docs

cquiroz (approved):
Looks great, thanks a lot this was a hard one to fix

common/src/main/scala/explore/components/Tile.scala
rpiaggio: `useRef*` has its uses, it's just that it doesn't trigger a rerender like `useState*` does.

common/src/main/scala/explore/components/Tile.scala
rpiaggio: Here we are using a "callback ref", documented here for React: https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

Basically, when a DOM ref is set, a callback is invoked with the value instead of saving it in a mutable var.

The only way I found to have this behavior in scalajs-react is via "untyped ref": https://github.com/japgolly/scalajs-react/blob/master/doc/REFS.md#untyped

PR #1467, opened 2022-01-19
Small updates to user preferences

cquiroz:

### DIFF
--- a/common/src/main/scala/explore/components/TileController.scala
+++ b/common/src/main/scala/explore/components/TileController.scala
@@ -7,6 +7,7 @@ import cats.effect.IO
 import cats.syntax.all._
 import crystal.react.View
 import crystal.react.implicits._
+import crystal.react.hooks._
 import crystal.react.reuse._
 import eu.timepit.refined.auto._
 import explore.common.UserPreferencesQueries._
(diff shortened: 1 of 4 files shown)


common/src/main/scala/explore/components/TileController.scala
```diff
@@ -37,22 +37,25 @@ final case class TileController(
...
   ): Callback =
-    UserGridLayoutUpsert
-      .storeLayoutsPreference[IO](userId, GridLayoutSection.ObservationsLayout, layouts)
-      .runAsync
-      .debounce(1.second)
+    Callback.log(s"store $userId $section") *>
+      UserGridLayoutUpsert
+        .storeLayoutsPreference[IO](userId, section, layouts)
+        .runAsync
+        .handleErrorWith(r => Callback.log(r.getMessage))
```
rpiaggio, line 57: Since we have `AppContextIO` already in implicit context, we can use `Logger` here (and move `.runAsync` after error handling).

common/src/main/scala/explore/components/TileController.scala
```diff
@@ -37,22 +37,25 @@ final case class TileController(
...
+        .debounce(1.second)
```
rpiaggio, line 58: I don't think `debounce` is actually working here (or in the other places we are using it in the codebase).

Its Scaladoc says: `Save the result of this as a `val` somewhere because it relies on internal state that must be reused.`. Unless we store the debounced `Callback` in a stable value, the internal state is recreated each time, voiding the effect.

Furthermore, if we were to actually use a stable value, it wouldn't cancel the previous invocation since it's a `IO` run asynchronously and we have no reference to it.

I thought maybe we could use `useSingleEffect` for this purpose, but it's not quite the same: here we want a timeout to happen only on re-invocations. We can probably create a `useDebounce` hook, and it can share some (most?) of its code with `useSingleEffect`.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/TileController.scala
cquiroz: What we want is th that if we have n updates in t seconds only one is executed

rpiaggio: Yup. `useSingleEffect` makes sure that an effect is only executed once at a given moment, cancelling the previous instance if reinvoked. We just need to add to it a parameter specifying an optional effect (which can be the timeout) to run between said cancellation and re-invocation.

PR #1469, opened 2022-01-19
Port SkyPlotSection and update sbt-lucuma

cquiroz:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,7 @@ val kindProjectorVersion = "0.13.2"
 
 ThisBuild / Test / bspEnabled                                        := false
 ThisBuild / ScalafixConfig / bspEnabled.withRank(KeyRanks.Invisible) := false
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -10,6 +10,7 @@ val kindProjectorVersion = "0.13.2"
 
 ThisBuild / Test / bspEnabled                                        := false
 ThisBuild / ScalafixConfig / bspEnabled.withRank(KeyRanks.Invisible) := false
+ThisBuild / scalafixDependencies += "com.github.liancheng" %% "organize-imports" % "0.6.0"
```
rpiaggio, line 13: Isn't this set by `sbt-lucuma` ?

rpiaggio (approved):
LGTM

build.sbt
cquiroz: It should be but it isn't. Something changed as it was broken even before

PR #1470, opened 2022-01-20
Debouce via hook.

rpiaggio:


cquiroz (approved):
LGTM

rpiaggio:
Only functional components were migrated to hook debounce mechanism. As we migrate other components to functional, we can migrate calls to `callback.debounce` to the hook.

PR #1502, opened 2022-02-04
Relocate elevation plot to its own panel

cquiroz:
This is a fairly big PR . It moves the elevation plot to its own tile and makes the target tab use grid layout too. 
Several components were converted to functional and styling made more flexible to e.g. allow the graph to be resized

There is still some work to do but this would unblock other developers and let me get some feedback

It is also an opportunity to test the new bulid

### DIFF
--- a/common-graphql/src/main/scala/explore/common/ObsQueriesGQL.scala
+++ b/common-graphql/src/main/scala/explore/common/ObsQueriesGQL.scala
@@ -28,6 +28,14 @@ object ObsQueriesGQL {
               asterism {
                 id
                 name
+                sidereal {
+                  ra {
+                    microarcseconds
+                  }
+                  dec {
+                    microarcseconds
+                  }
+                }
               }
             }
             constraintSet {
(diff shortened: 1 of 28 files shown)


cquiroz:
@armanbilge There i this problem:
image

.npmrc
```diff
@@ -1,2 +0,0 @@
-@fortawesome:registry=https://npm.fontawesome.com/
-//npm.fontawesome.com/:_authToken=${FONTAWESOME_NPM_AUTH_TOKEN}
```
armanbilge, line 2: It's because your PR delete the `.npmrc` file. Was it intentional?

cquiroz: Not that I recall

cquiroz: The commit from the linter action did
https://github.com/gemini-hlsw/explore/pull/1502/commits/a8607aaffd70738e3f085dc44f24a5bca71525aa

armanbilge: Yes I just realize that too. Huh.

cquiroz:
The linter action is working again!!

cquiroz:
Some screenshots
image
image

.npmrc
armanbilge: Oh, I know the problem. It's because the `npm install` step has a `rm .npmrc` sub-step in there. I copied it from the old build but I wasn't sure why you need it.

toddburnside (approved):
LGTM

PR #1503, opened 2022-02-04
Don't delete `.npmrc` in CI

armanbilge:
This caused problems in https://github.com/gemini-hlsw/explore/pull/1502.

I actually don't know why it's like this—I just copied the old CI. But now I'm suspicious, so I removed the code that doesn't make sense. Let's see if it works.

Edit: lol that didn't take long...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -259,7 +259,10 @@ lazy val npmInstall = WorkflowStep.Run(
 lazy val npmBuild = WorkflowStep.Run(
   List("npm run build"),
   name = Some("Build application"),
-  env = Map("NODE_OPTIONS" -> "--max-old-space-size=6144")
+  env = Map(
+    "NODE_OPTIONS"               -> "--max-old-space-size=6144",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -251,12 +251,7 @@ lazy val sbtStage = WorkflowStep.Sbt(List("stage"), name = Some("Stage"))
 
 // https://stackoverflow.com/a/55610612
 lazy val npmInstall = WorkflowStep.Run(
-  List(
-    "rm .npmrc",
-    """npm config set "@fortawesome:registry" https://npm.fontawesome.com/""",
-    """npm config set "//npm.fontawesome.com/:_authToken" ${{ secrets.FONTAWESOME_NPM_AUTH_TOKEN }}""",
-    "npm install"
-  ),
+  List("npm install"),
```
cquiroz, line 254: Do we need to restore `.npmrc`?

armanbilge: No, I don't think we should delete it. I think all these steps were a bad hack.

armanbilge:
Ok, I think it's fixed 👍

cquiroz:
Cool, thanks

cquiroz (approved):

PR #1505, opened 2022-02-04
Fixes an Eq error that sometimes manifested in tests

cquiroz:


common/src/main/scala/explore/model/RootModel.scala
```diff
@@ -70,6 +70,7 @@ object RootModel {
     Eq.by(m =>
       (m.vault,
        m.tabs,
+       m.localPreferences,
```
toddburnside, line 73: Ugh. Hopefully someday we'll be able to derive Eq and avoid this. :(

toddburnside (approved):
LGTM. Thanks for fixing this.

Issue #1504, opened 2022-02-04
Firebase failure on merge


cquiroz:
See

https://github.com/gemini-hlsw/explore/runs/5065315670?check_suite_focus=true

cquiroz:
This on is on build

https://github.com/gemini-hlsw/explore/runs/5066475173?check_suite_focus=true

PR #1506, opened 2022-02-04
Make fontawesome auth token secret env global to workflow

armanbilge:
Fixes https://github.com/gemini-hlsw/explore/issues/1504.

Btw, we could also just add this secret env globally to the whole workflow. Less secure, but then it's automatically applied to every step.


cquiroz:
Less secure in the sense that a rogue action could steal it?

armanbilge:
Yes, exactly.

cquiroz:
I guess that's ok, we could put it on the whole workflow

cquiroz (approved):
Thanks LGTM

PR #1513, opened 2022-02-07
Return of the target summary table

toddburnside:
image

A future PR will make the id a link for editing the individual target, and allow for deleting targets.


cquiroz (approved):
Great to get this back

PR #1517, opened 2022-02-11
Allow selection and editing of individual targets on target tab.

toddburnside:
Also links the target id in the target summary table to select and edit that target.

### DIFF
--- a/explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
+++ b/explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
@@ -1,0 +1,60 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.tabs
+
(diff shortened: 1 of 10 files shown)


explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
```diff
@@ -0,0 +1,60 @@
...
+import explore.undo.UndoStacks
+import japgolly.scalajs.react.vdom.html_<^._
+import lucuma.core.model.Target
+import lucuma.core.model.User
+import lucuma.ui.reusability._
+import react.common._
+
+object SiderealTargetEditorTile {
+
+  def sideralTargetEditorTile(
```
cquiroz, line 25: Where do we use this one?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
toddburnside: When editing an individual target. ie. not editing an asterism.

PR #1530, opened 2022-02-16
Keep target to observation id data up to date on targets tab

toddburnside:
During drag and drop and asterism editing, the mapping between targets and observations (in TargetGroupList) needs to be updated. Currently it isn't important since it is only used in the target summary table and gets updated via round trip. However, soon we will be needing this information when editing asterisms so we know if we are editing the target in a subset of the observations and need to warn/prevent/clone or however we are going to handle this situation.

### DIFF
--- a/model/shared/src/main/scala/explore/model/TargetGroup.scala
+++ b/model/shared/src/main/scala/explore/model/TargetGroup.scala
@@ -1,0 +1,34 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats.Eq
+import cats.implicits._
+import io.circe.Decoder
+import io.circe.Decoder._
+import lucuma.core.model.Observation
+import monocle.Focus
+import monocle.Lens
+
+import scala.collection.immutable.SortedSet
+
+final case class TargetGroup(obsIds: SortedSet[Observation.Id], targetWithId: TargetWithId) {
+  def addObsIds(ids: ObsIdSet): TargetGroup    = TargetGroup.obsIds.modify(_ ++ ids.toSortedSet)(this)
+  def removeObsIds(ids: ObsIdSet): TargetGroup =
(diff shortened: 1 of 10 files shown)


model/shared/src/main/scala/explore/model/TargetGroup.scala
```diff
@@ -0,0 +1,34 @@
...
+import cats.implicits._
+import io.circe.Decoder
+import io.circe.Decoder._
+import lucuma.core.model.Observation
+import monocle.Focus
+import monocle.Lens
+
+import scala.collection.immutable.SortedSet
+
+final case class TargetGroup(obsIds: SortedSet[Observation.Id], targetWithId: TargetWithId) {
```
rpiaggio, line 16: A `TargetGroup` is a target with a set observation ids? Why is it called `TargetGroup`?

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/model/TargetGroup.scala
toddburnside: Only because that is what it is called in the API. Why is it called that in the API? 🤷‍♂️ 
Here, I simply made it a case class instead of using the type created by clue. Mostly because I wanted a SortedSet of observation ids (instead of a list) in order to simplify the code and interactions with ObsIdSets. 

But, it might also be useful when dealing with the "editing a subset of observations" problem in the asterism editor. I haven't fully thought through that issue yet. If it is reused there, I may rename it.

swalker2m: There are a series of these, which are actually _Observation_ groups I suppose.  They differ by how the observations are grouped.  In this case by those that share an individual target.  Perhaps `ObsGroupByTarget`, `ObsGroupByConstraintSet`, `ObsGroupByScienceRequirements`, etc. would be better?  Other ideas?

rpiaggio: `ObsGroupBy*` sounds good to me 👍 . Another option would be `*ObsGroup`, like in `TargetObsGroup`, `ConstraintSetObsGroup`, etc.

PR #1532, opened 2022-02-18
Fix problem with Elevation Plot tile collapsing

toddburnside:
The elevation plot tile was at time collapsing down to just the title. This happened mostly on the targets tab, but occasionally on the observation tab.  
image

It was easy to reproduce on the targets tab: Select an asterism with no targets, then go back to an asterism with targets. The underlying cause of this issue was due to the elevation plot tile being conditionally included in the TileController. The grid layout was being updated without the elevation plot tile, and when the elevation plot tile was reintroduced the layout was messed up. Since the layout is saved in user preferences, the problem was persistent. It was possible to fix it without just logging in as a different user, but it was not obvious how to do so. I tried to fix this issue, but after significant time investment chose this solution instead.

On the observations tab, I was unable to consistently reproduce the issue, but it would happen on occasion.

The solution I came up with isn't ideal. The Elevation Plot tile is always displayed, whether there is a target selected or not. However, this matches the behavior of the asterism editor. Also, NOT having a target in an asterism should be the exception, not the norm, so maybe this behavior is acceptable.

### DIFF
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -517,7 +517,6 @@ object TargetTabContents {
                                     (Constants.InitialTreeWidth.toInt, defaultLayout)
               )
               .attempt
-              .flatTap(x => IO(println(s"Query result: $x")))
               .flatMap {
                 case Right((w, l)) =>
                   (panels
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -383,16 +383,14 @@ object TargetTabContents {
         )
 
       val skyPlotTile =
-        selectedCoordinates.flatten.map(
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -519,6 +517,7 @@ object TargetTabContents {
                                     (Constants.InitialTreeWidth.toInt, defaultLayout)
               )
               .attempt
+              .flatTap(x => IO(println(s"Query result: $x")))
```
rpiaggio, line 520: Debug logging?

toddburnside: Ooops. Yep.

toddburnside:
> Maybe we want to display something like "No target selected" in the plot tile when there's no target?

I started out with that, but took it out to match the blankness of the target editor. I was thinking it would be better to be consistent in both, but maybe not.

PR #1535, opened 2022-02-21
Synchronize elevation plot with target selected in asterism editor

toddburnside:
Second commit adds deep linking to targets and "unifies" the focused target into the RootModel. [sc-1203]


explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -79,23 +80,24 @@ object AsterismEditor {
...
-      .useStateBy(_.asterism.get.headOption.map(_.id))
       // adding
       .useState(false)
       // reset "loading" for add button when science targets change, which indicates server roundtrip is over
-      .useEffectWithDepsBy((props, _, _) => props.asterism.get)((_, _, adding) =>
+      .useEffectWithDepsBy((props, _) => props.asterism.get)((_, adding) =>
         _ => adding.setState(false)
       )
-      .renderWithReuse { (props, selectedTargetIdState, adding) =>
+      .useEffectWithDepsBy((props, _) => (props.asterism.get, props.selectedTargetId.get))(
```
toddburnside, line 89: We could potentially just use `useEffectOnMountBy` to set the selected target to the first one in the list when the component is mounted. Doing it with `useEffectWithDepsBy` has the effect of trying to keep whatever target was last selected to still be selected. This happens even if that target was in a different asterism, which I haven't decided if I like. On the other hand, with a little more expansion of the code, we could select a particular target in an asterism from the target summary table - which I think I do like.

rpiaggio (approved):
LGTM

toddburnside:
> It seems we have a selected target for the observation tab and another one for the target tab. Would it be desireable to move it to `RootModel` and share it?

Yeah. It might be. It seems this is related target deep links in https://app.shortcut.com/lucuma/story/1203. However, it seems like the old implementation will not really cover our needs.

So, perhaps we implement the `Focused` stuff a little differently than it was because the target tab has both the possibility of selecting an observation or a target. We have both a `FocusedObs` and a `FocusedTarget`, which are both optional ids. If there is a `FocusedObs` it takes precedence and on the target tab the observation is selected in the asterism list. In this case, the `FocusedTarget`, if it exists, would be selected in the asterism editor.

If there is no `FocusedObs` but there is a `FocusedTarget`, then on the target tab the target would be selected in the Target Library. 

One question is: Do we need the target tab URL to reflect both the observation id AND the target id? Or, if there is a `FocusedObs` would just a link to the observation be sufficient?

common/src/main/scala/explore/model/RootModel.scala
```diff
@@ -26,6 +26,7 @@ case class RootModel(
   tabs:                           EnumZipper[AppTab],
   localPreferences:               ExploreLocalPreferences,
   focusedObs:                     Option[FocusedObs] = none,
+  focusedTarget:                  Option[Target.Id] = none,
```
toddburnside, line 29: I couldn't decide whether I should "unwrap" the `FocusedObs` or wrap the focused target to make them consistent. I know we had some discussion about `FocusedObs` when we adopted unshared targets and got rid of the focused ADT in a previous incarnation of the target model, but I don't remember why we decided to keep the wrapper. At this point, it seems like overkill to me. So, at the moment the focused observation id is wrapped in `FocusedObs` and the focused target id is used directly.

rpiaggio: `FocusedObs` may be vestigial from the old ADT, maybe we should remove it too and make things uniform?

explore/src/main/scala/explore/Routing.scala
```diff
@@ -85,7 +87,8 @@ object Routing {
       import dsl._
 
       def id[Id](implicit gid: Gid[Id]): StaticDsl.RouteB[Id] =
-        string(gid.regexPattern).pmapL(gid.fromString)
+        // We can't use gid.regexPattern because capture groups don't work here.
```
rpiaggio, line 90: Is this what was causing trouble in URLs with multiple ids?

toddburnside: Yep. That was the problem. Seemed to work with one id, probably when it was the last think in the url. But, failed to parse the url correctly when there was a second one.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/model/RootModel.scala
toddburnside: I agree. I'll remove it.

PR #1536, opened 2022-02-21
Source profile editor.

rpiaggio:
This is far from done but I'd like to merge before diverging too much since there are some fundamental changes to the target editor.

This PR adds functionality to edit a target's source profile to the UI. Layout/styling is not final.

Not yet supported:
- Editing gaussian `fwhm` angle parameter.
- Editing emission lines.
- Editing power law SED index.
- Editing black body SED temperature.
- User defined SEDs.

Supported:
- Changing between Point/Uniform/Gaussian profiles, with brightness units automatically changed between integrated<->surface.
- Specifying the spectrum for all other SEDs.
- Undo.

New functionality relies heavily on `RemoteSyncUndoableF`. This is a structure that allows modifying the model while keeping track of changes needed in the `*Input` structures in order to replicate the changes remotely via the API. It allows zooming into sub-properties in order to compose editor components. The target editor was refactored to make use of this this approach.

`TargetQueries.scala` was removed since it is no longer needed. `*Input` modifications are composable now, so they are done in each component.


explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -61,7 +62,12 @@ object AsterismEditor {
     selectedTarget: View[Option[Target.Id]]
   )(implicit ctx:   AppContextIO): Callback = {
     val (targetId, createTarget) = oTargetId.fold(
-      (newId, ((tid: Target.Id) => TargetQueries.createSiderealTarget[IO](tid, target)))
+      (newId,
```
rpiaggio, line 65: This was the only thing left in `TargetQueries.scala`, so I just moved it over here and deleted that file.

common/src/main/scala/explore/common/RemoteSyncUndoableF.scala
```diff
@@ -0,0 +1,179 @@
...
+ * API.
+ *
+ * We can `zoom` into a property of the model while providing a parallel way to zoom into the delta
+ * structure. Potential modifications to the the delta structure are stored in a coyoneda-like
+ * fashion, but are not executed until the model is modified.
+ *
+ * The idea is to keep code to edit the model as simple as possible, while providing a way to
+ * compose components that edit different parts of the same model object in an undo context.
+ */
+trait RemoteSyncUndoableF[F[_], B, S] { self =>
```
rpiaggio, line 33: I really struggled with finding a name for this. I'm open to suggestions.

explore/src/main/scala/explore/targeteditor/SpectralDefinitionEditor.scala
```diff
@@ -0,0 +1,375 @@
...
+
+    // val blackBodyTemperatureRSUOpt: Option[
+    //   RemoteSyncUndoable[Quantity[PosBigDecimal, Kelvin], Input[BigDecimal]]
+    // ] =
+    //   props.sedRSUOpt.flatMap(
+    //     _.zoomOpt(
+    //       UnnormalizedSED.blackBody.andThen(UnnormalizedSED.BlackBody.temperature),
+    //       UnnormalizedSedInput.blackBodyTempK.modify
+    //     )
+    //   )
```
rpiaggio, line 218: This code left commented out on purpose. It will be used soon.

common/src/main/webapp/less/style.less
```diff
@@ -1304,6 +1304,7 @@ img.partner-split-flag {
   display: flex;
   flex-direction: column;
   border: none;
+  height: 300px; // This is temporary
```
rpiaggio, line 1307: This will be taken care of in a future PR.

common/src/main/scala/explore/common/RemoteSyncUndoableF.scala
```diff
@@ -0,0 +1,179 @@
...
+   */
+  def viewMod(toInput: B => S => S)(implicit
+    F:                 MonadError[F, Throwable],
+    dispatcher:        Effect.Dispatch[F],
+    logger:            Logger[F]
+  ): View[B] =
+    _undoCtx
+      .undoableView(_modelGet, _modelMod)
+      .withOnMod(value =>
+        Callback.log(_remoteMod(toInput(value))(_remoteBaseInput)) >>
```
toddburnside, line 63: Was this for development. If not, should we log more information?

common/src/main/scala/explore/common/RemoteSyncUndoableF.scala
toddburnside: This is a neat way to bundle the updates/undo. The naming is hard. I haven't been able to come up with anything better.

toddburnside (approved):
LGTM

common/src/main/scala/explore/common/RemoteSyncUndoableF.scala
rpiaggio: Good catch. Yes, it was for debugging, I removed it now.

PR #1543, opened 2022-02-22
Disable Aladin.

rpiaggio:
Dev tools in Chrome, Firefox and Safari hang when Aladin is displayed. I'm temporarily disabling it in order to be able to debug while we investigate further.


toddburnside:
Since this appears to be a development issue, and currently only being experienced by one person, do we want to disable Aladin in staging? We could just disable locally as needed until we find a fix. Like I disable the Reusability Overlay when it is in the way and I don't need it. But, I guess that doesn't allow you to use the dev tools on the staging version. 🤔

rpiaggio:
I thought you were experiencing this too, just saw your messages in Slack telling me that was not the case.

I agree we shouldn't disable Aladin if it's only happening to me, I can do it locally when I need to debug as you say.

PR #1548, opened 2022-02-23
Keep target table in asterism editor from disappearing

toddburnside:
This may not be the final answer to the asterism editor layout, but it does allow the table to be displayed. We may want to consider virtualizing the TargetTable so that the table header doesn't scroll, but that would require more work to get it to behave as we want. Right now I size the table based on the number of targets. Without this, table rows would "spread out" to cover the full height, which looks strange and takes up more space than needed. After 2 targets, a partial 3rd row is shown and the whole table - including the header - scrolls.

One issue with the asterism editor is that there are a lot of elements that can react to scrolling, so sometimes you end up scrolling something other than you intended - like the Aladin control. I don't know if there is a way around this, though.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
+++ b/explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
@@ -111,7 +111,7 @@ object AsterismEditor {
           case 2 => 91
           case _ => 110
         }
-        val editorHeight      = resize.height.map(h => math.max(0, h - targetTableHeight)).getOrElse(0)
+        val editorHeight      = resize.height.foldMap(h => math.max(0, h - targetTableHeight))
 
         React.Fragment(
           props.renderInTitle(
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -103,6 +105,14 @@ object AsterismEditor {
             (mod, _) => selectedTargetIdState.modState(mod)
           )
 
+        val targetTableHeight = props.asterism.get.length match {
+          case 0 => 27
+          case 1 => 57
+          case 2 => 91
+          case _ => 110
+        }
+        val editorHeight      = resize.height.map(h => math.max(0, h - targetTableHeight)).getOrElse(0)
```
rpiaggio, line 114: foldMap?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
toddburnside: Ah, yes. Good call.

PR #1561, opened 2022-03-01
mkcert

cquiroz:
This PR adds using `mkcert` to generate the ssl certificates which will be included in chrome and FF thus removing the warning and improving caching efficiency
It also supports nix on M1 macs


toddburnside (approved):
LGTM - not that I really understand nix or vite...

PR #1565, opened 2022-03-02
Source profile/SED editor

rpiaggio:
image

image

image


common/src/main/scala/explore/utils/pprinter.scala
```diff
@@ -0,0 +1,64 @@
...
+  def apply(x: Any): String =
+    pprinterInstance(x, initialOffset = 4).toString
+}
+
+class prodPPrinter extends pprinter {
+  def apply(x: Any): String = x.toString
+}
+
+object pprinter {
+  private val printer: pprinter =
```
cquiroz, line 30: Is `pprinter` not working in prod? or is it to minimize bundle size?

rpiaggio: Right, in order to minimize the bundle size in prod, pprinter defaults to toString in prod.

explore/src/main/scala/explore/targeteditor/BrightnessesEditor.scala
```diff
@@ -217,22 +200,19 @@ sealed abstract class BrightnessesEditorBuilder[T, Props <: BrightnessesEditor[T
             }
           )
 
-        // Put it inside a form to get the SUI styles right
-        Form(as = <.div, size = Small)(
```
cquiroz, line 221: I like this. putting non Form stuff in a form is a kludge. we can get styles correct with plain css

cquiroz (approved):
Looks great

explore/src/main/scala/explore/targeteditor/BrightnessesEditor.scala
rpiaggio: Actually, this was removed because the code here is expected to be in a Form already (the one in SiderealTargetEditor).

Which, I think, may not have been the correct decision, it may be breaking the grid layout in large screens.

rpiaggio: I'll check this and address in a future PR.

PR #1573, opened 2022-03-04
Allow editing of targets for observation subsets (cloning targets)

toddburnside:
The second commit improves the styling of checkboxes somewhat. It may not be the last word in checkbox styling, but it does allow checks to be seen in both light and dark modes, in both focused and unfocused states.


common/src/main/webapp/less/variables-dark.less
```diff
@@ -282,6 +282,7 @@
   --checkbox-focus-background: @gray-dark;
   --checkbox-active-background: var(--control-background-color);
   --checkbox-active-focus-background: var(--control-background-color);
+  --checkbox-active-focus-check-color: @selectedTextColor;
```
cquiroz, line 285: Would you use a variable here too?

explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -154,26 +167,57 @@ object AsterismEditor {
                 val optional =
                   Optional[List[TargetWithId], Target](_.find(_.id === targetId).map(_.target))(
                     target =>
-                      _.map(twid =>
+                      _.map { twid =>
+                        println(s"targetId: $targetId selectedTargetId: ${selectedTargetId.get}")
```
cquiroz, line 171: println?

explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
```diff
@@ -73,18 +77,51 @@ object SiderealTargetEditor {
 
   implicit val propsReuse: Reusability[Props] = Reusability.derive
 
+  def readonlyView[A](view: View[A]): View[A] = {
```
cquiroz, line 80: Should this go to crystal?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
toddburnside: It's possible. I thought we could run it by Raul. There may be a better way to do it.

explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
toddburnside: Yeah. That's a special production ready `println`. 😬 I'll get rid of it.

common/src/main/webapp/less/variables-dark.less
toddburnside: I don't know. There is currently a mix of `@` and `var`. I'm not really sure when it's worth creating a new `var` from an `@`.

PR #1580, opened 2022-03-07
Allow guest on test apps

cquiroz:
I think this would allow you to use guest on test apps


explore/src/main/scala/explore/Explore.scala
```diff
@@ -170,7 +170,12 @@ object ExploreMain extends IOApp.Simple {
           _                    <- logger.info(s"Config: ${appConfig.show}")
           ctx                  <-
             AppContext.from[IO](appConfig, reconnectionStrategy, pageUrl, setPage)
-          r                    <- (ctx.sso.whoami, setupDOM(), showEnvironment(appConfig.environment)).parTupled
+          r                    <- (ctx.sso.whoami.handleError(_ =>
+                                     none
+                                   ), // recover to enable guest users on unauthorized domains
+                                   setupDOM(),
+                                   showEnvironment(appConfig.environment)
+                                  ).parTupled
```
toddburnside, line 178: Does this work? Isn't the problem with SSO authorization?

cquiroz: So, we cannot auth because of the domain but this change will capture the error and set auth to none, so you get the login view.
Still now testing even guest isn't working, I need to do some extras I think

cquiroz:
Ok, this doesn't solve the problem but at least it shows a nicer UI that the previous error message

toddburnside:
> Ok, this doesn't solve the problem but at least it shows a nicer UI that the previous error message

Yeah, I think without the auth we are not able to use any of the lucuma services.

explore/src/main/scala/explore/Explore.scala
rpiaggio: What happens if there's a legit error calling SSO? (It's down for example).

cquiroz:
Let's close this one

PR #1581, opened 2022-03-07
Try to run linter in node 16

cquiroz:


common/src/main/webapp/less/style.less
```diff
@@ -998,7 +999,7 @@ tfoot {
 }
 
 .aladin-detail-text {
-  color: white;
+  color: hsl(0, 0%, 100%);
```
rpiaggio, line 1002: I'm guessing this is unrelated?

rpiaggio (approved):
LGTM

common/src/main/webapp/less/style.less
cquiroz: This was to Truly test the fixing part.I'll revert it

PR #1584, opened 2022-03-08
Set a key to update the view when the coords change

cquiroz:


build.sbt
```diff
@@ -333,9 +333,6 @@ def runLinters(mode: String) = WorkflowStep.Use(
   name = Some(s"Run linters in $mode mode"),
   params = Map(
     "github_token"         -> "${{ secrets.GITHUB_TOKEN }}",
-    "auto_fix"             -> "true",
```
toddburnside, line 336: So, this means that it won't add the commit to fix linting errors? Will the build fail, instead?

cquiroz: Yes, the auto fix unfortunately seems flaky. You'd need to fix the errors locally, like with `scalafmt`

toddburnside (approved):
LGTM

PR #1601, opened 2022-03-16
Extract ConstraintSet, ElevationRange to lucuma-core

rpiaggio:
This is complete and reviewable but it's a draft since we are waiting for the lucuma-sso build to get unstuck.


toddburnside (approved):
And back into core it goes. 😆 LGTM

PR #1604, opened 2022-03-18
Refactor to use Reuse[View]

rpiaggio:
This is large-ish PR implementing the use of `ReuseView` everywhere instead of plain `View`. This fixes a couple of issues we were having, like incorrectly cloning shared targets when brightnesses were edited for all of them, or target fields briefly changing back to its previous values due to stale mod functions.

Also:
- `ReactTableHelpers` was substituted in favor of directly using components at use sites.
- `Tile` logic for handling ref to title strip was modified so that it's more resilient to reusability issues.


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -58,9 +56,11 @@ object Tile {
 
   type RenderInTitle = VdomNode ==> VdomNode
 
-  implicit val propsReuse: Reusability[Tile] = Reusability.derive && Reusability.by(_.render)
+  implicit val propsReuse: Reusability[Tile] = // Reusability.always
+    Reusability.derive && Reusability.by(_.render)
 
-  implicit val htmlElementReuse: Reusability[html.Element] = Reusability.byRefOr_==
+  implicit val htmlElementReuse: Reusability[html.Element] = // Reusability.always
+    Reusability.byRefOr_==
```
toddburnside, line 63: Commented code

toddburnside (approved):
LGTM. It turned into a bigger change than I expected. Thanks for taking care of it.

PR #1614, opened 2022-03-22
Fix target table height in AsterismEditor

toddburnside:


explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -114,20 +113,11 @@ object AsterismEditor {
             .fold(props.selectedTargetId.set(asterism.headOption.map(_.id)))(_ => Callback.empty)
         }
       )
-      .useResizeDetector()
```
toddburnside, line 117: GItHub is doing a poor job of creating a diff. The only real changes are removing the resizeDetector and the sizing code. The fix is in the css by removing the `flex`, etc.

rpiaggio (approved):
LGTM, nice

PR #1617, opened 2022-03-22
Aladin Fn components and bug fix

cquiroz:
I tried to port AladinContainer to a functional component and ended up reproducing an issue reported where aladin would freeze explore in some cases.

This was tracked to an error when the height of the container has not been well defined and that produces an attempt too many remote tiles which blocks the browser.

The port to function components is still missing some visualization related features but they are being worked on in parallel and I'd rather get feedback on the aladin fix


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -214,51 +101,98 @@ object AladinContainer {
...
-      .configure(Reusability.shouldComponentUpdate)
-      .build
+        <.div(
+          ExploreStyles.AladinContainerBody,
+          // This is a bit tricky. Sometimes the height can be 0, this happens if
+          // during a second render. If we let the height to be zero, aladin will
+          // take it as 1 and miss calculate the amount of tiles needed.
+          // This can endup requesting a large amount of tiles and would
+          // freeze explore
+          if (resize.height.exists(_ > 0))
```
toddburnside, line 182: Good detective work! Is there any chance the value could actually be 1 and cause the problem? Is 1 a magic number in aladin?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
rpiaggio: Thank you for tracking down and fixing this

rpiaggio (approved):
LGTM

PR #1629, opened 2022-03-24
Create observation and target id via API, not locally

toddburnside:
For observations, this also means we use the observation created by the API instead of creating a temporary "default" one locally. So, defaults won't get out of sync between the UI and API.

Also converted ObsList to a function component.

### DIFF
--- a/explore/src/main/scala/explore/observationtree/ObsList.scala
+++ b/explore/src/main/scala/explore/observationtree/ObsList.scala
@@ -63,76 +63,17 @@ object ObsList {
   )(implicit
     ctx:        AppContextIO
   ): IO[Unit] =
-    adding.set(true).to[IO] >>
-      createObservation[IO]().flatMap {
-        _.fold(IO.unit) { obs =>
-          ObsListActions
-            .obsExistence(obs.id, focusedObs)
-            .mod(undoCtx)(obsListMod.upsert(obs, pos))
-            .to[IO]
+    adding.async.set(true) >>
+      createObservation[IO]()
+        .flatMap {
+          _.foldMap { obs =>
+            ObsListActions
+              .obsExistence(obs.id, focusedObs)
+              .mod(undoCtx)(obsListMod.upsert(obs, pos))
+              .to[IO]
+          }
         }
-      } >> adding.set(false).to[IO]
-
-  def render(props: Props, adding: ReuseView[Boolean]): VdomNode =
-    AppCtx.using { implicit ctx =>
-      val undoCtx      = UndoContext(props.undoStacks, props.observations)
-      val observations = props.observations.get.toList
-
-      <.div(ExploreStyles.ObsTreeWrapper)(
-        <.div(ExploreStyles.TreeToolbar)(
-          Button(
-            size = Mini,
-            compact = true,
-            icon = Icons.New,
-            content = "Obs",
-            disabled = adding.get,
-            loading = adding.get,
-            onClick = insertObs(observations.length, props.focusedObs, undoCtx, adding).runAsync
-          ),
-          UndoButtons(undoCtx, size = Mini, disabled = adding.get)
-        ),
-        <.div(ExploreStyles.ObsTree)(
-          <.div(ExploreStyles.ObsScrollTree)(
-            <.div(
-              Button(onClick = props.focusedObs.set(none), clazz = ExploreStyles.ButtonSummary)(
-                Icons.ListIcon.clazz(ExploreStyles.PaddedRightIcon),
-                "Observations Summary"
-              )
-            ),
-            observations.toTagMod { obs =>
-              val focusedObs = obs.id
-              val selected   = props.focusedObs.get.exists(_ === focusedObs)
-              <.a(
-                ^.href := ctx.pageUrl(AppTab.Observations,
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -60,121 +55,112 @@ object ObsList {
...
-                observations.toList.get(math.min(idx, observations.length - 1).toLong)
-              }
-              .map(_.id)
-          case other                              => other.some
-        })
+    ScalaFnComponent
+      .withHooks[Props]
+      // Saved index into the observation list
+      .useState(none[Int])
+      .useEffectWithDepsBy((props, _) => (props.focusedObs, props.observations)) {
```
toddburnside, line 142: This replaces both componentDidMount and componentDidUpdate. The logic is quite different, but it seems to be working properly...

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -60,121 +55,112 @@ object ObsList {
...
-      )
+  protected def insertObs(
+    pos:        Int,
+    focusedObs: ReuseView[Option[Observation.Id]],
+    undoCtx:    UndoContext[ObservationList],
+    adding:     ReuseView[Boolean]
+  )(implicit
+    ctx:        AppContextIO
+  ): IO[Unit] =
+    adding.set(true).to[IO] >>
```
rpiaggio, line 66: We can also use `adding.async.set` here.

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -60,121 +55,112 @@ object ObsList {
...
+      createObservation[IO]().flatMap {
+        _.fold(IO.unit) { obs =>
```
rpiaggio, line 68: `foldMap` ?

toddburnside: Indeed. For some reason I frequently forget about that one.

explore/src/main/scala/explore/observationtree/ObsList.scala
toddburnside: That seems a little cleaner. I'll change that in a few places.

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -60,121 +55,112 @@ object ObsList {
...
+          ObsListActions
+            .obsExistence(obs.id, focusedObs)
+            .mod(undoCtx)(obsListMod.upsert(obs, pos))
+            .to[IO]
+        }
+      } >> adding.set(false).to[IO]
```
rpiaggio, line 74: We can use `guarantee` here to make sure the UI doesn't get stuck in case there's an error communicating with the ODB.

This is actually a more general concern that we should address at some point, ie recovering from errors and showing error messages.

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -60,121 +55,112 @@ object ObsList {
...
 
-      newObs.flatMap { obs =>
-        ObsListActions
-          .obsExistence(obs.id, focusedObs)
-          .mod(undoCtx)(obsListMod.upsert(obs, pos))
-      }
-    }
+  def render(props: Props, adding: ReuseView[Boolean]): VdomNode =
```
rpiaggio, line 76: We can make this `private`, or actually move it into the `renderWithReuse` below.

explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -123,22 +111,28 @@ object AsterismEditor {
         React.Fragment(
           props.renderInTitle(
             TargetSelectionPopup(
-              trigger = Reuse.by(adding.value)(
+              trigger = Reuse.by(adding.get)(
```
rpiaggio, line 114: We can use `adding.map( a => Button(...))` now.

toddburnside: Ah, yeah. nice.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/observationtree/ObsListActions.scala
```diff
@@ -72,10 +72,8 @@ object ObsListActions {
         elemWithIndexOpt.fold {
           ProgramDeleteObservation.execute[IO](obsId).void
         } { case (obs, _) =>
-          ProgramCreateObservation
-            .execute[IO](CreateObservationInput(programId = "p-2", observationId = obs.id.assign))
-            .void >>
-            focusedObs.set(obs.id.some).to[IO]
+          // Not much to do here, the observation must be created before we get here
```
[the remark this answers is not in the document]
toddburnside: It's not straight forward because we need the result of the insertion. Even if we committed to a longer wait for the server round trip, if we don't have the new id, we won't be able to easily select it. We could look for a "new observation", but that seems a bit sketchy to me. We do have the possibility of other users. Plus, that'd be a lot more state to look after.

PR #1633, opened 2022-03-25
Avoid rendering Aladin if too small.

rpiaggio:
This is a followup to #1617, the freeze was still happening when height was > 0 but it was very small.


cquiroz (approved):
LGTM

PR #1639, opened 2022-03-28
Don't need to setup process env variable

cquiroz:
Do we still need this after?
https://github.com/typelevel/cats-effect/pull/2642

### DIFF
--- a/explore/src/main/webapp/main.jsx
+++ b/explore/src/main/webapp/main.jsx
@@ -10,18 +10,6 @@ import "react-circular-progressbar/dist/styles.css";
 
 import { Explore } from "@sjs/main.js";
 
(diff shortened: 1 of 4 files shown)


rpiaggio:
IIRC (which I may not), https://github.com/typelevel/cats-effect/pull/2642 made everything work as expected when tracing is disabled, but we still need to explicitly disable it to avoid the performance hit in dev. 

Are you experiencing regular performance even when removing the setting?

@armanbilge can you refresh things for us?

armanbilge:
> Are you experiencing regular performance even when removing the setting?

Yes, this is the important question :)

Tracing does cause a non-trivial performance impact (this may vary by browser) so it depends on the project whether or not it they should disable it in development. It is always disabled in production.

cquiroz:
Ok, we need every ounce of performance in dev

PR #1644, opened 2022-03-30
Avoid exception on ITC errors + show ITC rows again

rpiaggio:
An exception was being thrown to the console when logging an ITC call that resulted in an error. Errors are now logged via `Logger.debug`.

Should we log these errors via `Logger.error`? Or at least some of them?

The 2 types of errors I've seen so far are:
```
ITC: Result for mode Mode: GMOS-S, B600_G5323, Some(GPrime), IfuBlue: List(ERRORS: Chain(GenericError(Error calculating itc org.http4s.client.UnexpectedStatus: unexpected HTTP status: 400 Bad Request for request POST https://gemini-new-itc.herokuapp.com/json)))
```

and:
```
ITC: Result for mode Mode: GMOS-S, B600_G5323, None, Ifu2Slits: List(ERRORS: Chain(GenericError(Error calculating itc lucuma.itc.ItcCalculationError: S/N obtained is 0)))
```

### DIFF
--- a/common/src/main/scala/explore/components/ui/ExploreStyles.scala
+++ b/common/src/main/scala/explore/components/ui/ExploreStyles.scala
@@ -263,6 +263,7 @@ object ExploreStyles {
   val ConfigurationFilter: Css       = Css("explore-configuration-filter")
   val ConfigurationFilterItem: Css   = Css("explore-configuration-filter-item")
   val ExploreTable: Css              = Css("explore-table")
+  val Virtualized: Css               = Css("virtualized")
   val TableRowSelected: Css          = Css("explore-table-row-selected")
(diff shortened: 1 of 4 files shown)


cquiroz:
Thanks for finding this. We should log the errors as they may help what's causing them

rpiaggio:
> Thanks for finding this. We should log the errors as they may help what's causing them

Ok, I changed logging errors to `Logger.error`.

rpiaggio:
Also added a fix to show ITC rows again. Basically, we need to add a CSS class when displaying virtualized tables (which would break them when they are regular `<table>`s).

cquiroz (approved):
LGTM, thanks

PR #1646, opened 2022-03-31
Add selected obs/target to asterismView reuse.

rpiaggio:
The issue here was that the asterism view was being reused when the observation changed, since the asterism view was zoomed in from a ReuseView containing all targets for all observations, and that wasn't changing.

In a way, it's the inverse case of what we had before: now we reuse scope is too wide instead of being too narrow.

This issue makes me worry that we may be comparing a lot of unnecessary things with the new ReuseView everywhere. I've also been profiling the reusability comparisons that are performed and there are way more of these than I imagined.

Moving forward we might need to dial back on the use of Reusability in some places, especially outside of the scope of undo contexts.

### DIFF
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -145,15 +145,15 @@ object TargetTabContents {
     targetGroupMap.get(targetId).fold(0)(tg => (tg.obsIds -- obsIds.toSortedSet).size)
 
   protected def renderFn(
-    props:                Props,
-    panels:               ReuseView[TwoPanelState[TargetOrObsSet]],
-    options:              ReuseView[TargetVisualOptions],
-    defaultLayouts:       LayoutsMap,
-    layouts:              ReuseView[LayoutsMap],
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -356,10 +359,13 @@ object TargetTabContents {
...
-        _.withOnMod(onModAsterismsWithObs(groupIds, idsToEdit))
-          .zoom(getAsterism)(modAsterism)
-      )
+      val asterismView: ReuseView[List[TargetWithId]] =
+        asterismGroupsWithObs
+          .map(
+            _.withOnMod(onModAsterismsWithObs(groupIds, idsToEdit))
+              .zoom(getAsterism)(modAsterism)
+          )
+          .addReuseByFrom(panels)
```
rpiaggio, line 368: This is the actual fix. It adds the selected panel information to reusability.

The rest is format/renaming.

toddburnside (approved):
👍

cquiroz (approved):
LGTM

PR #1661, opened 2022-04-01
z-indices and text-transforms.

rpiaggio:
General revision of z-index and make table header text-transform override SUI's settings.


common/src/main/webapp/less/style.less
```diff
@@ -1057,7 +1057,7 @@ tfoot {
 .edit-warning {
   padding: 5px !important; // important to override SUI's style.
   position: absolute !important; // important to override SUI's style.
-  z-index: 99999;
+  z-index: 2000;
```
cquiroz, line 1060: Yeah let's give some breathing ground to other rules 😜

cquiroz (approved):
LGTM

PR #1676, opened 2022-04-05
New DB API, BigDecimal brightness value.

rpiaggio:
- Upgrade to latest API.
- Support smaller brightness values.


common-graphql/src/main/scala/queries/common/SequenceStepsGQL.scala
```diff
@@ -188,7 +185,7 @@ object SequenceStepsGQL {
       val wavelength: math.Wavelength
     }
     trait SeqFpu[Site <: SeqSite]              {
-      val builtin: Site#Fpu
+      val builtin: Option[Site#Fpu]
```
cquiroz, line 188: When is this none? before setting the config?

model/shared/src/main/scala/explore/model/display.scala
```diff
@@ -76,4 +79,18 @@ object display {
...
 
+  // Not implicit. When we have opaque types we may define a BrightnessValue.
+  // Scientific notation only kicks in when more than 3 decimal digits would be needed.
+  val displayBrightness: Display[BigDecimal] = Display.byShortName { x =>
+    if (x.scale <= 3 && x.precision <= 5)
+      x.toString
+    else {
+      // TODO The same scientific notation formatting is also defined in lucuma-ui.
+      // Should we move it to a Format[String, BigDecimal] in lucuma-core?
+      val formatter = new DecimalFormat("0.0E0")
```
cquiroz, line 90: are you using the formatter from `scala-java-locales`?

cquiroz (approved):
Nice simplification

common-graphql/src/main/scala/queries/common/SequenceStepsGQL.scala
rpiaggio: IDK, it changed to optional in the latest version of the API. There's now a `customMask` besides `builtin`, so I'm assuming it's one or the other. In any case, we should contemplate the `customMask` case, which I was planning to do in a future PR (it's not clear to me how to show it in the table, it has a `fileName` and `slitWidth`).

@swalker2m  ?

model/shared/src/main/scala/explore/model/display.scala
rpiaggio: Yes

toddburnside (approved):
LGTM

common-graphql/src/main/scala/queries/common/SequenceStepsGQL.scala
swalker2m: Yes, when specified at all the FPU is either builtin or custom mask.  I believe the filename is actually [standardized and based on the program id](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/util/MdfConfigRule.scala#L26) (at least in today's program format).  We may need to make corresponding restrictions in the API in which case it might be sufficient to indicate "custom" and a sequence number.

PR #1681, opened 2022-04-08
Add the program id to the URL

toddburnside:
Currently, there is no validation of the program id. If you specify a URL without a program id, it redirects to "p-2". If you specify an invalid URL, it will show tabs without data, but you will not be able to add any observations, etc.

The next step is to see how we can perform the validation. Raul made the good suggestion that we should trap "invalid program" errors from the API and redirect when we get them. This would eliminate the need to try to validate the program id before loading a page, and would catch the situation where a program becomes invalid for a user due to permissions, etc. However, the API currently does not emit an error if you ask for data for an invalid program.

The page for choosing a program also needs to be created. This will be used in the case of an invalid program id in the URL, or if the user chooses to change the program. Since it does not require changes to the API, I may work on that first before validation.

Note: invalid program in this case refers to either a missing program, or one for which the user does not have authorization.


swalker2m:
> However, the API currently does not emit an error if you ask for data for an invalid program.

[Switching to `unsafeSelect` instead of `select` in `ProgramQuery`](https://github.com/gemini-hlsw/lucuma-tmp-api/blob/master/modules/core/src/main/scala/lucuma/odb/api/schema/ProgramQuery.scala#L47) should generate a missing reference error for an unknown program id.  (Currently no permissions checks are done of course.). Is this what we want to do?  An optional result seems arguably correct, like calling `get(key)`  on a `Map`.

toddburnside:
> [Switching to `unsafeSelect` instead of `select` in `ProgramQuery`](https://github.com/gemini-hlsw/lucuma-tmp-api/blob/master/modules/core/src/main/scala/lucuma/odb/api/schema/ProgramQuery.scala#L47) should generate a missing reference error for an unknown program id. (Currently no permissions checks are done of course.). Is this what we want to do? An optional result seems arguably correct, like calling `get(key)` on a `Map`.

It seems like it would be good to get an error if the user is trying to access a program that either doesn't exist, or they don't have authorization for. But, currently, we aren't handling errors in any coherent way in Explore, anyway. Before we make any changes to the API, I need to see how clue is, or can, handle errors. We should probably also discuss within the UI team how we are going to handle errors in general. So, don't make any changes yet.

common-graphql/src/main/scala/queries/common/AsterismQueriesGQL.scala
```diff
@@ -202,7 +202,7 @@ object AsterismQueriesGQL {
           }
         }
 
-        observations(programId: "p-2") {
+        observations(programId: $programId) {
```
cquiroz, line 205: Nice that this is not hardcoded anymore

explore/src/main/scala/explore/config/SequenceEditor.scala
```diff
@@ -10,33 +10,38 @@ import explore.implicits._
...
+
   private def renderFn(config: Option[SequenceSteps.Data.Observations.Nodes.Config]): VdomNode =
     config.fold[VdomNode](<.div("Default observation not found"))(SequenceTable.apply)
 
   val component =
-    ScalaComponent
-      .builder[Props]
-      .render(_ =>
+    ScalaFnComponent
+      .withHooks[Props]
```
cquiroz, line 32: I think this can be done with `withReuse[Props]`

cquiroz (approved):
LGTM

PR #1692, opened 2022-04-12
Editable observation subtitle in badges.

rpiaggio:
Kapture 2022-04-12 at 19 59 40


toddburnside (approved):
Cool. Looks good to me, other than tests not compiling. 😄  You could also run `fix imports`,

PR #1693, opened 2022-04-13
Improvements to EditableLabel.

rpiaggio:
Kapture 2022-04-13 at 14 34 17

A couple of temporary fixes to avoid rerender upon editing/navigating.


cquiroz:
Does it work on pressing enter?

common/src/main/scala/explore/common/AsterismQueries.scala
```diff
@@ -118,6 +118,9 @@ object AsterismQueries {
       }
     )
 
+  import eu.timepit.refined.auto._
```
cquiroz, line 121: Maybe imports at  the top?

rpiaggio: This will be removed shortly when the reload fix is complete.

PR #1694, opened 2022-04-13
Update some libs

cquiroz:


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -67,7 +67,7 @@ object AladinContainer {
         }
       )
       .useResizeDetector()
-      .render { (props, svg, aladinRef, resize) =>
+      .renderWithReuse { (props, svg, aladinRef, resize) =>
```
cquiroz, line 70: This is enable by `scalajs-react` 2.1.0

rpiaggio (approved):
LGTM

PR #1698, opened 2022-04-13
Fix reload issues and ObsBadge match error

toddburnside:
The value inside of `SelectedPanel.Editor(value)` is no longer used. The state update came too late and was causing problems because the editor was being rendered based on old values. On the target tab, this could result in the URL being "rewritten" by the AsterismEditor. Now the focused obs/target values from props are used directly. `SelectedPanel` can be cleaned up and simplified, but I wanted to get this PR into staging because it fixes problems. I will clean up `SelectedPanel` and `TwoPanelState` in a separate PR.

Update: I pushed another commit removing the type parameter from `TwoPanelState`.

### DIFF
--- a/common/src/main/scala/explore/common/AsterismQueries.scala
+++ b/common/src/main/scala/explore/common/AsterismQueries.scala
@@ -109,11 +109,13 @@ object AsterismQueries {
     ScalaFnComponent[ReuseView[AsterismGroupsWithObs] ==> VdomNode](render =>
       AppCtx.using { implicit appCtx =>
         LiveQueryRenderMod[ObservationDB, AsterismGroupObsQuery.Data, AsterismGroupsWithObs](
-          AsterismGroupObsQuery.query(programId).reuseAlways,
+          Reuse.currying(programId).in(pid => AsterismGroupObsQuery.query(pid)),
           (AsterismGroupObsQuery.Data.asAsterismGroupWithObs.get _).reuseAlways,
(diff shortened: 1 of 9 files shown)


cquiroz (approved):
LGTM

rpiaggio (changes requested):
As a general rule we shouldn't define components dynamically. If we do, it should be done once and store the resulting component for future use (like we do with context/state providers and tables).

In this case, although using reusability avoids rerendering with a new component definition during the lifetime of the parent, a new component is defined whenever the parent remounts. If we need to pass to the component things that change during runtime we probably want to unify this through the mechanism of component properties.

This was part of the reason for substituting ReactTableHelpers with direct usage of the controls.

toddburnside:
> As a general rule we shouldn't define components dynamically. If we do, it should be done once and store the resulting component for future use (like we do with context/state providers and tables).
> 
> In this case, although using reusability avoids rerendering with a new component definition during the lifetime of the parent, a new component is defined whenever the parent remounts. If we need to pass to the component things that change during runtime we probably want to unify this through the mechanism of component properties.
> 
> This was part of the reason for substituting ReactTableHelpers with direct usage of the controls.

Hmmm. I see. I guess I can rewrite the LiveQueries into more complex components so they can take the program id as a property value as well as the `A => VdomNode`.

rpiaggio (commented):
LGTM :+1: Thank you. I know this is merged already, but just in case ;)

PR #1699, opened 2022-04-14
Refactoring of aladin visualization rendering

cquiroz:
This brings some improvements tested on react-aladin on the way visualization is handled on explore

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
@@ -3,21 +3,23 @@
 
 package explore.targeteditor
 
-import cats.syntax.all._
 import crystal.react.ReuseView
 import crystal.react.reuse._
+import crystal.react.hooks._
 import explore.components.ui.ExploreStyles
 import explore.model.TargetVisualOptions
 import explore.model.enum.Visible
-import explore.model.reusability._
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -37,26 +39,74 @@ final case class AladinContainer(
...
 
 object AladinContainer {
-  type Props = AladinContainer
+  type Props       = AladinContainer
+  type World2PixFn = Coordinates => Option[(Double, Double)]
+  val DefaultWorld2PixFn: World2PixFn = (_: Coordinates) => None
 
   protected implicit val propsReuse: Reusability[Props] =
-    Reusability.by(x => (x.target, x.options, x.fov))
+    Reusability.by((x: Props) => (x.target, x.options, x.fov))
```
toddburnside, line 47: Are there reasons for not including the other properties in the Reuseability?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
cquiroz: I guess I should add the callbacks, I just fear `Reuse` 😃

PR #1707, opened 2022-04-19
Config queries

cquiroz:
We are going to need the selected config on the targets for AGS

### DIFF
--- a/common-graphql/src/main/scala/queries/common/ConstraintGroupQueriesGQL.scala
+++ b/common-graphql/src/main/scala/queries/common/ConstraintGroupQueriesGQL.scala
@@ -59,6 +59,24 @@ object ConstraintGroupQueriesGQL {
           	    microseconds
           	  }
           	}
+            scienceConfiguration {
+              gmosNorthLongSlit {
+                filter
+                disperser
+                fpu
+                slitWidth {
+                  microarcseconds
+                }
+              }
+              gmosSouthLongSlit {
+                filter
+                disperser
+                fpu
+                slitWidth {
+                  microarcseconds
+                }
+              }
+            }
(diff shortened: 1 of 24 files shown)


common-graphql/src/main/scala/queries/common/ConstraintGroupQueriesGQL.scala
```diff
@@ -59,6 +59,24 @@ object ConstraintGroupQueriesGQL {
...
+              }
+              gmosSouthLongSlit {
+                filter
+                disperser
+                fpu
+                slitWidth {
+                  microarcseconds
+                }
+              }
+            }
```
toddburnside, line 79: Do we actually need this information for the constraints tab? Or, is it just here because of the changes to ObsSummary?

cquiroz: We are only showing the disperser, so it could be simplified at the cost of adding more classes to only hold that field

toddburnside (approved):
LGTM

common-graphql/src/main/scala/queries/common/ConstraintGroupQueriesGQL.scala
toddburnside: Sorry, didn't see your response and deleted my original comment. I saw it was being used for the ObsBadge display. I don't think the cost of modifying ObsSummary would be worth it to remove a few fields.

cquiroz: Right, there is a fine line on just reusing code or having more specific queries

PR #1709, opened 2022-04-20
Set the FOV from the user preferences

cquiroz:
A while ago we setup saving the fov on a targetid basis but the value wasn't really used
This PR restores that functionality. While this is not a very pressing it sets the ground work to do a few other fixes related to the aladin component

I'm not too happy with a few aspects of the code. I'll add comments to further discuss

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinCell.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinCell.scala
@@ -68,7 +68,7 @@ object AladinCell extends ModelOptics {
         UserTargetPreferencesQuery
           .queryWithDefault[IO](props.uid, props.tid, Constants.InitialFov)
           .flatMap(fov =>
-            options.setState(TargetVisualOptions.Default.copy(fov = fov.some).ready).to[IO]
+            options.setState(TargetVisualOptions.Default.copy(fov = fov).ready).to[IO]
           )
           .runAsyncAndForget
       }
@@ -77,16 +77,16 @@ object AladinCell extends ModelOptics {
           ((c: Coordinates) => mouseCoords.setState(c)).reuseAlways
 
         val fov: Option[Fov] =
-          options.value.fold(_ => none, _ => none, tv => tv.fov.map(a => Fov(a, a)))
+          options.value.fold(_ => none, _ => none, tv => Fov(tv.fov, tv.fov).some)
 
         def fovSetter(props: Props, newFov: Fov): Callback =
           if (newFov.x.toMicroarcseconds === 0L) Callback.empty
           else {
             implicit val ctx = props.ctx
-            options.modState(_.map(_.copy(fov = newFov.x.some))) *>
+            options.modState(_.map(_.copy(fov = newFov.x))) *>
               UserTargetPreferencesUpsert
                 .updateFov[IO](props.uid, props.tid, newFov.x)
-                .runAsyncAndForget
+                .runAsync
                 .debounce(1.seconds)
           }
 
--- a/explore/src/main/scala/explore/targeteditor/AladinCell.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinCell.scala
@@ -14,19 +14,21 @@ import explore.Icons
 import explore.common.UserPreferencesQueries._
 import explore.components.ui.ExploreStyles
 import explore.implicits._
+import explore.model.Constants
 import explore.model.ScienceConfiguration
 import explore.model.TargetVisualOptions
 import explore.model.reusability._
 import explore.optics.ModelOptics
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
+import lucuma.core.math.Angle
 import lucuma.core.math.Coordinates
 import lucuma.core.model.Target
 import lucuma.core.model.User
 import lucuma.ui.reusability._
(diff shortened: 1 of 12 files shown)


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -24,14 +23,12 @@ import react.aladin.reusability._
...
 
-import scala.annotation.nowarn
 import scala.concurrent.duration._
 
 final case class AladinContainer(
   target:                 ReuseView[Coordinates],
-  options:                TargetVisualOptions,
   configuration:          Option[ScienceConfiguration],
-  fov:                    ReuseView[Fov],
+  fov:                    Fov,
```
cquiroz, line 31: I guess we should try to combine `fov` and `updateFov`?

model/shared/src/main/scala/explore/model/TargetVisualOptions.scala
```diff
@@ -10,31 +10,31 @@ import lucuma.core.math.syntax.int._
 import monocle.Focus
 
 final case class TargetVisualOptions(
-  fov:      Visible,
-  fovAngle: Angle,
+  ccd:      Visible,
+  fov:      Option[Angle],
```
cquiroz, line 14: I think this could better be a `Pot[Angle]` however I start having problems with the combinations of `ReuseView` and `Pot` especially if I need to zoom on the options to get the the angle

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -52,44 +53,59 @@ object AladinCell extends ModelOptics {
...
-              props.target,
-              props.options.get,
-              props.configuration,
-              fov,
-              coordinatesSetter,
-              Reuse.currying(props).in(fovSetter _),
-              center
-            ).withKey(props.aladinCoords.toString),
-            AladinToolbar(fov.get, mouseCoords.value),
+            fov.get.map(fov =>
```
cquiroz, line 98: Only render after fov has been read

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -52,44 +53,59 @@ object AladinCell extends ModelOptics {
 
   implicit val propsReuse: Reusability[Props] = Reusability.derive[Props]
 
+  val angleFovLens: Lens[Angle, Fov] = Lens[Angle, Fov](a => Fov(a, a))(f => (_ => f.x))
```
rpiaggio, line 56: Is this a lawful `Lens`?

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -52,44 +53,59 @@ object AladinCell extends ModelOptics {
...
+      .useStateViewWithReuse(TargetVisualOptions.Default)
       // flag to trigger centering. This is a bit brute force but
       // avoids us needing a ref to a Fn component
       .useStateViewWithReuse(false)
-      .renderWithReuse { (props, mouseCoords, fov, center) =>
+      .useEffectWithDepsBy((p, _, _, _) => (p.uid, p.tid)) { (props, _, options, _) => _ =>
+        implicit val ctx = props.ctx
+        UserTargetPreferencesQuery
+          .queryWithDefault[IO](props.uid, props.tid, Constants.InitialFov)
+          .flatMap(fov => options.mod(_.copy(fov = fov.some)).to[IO])
```
rpiaggio, line 74: You should be able to `options.zoom(TargetVisualOptions.fov).async.set(fov.some)`.

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -52,44 +53,59 @@ object AladinCell extends ModelOptics {
...
+          .runAsyncAndForget
```
rpiaggio, line 75: If you `runAsync` instead (which requires a `Logger` but we already have that in context here), it will log any errors. Otherwise they will be silenced.

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -52,44 +53,59 @@ object AladinCell extends ModelOptics {
...
+
   val component =
     ScalaFnComponent
       .withHooks[Props]
       // mouse coordinates, starts on the base
       .useStateBy(_.aladinCoords)
-      // field of view
-      .useStateViewWithReuseBy((p, _) => Fov(p.options.get.fovAngle, p.options.get.fovAngle))
+      // target options, will be read from the user preferences
+      .useStateViewWithReuse(TargetVisualOptions.Default)
```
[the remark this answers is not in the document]
cquiroz: That's what I tried initially but (probably due to my own stupidity failed). The combination of `ReuseView[Pot[TargetVisualOption]]` was kind of unwidely to use and also the rendering part was kind of hard

I welcome suggestions

explore/src/main/scala/explore/targeteditor/AladinCell.scala
cquiroz: I guess I should make it a getter

rpiaggio (approved):
LGTM

PR #1711, opened 2022-04-20
Prettify style.less

cquiroz:
This only runs less over prettify. It shouldn' have any effect as it is just formatting


toddburnside (approved):
Seems fine to me

PR #1712, opened 2022-04-20
Modified styles to align target edit form

diego-ge:
Changed styles to align form.
No HTML architecture change required, just set the same size of first column for all grid templates in form.

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -1089,18 +1089,23 @@ tfoot {
 @mobile-responsive-cutoff: 670px;
 
 .target-form {
-  grid-template-columns: [label] minmax(100px, 1fr) [field] 6fr;
+  grid-template-columns: [label] minmax(100px, 10%) [field] 1fr;
   justify-items: stretch;
 }
 
 .target-search-form {
   grid-column: 1 / span 2;
   display: grid;
-  grid-template-columns: [label] minmax(100px, 1fr) [field] 6fr;
+  grid-template-columns: [label] minmax(100px, 10%) [field] minmax(calc(100% - 100px - 0.3em), calc(90% - 0.3em));
   grid-gap: 0.3em;
 
   .field {
     margin-bottom: 0 !important;
+  }
+
+  .input {
+    display: grid !important;
+    grid-template-columns: [input] calc(100% - 20px) [svg] 20px;
   }
 }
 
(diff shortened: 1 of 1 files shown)


cquiroz:
Could you post a screenshot?

common/src/main/webapp/less/style.less
```diff
@@ -1089,19 +1089,24 @@ tfoot {
...
-  grid-template-columns: [label] minmax(100px, 1fr) [field] 6fr;
+  grid-template-columns: [label] minmax(100px, 10%) [field] 1fr;
   justify-items: stretch;
 }
 
 .target-search-form {
   grid-column: 1 / span 2;
   display: grid;
-  grid-template-columns: [label] minmax(100px, 1fr) [field] 6fr;
+  grid-template-columns: [label] minmax(100px, 10%) [field] minmax(calc(100% - 100px - 0.3em), calc(90% - 0.3em));
```
cquiroz, line 1099: Where does the `100px` come from? Should we convert it to a variable?

diego-ge: Sure, here you have 2 screenshots with a small and wide screen to show it is always aligned and also the magnifier icon is not hidden behind aladin div.
image
image

cquiroz: Very nice

cquiroz (approved):
Thanks, LGTM

common/src/main/webapp/less/style.less
```diff
@@ -1089,19 +1089,24 @@ tfoot {
...
   grid-gap: 0.3em;
 
   .field {
     margin-bottom: 0 !important;
   }
+
+  .input {
```
rpiaggio, line 1106: Is this the `input` class from SUI? Should we specify that in the selector as `.ui.input` ?

cquiroz: Right. I wonder if we can avoid the `important`

rpiaggio (approved):
LGTM. Also looks like magic to me :D Thank you!

common/src/main/webapp/less/style.less
diego-ge: Yes, it is the `input` class from SUI, that's why I have modified it inside the `target-search-form` class, to avoid breaking anything else. If you think it can be a breaking change a new class can be created for that specific component or use `.ui.input` as @rpiaggio mentioned.
Also the `important` can be omitted in the `display: grid`.
I will apply both suggestions.

common/src/main/webapp/less/style.less
```diff
@@ -1103,8 +1103,8 @@ tfoot {
     margin-bottom: 0 !important;
   }
 
-  .input {
-    display: grid !important;
+  .ui.input {
+    display: grid;
```
cquiroz, line 1107: Great

PR #1713, opened 2022-04-20
Update react-aladin

cquiroz:


explore/src/main/scala/explore/targeteditor/GmosGeometry.scala
```diff
@@ -41,27 +43,36 @@ object GmosGeometry {
   val port: PortDisposition =
     PortDisposition.Side
 
+  // Move to react common
+  private implicit val cssOrder: Order[Css] = Order.by(_.htmlClass)
```
toddburnside, line 47: This is needed for the NonEmptyMaps?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/GmosGeometry.scala
cquiroz: Yes

PR #1715, opened 2022-04-20
Magnitude tables

cquiroz:
I'm trying to improve the layout of the target tile and one thing we need is that the brightness table is more responsive

This PR converts the table to virtualized and through some style changes it manages to adapt to the size and add scrolling if needed

There is still more work to do but this is a first step
screencast 2022-04-20 19-01-19

### DIFF
--- a/common/src/main/scala/explore/components/ui/ExploreStyles.scala
+++ b/common/src/main/scala/explore/components/ui/ExploreStyles.scala
@@ -130,6 +130,10 @@ object ExploreStyles {
 
   val DraggingOver: Css = Css("dragging-over")
 
+  val TargetTileBody: Css            = Css("target-tile-body")
+  val TargetTileEditor: Css          = Css("target-tile-editor")
+  val TargetSourceProfileEditor: Css = Css("target-source-profile-editor")
+
   val SkyPlotTile: Css               = Css("sky-plot-tile")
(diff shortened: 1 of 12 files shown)


toddburnside (approved):
LGTM - nice work.

rpiaggio:
This table used to be virtualized and I switched it to regular table in order to not have to provide a height and that way keep the table footer (which has the "add new band" controls) together with the table lines, otherwise it can be too far away in big screens.

Also in small screens we want the height to grow with the contents (probably with a minimum height), and that is not easy to do with virtualized. With the new changes in small screens you only see part of a line by default:
image

The tile size can be increased to display more, but you have to increase quite a bit to get more info, which is an issue in small screens.

I propose we switch back to a regular table which IMHO has more desirable semantics for this case, and try to style it so that it's responsive as needed.

cquiroz:
This is not providing a height actually, it gets it from expanding on the grid. 
I think we should limit how small the tile can be to avoid making the table too small, WDYT?

rpiaggio:
Well, we are providing a 100% height. In other words, height is depending on external factors and not on the contents.

We can definitely add a `min-height` which will help in small screens, but I don't see a way to avoid the "new line" controls from getting detached from the rows.

PR #1722, opened 2022-04-22
Fixed autofill unreadable fields and target table hidden on small sizes

diego-ge:
Fixed autofill unreadable parameters and preventing target table from getting hidden if tile size too small

https://user-images.githubusercontent.com/44274704/164801643-7e681eb7-f615-4b5a-b886-21b775085407.mp4

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -1135,6 +1135,7 @@ tfoot {
 .explore-table.target-asterism-table {
   height: auto;
   max-height: 150px;
+  min-height: 63px;
   overflow-y: scroll;
 }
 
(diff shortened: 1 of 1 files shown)


common/src/main/webapp/less/style.less
```diff
@@ -2243,3 +2244,25 @@ th.sticky-header,
...
+.ui.input input:-webkit-autofill:focus,
+.ui.input input:-webkit-autofill:active,
+.ui.form input:-webkit-autofill,
+.ui.form input:-webkit-autofill:hover,
+.ui.form input:-webkit-autofill:focus,
+.ui.form input:-webkit-autofill:active,
+input:-webkit-autofill,
+input:-webkit-autofill:hover,
+input:-webkit-autofill:focus,
+input:-webkit-autofill:active
```
cquiroz, line 2263: Does this only afect safari?

cquiroz (approved):
LGTM

common/src/main/webapp/less/style.less
diego-ge: No, it shouldn't, actually the video was done using chrome.

cquiroz: Is this available as a standard field. In theory we should be able to generate the browser specific suffixes via postcss

common/src/main/webapp/less/style.less
```diff
@@ -2243,3 +2244,25 @@ th.sticky-header,
...
+{
+  box-shadow: 0 0 0 30px var(--form-input-background) inset !important;
+  -webkit-box-shadow: 0 0 0 30px var(--form-input-background) inset !important;
```
cquiroz, line 2266: Could we investigate https://github.com/postcss/autoprefixer to avoid these?

PR #1723, opened 2022-04-25
Some improvements to brightness/emission table width

cquiroz:
The brightness table tends to be too wide. This PR brings some improvements to reduce the need to have horizontal scrolling though still if it is too narrow there is no way around


toddburnside (approved):
LGTM

PR #1724, opened 2022-04-25
Improvements to the Target grid layout

cquiroz:
Currently the grid layouts has some issues when resizing and on resetting the value stored in user preferences.
This PR brings several improvements:

* Remove the attempt to scale the tiles depending on the height. Though it sounds nice in practice it is difficult to implement
* Add some debugging aids to show the tiles' properties
* Put limits on how small tiles can be. In practice very small tiles are hard to layout

This is a first step we still need to improve the obs tab layout and some other improvements when resizing

### DIFF
--- a/common/src/main/scala/explore/components/TileController.scala
+++ b/common/src/main/scala/explore/components/TileController.scala
@@ -31,6 +31,7 @@ import react.gridlayout._
 import scala.concurrent.duration._
 import scala.scalajs.js.JSConverters._
 import scala.scalajs.js.|
+import scala.scalajs.LinkingInfo.developmentMode
 
 final case class TileController(
   userId:           Option[User.Id],
@@ -57,10 +58,7 @@ object TileController {
     ctx:       AppContextIO
   ): Callback =
     debouncer
-      .submit(
-        UserGridLayoutUpsert
-          .storeLayoutsPreference[IO](userId, section, layouts)
-      )
+      .submit(UserGridLayoutUpsert.storeLayoutsPreference[IO](userId, section, layouts))
       .runAsyncAndForget
 
   val itemHeght = layoutItems.andThen(layoutItemHeight)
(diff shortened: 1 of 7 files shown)


common/src/main/scala/explore/components/TileController.scala
```diff
@@ -126,20 +127,50 @@ object TileController {
...
+                          ^.cls := "rgl-tile-overlay",
+                          s"id: ${i.i} x: ${i.x} y: ${i.y} w: ${i.w} h: ${i.h}${i.minH.toOption
+                              .foldMap(m => s" minH: $m")}${i.maxH.toOption
+                              .foldMap(m => s" maxH: $m")}${i.minW.toOption
+                              .foldMap(m => s" minW: $m")}${i.maxW.toOption.foldMap(m => s" maxW: $m")}"
+                        )
+                      )
+                    }
+                }
+                .getOrElse(EmptyVdom),
```
toddburnside, line 173: This debug information will always show in dev mode?

cquiroz: Nope, it is hidden via css too. I don't expect we'd use this very often

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -529,27 +528,28 @@ object ObsTabContents {
...
-                (panels
-                  .mod(
-                    TwoPanelState.treeWidth.replace(w.toDouble)
-                  ) *> layout.mod(o => mergeMap(o, l)))
-                  .to[IO]
-              case Left(_)       => IO.unit
-            }
-            .runAsync
-      }
+      // TODO Rework the obs tab layout
```
toddburnside, line 531: A separate PR?

toddburnside (approved):
LGTM

common/src/main/scala/explore/components/TileController.scala
toddburnside: Ah. OK. It seemed pretty special purpose. :)

explore/src/main/scala/explore/tabs/ObsTabContents.scala
cquiroz: Yes, some day

PR #1728, opened 2022-04-26
Add Program Management modal popup

toddburnside:
Adds the ability to select/add/delete/edit programs from the menu, or in the case where a program is not specified in the URL. In the latter case, if the user only has one program, it will select that program without showing the popup. This does not yet handle the case of an invalid ID in the URL. 

Also, at this point editing the program name does not persist because the API version in development and staging does not support it. It is supported in API master, but upgrading Explore to use it is not straight forward since API master also contains change to the sequences. I think it will be better to wait for Raul's changes.

I opted for the ability to show deleted programs over undo/redo because I think it could be important for a user to be able to undelete a program even if they do something like reload the page. Undo/redo could be added as well if it becomes regarded as important for name editing.

image

### DIFF
--- a/common/src/main/scala/explore/model/reusability.scala
+++ b/common/src/main/scala/explore/model/reusability.scala
@@ -11,6 +11,7 @@ import japgolly.scalajs.react.Reusability
 import lucuma.catalog.AngularSize
 import lucuma.catalog.CatalogTargetResult
 import lucuma.core.model.Target
+import lucuma.schemas.ObservationDB.Enums.Existence
 import lucuma.ui.reusability._
 
 import scala.collection.immutable.TreeSeqMap
@@ -70,4 +71,6 @@ object reusability {
   implicit val catalogTargetResultReuse: Reusability[CatalogTargetResult]          = Reusability.derive
   implicit val scienceConfigurationnResultReuse: Reusability[ScienceConfiguration] =
     Reusability.byEq
+
+  implicit val existenceReuse: Reusability[Existence] = Reusability.derive
 }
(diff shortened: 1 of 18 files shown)


common/src/main/scala/explore/model/reusability.scala
```diff
@@ -70,4 +71,6 @@ object reusability {
   implicit val catalogTargetResultReuse: Reusability[CatalogTargetResult]          = Reusability.derive
   implicit val scienceConfigurationnResultReuse: Reusability[ScienceConfiguration] =
     Reusability.byEq
+
+  implicit val existenceReuse: Reusability[Existence] = Reusability.derive
```
toddburnside, line 75: I'm not sure if there is a better place for this?

cquiroz: This is probably the right place

explore/src/main/scala/explore/programs/ProgramsPopup.scala
```diff
@@ -0,0 +1,170 @@
...
+          ProgramTable(
+            props.currentProgramId,
+            programs,
+            showDeleted.get,
+            selectProgram = Reuse.currying(isRequired, isOpen, props.undoStacks).in(selectProgram _)
+          ),
+          <.div(
+            Button(
+              clazz = ExploreStyles.ProgramAdd,
+              // size = Small,
```
cquiroz, line 128: Commentted code

explore/src/main/scala/explore/programs/ProgramsPopup.scala
```diff
@@ -0,0 +1,170 @@
...
+            )
+          )
+        )
+      )
+    )
+  }
+
+  protected val component = ScalaFnComponent
+    .withHooks[Props]
+    .useStateViewWithReuse(false) // isOpen
```
cquiroz, line 148: I hope when we have opaque types we could make these so it is clearer which one is doing what, but that's unrelated to this PR

toddburnside: Right. There are a lot of Booleans. ☹️

cquiroz:
Yeah, the eraser looks nice

PR #1732, opened 2022-04-26
Don't change FOV on tile resizing

cquiroz:
This fixes a bug where aladin's fov changes when the tile gets resized

Still there is some fov change when the aladin section is hidden and redisplayed but that will be addressed separetly


toddburnside (approved):
👍

PR #1755, opened 2022-04-29
Add position angle control

cquiroz:
This is not yet on the backend model but we can keep it memory

image


explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -93,12 +105,40 @@ object ConfigurationPanel {
       op.runS(SpectroscopyRequirementsData()).value
     }
 
+  // Unsafe due to the angle truncation but usabe for the UI
+  private val unsafePosAngleAngleLens: Lens[PosAngle, TruncatedPA] =
```
rpiaggio, line 109: Can this be a `SplitEpi`?

rpiaggio:
It seems to me the logic could be streamlined by defining `case class PosAngle(tpe: PosAngleOptions, value: Angle)` (We woudn't need the conversions between the enum and ADT). Have you already considered that path and discarded it?

Also, maybe we can rename `PosAngleOptions` -> `PosAngleType` ?

cquiroz:
I expect PosAngle to be in core and would be used by eg ags and the api while the options are only for the UI to be used on a select widget

explore/src/main/scala/explore/config/ConfigurationPanel.scala
cquiroz: Can you pass a `SplitEpi` to `View`?

toddburnside: There is a `zoomSplitEpi` method in `lucuma.ui.implicits`

cquiroz: This maybe going beyond what a `SplitEpi` can do. `SplitEpi` would require a function `Angle => PosAngle` and there isn't enough information to build that info as `PosAngle` is an ADT

rpiaggio:
> I expect PosAngle to be in core and would be used by eg ags and the api while the options are only for the UI to be used on a select widget

I see, thank you

model/shared/src/main/scala/explore/model/PosAngle.scala
```diff
@@ -0,0 +1,38 @@
...
+object PosAngle {
+  case class Fixed(angle: Angle)               extends PosAngle {
+    override def toString: String = s"Fixed(${angle.toDoubleDegrees})"
+  }
+  case class AllowFlip(angle: Angle)           extends PosAngle {
+    override def toString: String = s"AllowFlip(${angle.toDoubleDegrees})"
+  }
+  case class AverageParallactic(angle: Angle)  extends PosAngle {
+    override def toString: String = s"AverageParallactic(${angle.toDoubleDegrees})"
+  }
```
swalker2m, line 23: Should we be storing an `Angle` in the case of average parallactic?  It will depend on the time of day.

explore/src/main/scala/explore/config/ConfigurationPanel.scala
rpiaggio: Oh ok. I was trying to avoid the unsafe lens and thinking we can provide an update function for the `Angle`. This would require turning things around, with a safe lens into the Angle and then a SplitEpi Angle <-> TruncatedAngle, but then we would be stuck with a TruncatedAngle in the model, which we probably don't want.

Maybe we can just zoom using a get and mod function instead of a lens?

model/shared/src/main/scala/explore/model/PosAngle.scala
swalker2m: Screen Shot 2022-04-29 at 17 09 02

cquiroz: My understanding (which is vague) is that the angle in `AverageParallactic` will be calculated, that's why it is read only on the UI. I'm not yet clear how is it going to be calculated or the parameters used to do so

cquiroz:
I updated the `PosAngle` `ADT` removing angle from `AverageParallactic` and adding `Unconstrained`. The UI code is also improved by zooming directly to the angle when appropriate

swalker2m:
> I updated the PosAngle ADT removing angle from AverageParallactic and adding Unconstrained. The UI code is also improved by zooming directly to the angle when appropriate

Looks good. This will eventually migrate to `lucuma-core` then?

cquiroz:
@swalker2m Yes I expect it will, I'm yet uncertain on whether it will be a part of config or observation?

swalker2m:
> I'm yet uncertain on whether it will be a part of config or observation?

Me either, but my guess is that to make it easier to group by science mode and bulk edit, we should add it directly to the observation.  For example, the same GMOS North long slit mode could apply to multiple observations but at different position angles.

cquiroz:
Right, in that case we should probably move it to `lucuma-core` and use it there or on the api as part of observation

cquiroz:
Alongside PA, I'm going to add a control for "observation time". This will be used for PM correction and we should consider if it belongs to observation too

rpiaggio (approved):
LGTM

PR #1766, opened 2022-04-30
Display generated sequences + ODB updates + Reusability.byEq

rpiaggio:
Only future steps are shown for the moment (past visits will be addressed in the future).

No generation input parameters yet, also will be addressed in future PR.

image


common-graphql/src/main/scala/queries/common/ManualSequenceGQL.scala
```diff
@@ -0,0 +1,269 @@
...
+
+// gql: import lucuma.schemas.decoders._
+// gql: import lucuma.ui.reusability._
+
+object ManualSequenceGQL {
+
+  @GraphQL
+  trait SequenceSteps extends GraphQLOperation[ObservationDB] {
+    val document = """
+      query($programId: ProgramId!) {
```
cquiroz, line 22: general question. this queries are validated by the scalafix steps right? Just asking to confirm these are validated at compile time

explore/src/main/scala/explore/config/ManualSequenceTables.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.vdom.html_<^._
+import lucuma.core.model.sequence._
+import lucuma.ui.reusability._
+import react.common._
+import react.semanticui.elements.header.Header
+import react.semanticui.elements.segment.Segment
+
+final case class ManualSequenceTables(config: ManualConfig)
```
cquiroz, line 14: This looks very similar to `GeneratedSequenceTable` could it be generated?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -310,32 +310,29 @@ object SpectroscopyModesTable {
         .setSortType(DefaultSortTypes.number)
     ).filter { case c => (c.id.toString) != FPUColumnId.value || fpu.isEmpty }
 
-  protected def rowToConf(row: SpectroscopyModeRow): Option[ScienceConfiguration] =
+  protected def rowToConf(row: SpectroscopyModeRow): Option[ScienceModeBasic] =
```
cquiroz, line 313: Great to see the ITC queries updated

cquiroz (approved):
Great work, just a few comments

common-graphql/src/main/scala/queries/common/ManualSequenceGQL.scala
rpiaggio: Right, they are checked against the schema by the scalafix rule, which is run whenever we compile.

explore/src/main/scala/explore/config/ManualSequenceTables.scala
rpiaggio: They are very similar indeed. I expect them to diverge in the future. I'll unify them nonetheless in a next PR.

PR #1774, opened 2022-05-02
Measure the size of the surrounding div for the obs and target areas

cquiroz:
We need to measure the size of the areas of targets and observations but for some reason it is not working all the time.
This PR wraps each on a div that is being measured reliably.


toddburnside (approved):
LGTM

rpiaggio (approved):
LGTM

PR #1777, opened 2022-05-02
Small improvements to the obs grid layout

cquiroz:
I did some improvements to the layout on the target tab, this brings the same to the obs layout. A big change is restoring the layout from the user preferences


toddburnside (approved):
LGTM

PR #1779, opened 2022-05-03
Add time control for the time of observation

cquiroz:
Like `posAngle` it is not currently stored in the backend

image

To better style the date time picker the styles was copied from the original project and customized


rpiaggio (approved):
LGTM

PR #1780, opened 2022-05-03
Show gmos fpu

cquiroz:
Now with PosAngle we can display the fpu if selected

https://user-images.githubusercontent.com/3615303/166562678-9a86b998-a15b-4717-a0e6-c0ee39740a1e.mp4

### DIFF
--- a/model/shared/src/main/scala/explore/model/ObsSummary.scala
+++ b/model/shared/src/main/scala/explore/model/ObsSummary.scala
@@ -131,7 +131,17 @@ object ObsSummaryWithTitleConstraintsAndConf {
   val activeStatus = Focus[ObsSummaryWithTitleConstraintsAndConf](_.activeStatus)
 
   implicit val eqObsSummaryWithTitleConstraintsAndConf: Eq[ObsSummaryWithTitleConstraintsAndConf] =
-    Eq.by(o => (o.id, o.title, o.subtitle, o.constraints, o.status, o.activeStatus, o.duration))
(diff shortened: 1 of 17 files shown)


model/shared/src/main/scala/explore/model/ObsSummary.scala
```diff
@@ -131,7 +131,17 @@ object ObsSummaryWithTitleConstraintsAndConf {
...
-    Eq.by(o => (o.id, o.title, o.subtitle, o.constraints, o.status, o.activeStatus, o.duration))
+    Eq.by(o =>
+      (o.id,
+       o.title,
+       o.subtitle,
+       o.constraints,
+       o.status,
+       o.activeStatus,
+       o.duration,
+       o.scienceMode
```
toddburnside, line 142: Do you know if this `scienceMode` is used elswewhere? Or is it only for the purpose of sending to the AladinContainer?

toddburnside (approved):
Looks good for Andy to check out. I'll try to fix the API/local sync problem in a PR tomorrow.

model/shared/src/main/scala/explore/model/ObsSummary.scala
cquiroz: It is (or will be used for step generation)

toddburnside: The value in the ObsSummary will be used? It's also in the ObsEditQuery.

PR #1782, opened 2022-05-04
Fix reuse issue on target tab.

toddburnside:
Fixes issue where switching observation on the target tab would change the URL and the sidereal target editor, but the target table in the asterism editor wouldn't update.


cquiroz (approved):
LGTM

PR #1783, opened 2022-05-04
Dramatically reduce "live query" rendering complexity by using the power of Streams and Resources

rpiaggio:
All the `LiveQueryRender*` and `SubscriptionRender*` components are gone, as well as their complex shared logic.

The same behavior can be achieved by making subscriptions return `Resource[Stream]`, couple of helper methods, and the new hooks and much simpler components in crystal.


explore/src/main/scala/explore/config/GeneratedSequenceViewer.scala
```diff
@@ -32,15 +33,16 @@ object GeneratedSequenceViewer {
...
-            Option[FutureExecutionConfig]
-          ](
-            SequenceSteps.query(props.obsId).reuseAlways,
-            ((_: SequenceSteps.Data).observation.flatMap(_.execution.config)).reuseAlways,
-            List.empty.reuseAlways
-          )(
+          StreamResourceRenderer[Option[FutureExecutionConfig]](
+            Resource.eval(
+              IO(
+                fs2.Stream.eval(
```
cquiroz, line 39: Neat

cquiroz (approved):
Looks like a great improvement

PR #1788, opened 2022-05-05
Avoid LocalDateTime.now()

rpiaggio:
`LocalDateTime.now()` doesn't seem to work outside of the loaded time zones (which are `Pacific/Honolulu` and `America/Santiago`).

This PR implements a workaround by explicitly getting the local date time from the JS `Date` and converting it into a `LocalDateTime`.

### DIFF
--- a/explore/src/main/scala/explore/config/ConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/ConfigurationPanel.scala
@@ -55,6 +55,7 @@ import react.semanticui.collections.form.Form
 import react.semanticui.elements.button.Button
 import react.semanticui.sizes._
 
+import java.time.Instant
 import java.time.LocalDateTime
 import java.time.ZoneOffset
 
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -532,7 +532,7 @@ object ObsTabContents {
               .runAsync
       }
       // Shared obs conf (posAngle/obsTime)
-      .useStateViewWithReuse(ObsConfiguration(PosAngle.Default, LocalDateTime.now()))
+      .useStateViewWithReuse(ObsConfiguration(PosAngle.Default, new js.Date().toLocalDateTime))
```
cquiroz, line 535: This is a good observation. Now this time is always going to be in UTC, maybe it makes more sense to make that explicit?
In either case the use of `now` is not ideal, but it is done in the interim as a solution to not having this value coming from the odb

rpiaggio: > This is a good observation.

Actually, the app was crashing for me since I'm in a different time zone :D

> Now this time is always going to be in UTC

`new js.Date()` produces a `Date` with the local time actually.

rpiaggio: After our Slack discussion, I'm changing the value to store an `Instant`.

explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -149,26 +150,16 @@ object ConfigurationPanel {
...
-        localDate.getYear,
-        localDate.getMonthValue - 1,
-        localDate.getDayOfMonth,
-        localDate.getHour(),
-        localDate.getMinute()
-      )
+  implicit class InstantOps(val instant: Instant) extends AnyVal {
+    // DatePicker only works in local timezone, so we trick it by adding the timezone offset.
+    def toDatePickerJsDate: js.Date =
+      new js.Date(instant.toEpochMilli.toDouble + (new js.Date()).getTimezoneOffset() * 60000)
```
cquiroz, line 156: is `js.Date` always local?
Have you seen this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC

rpiaggio: Yup, I tried that first.

`js.Date` just stores millis since epoch, but the issue here is `DatePicker`, which will only show it as local time. See: https://github.com/Hacker0x01/react-datepicker/issues/1787 (open for almost 3 years and still unsolved).

cquiroz (approved):
LGTM

PR #1796, opened 2022-05-06
Setup workers

cquiroz:
This PR setups the infrastructure to create and use web workers. There is no real activity done by them but this is a springboard to start using them

Notes:
* workers will live in a separate module. The main reason is to avoid calling invalid apis in particular dom related apis
* workers will produce a separate scala.js output file from explore but the build takes care of doing them always together
* A wrapper `WebWorkerF` was created to encapsulate the operation
* I added a worker to the `AppContext` to make it available everywhere, maybe it is not the best approach. we can revesit
* I added a sample of how to send/receive messages to the worker

### DIFF
--- a/common/src/main/scala/explore/model/AppContext.scala
+++ b/common/src/main/scala/explore/model/AppContext.scala
@@ -29,9 +29,11 @@ import lucuma.schemas._
 import org.http4s._
 import org.http4s.dom.FetchClientBuilder
 import org.http4s.implicits._
+import org.scalajs.dom
 import org.typelevel.log4cats.Logger
 import queries.schemas._
 import retry._
+import workers.WebWorkerF
 
 import scala.concurrent.duration._
 
@@ -116,7 +118,8 @@ case class AppContext[F[_]](
     Option[Target.Id],
     SetRouteVia
   ) => Callback,
-  environment: ExecutionEnvironment
+  environment: ExecutionEnvironment,
+  worker:      WebWorkerF[F] // There will be a few workers in the future
 )(implicit
   val F:       Applicative[F],
   val logger:  Logger[F],
@@ -151,6 +154,17 @@ case class AppContext[F[_]](
   ): Callback = replacePage(appTab, programId, obsId.map(o => ObsIdSet.one(o)), targetId)
 }
 
+object WebWorkers {
+  def createIOWorker[F[_]: Async](file: String): F[WebWorkerF[F]] = Sync[F].delay {
+    val worker = new dom.Worker(file,
+                                new dom.WorkerOptions {
+                                  `type` = dom.WorkerType.module
+                                }
+    )
+    WebWorkerF[F](worker)
+  }
+}
+
 object AppContext {
   def from[F[_]: Async: FetchJSBackend: WebSocketBackend: Parallel: Effect.Dispatch: Logger](
     config:               AppConfig,
@@ -168,6 +182,7 @@ object AppContext {
       clients    <-
         Clients
           .build[F](config.odbURI, config.preferencesDBURI, config.itcURI, reconnectionStrategy)
+      worker     <- WebWorkers.createIOWorker[F]("/workers.js")
       staticData <- StaticData.build[F](uri"/instrument_spectroscopy_matrix.csv")
       version     = utils.version(config.environment)
       actions     = Actions[F]()
@@ -178,6 +193,7 @@ object AppContext {
                           SSOClient(config.sso),
                           pageUrl,
                           setPageVia,
-                          config.environment
+                          config.environment,
+                          worker
     )
 }
--- a/common/src/main/scala/explore/model/AppContext.scala
+++ b/common/src/main/scala/explore/model/AppContext.scala
@@ -155,13 +155,27 @@ case class AppContext[F[_]](
 }
 
 object WebWorkers {
-  def createIOWorker[F[_]: Async](file: String): F[WebWorkerF[F]] = Sync[F].delay {
-    val worker = new dom.Worker(file,
-                                new dom.WorkerOptions {
-                                  `type` = dom.WorkerType.module
-                                }
-    )
-    WebWorkerF[F](worker)
+  import scala.scalajs.js
+  import scala.scalajs.js.annotation.JSImport
+
+  /**
+   * This deserves an explanation:
+   *
+   * To make the webworker act correctly in both dev and production we shoud import it as a module
+   * rather than just doing a direct consructor call.
+   *
+   * Doing the import with the "worker" param gives a constructor for the worker which we can wrap
+   * inline lets us save some space keeping a single chunk More info see:
+   * https://vitejs.dev/guide/features.html#import-with-query-suffixes=
+   */
+  @js.native
+  @JSImport("/workers.js?worker&inline", JSImport.Default)
+  object TestWorker extends js.Object {
+    def apply(): dom.Worker = js.native
+  }
+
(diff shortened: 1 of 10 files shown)


explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -499,10 +499,15 @@ object ObsTabContents {
       // Shared obs conf (posAngle/obsTime)
       .useStateViewWithReuse(ObsConfiguration(PosAngle.Default, Instant.now))
       .useSingleEffect(debounce = 1.second)
+      // worker usage demo Delete soon
+      .useEffectOnMountBy((props, _, _, _, _, _, _) =>
+        props.ctx.worker.stream.evalMap(msg => IO.println(msg.data.toString)).compile.drain
```
cquiroz, line 504: To get messages just get values from the stream

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -499,10 +499,15 @@ object ObsTabContents {
...
+      )
       .renderWithReuse {
         (props, twoPanelState, resize, layouts, defaultLayout, obsConf, debouncer) =>
           implicit val ctx = props.ctx
           <.div(
+            <.button("test", ^.onClick --> ctx.worker.postMessage(1000)).when(true), // TODO delete
```
cquiroz, line 510: This is a sample of sending a message

project/Settings.scala
```diff
@@ -347,5 +348,12 @@ object Settings {
         "com.lihaoyi" %%% "pprint"
       )(pprint)
     )
+
+    val ScalaWebAppUtil = Def.setting(
```
cquiroz, line 352: Will be used later on for IndexedDB

common/src/main/scala/explore/model/AppContext.scala
```diff
@@ -151,6 +154,31 @@ case class AppContext[F[_]](
...
+   *
+   * To make the webworker act correctly in both dev and production we shoud import it as a module
+   * rather than just doing a direct consructor call.
+   *
+   * Doing the import with the "worker" param gives a constructor for the worker which we can wrap
+   * inline lets us save some space keeping a single chunk More info see:
+   * https://vitejs.dev/guide/features.html#import-with-query-suffixes=
+   */
+  @js.native
+  @JSImport("/workers.js?worker&inline", JSImport.Default)
```
rpiaggio, line 172: This is vite specific then? Why `inline`?

common/src/main/scala/workers/WebWorkerF.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Terminate the web worker
+   */
+  def terminate: F[Unit]
+
+  /**
+   * Streams of events from the worker
+   */
+  def stream: Stream[F, dom.MessageEvent]
```
rpiaggio, line 34: Awesome that it integrates with fs2.

common/src/main/scala/workers/WebWorkerF.scala
```diff
@@ -0,0 +1,60 @@
...
+}
+
+object WebWorkerF {
+
+  def apply[F[_]: Async](worker: dom.Worker) = new WebWorkerF[F] {
+    def postMessage(message: js.Any): F[Unit] = Sync[F].delay(worker.postMessage(message))
+
+    def terminate: F[Unit] = Sync[F].delay(worker.terminate())
+
+    def stream: Stream[F, dom.MessageEvent] =
```
rpiaggio, line 44: We can pass the `Dispather` implicitly and reuse the one we already have in explore's context.

rpiaggio: Calling `stream` multiple times will allocate multiple `Channel`s and register multiple `onmessage` callbacks. Is that what we want?

Otherwise we can set up the `Channel` and register a callback in the enclosing `WebWorkerF` constructor.

explore/src/main/scala/explore/tabs/ObsTabContents.scala
rpiaggio: Awesome

common/src/main/scala/workers/WebWorkerF.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Streams of events from the worker
+   */
+  def stream: Stream[F, dom.MessageEvent]
+}
+
+object WebWorkerF {
+
+  def apply[F[_]: Async](worker: dom.Worker) = new WebWorkerF[F] {
```
rpiaggio, line 39: Can we make this return a `Resource[WebWorker[F]]` that calls `terminate` when closing? That way clients are freed from the onus of manually terminating.

This won't make much of a difference if we use global web workers, but if we ever use component-local ones we can easily use them with the `useResource` hook.

workers/src/main/scala/workers/IDBWorker.scala
```diff
@@ -0,0 +1,32 @@
...
+import js.annotation._
+
+import cats.effect.unsafe.implicits._
+
+/**
+  * Sample web worker, extremely simple just accept messages, prints them
+  * and answers a number
+  */
+@JSExportTopLevel("IDBWorker", moduleID = "worker")
+object IDBWorker {
```
rpiaggio, line 20: Neat.

We could define a `trait IOWebWorker` where you just have to define a `def onMessage(msg: dom.MessageEvent): IO[Unit]`.

rpiaggio (approved):
This looks great, thank you!

LGTM, with a couple of suggestions.

I'm OK with merging this with the demo until we have a real use case implemented.

toddburnside:
> I'm OK with merging this with the demo until we have a real use case implemented.

Plus, it gives Carlos something to demo next week.

toddburnside (approved):
Cool stuff!

common/src/main/scala/explore/model/AppContext.scala
cquiroz: It is vite specific. `inline` means it wont get built into a separate chunk. The extra chunk in my local tests was fairly big so this seems a good option which also will work well in terser. We can further test this in the future

common/src/main/scala/workers/WebWorkerF.scala
cquiroz: Would it be enough to convert this to a lazy val?

common/src/main/scala/workers/WebWorkerF.scala
cquiroz: Good idea

workers/src/main/scala/workers/IDBWorker.scala
cquiroz: Right, this is still more on setting up the infrastructure, I thought about passing a partial function for message events too `PartialFunction[js.Any, IO[Unit]]`

cquiroz:
There isn't anything to demo yet but there are concrete use cases:

* loading catalog data
* Doing cache via indexedb
* Calculation of the sky plots
* Anything in general that is CPU intensive

common/src/main/scala/workers/WebWorkerF.scala
rpiaggio: I think that would work, yes.

toddburnside:
> There isn't anything to demo yet but there are concrete use cases:

If there is a button, there is something to demo. 😆

cquiroz:
Tested the "demo" in FF/Chrome/Safari in OSX

PR #1797, opened 2022-05-07
Fix small issue on the target layout

cquiroz:
Fixes an issue when the target edit warning indicator makes the target layout getsa bit too tall.

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -651,6 +651,8 @@ tfoot {
 
 .target-tile-editor {
   flex-grow: 2;
+  display: flex;
+  flex-direction: column;
 }
 
 .explore-form.target-source-profile-editor {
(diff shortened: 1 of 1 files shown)


common/src/main/webapp/less/style.less
```diff
@@ -2035,6 +2037,12 @@ html {
   }
 }
 
+.tile-xs-width {
+  .explore-shared-edit-warning {
+    display: none;
```
toddburnside, line 2042: No warning for very narrow displays?

toddburnside (approved):
LGTM

common/src/main/webapp/less/style.less
cquiroz: So, and this maybe a bad take. We are hiding the editors for narrow screens thus the warning isnt necessarily. If we were going to show the editors we should certainly restore the warning

toddburnside: That probably makes sense. It would be pretty hard to edit a target on, say, a phone.

PR #1800, opened 2022-05-10
Don't save fov foor very small changes

cquiroz:
Due to approximations done in aladin the fov maybe overwritten more often than necessary and it leads to a slow non user performed drift

This PR simply doesn't save for changes under 1 arcsec


toddburnside (approved):
LGTM

PR #1802, opened 2022-05-12
Apply proper motion on aladin target overlay

cquiroz:
https://user-images.githubusercontent.com/3615303/168156538-3e16d08c-2ed8-4712-80f9-0d8ab7be60f7.mp4

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
@@ -124,8 +124,8 @@ object AladinContainer {
     ScalaFnComponent
       .withHooks[Props]
       // Base coordinates with pm correction if possible
-      .useStateBy { p =>
-        p.obsConf.map(o => p.target.get.at(o.obsInstant)).getOrElse(p.target.get.baseCoordinates)
+      .useMemoBy(_.obsConf.map(_.obsInstant)) { p => i =>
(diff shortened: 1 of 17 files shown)


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -223,26 +226,48 @@ object AladinContainer {
...
+                  target = baseCoordinatesForAladin,
+                  fov = props.options.fovAngle,
+                  showGotoControl = false,
+                  customize = includeSvg _
+                )
+              }
+            )
+          }
+
+          // )
```
rpiaggio, line 270: Commented line

rpiaggio (approved):
LGTM

rpiaggio:
Looks great!

I wonder if we can put the observation time picker on a new header for the whole observations (above the tiles). Maybe an empty tile will do?

swalker2m:
Very nice!  Will the time picker have a "Now" option perhaps?  And operators will want a "Now + Setup" eventually.

cquiroz:
> Looks great!
> 
> I wonder if we can put the observation time picker on a new header for the whole observations (above the tiles). Maybe an empty tile will do?

That was an option and it make sense as to apply to every part of the observation

cquiroz:
> Very nice! Will the time picker have a "Now" option perhaps? And operators will want a "Now + Setup" eventually.

I certainly miss a `Now` option. what `Now + Setup` means?

PR #1803, opened 2022-05-13
Interim solution to store the obstime locally

cquiroz:
While we don't have obs time or pos angle in the API doing demos is a bit hard. This PR stores them locally so each obs gets its own time

Note this is temporal as whenever they get out of sync with db reboots the saved values need to be cleared. A facility for  doing so was added to the top menu


rpiaggio (approved):
LGTM

PR #1804, opened 2022-05-14
Option to hide reusability overlays.

rpiaggio:
Kapture 2022-05-14 at 16 27 53

### DIFF
--- a/explore/src/main/scala/explore/TopBar.scala
+++ b/explore/src/main/scala/explore/TopBar.scala
@@ -54,24 +54,24 @@ final case class TopBar(
 object TopBar {
   type Props = TopBar
 
-  implicit val propsReuse: Reusability[Props] = Reusability.by(p => (p.user, p.programId))
+  implicit private val propsReuse: Reusability[Props] = Reusability.by(p => (p.user, p.programId))
 
-  def currentTheme: Theme =
-    if (dom.document.body.classList.contains(Theme.Light.clazz.htmlClass))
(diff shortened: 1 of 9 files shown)


toddburnside (approved):
Nice. Will eliminate the need for me to comment out the overlays in the scala code all the time. Thanks.

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -177,19 +178,25 @@ object TopBar {
...
                       // DELETME
-                      DropdownItem(text = "Reset Local Obs",
-                                   icon = Icons.SkullCrossBones.fixedWidth(),
-                                   onClick = ExploreLocalPreferences.cleanObsConfig[IO].runAsync
+                      DropdownItem(
+                        text = "Reset Local Obs",
+                        icon = Icons.SkullCrossBones.fixedWidth(),
+                        onClick = ExploreLocalPreferences.cleanObsConfig[IO].runAsync
+                      )
+                        .when(appCtx.environment === ExecutionEnvironment.Development),
```
cquiroz, line 195: Only dev, great

cquiroz (approved):
This is going to be useful, thanks

cquiroz:
Maybe we should hide the indicator by default and when needed activate it?

toddburnside:
> 



> Maybe we should hide the indicator by default and when needed activate it?

That would work for me.

PR #1805, opened 2022-05-15
Update ODB API

rpiaggio:

PR #1809, opened 2022-05-16
Advanced configuration panel, only enum fields.

rpiaggio:
There's still a long way to go in the advanced configuration panel, but I'd like to merge now that some fields are working.

Note: this PR includes #1805 

Kapture 2022-05-15 at 23 03 59


common/src/main/scala/explore/implicits.scala
```diff
@@ -106,11 +107,19 @@ trait RefinedImplicits {
     f(v.value)
 }
 
+trait EnumeratedImplicits {
+  implicit def combinedEnum[A: Enumerated, B: Enumerated]: Enumerated[(A, B)] =
```
cquiroz, line 111: zip?

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -0,0 +1,256 @@
...
+              )
+            ),
+            Button(
+              size = Small,
+              compact = true,
+              clazz = ExploreStyles.VeryCompact,
+              content = "Simple Configuration",
+              icon = Icons.ChevronsLeft,
+              onClick = props.onShowBasic.value
+            )(^.tpe := "button")
```
cquiroz, line 136: Could this be done with the `as` mechanism? It is just a detail

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -0,0 +1,256 @@
...
+      ScienceModeAdvanced.GmosNorthLongSlit.explicitYBin
+    @inline protected val explicitReadMode                  =
+      ScienceModeAdvanced.GmosNorthLongSlit.explicitAmpReadMode
+    @inline protected val explicitGain                      =
+      ScienceModeAdvanced.GmosNorthLongSlit.explicitAmpGain
+    @inline protected val explicitRoi                       =
+      ScienceModeAdvanced.GmosNorthLongSlit.explicitRoi
+  }
+
+  // Gmos Soutth Long Slit
```
cquiroz, line 216: Typo

explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -70,190 +45,78 @@ final case class ConfigurationPanel(
...
-    )
+  protected implicit val propsReuse: Reusability[Props] = Reusability.derive
 
   protected val component =
     ScalaFnComponent
       .withHooks[Props]
-      .useStateViewWithReuse[ScienceMode](ScienceMode.Spectroscopy)
-      .useStateViewWithReuse[ImagingConfigurationOptions](ImagingConfigurationOptions.Default)
-      .renderWithReuse { (props, mode, imaging) =>
+      .useStateViewWithReuse(false) // showAdvanced
```
cquiroz, line 53: We need opaque types 😝

cquiroz (approved):
LGTM

common/src/main/scala/explore/implicits.scala
rpiaggio: :+1:

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
rpiaggio: Hmmm `as` is for changing the tag, but in this case we want to add an attribute (so that the button does not submit the form).

PR #1816, opened 2022-05-18
Empty FWHM and fix brightness table in gaussian.

rpiaggio:
Fixes:
- [sc-1362] Cannot backspace over first digit of profile FWHM
- [sc-1364] Brightness table size with Gaussian profile


toddburnside (approved):
LGTM

PR #1822, opened 2022-05-19
LiveQuery with holdAfterMod.

rpiaggio:
When I scrapped `LiveQueryMod` and friends, I forgot to carry over setting the `holdAfterMod` parameter, which prevents input fields from briefly flickering back to their old value when they are modified.

This PR takes care of that by creating a proxy `LiveQuery` component that just calls `StreamResourceRendererMod` with `potRender` and `holdAfterMod = 2.seconds`.


toddburnside (approved):
LGTM

PR #1825, opened 2022-05-19
Read catalog stars in explore

cquiroz:
This PR adds support for reading catalog stars in explore. Given this is an expensive operation it is offloaded to a `WebWorker` which will keep a local cache using `IndexedDb`. I'm using some features to improve performance like storing data binary encoded and passing message from/to the worker using `Transferrable` objects. This was tested in Chrome, FF and Safari

The cache will be expired after a while or on demand.

The stars are not yet being displayed on the UI but that will come soon enough

The infrastructure developed for this feature can be used for other similar cases, specifically for ITC and elevation plots

I wrote a bit about the way the cache is structured at
https://github.com/gemini-hlsw/explore/wiki/IndexedDB-Cache

### DIFF
--- a/common/src/main/scala/workers/WebWorkerF.scala
+++ b/common/src/main/scala/workers/WebWorkerF.scala
@@ -3,15 +3,18 @@
 
 package workers
 
+import boopickle.Default._
 import cats.effect.Async
 import cats.effect.Resource
 import cats.effect.Sync
 import cats.effect.std.Dispatcher
+import explore.model.boopickle._
 import fs2.Stream
 import fs2.concurrent.Channel
 import org.scalajs.dom
 
 import scala.scalajs.js
+import scala.scalajs.js.typedarray._
 
 /**
  * WebWorker abstraction running on F. it is possible to post messages and get a stream of events
@@ -23,6 +26,16 @@ trait WebWorkerF[F[_]] {
    * Post a message on effect F
    */
   def postMessage(message: js.Any): F[Unit]
+
+  /**
+   * Post a transferrable message on effect F
+   */
+  def postTransferrable(buffer: Int8Array): F[Unit]
+
+  /**
+   * Post a boopickle encoded message on effect F
+   */
+  def postTransferrable[A: Pickler](a: A): F[Unit]
 
   /**
    * Terminate the web worker
@@ -44,6 +57,12 @@ object WebWorkerF {
     Resource.make(Sync[F].delay(new WebWorkerF[F] {
       def postMessage(message: js.Any): F[Unit] =
         Sync[F].delay(worker.postMessage(message))
+
+      def postTransferrable(buffer: Int8Array): F[Unit] =
+        Sync[F].delay(worker.postMessage(buffer, js.Array(buffer.buffer: dom.Transferable)))
+
+      def postTransferrable[A: Pickler](a: A): F[Unit] =
+        postTransferrable(asTransferable(a))
 
       def terminate: F[Unit] =
         Sync[F].delay(worker.terminate())
--- a/common/src/main/scala/workers/WebWorkerF.scala
+++ b/common/src/main/scala/workers/WebWorkerF.scala
@@ -35,7 +35,8 @@ trait WebWorkerF[F[_]] {
   /**
    * Post a boopickle encoded message on effect F
    */
-  def postTransferrable[A: Pickler](a: A): F[Unit]
+  def postTransferrable[A: Pickler](a: A): F[Unit] =
+    postTransferrable(asTransferable(a))
 
   /**
    * Terminate the web worker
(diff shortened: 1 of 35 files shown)


common/src/main/scala/workers/WebWorkerF.scala
```diff
@@ -45,6 +58,12 @@ object WebWorkerF {
       def postMessage(message: js.Any): F[Unit] =
         Sync[F].delay(worker.postMessage(message))
 
+      def postTransferrable(buffer: Int8Array): F[Unit] =
+        Sync[F].delay(worker.postMessage(buffer, js.Array(buffer.buffer: dom.Transferable)))
+
+      def postTransferrable[A: Pickler](a: A): F[Unit] =
+        postTransferrable(asTransferable(a))
```
rpiaggio, line 65: This seems like a good general default implementation. We can put it in the `trait WebWorkerF`.

explore/src/main/scala/explore/Explore.scala
```diff
@@ -207,7 +207,7 @@ object ExploreMain extends IOApp.Simple {
 
     (for {
       dispatcher <- Dispatcher[IO]
-      worker     <- WebWorkerF[IO](WebWorkers.TestWorker(), dispatcher)
+      worker     <- WebWorkerF[IO](WebWorkers.CacheIDBWorker(), dispatcher)
```
rpiaggio, line 210: Are we planning to have more workers? If so, maybe we can give this a more specific name.

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -66,7 +71,28 @@ object AladinCell extends ModelOptics {
...
+      // Lisen on web worker for messages with catalog candidates
+      .useEffectOnMountBy((props, _, _, _, gs) =>
+        props.ctx.worker.stream
+          .evalMap(m =>
+            decodeFromTransferable[IO, CatalogResults](m)(gsl => gs.setState(gsl.candidates).to[IO])
+          )
+          .compile
+          .drain
+          .whenA(props.obsConf.isDefined)
+      )
```
rpiaggio, line 85: Can we use a `StreamRenderer` (from crystal) ?

`StreamRenderer` does exactly this: reads from a stream and sets a state whenever a new value is read.

rpiaggio: It will also cancel the fiber on unmount.

model/shared/src/main/scala/explore/model/CatalogResults.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import cats.Eq
+
+/**
+ * Holds a set of candidate guide stars
+ */
+final case class CatalogResults(candidates: List[GuideStarCandidate])
+
+object CatalogResults {
+  def empty = CatalogResults(Nil)
```
rpiaggio, line 14: This can be `val Empty`.

model/shared/src/main/scala/explore/model/boopickle/package.scala
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+package object boopickle extends BoopicklePlatform {}
```
rpiaggio, line 6: Why do we need this an the empty `BoopicklePlatform`?

cquiroz: I want to have a few facilities that are js only. To convert binary blobs to Transferrable objects

explore/src/main/scala/explore/targeteditor/AladinCell.scala
cquiroz: Ah, nice tip

explore/src/main/scala/explore/Explore.scala
cquiroz: Maybe. I called it `CatalogWorker` originally but then became more general. Given the extra size we get per worker it could make sense to have only one, though for expensive actions (Semester skyplots) maybe we want separte ones

explore/src/main/scala/explore/targeteditor/AladinCell.scala
cquiroz: Ah but Stream renderer creates a component right? whil the code above only maniulates the state

rpiaggio: Right, you can save the component in state.

Or, on second thought, we should have a `useStream` hook. I'll check that approach.

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/model/boopickle/package.scala
cquiroz: I think we could move it

cquiroz:
Ah ok now I know why `boopickle` is in model and not in `common`. This lives on WebWorker land and we cannot have there dependencies that are for UI like react-table or you could easily call a forbidden dom method.
`common` has several of those while `model` has not.
I prefer if the `workers` project depends on `model` rather than `common`

rpiaggio:
> This lives on WebWorker land and we cannot have there dependencies that are for UI like react-table or you could easily call a forbidden dom method.

I see. It seems like we should have a `common` and a `common-ui`.

cquiroz:
Could be but I'm happy with this solution so far

PR #1827, opened 2022-05-20
Avoid overflowing with the active status toggle

cquiroz:
Before:
image

After:
image

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -1804,6 +1804,7 @@ html {
 .obs-badge-description {
   display: flex;
   justify-content: space-between;
+  text-overflow: ellipsis;
 }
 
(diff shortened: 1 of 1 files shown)


toddburnside:
When did these get smooshed together?
image
It seems a bit hard to read.

toddburnside (approved):
LGTM

cquiroz:
Yeah I didn't want to reduce the size of the text. but tested it making it very small and the text will overflow to the next line.

toddburnside:
> Yeah I didn't want to reduce the size of the text. but tested it making it very small and the text will overflow to the next line

Actually, I meant the fact that `GMOS-S B600` is smooshed together with `<0.1" <0.1 mag Dark Wet`. It's nothing that changed in this PR, but I just noticed it when looking at the images.

PR #1840, opened 2022-05-22
Layout saving improvements

cquiroz:
This fixes some bugs saving the position of the tiles and reduces a bit the amount of data stored in user preferences

### DIFF
--- a/common/src/main/scala/explore/components/TileController.scala
+++ b/common/src/main/scala/explore/components/TileController.scala
@@ -3,9 +3,9 @@
 
 package explore.components
 
+import cats.Eq
 import cats.effect.IO
 import cats.syntax.all._
-import crystal.react.ReuseView
 import crystal.react.hooks._
 import crystal.react.implicits._
 import crystal.react.reuse._
@@ -36,7 +36,7 @@ final case class TileController(
   userId:           Option[User.Id],
   gridWidth:        Int,
   defaultLayout:    LayoutsMap,
-  layoutMap:        ReuseView[LayoutsMap],
+  layoutMap:        LayoutsMap,
   tiles:            List[Tile],
   section:          GridLayoutSection,
   clazz:            Option[Css] = None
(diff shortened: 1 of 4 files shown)


common/src/main/scala/explore/components/TileController.scala
```diff
@@ -139,24 +142,40 @@ object TileController {
           rowHeight = Constants.GridRowHeight,
           draggableHandle = s".${ExploreStyles.TileTitleMenu.htmlClass}",
           onBreakpointChange = (bk: BreakpointName, _: Int) => bn.setState(bk),
-          onLayoutChange = (_: Layout, b: Layouts) =>
+          onLayoutChange = (_: Layout, b: Layouts) => {
+            // We need to sort the layouts to do proper comparision
```
rpiaggio, line 146: Typo in the comment

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -392,21 +392,24 @@ object ObsTabContents {
...
+                notesTile,
+                skyPlotTile,
+                constraintsTile,
+                configurationTile
+              ),
+              GridLayoutSection.ObservationsLayout,
+              clazz = ExploreStyles.ObservationTiles.some
+            )
+
+          potRender[LayoutsMap](rglRender.reuseAlways)(layouts.get)
```
rpiaggio, line 412: Does this rerender when the values that `rglRender` uses change? `props.userId`, `coreWidth`, the tiles, etc? Otherwise, we may want to make reusability depend on those.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/tabs/ObsTabContents.scala
cquiroz: Excellent question. I'm not sure, This is inside another `Reuse.by` at line
https://github.com/gemini-hlsw/explore/blob/master/explore/src/main/scala/explore/tabs/ObsTabContents.scala#L305=

Do we need it twice? I get lost with nested reusabilities

PR #1845, opened 2022-05-23
Show indicator while loading guidestars

cquiroz:
image


explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -79,13 +76,15 @@ object AladinCell extends ModelOptics {
       // flag to trigger centering. This is a bit brute force but
       // avoids us needing a ref to a Fn component
       .useStateViewWithReuse(false)
-      // GuideStar candidates
-      .useState(List.empty[GuideStarCandidate])
+      // GuideStar candidates (loading, list)
+      .useState((true, List.empty[GuideStarCandidate]))
```
rpiaggio, line 80: Maybe it's more readable to use 2 separate states?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinCell.scala
cquiroz: I thought about that but they really change together, When loadig goes to false the list of candidates gets populated. I thought about a utility class but maybe that's overkill

PR #1846, opened 2022-05-23
Changed label text in elevation plot

diego-ge:


explore/src/main/scala/explore/targeteditor/SkyPlotNight.scala
```diff
@@ -351,7 +351,7 @@ object SkyPlotNight {
                 .setLabels(YAxisLabelsOptions().setFormat("{value}°")),
               YAxisOptions()
                 .setOpposite(true)
-                .setTitle(YAxisTitleOptions().setText("Brightness (mags/arcsec²)"))
+                .setTitle(YAxisTitleOptions().setText("Sky Brightness [V] (mags/arcsec²)"))
```
cquiroz, line 354: What does the `[V]` mean?

cquiroz: /cc @andrewwstephens

andrewwstephens: The [V] is to indicate that the sky brightness value is in the V-band.
If it's not too much trouble, could you change "mags/arcsec^2" to "mag/arcsec^2" (without the "s")?

diego-ge: No problem at all.
Will do it right now.

PR #1847, opened 2022-05-24
Constraints tile improvements

cquiroz:
Improves responsiveness of the tile for constraints adjusting the layout and the undo buttons.
It also has a few small bug fixes

https://user-images.githubusercontent.com/3615303/170076170-08fffe48-8524-4a23-a1ef-ec250713c6bd.mp4

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -36,7 +36,8 @@ final case class Tile(
   key:               js.UndefOr[Key] = js.undefined,
   controllerClass:   Option[Css] = None, // applied to wrapping div when in a TileController.
   bodyClass:         Option[Css] = None, // applied to tile body
-  tileClass:         Option[Css] = None  // applied to the tile
+  tileClass:         Option[Css] = None, // applied to the tile
+  tileTitleClass:    Option[Css] = None  // applied to the title
 )(val render:        Tile.RenderInTitle ==> VdomNode)
     extends ReactFnProps[Tile](Tile.component) {
   def showMaximize: Boolean =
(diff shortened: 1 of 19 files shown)


common/src/main/scala/explore/utils/package.scala
```diff
@@ -72,6 +72,13 @@ package object utils {
         pot.fold(pendingRender, errorRender, valueRender)
     )
 
+  def potRenderBasic[A](
```
cquiroz, line 75: @rpiaggio WDYT, a potRender for dummies (or just me if you don't want to make it plural)

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -109,29 +110,36 @@ object Tile {
...
-              MenuItem(as = <.a)(p.title)
-            ),
-            p.control.map(b => <.div(ExploreStyles.TileControl, b)),
-            <.span(^.key := "tileTitle", ^.untypedRef(setInfoRef))(
-              ExploreStyles.TileTitleStrip,
-              ExploreStyles.FixedSizeTileTitle.when(!p.canMinimize && !p.canMaximize)
-            ),
-            minimizeButton.when(p.showMinimize),
-            maximizeButton.when(p.showMaximize)
+          // Title title, set classes based on size
```
rpiaggio, line 115: Did you mean `Tile title` ?

common/src/main/scala/explore/utils/package.scala
rpiaggio: Sounds good if we are going to attempt moving to a world with less reuse.

Maybe this can be `potRender` and we can rename the other one to `potRenderWithReuse`.

PR #1860, opened 2022-05-26
Full migration to hooks for streams and resources.

rpiaggio:
`LiveQuery` is gone, as well as crystal's `Stream(Resource)Renderer(Mod)`, all of their uses replaced by the new hooks.

I had to reorganize the `ProgramsPopup` a bit.

Also, the observation tab tiles were refactored into a separate component.


toddburnside (approved):
LGTM

PR #1861, opened 2022-05-27
More migration to new hooks.

rpiaggio:


explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -76,29 +76,28 @@ object AladinCell extends ModelOptics {
       // flag to trigger centering. This is a bit brute force but
       // avoids us needing a ref to a Fn component
       .useStateViewWithReuse(false)
-      // GuideStar candidates (loading, list)
-      .useState((true, List.empty[GuideStarCandidate]))
```
cquiroz, line 80: Why is this gone?it seems we'd need to store the state somewhere?

rpiaggio: The state is within the `useStreamBy` hook.

explore/src/main/scala/explore/config/GeneratedSequenceViewer.scala
```diff
@@ -28,16 +28,12 @@ object GeneratedSequenceViewer {
   val component =
     ScalaFnComponent
       .withHooks[Props]
-      .useStreamOnMountBy { props =>
+      .useEffectResultOnMountBy { props =>
```
cquiroz, line 31: I wonder if we'd need a hook that would take deps and be reactivated if the deps change

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -76,28 +75,24 @@ object AladinCell extends ModelOptics {
...
-            decodeFromTransferable[IO, CatalogResults](m)(gsl =>
-              gs.setState((false, gsl.candidates)).to[IO]
+      // Listen on web worker for messages with catalog candidates
+      .useStreamBy((props, _, _, _) => (props.target.get, props.obsConf))((props, _, _, _) =>
+        deps =>
+          props.ctx.worker.stream
+            .map(decodeFromTransferable[CatalogResults])
+            .unNone
+            .map(_.candidates)
+            .merge(
```
cquiroz, line 85: Looks longger than the original

cquiroz (commented):
Some comments

explore/src/main/scala/explore/config/GeneratedSequenceViewer.scala
rpiaggio: There's a version that does that: `useEffectResultWithDepsBy`

explore/src/main/scala/explore/targeteditor/AladinCell.scala
rpiaggio: Well, it was 20 lines before and now it's 16.

cquiroz (approved):
LGTM

PR #1862, opened 2022-05-28
Asterism model

cquiroz:
I'm introducing an `Asterism` class to replace `List[TargetWithId]` and also restrictting the asterism to have a non empty list.
The main reason is to have a `baseTarget` element which now is the first but it should instead be calculated as the center.
It also has some improvements to the UI specifically hiding the ttarget table if no targets are present

image

### DIFF
--- a/model/shared/src/main/scala/explore/model/Asterism.scala
+++ b/model/shared/src/main/scala/explore/model/Asterism.scala
@@ -9,9 +9,22 @@ import cats.syntax.all._
 import lucuma.core.model.Target
 import monocle._
 
-final case class Asterism(targets: NonEmptyList[TargetWithId]) {
+final case class Asterism private (private val targets: NonEmptyList[TargetWithId]) {
   def toSidereal: List[SiderealTargetWithId] =
     targets.traverse(_.toSidereal).foldMap(_.toList)
+
+  def asList: List[TargetWithId] = targets.toList
+
+  def add(t: TargetWithId): Asterism =
+    Asterism.isoTargets.reverse.modify(_ :+ t)(this)
+
+  def ids: NonEmptyList[Target.Id] = targets.map(_.id)
+
+  def remove(id: Target.Id): Option[Asterism] =
+    if (hasId(id)) {
+      val filtered = targets.filter(_.id =!= id)
+      Asterism.fromTargets(filtered)
+    } else this.some
 
   // This should be calculatedd from the other targets or manually overriden
   def baseTarget: TargetWithId = targets.head
@@ -34,8 +47,8 @@ object Asterism {
   def fromTargets(targets: List[TargetWithId]): Option[Asterism] =
     NonEmptyList.fromList(targets).map(Asterism.apply)
 
-  def of(targets: TargetWithId): Asterism =
-    Asterism(NonEmptyList.of(targets))
+  def one(targets: TargetWithId): Asterism =
+    Asterism(NonEmptyList.one(targets))
 
   def targetOptional(targetId: Target.Id): Optional[Option[Asterism], Target] =
     Optional[Option[Asterism], Target](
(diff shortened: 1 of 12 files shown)


model/shared/src/main/scala/explore/model/Asterism.scala
```diff
@@ -0,0 +1,56 @@
...
+  def hasId(id: Target.Id): Boolean = targets.exists(_.id === id)
+}
+
+object Asterism {
+  val isoTargets: Iso[NonEmptyList[TargetWithId], Asterism] =
+    Iso[Asterism, NonEmptyList[TargetWithId]](_.targets)(Asterism.apply).reverse
+
+  val targets: Lens[Asterism, NonEmptyList[TargetWithId]] = isoTargets.reverse.asLens
+
+  val fromTargets: Lens[NonEmptyList[TargetWithId], Asterism] = isoTargets.asLens
```
rpiaggio, line 28: I think we can omit these two lenses, they can be created at usage site from the Iso, if need be.

model/shared/src/main/scala/explore/model/Asterism.scala
```diff
@@ -0,0 +1,56 @@
...
+
+package explore.model
+
+import cats.Eq
+import cats.data.NonEmptyList
+import cats.syntax.all._
+import lucuma.core.model.Target
+import monocle._
+
+final case class Asterism(targets: NonEmptyList[TargetWithId]) {
```
rpiaggio, line 12: Would it make sense to internally use a `Map[Target.Id, Target]`? We should use a `Map` implementation that preserves insertion order, like `VectorMap` or  `TreeSeqMap`. However, I'm not sure there are "`NonEmpty`" versions of those out of the box.

The `targetOptional` method for building `Optional` wouldn't be needed then, it would just be the standard `index` optic.

model/shared/src/main/scala/explore/model/Asterism.scala
```diff
@@ -0,0 +1,56 @@
...
+      case _                       => Nil
+    }
+
+  def fromTargets(targets: List[TargetWithId]): Option[Asterism] =
+    NonEmptyList.fromList(targets).map(Asterism.apply)
+
+  def of(targets: TargetWithId): Asterism =
+    Asterism(NonEmptyList.of(targets))
+
+  def targetOptional(targetId: Target.Id): Optional[Option[Asterism], Target] =
```
rpiaggio, line 44: Is this lawful?

rpiaggio (approved):
I think we just explore the `Map` version. But I don't want that to hold the implementation, so LGTM.

model/shared/src/main/scala/explore/model/Asterism.scala
cquiroz: 👍

model/shared/src/main/scala/explore/model/Asterism.scala
cquiroz: Right, there is a NonEmptyMap but I don't think it cares about order. Maybe a NonEmptySet to ensure ids are unique?

model/shared/src/main/scala/explore/model/Asterism.scala
cquiroz: That's a good question, I moved this here because it was duplicated in two places across the code. It would be great to test it but it takes a param. Maybe I can put the checkAll inside a forAll

cquiroz: Actually you can do the test and seem to work, it is a bitslow but that's fine

model/shared/src/main/scala/explore/model/Asterism.scala
cquiroz: I'm going to punt in this question and perhaps address it later. In the mean time I'm hidding the detail that we are using `NonEmptyList` thus changing it will be easier in the future

PR #1863, opened 2022-05-28
If we are not focused on a target, use the base of the asterism

cquiroz:
This fixes the bug where the sky plot is blank if no target is selected


toddburnside (approved):
LGTM

PR #1864, opened 2022-05-28
Rename SkyPlot to ElevationPlot

cquiroz:
I propose renaming `SkyPlot` to `ElevationPlot`. We have a mix of both terms and in general we are talking about an elevation plot


toddburnside (approved):
If it is called an Elevation Plot in the UI, it might as well be the same in the code. 👍

PR #1869, opened 2022-05-31
Update for Shane's latest API reorganization.

toddburnside:


common/src/main/scala/explore/common/ScienceQueries.scala
```diff
@@ -124,9 +128,11 @@ object ScienceQueries {
       extends AnyVal {
     def toInput: GmosNorthLongSlitAdvancedConfigInput =
       GmosNorthLongSlitAdvancedConfigInput(
+        Input.ignore,
```
toddburnside, line 131: There are some new fields in the `*AdvancedConfigInput`s. I'm just ignoring them for now.

PR #1870, opened 2022-06-01
Obliterate Reusability

rpiaggio:


common/src/main/scala/explore/components/TileController.scala
```diff
@@ -105,7 +103,7 @@ object TileController {
       }
       // Store the current layout
       .useStateViewWithReuseBy((p, _, _) => updateResizableState(p.layoutMap))
-      .renderWithReuse { (p, debouncer, bn, currentLayout) =>
+      .render { (p, debouncer, bn, currentLayout) =>
```
cquiroz, line 106: In this case for example, it seems the Reuse does not complicate the code a lot

explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -27,89 +26,83 @@ import scala.scalajs.js
...
-  val view: ReuseView[RootModel]
+  val view: View[RootModel]
 ) extends ReactProps[ExploreLayout](ExploreLayout.component)
 
 object ExploreLayout {
   type Props = ExploreLayout
 
-  implicit val resolutionWithPropsReuse
-    : Reusability[ResolutionWithProps[Page, ReuseView[RootModel]]] =
+  implicit val resolutionWithPropsReuse: Reusability[ResolutionWithProps[Page, View[RootModel]]] =
```
toddburnside, line 37: Do we still need this?

common/src/main/scala/explore/components/TileController.scala
rpiaggio: I started by removing it from everywhere, and have a working version of Explore with that.

We can add it back in the places where it's needed, but I think we should have a uniform criterion.

For example, I notice the performance is not the same as before when resizing. We should probably add it back in the situations where it does have an impact.

cquiroz (approved):
LGTM

PR #1879, opened 2022-06-02
Update the elevation plot's site if the conf changes

cquiroz:


explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
```diff
@@ -60,6 +60,7 @@ object ElevationPlotSection {
     ScalaFnComponent
       .withHooks[Props]
       .useStateBy[Site](preferredSiteFor)
+      .useEffectWithDepsBy((p, _) => p)((_, s) => p => s.setState(preferredSiteFor(p)))
```
cquiroz, line 63: I thought this could be done with a plain `useEffectBy` but it started giving me errors in runtime

rpiaggio: `useEffect` without dependencies runs after each render. Changing state forces a rerender, so that creates an infinite loop (unless the rerender is stopped by Reusability, which we don't use anymore).

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
cquiroz: Right, that's what I'm seeing, thus the approach above is right

PR #1883, opened 2022-06-04
Persist elevation plot settings

cquiroz:


model/shared/src/main/scala/explore/model/ElevationPlotOptions.scala
```diff
@@ -0,0 +1,28 @@
...
+
+object ElevationPlotOptions {
+  val site  = Focus[ElevationPlotOptions](_.site)
+  val range = Focus[ElevationPlotOptions](_.range)
+  val time  = Focus[ElevationPlotOptions](_.time)
+
+  val Default =
+    ElevationPlotOptions(Site.GS, PlotRange.Night, TimeDisplay.Site)
+
+  implicit val targetVisualOptionsEq: Eq[ElevationPlotOptions] =
```
toddburnside, line 26: Someday names won't be necessary. But for now this should probably match the class. 😄

toddburnside (approved):
LGTM

model/shared/src/main/scala/explore/model/ElevationPlotOptions.scala
cquiroz: Some day as in scala 3 auto derivation?

toddburnside: Well, that too. But I don't think `given`s have to have names in scala 3

PR #1894, opened 2022-06-06
Show on the UI when we can't read the catalog

cquiroz:
image

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinToolbar.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinToolbar.scala
@@ -27,7 +27,7 @@ import react.semanticui.sizes._
 final case class AladinToolbar(
   fov:                 Fov,
   current:             Coordinates,
-  loadingGSCandidates: Either[String, Boolean],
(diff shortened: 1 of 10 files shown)


explore/src/main/scala/explore/targeteditor/AladinToolbar.scala
```diff
@@ -27,7 +27,7 @@ import react.semanticui.sizes._
 final case class AladinToolbar(
   fov:                 Fov,
   current:             Coordinates,
-  loadingGSCandidates: Boolean,
+  loadingGSCandidates: Either[String, Boolean],
```
toddburnside, line 30: Not a big deal, but this seems like it could be `Option[Boolean]`.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinToolbar.scala
cquiroz: I guess that's right. I'm not really showing the error msg

PR #1895, opened 2022-06-07
Added onBlur to store target name change

diego-ge:
It was added an onBlur callback to store the target name.
Target name is being saved to DB on search and blur, but left side menu is not being updated properly.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/SearchForm.scala
+++ b/explore/src/main/scala/explore/targeteditor/SearchForm.scala
@@ -24,11 +24,15 @@ import react.semanticui.collections.form.Form.FormProps
 import react.semanticui.collections.form._
 import react.semanticui.elements.label.LabelPointing
 import react.semanticui.shorthand._
+import cats.data.ValidatedNec
+import cats.data.Validated.Valid
+import cats.data.Validated.Invalid
 
 import scalajs.js.JSConverters._
 
 final case class SearchForm(
   id:          Target.Id,
+  targetView:  View[Target.Sidereal],
   name:        NonEmptyString,
   searching:   View[Set[Target.Id]],
   searchAndGo: SearchCallback => Callback
@@ -119,6 +123,11 @@ object SearchForm {
           disabled = disabled,
           errorClazz = ExploreStyles.InputErrorTooltipBelow,
           errorPointing = LabelPointing.Above,
+          onBlur = (r: ValidatedNec[NonEmptyString, NonEmptyString]) =>
+            r match {
+              case Valid(name)   => props.targetView.zoom(Target.Sidereal.name).set(name)
+              case Invalid(name) => Callback.log(name)
+            },
           onTextChange = _ => $.setStateL(State.searchError)(none),
           onValidChange = valid => $.setStateL(State.searchEnabled)(valid),
           icon = searchIcon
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
```diff
@@ -273,6 +273,7 @@ object SiderealTargetEditor {
                   // SearchForm doesn't edit the name directly. It will set it atomically, together
                   // with coords & magnitudes from the catalog search, so that all 3 fields are
                   // a single undo/redo operation.
+                  allView,
```
cquiroz, line 276: maybe `allView` can replace the use of `targetView` as well, but we can take a look later

explore/src/main/scala/explore/targeteditor/SearchForm.scala
```diff
@@ -119,6 +123,11 @@ object SearchForm {
           disabled = disabled,
           errorClazz = ExploreStyles.InputErrorTooltipBelow,
           errorPointing = LabelPointing.Above,
+          onBlur = (r: ValidatedNec[NonEmptyString, NonEmptyString]) =>
+            r match {
+              case Valid(name)   => props.targetView.zoom(Target.Sidereal.name).set(name)
+              case Invalid(name) => Callback.log(name)
+            },
```
toddburnside, line 130: Could this be done with a `withOnMod` on the view being used for the `value`?

toddburnside: I don't follow...

explore/src/main/scala/explore/targeteditor/SearchForm.scala
```diff
@@ -119,6 +123,11 @@ object SearchForm {
           disabled = disabled,
           errorClazz = ExploreStyles.InputErrorTooltipBelow,
           errorPointing = LabelPointing.Above,
+          onBlur = (r: ValidatedNec[NonEmptyString, NonEmptyString]) =>
+            r match {
+              case Valid(name)   => props.targetView.zoom(Target.Sidereal.name).set(name)
+              case Invalid(name) => Callback.log(name)
```
toddburnside, line 129: Is there a value to logging just the name when it is invalid?

explore/src/main/scala/explore/targeteditor/SearchForm.scala
cquiroz: In the props we don't have a `View[NonEmptyString]` of the name if we wanted to set it, only the concrete name

toddburnside: I was thinking of something like `value = View.fromState($).zoom(State.searchTerm).withOnMod(name => props.targetView.zoom(Target.Sidereal.name).set(name))`. Or something like that. I think it does the same thing as you have here. IIRC, the view is only changed on Valid so you don't need to worry about that. But maybe it isn't any better?

cquiroz: I see, that's right  we could that

toddburnside (approved):
LGTM

diego-ge:
The just the name field will be put in the undo stack and will be sent to the back end as it was done with the onBlur.
In case of a submit and a successful search the whole target will be put in the undo stack and will be sent to the backend.

If the name is modified and the user push enter or clicks on the submit button, both events (onBlur and submit) will be triggered, then the undo stack will have the whole target, as the last change and before that, will also have just the target name.

PR #1906, opened 2022-06-09
Store posangle in the db

cquiroz:


explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -81,16 +80,18 @@ object ConfigurationPanel {
         val showAdvancedCB: Option[Callback] =
           optModeView.get.map(_ => showAdvanced.set(true))
 
+        val posAngleView: View[Option[PosAngleConstraint]] =
+          props.scienceData.undoableView(ObservationData.posAngleConstraint)
```
cquiroz, line 84: I followed @rpiaggio recommendation about `undoableView` and seems to work quite well

common/src/main/scala/explore/schemas/odb/implicits.scala
```diff
@@ -240,6 +242,28 @@ object implicits {
       }
   }
 
+  implicit class PosAngleConstraintOps(val p: PosAngleConstraint) extends AnyVal {
+    def toInput: PosAngleConstraintInput =
```
cquiroz, line 246: @toddburnside I guess this is why we could remove the encoder?

swalker2m: That's my understanding at least.  This is doing what the encoder would have done.  Deleting the constraint when necessary is handled at a higher level?

cquiroz: In explore it is an `Option[PosAngl]` when set to none (In the UI Unconstrained) it gets removed on the db via an `Input.orUnassign`

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/model/syntax/package.scala
```diff
@@ -3,29 +3,30 @@
...
+      case PosAngleOptions.AllowFlip           => PosAngleConstraint.AllowFlip(a).some
+      case PosAngleOptions.AverageParallactic  => PosAngleConstraint.AverageParallactic.some
+      case PosAngleOptions.ParallacticOverride => PosAngleConstraint.ParallacticOverride(a).some
+      case PosAngleOptions.Unconstrained       => none
     }
   }
 
-  implicit class PosAngleOps(val pa: PosAngle) extends AnyVal {
+  implicit class PosAngleOps(val pa: Option[PosAngleConstraint]) extends AnyVal {
     def toPosAngleOption: PosAngleOptions = pa match {
```
rpiaggio, line 24: Ultra-picky comment: maybe call this `toPosAngleOptions` ? At first glance of `toPosAngleOption` I was expecting the return type to be `Option[PosAngle]`.

common/src/main/scala/explore/schemas/odb/implicits.scala
toddburnside: > @toddburnside I guess this is why we could remove the encoder?

Yes, exactly. We can never just encode directly to the UI because the mutations require the *Inputs.

PR #1909, opened 2022-06-10
Edit dithers and offsets.

rpiaggio:
Can be edited in Advanced Configuration Panel and within Sequence Generator popup.

image

image

Also, the generated sequence is re-queried when any observation parameter changes, locally or remotely.

Finally, I rearranged code a bit:
- Put all scattered `ValidFormatInput` into a `validators.scala`.
- Renamed `TruncatedPA` to `TruncatedAngle`.


common/src/main/scala/explore/model/validators.scala
```diff
@@ -0,0 +1,67 @@
...
+import lucuma.core.math.Angle
+import lucuma.core.math.Axis
+import lucuma.core.math.Offset
+import lucuma.ui.optics.ValidFormatInput
+
+import scala.util.Try
+
+object validators {
+
+  // TODO TEST ALL THESE!!!!
```
rpiaggio, line 26: Testing will be done in upcoming PR. We need to first publish a testkit for lucuma-ui's ValidFormat.

PR #1917, opened 2022-06-13
Proposal Class editing via API

toddburnside:
A couple of things are left to do:

1. If there is no proposal for the program, logged in users need to have the ability to create a proposal, while guest users will not. I will read the specs and work with Andy to determine how this should work.
2. The textbox for the `abstract` field has some problems. Right now it updates the view with every keystroke, which also entails adding an entry to the undo for every keystroke. I'm going to create a `FormTextBoxEV` in `lucuma-ui` rather than dealing with all the state here. The CSS also needs to be modified so it's focused state matches that of the inputs. I will do that when I update the TextBox.


common/src/main/scala/explore/model/reusability.scala
```diff
@@ -86,5 +89,10 @@ object reusability {
   implicit val agsParamsReuse: Reusability[AgsParams]                           = Reusability.byEq
   implicit val agsAnalysisReuse: Reusability[AgsAnalysis]                       = Reusability.byEq
 
+  implicit val partnerSplitsReuse: Reusability[SortedMap[Partner, IntPercent]] =
+    Reusability.by((_: SortedMap[Partner, IntPercent]).toMap)(Reusability.map)
+  implicit val proposalClassReuse: Reusability[ProposalClass]                  = Reusability.derive
+  implicit val proposalReuse: Reusability[Proposal]                            = Reusability.derive
```
rpiaggio, line 95: Can we use `byEq`?

common/src/main/scala/explore/schemas/odb/implicits.scala
```diff
@@ -264,6 +267,60 @@ object implicits {
       }
   }
 
+  implicit class NonNegDurationOps(val nnd: NonNegDuration) {
```
rpiaggio, line 270: `extends AnyVal` ?

explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -0,0 +1,495 @@
...
+  def renderFn(
+    programId:         Program.Id,
+    proposal:          View[Proposal],
+    undoStacks:        View[UndoStacks[IO, Proposal]],
+    totalHours:        View[Hours],
+    minPct2:           View[IntPercent],
+    proposalClassType: View[ProposalClassType],
+    showModal:         View[Boolean],
+    splitsList:        View[List[PartnerSplit]],
+    // state:         View[State],
```
rpiaggio, line 374: Commented code

common/src/main/scala/explore/model/reusability.scala
toddburnside: Ah. That's right. We should be able to.

common/src/main/scala/explore/schemas/odb/implicits.scala
toddburnside: Unfortunately, the compiler does not allow that. `value class may not wrap another user-defined value class`

PR #1922, opened 2022-06-14
npm dependency upgrades

toddburnside:


toddburnside:
Upgrading stylelint to version 14 appears to be impractical since they stopped supporting less natively.

PR #1926, opened 2022-06-15
Use FormTextAreaEV for proposal abstract and fix textarea styling

toddburnside:


explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -211,6 +210,9 @@ object ProposalEditor {
 
     val abstractView = abstractAligner.view(_.orUnassign)
 
+    val NESIso: Iso[Option[NonEmptyString], String] =
+      Iso[Option[NonEmptyString], String](_.foldMap(_.value))(s => NonEmptyString.from(s).toOption)
```
toddburnside, line 214: This seems to be a lawful ISO to me, but I haven't tested it. I'm wondering if it might be generally useful, but I'm not sure where it should live, if so.

rpiaggio: `explore.optics.ModelOptics`?

Or we could generalize it as a `Iso[Option[A Refined P]], A]` for any `P` and put it into `lucuma.core.optics`, where we already have a `refinedPrism`.

rpiaggio (approved):
LGTM with a couple of comments

explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -350,13 +352,12 @@ object ProposalEditor {
...
-        FormTextArea( // TODO: Every keystroke generates a new Undo....but no blur event
+        FormTextAreaEV(
+          id = "abstract",
           label = "Abstract",
           rows = 10,
-          value = abstractView.get.foldMap(_.value),
-          onChangeE = (_: TextArea.ReactChangeEvent, tap: TextArea.TextAreaProps) =>
-            abstractView.set(NonEmptyString.from(tap.value.asInstanceOf[String]).toOption)
-        ).addModifiers(Seq(^.id := "abstract"))
+          value = abstractView.zoom(NESIso.asLens)
```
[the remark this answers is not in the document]
toddburnside: Good to know.

explore/src/main/scala/explore/proposal/ProposalEditor.scala
toddburnside: I don't think it can be generalized that much. With NonEmptyString, the only invalid string is the empty one. With other refinements, there can be more than one A that would be invalid when refined.

rpiaggio: Yup, you're right, a general `Iso` is not actually possible.

PR #1930, opened 2022-06-17
Small fixes related to saving obsTime

cquiroz:
Two fixes:
* Reduces the amount of data stored in undo as we are not supporting undo for vizTime
* Properly set the viz time as now if not defined in the db

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
+++ b/explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
@@ -267,7 +267,9 @@ object SiderealTargetEditor {
                 props.uid,
                 props.id,
                 props.posAngle,
-                props.vizTime,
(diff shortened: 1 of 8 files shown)


explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
```diff
@@ -267,7 +267,9 @@ object SiderealTargetEditor {
                 props.uid,
                 props.id,
                 props.posAngle,
-                props.vizTime,
+                // This may need to go into a effect but it complicates the code quite a bit, and
+                // we need precission of months/years
```
rpiaggio, line 271: You can `useEffectResult` and then `potRender` (or just `.toOption`).

cquiroz: Excellent tip. Updated

rpiaggio (approved):
LGTM

PR #1951, opened 2022-06-22
Rework on the way visualizations are created with a pure React solution

cquiroz:


toddburnside:
> There is a nice reduction on size mostly by getting rid of the `svg.js` library

Excellent! 🎉

toddburnside (approved):
LGTM - It's nice to have one fewer JS libraries!

PR #1984, opened 2022-07-06
Add types to arbitraries

cquiroz:
This is coming from the scala 3 port


toddburnside (approved):
LGTM. But, why does scala 3 require the type ascriptions? It seems strange.

PR #1986, opened 2022-07-06
Update itc schema to get the version

cquiroz:
Read the version from the itc.
This is mostly a sanity check now that ITC runs on scala 3


rpiaggio (approved):
LGTM

PR #2005, opened 2022-07-12
Display coordinates in target table at observation

cquiroz:
For a review please focus on the second commit

image

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/TargetTable.scala
+++ b/explore/src/main/scala/explore/targeteditor/TargetTable.scala
@@ -5,7 +5,9 @@ package explore.targeteditor
 
 import cats.effect.IO
 import cats.syntax.all._
+import crystal.Ready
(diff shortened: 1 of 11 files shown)


explore/src/main/scala/explore/targeteditor/TargetTable.scala
```diff
@@ -110,9 +108,16 @@ object TargetTable {
             .BaseColumnBuilder(TargetTable)(_.target.some)
             .allColumns
       }
+      // If vizTime is not set, change it to now
+      .useEffectResultWithDepsBy((p, _) => p.vizTime) { (_, _) => vizTime =>
```
toddburnside, line 112: Could this be `useMemoBy` instead of `useEffectResultWithDepsBy`?

toddburnside (approved):
LGTM w/ one comment/question.

PR #2006, opened 2022-07-12
Improvements to pwa caching

cquiroz:
It seems we should be able to enable PWA in production.
Also add more caching for ITC and aladin

### DIFF
--- a/explore/src/main/webapp/main.jsx
+++ b/explore/src/main/webapp/main.jsx
@@ -15,10 +15,7 @@ import { registerSW } from "virtual:pwa-register";
 
 // Setup the Service Worker
 if ("serviceWorker" in navigator
-  // && !/localhost/.test(window.location) && !/local.lucuma.xyz/.test(window.location)) {
-    // Let's not do this in production yet
-  && !/local.lucumay.xyz/.test(window.location)
-  && !/explore.gemini.edu/.test(window.location) && !/gemini.edu/.test(window.location)) {
+  && !/local.lucumay.xyz/.test(window.location)) {
     ExplorePWA.runServiceWorker();
 }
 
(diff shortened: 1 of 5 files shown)


explore/src/main/webapp/main.jsx
```diff
@@ -15,10 +15,7 @@ import { registerSW } from "virtual:pwa-register";
 
 // Setup the Service Worker
 if ("serviceWorker" in navigator
-  // && !/localhost/.test(window.location) && !/local.lucuma.xyz/.test(window.location)) {
-    // Let's not do this in production yet
-  && !/local.lucumay.xyz/.test(window.location)
-  && !/explore.gemini.edu/.test(window.location) && !/gemini.edu/.test(window.location)) {
+  && !/local.lucumay.xyz/.test(window.location)) {
```
rpiaggio, line 18: Is this to avoid PWA in development? There is (and was) an extra `y` there, which leads me to believe this is not doing what we wanted it to do.

workers/src/main/scala/workers/CatalogCache.scala
```diff
@@ -41,7 +41,7 @@ import lucuma.core.math.Wavelength
 import explore.model.Constants
 
 trait CatalogQuerySettings {
-  val proxy = uri"https://lucuma-cors-proxy.herokuapp.com"
+  val proxy = uri"https://cors-proxy.lucuma.xyz"
```
rpiaggio, line 44: Nice

rpiaggio (approved):
LGTM

explore/src/main/webapp/main.jsx
cquiroz: ah interesting I'll fix it, to use pwa in dev you need to also setup the pwa plugin in dev mode so this maybe redundant

PR #2007, opened 2022-07-12
Stop manage programs dialog from flickering when toggling "show deleted"

rpiaggio:
Moved much of the logic of `ProgramsPopup` into `ProgramTable`.

Also required bug fixes in `clue` and `crystal`.


toddburnside (approved):
LGTM

PR #2010, opened 2022-07-12
Filter the allowed configurations depending on the asterism declination

cquiroz:


model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
```diff
@@ -360,7 +370,12 @@ final case class SpectroscopyModesMatrix(matrix: List[SpectroscopyModeRow]) {
         wavelength.forall(w => w >= r.minWavelength.w && w <= r.maxWavelength.w) &&
         resolution.forall(_ <= r.resolution) &&
         coverage.forall(_ <= r.wavelengthCoverage) &&
-        slitWidth.forall(_.toMicroarcseconds <= r.slitLength.size.toMicroarcseconds)
+        slitWidth.forall(_.toMicroarcseconds <= r.slitLength.size.toMicroarcseconds) &&
+        declination.forall {
+          case d if d.toAngle.toSignedDoubleDegrees < -40.0 => r.instrument.site === Site.GS
+          case d if d.toAngle.toSignedDoubleDegrees > 30.0  => r.instrument.site === Site.GN
```
rpiaggio, line 376: Maybe the cutoff declinations should be in the `Site` enum instead of hardcoded here?

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
cquiroz: This is more of a UI property that may not be appropriate to go to core, but perhaps we should define it on explore's model

PR #2013, opened 2022-07-13
Update Wavelength Inputs to only allow valid Wavelengths

toddburnside:


cquiroz (approved):
Uh, looks like a nice simplification

PR #2017, opened 2022-07-15
Disable linting of `less` files in CI

toddburnside:
I will look into switching to linting of scss files. If it isn't too painful, I'll implement that in a later PR. If it is too painful, I'll removing css linting altogether.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -280,7 +280,7 @@ jobs:
         uses: wearerequired/lint-action@v1.11.1
         with:
           stylelint: true
-          stylelint_extensions: css,less
+          stylelint_extensions: css
           stylelint_args: common/src/main/webapp/less
           github_token: ${{ secrets.GITHUB_TOKEN }}
           stylelint_dir: common/src/main/webapp/less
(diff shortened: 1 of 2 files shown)


cquiroz (commented):
I'd think you need to remove it form the CI tasks too?

toddburnside:
> 



> I'd think you need to remove it form the CI tasks too?

Which CI tasks are you referring to? Do you mean remove the linting actions altogether, not just eliminate `.less` from the extension list?
As far as I can tell, it seems to be complaining that it didn't find any files to lint. But shouldn't `vars.css` be there?

cquiroz:
Like this one
https://github.com/gemini-hlsw/explore/blob/master/build.sbt#L247=

toddburnside:
> Like this one

I thought I could just remove linting `less` until I looked at scss more. But, I'm just going to rip it out. I can always pull it back in.

rpiaggio (approved):
Hope we can switch to SCSS and have linting for it. In the meantime, this LGTM.

cquiroz (approved):
LGTM

PR #2018, opened 2022-07-15
Add targets in target tab.

rpiaggio:
Allow creating Sidereal targets from the Target tab.

Also:
- Renamed `TargetGroup` to `TargetWithObs`, and similarly all references to `targetGroup`s or derived names.
- Introduced `Focused` class to replace `(focusedObsSet, focusedTarget)`.
- Converted `AsterismGroupObsList` to functional component (that's also where the new functionality was added).
- Removed `less` for stylelint.

image


explore/src/main/scala/explore/observationtree/AsterismGroupObsListActions.scala
```diff
@@ -229,42 +236,29 @@ object AsterismGroupObsListActions {
...
-                .toSortedMap(_.obsIds)
-              agwo.copy(targetGroups = newTargetGroups, asterismGroups = newAsterismGroups)
-            }
+  def targetExistence(targetId: Target.Id, setPage: Option[Target.Id] => IO[Unit])(implicit
+    c:                          TransactionalClient[IO, ObservationDB]
+  ): Action[AsterismGroupsWithObs, Option[TargetWithObs]] =
+    // can switch appCtx to TransactionalClient when logging not required
+    Action[AsterismGroupsWithObs, Option[TargetWithObs]](
+      getter = (agwo: AsterismGroupsWithObs) => agwo.targetsWithObs.get(targetId),
+      setter = (otwo: Option[TargetWithObs]) =>
```
toddburnside, line 245: Oi. I can't understand this stuff, anymore. 🙄 I'll have to trust that you tested it...

explore/src/main/scala/explore/observationtree/AsterismGroupObsListActions.scala
```diff
@@ -229,42 +236,29 @@ object AsterismGroupObsListActions {
...
-        otg.fold(
-          appCtx.logger.error("Creating targets in AsterismGroupObsListActions not yet supported")
-        )(_ => deleteTarget(targetId).void),
+      // DB creation is performed beforehand, in order to get id
+      onSet =
+        (_, otg) => otg.fold(deleteTarget(targetId) >> setPage(none))(_ => setPage(targetId.some)),
       onRestore = (_, otg) =>
-        otg.fold(deleteTarget(targetId).void) { _ =>
-          undeleteTarget(targetId).void
+        otg.fold(deleteTarget(targetId) >> setPage(none)) { _ =>
```
toddburnside, line 260: Does the target have to focused for us to delete it? If not, should we just leave the page as it is so we don't unfocus something else? If it was focused and is deleted, that should be fixed elsewhere.

toddburnside (approved):
LGTM, with one question.

explore/src/main/scala/explore/observationtree/AsterismGroupObsListActions.scala
rpiaggio: It took me a bit to remember how to use this.

Basically, these are instructions on how to get/set a `Target` with a certain `targetId` in an `AsterismGroupsWithObs` (which is the structure that holds asterisms, targets and observations).

Passing/getting `None` means deletion/absence, wheareas `Some` means adding/existence.

When a target is deleted or added with associated observations, the asterisms of those observations have to be adjusted.

toddburnside:
Does this just add an `empty` target? Couldn't we could reuse the target selection popup when they click the `+ Sidereal` button and just exclude the program targets from the search?

PR #2026, opened 2022-07-18
Fix brightnessValidWedge, again.

rpiaggio:


cquiroz (approved):
LGTM

model/shared/src/main/scala/explore/model/display.scala
```diff
@@ -78,11 +78,13 @@ trait DisplayImplicits {
...
-    // We don't want scientific notation to kick in for magnitude units: we want to keep 2 integers.
+    // We don't want scientific notation to kick in for magnitude units.
+    // We assume it's magnitudes when x.abs >= 0.00001, since other units are usually
+    // expressed in the order of e-18 or e-19.
     // We could make the format depend on the units, but that may be confusing for users
     // in case they want to type the value and then change the units.
-    if (0 <= x.scale && x.scale <= 3 && x.precision <= 5)
-      x.toString
+    if (x.abs >= 0.00001)
+      "%.3f".format(x)
```
cquiroz, line 87: Maybe 2 decimals is enough?

rpiaggio: The examples in the inception document use 3 decimal places, but we can change it if needed. @andrewwstephens ?

image

rpiaggio: I'll merge and we can change this afterwards.

andrewwstephens: :+1:  most photometric catalogs give 3 decimal places so let's stick with 3.

PR #2032, opened 2022-07-20
Visualization on full screen

cquiroz:
Aladin's full screen method is somewhat primitive and doesn't play well with visualization.
This PR adds an explore specific method to extend the size of the aladin cell to cover the whole tile

The state of full screen is stored on the user preferences

https://user-images.githubusercontent.com/3615303/180058494-c4cda69a-f6f3-4479-9b09-9952ad6a5a0d.mp4


rpiaggio (approved):
LGTM

PR #2039, opened 2022-07-21
Lint sass files

toddburnside:


rpiaggio (approved):
LGTM, thanks for taking care of this

common/src/main/webapp/less/variables-light.less
```diff
@@ -62,6 +71,20 @@
   --scrollbar-background: lighten(@lightGrey, 5%);
   --scrollbar-background-thumb: darken(@lightGrey, 15%);
   --scrollbar-background-thumb-hover: darken(@lightGrey, 20%);
+  --hud-color: hsla(
```
cquiroz, line 74: Is it detecting when a variable is defined in dark but not light?

common/src/main/webapp/sass/visualization.scss
```diff
@@ -25,12 +25,12 @@
   }
 
   .gmos-probe-arm {
-    stroke: hsl(0, 67.9%, 41.6%);
+    stroke: hsl(0deg 67.9% 41.6%);
```
cquiroz, line 28: Didn't know this notation

common/src/main/webapp/sass/visualization.scss
```diff
@@ -189,43 +190,49 @@
...
+  @include aladin-button-mixin;
+
   right: 10px;
   top: 50%;
   display: flex;
   flex-direction: column;
 }
 
-.ui.icon.button:not(.animated):not(.compact):not(.labeled).explore-aladin-fullscreen-button {
+.ui.icon.button:not(.animated, .compact, .labeled).explore-aladin-fullscreen-button {
```
cquiroz, line 216: Nice

cquiroz (approved):
LGTM

common/src/main/webapp/sass/visualization.scss
toddburnside: When `fix` switched from hex codes, it used this notation.

common/src/main/webapp/less/variables-light.less
toddburnside: It can only know that if the color is being used in the scss files being linted. This variable is used in `vizualization.scss`, so `lint-light` caught the omission.

common/src/main/webapp/sass/visualization.scss
toddburnside: Once again, the product of `fix`. 😄

common/src/main/webapp/less/variables-light.less
toddburnside: Unfortunately, since we are no longer linting `less` files, we will no longer know that for `styles.less`. 🤷

cquiroz: ok but still that seems very good

package.json
```diff
@@ -55,21 +55,21 @@
     "process": "^0.11.10",
     "rollup-plugin-visualizer": "^5.7.1",
     "sass": "^1.53.0",
-    "stylelint": "^13.13.1",
+    "stylelint": "^14.9.1",
```
cquiroz, line 58: 👍

PR #2043, opened 2022-07-22
Add "λ" readonly values to advanced config panel

toddburnside:


explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -181,11 +281,42 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
...
+            slitWidth = fpa,
+            resolution = res,
+            coverage = cov.map(_.micrometer.toValue[BigDecimal].toRefined[Positive])
+          )
+        }
+      }
+      // Try to find the readonly data from the spectroscopy matrix
+      .useMemoBy { (props, _, _, _) =>
+        // TODO: reuse on rows, not requirements? And narrow advanced comparison?
+        (props.scienceModeBasic, props.scienceModeAdvanced.get, props.spectroscopyRequirements)
```
toddburnside, line 307: If this looks OK to Andy, I should probably not base the reuse here on all of scienceModeAdvanced, but only the fields we use for the table lookup. 
I'm not sure if it it is worth basing the reuse on the `rows` from the previous `useMemoBy`. We'd need to add `Eq` instances to a whole bunch of classes in `SpectroscopyModesMatrix.scala`, and then it seems like we might as well just do the search rather than checking to see if the lists are equal. Thoughts?

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -160,6 +172,94 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
...
+    case _ => None
+  }
+
+  private def findMatrixData(
+    basic:        S,
+    advanced:     T,
+    requirements: SpectroscopyRequirementsData,
+    rows:         List[SpectroscopyModeRow]
+  ): Option[ReadonlyData] =
+    rows.flatMap(row => findMatrixDataFromRow(basic, advanced, requirements, row)).headOption
```
rpiaggio, line 261: I think we can avoid going through the whole list by using `collectFirst` (and building `ReadonlyData` afterwards).

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
rpiaggio: I'm not sure I understand what you are suggesting, but `useMemo` already returns a `Reusable` (which will be based in its own dependencies).

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
toddburnside: Good catch.

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
toddburnside: Right. But, instead of checking the filtered list or spectroscopy rows for reusability now, I'm just checking the spectroscopy requirements upon which is is based. Currently, there is no Eq for any of the many types in SpectroscopyModesMatrix, and I'm not sure it's worth scanning the filtered table for equality (reusability) each time.

toddburnside: Oh, I see what you are saying. I don't need a reuse of SpectroscopyModeRow, because the Reuse returned by `useMemo` is based on the dependecies of that `useMemo`, not the value of the memo. Nice.

toddburnside: Pushed a commit to utilize that.

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -76,7 +88,7 @@ sealed trait AdvancedConfigurationPanel[T <: ScienceModeAdvanced, S <: ScienceMo
 
 sealed abstract class AdvancedConfigurationPanelBuilder[
   T <: ScienceModeAdvanced,
-  S <: ScienceModeBasic,
+  S <: ScienceModeBasic: Eq,
```
rpiaggio, line 91: Do we use the `Eq` besides deriving `Reusability`? If not, can we require the `Reusability` directly? It may already be defined.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
toddburnside: We only use it for Reusability. Reusability is already defined on `ScienceModeAdvanced`, but not on all if it's subtypes. I guess I could define it for the subtypes, and then summon it here.

PR #2046, opened 2022-07-25
Add new empty target to observation.

rpiaggio:
Also:

- Some colors (all "Add" buttons as positive and all "Cancel" as negative).
- Differentiate when a new target is being created or linked to.
- Even spacing in tile title components.

image


toddburnside (approved):
LGTM

PR #2055, opened 2022-07-27
Exclude targets too close to the science target from the ags candidates

cquiroz:
The image below show it somehow in action

image


cquiroz:
Switched to show the discarded targets after consultation with users
image

PR #2065, opened 2022-07-27
Upgrade to new crystal

rpiaggio:
Use new version of `useStream` hooks that return `PotOption`.

### DIFF
--- a/common/src/main/scala/explore/utils/package.scala
+++ b/common/src/main/scala/explore/utils/package.scala
@@ -9,6 +9,7 @@ import cats.syntax.all._
 import clue.data._
 import clue.data.syntax._
 import crystal.Pot
+import crystal.PotOption
 import crystal.react.View
 import crystal.react.reuse._
 import eu.timepit.refined.types.string.NonEmptyString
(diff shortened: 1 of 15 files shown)


common/src/main/scala/explore/utils/package.scala
```diff
@@ -66,13 +67,14 @@ package object utils {
...
-  val DefaultPendingRender: Long => VdomNode = _ => Loader(active = true)
+  // TODO All the "potRender" methods should go in lucuma-ui
+  val DefaultPendingRender: VdomNode = Loader(active = true)
 
   val DefaultErrorRender: Throwable => VdomNode = t => Message(error = true)(t.getMessage)
 
   def potRenderWithReuse[A](
     valueRender:   A ==> VdomNode,
-    pendingRender: Long ==> VdomNode = Reuse.always(DefaultPendingRender),
+    pendingRender: => Reuse[VdomNode] = Reuse.always(DefaultPendingRender),
```
cquiroz, line 77: `Reuse`?

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -116,21 +117,17 @@ object AladinCell extends ModelOptics {
...
-        (props, _, _, _, _, _) => { case (c, _) =>
-          c.value
-            .map(candidatesAwait =>
-              candidatesAwait >>
-                props.ctx.worker.postTransferrable(
-                  CatalogRequest(props.target.get, props.obsConf.vizTime)
-                )
-            )
-            .orEmpty
+      .useEffectWithDepsBy((p, _, _, _, _, candidates) => (candidates, p.obsConf.vizTime))(
```
cquiroz, line 120: No await?

cquiroz (approved):
LGTM with some question

common/src/main/scala/explore/utils/package.scala
rpiaggio: I don't think we are using this method anymore. All parameters use `Reuse` here (`A ==> B` is type alias for `Reuse[A => B]`. Since we removed the `Long` for `Pending`, this would be the equivalent.

Anyway, maybe we can remove this function, IDK. It's a utility method that we may want to use some day.

explore/src/main/scala/explore/targeteditor/AladinCell.scala
rpiaggio: Right, no more await. React will run this effect every time `candidates` change. We only do something when it's in `ReadyNone` state, which means that the fiber has started, but nothing has been received on the stream yet.

toddburnside (approved):
LGTM

PR #2067, opened 2022-07-28
Topic and Resource[Stream] in WebWorker.

rpiaggio:
This PR brings changes that @cquiroz has done in another branch, where `WebWorkerF` uses now a `Topic` instead of a `Channel`, so that it can safely have multiple concurrent subscribers.

It also changes its interface so that the event stream must be accessed as a `Resource`. This allows us to synchronize upon its initialization and make sure we are actually processing the `Topic` messages before we send a query to the worker.

### DIFF
--- a/common/src/main/scala/workers/WebWorkerF.scala
+++ b/common/src/main/scala/workers/WebWorkerF.scala
@@ -11,7 +11,7 @@ import cats.effect.std.Dispatcher
 import cats.syntax.all._
 import explore.model.boopickle._
 import fs2.Stream
-import fs2.concurrent.Channel
+import fs2.concurrent.Topic
 import org.scalajs.dom
 
 import scala.scalajs.js
@@ -45,9 +45,9 @@ trait WebWorkerF[F[_]] {
   def terminate: F[Unit]
 
   /**
-   * Streams of events from the worker
+   * Resource with Stream of events from the worker
    */
-  def stream: Stream[F, dom.MessageEvent]
+  def streamResource: Resource[F, Stream[F, dom.MessageEvent]]
 }
 
 object WebWorkerF {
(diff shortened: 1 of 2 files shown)


common/src/main/scala/workers/WebWorkerF.scala
```diff
@@ -57,24 +57,31 @@ object WebWorkerF {
...
 
-          def postTransferrable(buffer: Int8Array): F[Unit] =
+          override def postTransferrable(buffer: Int8Array): F[Unit] =
             Sync[F].delay(worker.postMessage(buffer, js.Array(buffer.buffer: dom.Transferable)))
 
-          val terminate: F[Unit] = Sync[F].delay(worker.terminate())
+          override val terminate: F[Unit] = Sync[F].delay(worker.terminate())
 
-          val stream: Stream[F, dom.MessageEvent] = channel.stream
+          // We want to use `subscribeAwait` instead of `subscribe`:
```
cquiroz, line 78: Thanks for documenting this

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -89,73 +89,67 @@ object AladinCell extends ModelOptics {
       // flag to trigger centering. This is a bit brute force but
       // avoids us needing a ref to a Fn component
       .useStateView(false)
-      // to get faster reusability use a serial state, rather than check every candidate
-      .useSerialState(List.empty[GuideStarCandidate])
```
rpiaggio, line 93: It seems we don't need to do this anymore since we are not using `Reusability` and that we can use the `useStreamResource` result directly.

cquiroz (approved):
This is really nice, thanks a lot

explore/src/main/scala/explore/targeteditor/AladinCell.scala
cquiroz: Don't we need it for the `useMemoBy` below?

rpiaggio: Yes, you are right, I missed that.

However, I was just about to roll this back but I'm doubting if it is really necessary: the list of candidates is only updated once per query to the worker, right?

cquiroz: Right it should be only once. the list can be fairly large, omega cen has 17k target and doing a Reusability comparison would need to traverse the 17k items right?
That's why I thought about serial state but maybe we don't need it

rpiaggio: Wait, if one of the other `useMemo` dependencies changes then, yes, the canonic list reusability will be used for candidates.

I'll roll back.

PR #2068, opened 2022-07-29
Load configuration matrix on worker

cquiroz:
We want to move some ITC functionality to be done in the web worker
This is the first step where the configuration matrix is loaded on the web worker.
It doesn't add the functionality but adds a set of changes to improve the way ITC calculations are done in a future step


cquiroz:
Thanks to @rpiaggio for his help on supporting multiple clients for web workers

explore/src/main/scala/explore/config/ConfigurationPanel.scala
```diff
@@ -93,11 +104,38 @@ object ConfigurationPanel {
         bAssign.map(f)
       })
 
+  private implicit val matrixEq: Eq[SpectroscopyModesMatrix] = Eq.by(_.matrix.length)
```
toddburnside, line 107: If this is for reusability of the full modes matrix, will it ever really change?

toddburnside: And, if it doesn't change, could we just get rid of the `Eq` instance and go with `Reusability.always` below?

cquiroz: It only changes once, from empty to "loaded"
We need an  `Eq` for an equality check down in the code

toddburnside: I guess I'm missing where the equality check is below. But, for the reusability, don't you have essentially Not Ready or a full table? You never get a empty table from the stream, right? Or do I misunderstand the `useStreamResourceBy`? (Which is entirely possible)

toddburnside: If you do need the Eq, and you're only interested in empty or notEmpty, you could use `Eq.by(_.matrix.isEmpty)` to avoid traversing the list for the length.

cquiroz: you get a `PotOption[Matrix]` whose `Eq` and `Reusability` will depend on the typeclasses for `Matrix`
But you're right we could use `isEmpty`

PR #2077, opened 2022-08-03
Prevent duplicate target search results.

rpiaggio:
Pressing <ENTER> to submit the target selection form would not cancel the "start search after the user hasn't typed for 700ms" effect, resulting in both search processes running simultaneously and duplicate results.

### DIFF
--- a/explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
+++ b/explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
@@ -272,7 +272,13 @@ object TargetSelectionPopup {
             )
           )(
             ^.autoComplete.off,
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
```diff
@@ -272,7 +272,13 @@ object TargetSelectionPopup {
             )
           )(
             ^.autoComplete.off,
-            ^.onSubmit ==> (e => e.preventDefaultCB >> search(inputValue.get).runAsync)
+            ^.onSubmit ==> (e =>
+              e.preventDefaultCB >>
```
cquiroz, line 276: Neat

cquiroz (approved):
LGTM

PR #2079, opened 2022-08-04
More changes to the configuration panels

toddburnside:
https://user-images.githubusercontent.com/6035943/182866898-45016bec-4df2-40ad-8968-267058e41b4a.mov

The default values for most of the fields are not available at the moment. But, at some point they will be - probably provided by the ODB. At that point all the fields will display the default values.

### DIFF
--- a/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
@@ -66,17 +66,21 @@ import lucuma.ui.reusability._
 import lucuma.ui.syntax.all.*
 import lucuma.ui.syntax.all.given
 import monocle.Lens
+import mouse.boolean._
 import queries.schemas.implicits._
 import react.common.ReactFnProps
+import react.fa.IconSize
 import react.semanticui.collections.form.Form
 import react.semanticui.collections.form.FormInput
 import react.semanticui.elements.button.Button
+import react.semanticui.modules.popup.Popup
 import react.semanticui.shorthand._
 import react.semanticui.sizes._
 import spire.math.Bounded
 import spire.math.Interval
 
 import java.time.Duration
+import scalajs.js
 
 sealed trait AdvancedConfigurationPanel[T <: ScienceModeAdvanced, S <: ScienceModeBasic, Input] {
   val obsId: Observation.Id
@@ -86,7 +90,7 @@ sealed trait AdvancedConfigurationPanel[T <: ScienceModeAdvanced, S <: ScienceMo
   val scienceModeBasic: S
   val spectroscopyRequirements: SpectroscopyRequirementsData
   val potITC: View[Pot[Option[ITCSuccess]]]
-  val onShowBasic: Callback
+  val editState: View[ConfigEditState]
   val confMatrix: SpectroscopyModesMatrix
 
   implicit val ctx: AppContextIO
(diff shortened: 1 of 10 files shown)


explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -108,6 +113,9 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
 ] {
   type AA = Aligner[T, Input]
 
+  @inline protected def isCustomized(aligner:         AA)(implicit ctx: AppContextIO): Boolean
```
cquiroz, line 116: What does it mean `inline` when you'r defining an abstract method?

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -358,48 +403,74 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
...
+
           def onModeMod(modType: Option[ExposureTimeModeType]): Callback = {
             val optITC: Option[ITCSuccess] = props.potITC.get.toOption.flatten
             val oetm                       = modType.map {
               case ExposureTimeModeType.SignalToNoise =>
                 val sn = signalToNoiseView
                   .map(_.get)
+                  .orElse(props.spectroscopyRequirements.signalToNoise)
                   .orElse(optITC.map(_.signalToNoise))
                   .getOrElse(PosBigDecimal.unsafeFrom(BigDecimal(100)))
```
cquiroz, line 475: Can this be `BigDecimal(100).refined`?

explore/src/main/scala/explore/config/ConfigEditState.scala
```diff
@@ -0,0 +1,20 @@
...
+package explore.config
+
+import cats.Eq
+
+enum ConfigEditState {
+  case TableView, DetailsView, SimpleEdit, AdvancedEdit
+}
+
+object ConfigEditState {
+  implicit val eqConfigEditState: Eq[ConfigEditState] = Eq.instance {
```
cquiroz, line 13: We could experiment with `kittens` and this would become 
`enum ConfigEditState derives Eq`

WDYT?

cquiroz (approved):
Looks nice

explore/src/main/scala/explore/config/ConfigEditState.scala
toddburnside: I would like to be able to derive `Eq`.  I can pull in `kittens` and try it in another PR.

cquiroz: Kittens is still i milestone state thus I'd rather test it on explore before moving it to more core libs

I think it uses macros maybe worth test it the size of js doesn't explode using it

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
toddburnside: Yes, I guess that can work now with lucuma-refined. Although it needed a bit of type assistance.

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
toddburnside: Well, to be honest, I was just copying the pattern from the original AdvancedConfigurationPanel. I think (hope) it means that in the concrete classes the method will be inlined. But, I can't say if it actually works or not. The compiler doesn't complain. 🤷

PR #2080, opened 2022-08-04
Add typelevel kittens

toddburnside:
Just throwing this up to see if we want to go this route. I added kittens and did a couple of `Eq` derivations (and conversions to `enum`). The scala 3 version is only at milestone 4, but it is pretty nice, IMO. 😄

### DIFF
--- a/explore/src/main/scala/explore/itc/ItcColumn.scala
+++ b/explore/src/main/scala/explore/itc/ItcColumn.scala
@@ -4,43 +4,26 @@
 package explore.itc
 
 import cats.Eq
+import cats.derived.*
 import cats.syntax.all._
 import eu.timepit.refined.auto._
 
 import scala.concurrent.duration._
 
-sealed trait ItcQueryProblems extends Product with Serializable
-
-object ItcQueryProblems {
-  case object UnsupportedMode          extends ItcQueryProblems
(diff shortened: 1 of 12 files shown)


explore/src/main/scala/explore/itc/ItcColumn.scala
```diff
@@ -4,43 +4,26 @@
...
+enum ItcQueryProblems derives Eq {
+  case UnsupportedMode
+  case MissingWavelength
+  case MissingSignalToNoise
+  case MissingTargetInfo
+  case GenericError(msg: String)
 }
 
-sealed trait ItcResult extends Product with Serializable
+sealed trait ItcResult extends Product with Serializable derives Eq
```
toddburnside, line 21: This could be converted to an `enum` as well. But it appears that enum instances can't have bodies. So `duration` in `Result` below would need to be an extension method. This makes it a lot less clear of a improvement.

cquiroz:
Interesting there is a non trivial increase on js size

cquiroz:
Checked a bit the kittens code and it brings `shapless3` for the derivation. I wonder if that's a sourc of the increase and perhaps it is a one time cost and could be amortized across a bunch of derivations

PR #2082, opened 2022-08-05
Itc web worker

cquiroz:
This is a largish PR that moves requesting ITC data via the web worker. This doesn't really add new functionality but it would let us add much better local caching in a future PR.

A big change is a new intermediate project containing code to communicate with the graphql backend. This lets either the web worker or explore do backend queries


common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
```diff
@@ -40,6 +37,10 @@ import java.time.Instant
 import scala.collection.immutable.SortedMap
 
 object ObsQueries {
+  implicit class ListOps[A](val list: List[A]) {
```
toddburnside, line 40: extension method? 
We have this someplace common don't we? I assume it isn't visible from here?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -138,37 +149,33 @@ object SpectroscopyModesTable {
   val formatSlitLength: ModeSlitSize => String = ss =>
     f"${ModeSlitSize.milliarcseconds.get(ss.size).setScale(0, BigDecimal.RoundingMode.DOWN)}%1.0f"
 
-  def formatGrating(grating: InstrumentRow#Grating): String = grating match {
+  def formatGrating(grating: InstrumentRow#Grating): String = grating match
```
toddburnside, line 152: Are we dropping braces in some places?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -385,26 +389,48 @@ object SpectroscopyModesTable {
...
-      .useSerialStateView(
-        // .useState(
+      .useState {
         ItcResultsCache(
-          Map.empty[ITCRequestParams, EitherNec[ItcQueryProblems, ItcResult]]
+          Map.empty[ItcRequestParams, EitherNec[ItcQueryProblems, ItcResult]]
         )
+      }
+      // itcProgress
+      .useState(none[Progress])
```
toddburnside, line 400: I'm trying to understand how this works with a `worker` now. Before, with a single effect, this component would only get ITC results that it requested, right? What happens now if you submit a bunch of requests on one observation and switch to a new observation which submits a bunch of new requests? Will the new "observation" be processing responses to the requests from the old observation? Or does something prevent that?

common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
cquiroz: We do but it maybe hard to have it in common so I copied it

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: You can drop them in some places yes

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: Good question 🤔

toddburnside: Seems like that would throw off the Progress if nothing else.

cquiroz: Well progress gets reset if conditions change but I haven't thought about the case where we switch to another page
Maybe we need to index the cache by obsid

common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
```diff
@@ -40,6 +37,10 @@ import java.time.Instant
...
+    def toSortedMap[K: Ordering, V](getKey: A => K, getValue: A => V = identity[A](_)) =
```
rpiaggio, line 41: Why can't we use the one here https://github.com/gemini-hlsw/explore/blob/master/common/src/main/scala/explore/implicits.scala#L64 ?

toddburnside: I think right now the `common` module depends on this new module.

cquiroz: I moved it now to `model`

toddburnside (approved):
LGTM

PR #2083, opened 2022-08-06
Fixes size of zoow buttons

cquiroz:
Fix small regression to the zoom buttons' size

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
@@ -43,26 +43,22 @@ final case class AladinContainer(
   obsConf:                ObsConfiguration,
   options:                TargetVisualOptions,
   updateMouseCoordinates: Function1[Coordinates, Callback],
-  updateFov:              Function1[Fov,
-                       Callback
-  ], // TODO Move the functionality of saving the FOV in ALadincell here
-  updateViewOffset:    Function1[Offset, Callback],
-  centerOnTarget:      View[Boolean],
-  selectedGuideStar:   Option[AgsAnalysis],
-  guideStarCandidates: List[AgsAnalysis]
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -282,13 +281,15 @@ object AladinContainer {
                   ExploreStyles.AladinZoomControl,
                   Button(
                     size = Small,
+                    icon = true,
```
cquiroz, line 284: This is the core of the fix

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -43,26 +43,22 @@ final case class AladinContainer(
...
+  type Props = AladinContainer
 
   // This is used for screen coordinates, thus it doesn't need a lot of precission
-  implicit val doubleReuse: Reusability[Double]                  = Reusability.double(1.0)
-  implicit val selectedGSReuse: Reusability[Option[AgsAnalysis]] =
-    Reusability.by(_.map(_.target.id))
-  implicit val analysisReuse: Reusability[List[AgsAnalysis]]     = Reusability.by(_.length)
+  given Reusability[Double]              = Reusability.double(1.0)
+  given Reusability[Option[AgsAnalysis]] = Reusability.by(_.map(_.target.id))
+  given Reusability[List[AgsAnalysis]]   = Reusability.by(_.length)
```
toddburnside, line 61: It's so nice not to have to come up with names that are never used. 😄

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
cquiroz: Yes, it s

PR #2089, opened 2022-08-08
Disable dynamic loading for help

cquiroz:
The help is broken since we switched to scala 3. Traced to
https://github.com/lampepfl/dotty/issues/15701

We'll do static loading

### DIFF
--- a/explore/src/main/scala/explore/ExploreLayout.scala
+++ b/explore/src/main/scala/explore/ExploreLayout.scala
@@ -13,7 +13,7 @@ import explore.model._
 import explore.syntax.ui.*
 import explore.syntax.ui.given
 import japgolly.scalajs.react._
-import japgolly.scalajs.react.extra.router._
+import japgolly.scalajs.react.extra.router.ResolutionWithProps
 import japgolly.scalajs.react.vdom.html_<^._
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -27,12 +27,12 @@ import react.semanticui.modules.sidebar.SidebarPusher
 import react.semanticui.modules.sidebar.SidebarWidth
 
 import scala.scalajs.js
+import scala.scalajs.js.annotation.JSExportTopLevel
 
 final case class ExploreLayout(
-  c:        RouterCtl[Page],
```
toddburnside, line 32: Huh? We weren't using this anymore, eh? Good catch.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/ExploreLayout.scala
cquiroz: Yeah a bit surprising

cquiroz:
Size before and after are comparable. the index file increased on 138k but that was on the separate dynamic bundle which was around 140k

PR #2095, opened 2022-08-12
Use the chart type as an enum

cquiroz:


rpiaggio (approved):

model/shared/src/main/scala/explore/model/enum/ItcSeriesDataType.scala
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model.enums
+
+import lucuma.core.util.Enumerated
+
+enum ItcSeriesDataType:
+  case SignalData, BackgroundData, SingleS2NData, FinalS2NData, PixSigData, PixBackData
```
toddburnside, line 9: I guess this will need to change to `enum ItcSeriesDataType(tag: String)` when Rob gets rid of `Enumerated.of`. Unfortunately, that eliminates the wonderful conciseness.

cquiroz: It does

toddburnside (approved):
LGTM

PR #2096, opened 2022-08-12
Show an indicator when itc is being recalculated

cquiroz:


cquiroz:
https://user-images.githubusercontent.com/3615303/184443107-d2f0a6fe-e65f-4c3b-b36d-d68fcdd8a9cc.mp4

rpiaggio:
In the clip it seems that when the "recomputing" indicator is shown, the stale chart is rerendered in the background. Can we avoid this? Maybe don't show the chart when recomputing.

explore/src/main/scala/explore/itc/package.scala
```diff
@@ -22,3 +22,10 @@ def requiredForITC: TagMod =
       <.span(^.cls := "fa-layers-text fa-inverse", "ITC")
     )
   )("Required for ITC")
+
+opaque type PlotLoading = Boolean
+
+object PlotLoading:
+  def apply(b: Boolean): PlotLoading = b
+
+extension (p: PlotLoading) def boolValue: Boolean = p
```
rpiaggio, line 31: We just merged a PR that introudes a `trait NewType[A]` that implements this opaque type boilerplate.

cquiroz:
Actually I wated to show the existing data, to avoid a black box, WDYT?

rpiaggio:
> Actually I wated to show the existing data, to avoid a black box, WDYT?

IDK, is it useful to show the previous data?

If we show it, can we avoid the rerender?

cquiroz:
Hmm maybe me should only keep the axes while updating, just to show something on the background

cquiroz:
I changed this to use the facility of highchartss to indicate we're loading.
The redrawing is due to the large amount of points. I'll investigate in another PR the feasibility of using the "boost" option to get faster re renders

rpiaggio (approved):
LGTM

PR #2102, opened 2022-08-16
Optimize loading

cquiroz:
This PR updates explore to use the new ITC graph api which gives more metadata allowing us to retrieve less data thus saving on bandwidth


toddburnside (approved):
LGTM

PR #2104, opened 2022-08-17
More fluid semester plot.

rpiaggio:
- Move semester plot computations to worker.
- Avoid restarting computations on resize.
- Correctly resize tile when shrinking height.
- Fix issue on WorkerServer where cancellation messages were not processed if there was a heavy ongoing computation.

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -944,6 +944,7 @@ tfoot {
 .elevation-plot {
   flex-shrink: 1;
   flex-grow: 1;
+  height: 100%;
 
   .highcharts-xaxis .highcharts-axis-title {
     fill: hsl(0, 87%, 66%);
--- a/explore/src/main/scala/explore/tabs/ElevationPlotTile.scala
+++ b/explore/src/main/scala/explore/tabs/ElevationPlotTile.scala
@@ -31,8 +31,7 @@ object ElevationPlotTile {
       "Elevation Plot",
       canMinimize = true,
       bodyClass = ExploreStyles.ElevationPlotTileBody.some,
-      tileClass = ExploreStyles.ElevationPlotTile.some,
-      sizeStateCallback = s => Callback.log(s)
+      tileClass = ExploreStyles.ElevationPlotTile.some
     ) { (_: Tile.RenderInTitle) =>
       (uid, coordinates)
         .mapN { case (uid, c) =>
--- a/explore/src/main/scala/explore/tabs/ElevationPlotTile.scala
+++ b/explore/src/main/scala/explore/tabs/ElevationPlotTile.scala
@@ -9,6 +9,7 @@ import explore.components.ui.ExploreStyles
 import explore.implicits._
 import explore.model.ScienceMode
 import explore.targeteditor.ElevationPlotSection
+import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import lucuma.core.math.Coordinates
 import lucuma.core.model.Target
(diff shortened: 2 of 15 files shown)


common/src/main/webapp/less/style.less
```diff
@@ -944,6 +944,7 @@ tfoot {
 .elevation-plot {
   flex-shrink: 1;
   flex-grow: 1;
+  height: 100%;
```
cquiroz, line 947: heh, that was the fix 😃

explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
```diff
@@ -119,7 +122,7 @@ object ElevationPlotSection {
       // Actual date
       .useState(ZonedDateTime.now(Site.GS.timezone).toLocalDate.plusDays(1))
       .render { (props, _, options, date) =>
-        implicit val ctx = props.ctx
+        import props.given
```
cquiroz, line 125: Nice

workers/src/main/scala/workers/PlotServer.scala
```diff
@@ -0,0 +1,99 @@
...
+import java.time.Duration
+import java.time.Instant
+import java.time.LocalTime
+import scala.ContextFunction1
+import scala.collection.immutable.TreeMap
+import scala.scalajs.js.annotation.JSExport
+import scala.scalajs.js.annotation.JSExportTopLevel
+
+@JSExportTopLevel("PlotServer", moduleID = "exploreworkers")
+object PlotServer extends WorkerServer[IO, PlotMessage.Request] with IOApp.Simple {
```
cquiroz, line 35: Are we using `IOApp`? What about reloads?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/tabs/ElevationPlotTile.scala
```diff
@@ -30,7 +31,8 @@ object ElevationPlotTile {
       "Elevation Plot",
       canMinimize = true,
       bodyClass = ExploreStyles.ElevationPlotTileBody.some,
-      tileClass = ExploreStyles.ElevationPlotTile.some
+      tileClass = ExploreStyles.ElevationPlotTile.some,
+      sizeStateCallback = s => Callback.log(s)
```
toddburnside, line 35: Do you want to keep this logging here?

explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
```diff
@@ -43,17 +43,17 @@ import react.semanticui.elements.button.ButtonGroup
...
-  uid:              User.Id,
-  tid:              Target.Id,
-  scienceMode:      Option[ScienceMode],
-  coords:           Coordinates
-)(implicit val ctx: AppContextIO)
+  uid:           User.Id,
+  tid:           Target.Id,
+  scienceMode:   Option[ScienceMode],
+  coords:        Coordinates
+)(using val ctx: AppContextIO)
```
toddburnside, line 50: With the `import props.given` below, does this need a name anymore?

toddburnside (approved):
LGTM

workers/src/main/scala/workers/PlotServer.scala
rpiaggio: Good catch, that was just a test, let me revert.

rpiaggio:
> The semester plot used to have a line indicating progress, which would move along the calculation. Do we still have that? Is the worker sending messages along to inform progress to the UI?

Yes, that's still there. The worker sends each value as they become available, the UI sees a stream of values and adds each to the plot as soon as they are received, advancing the progress line.

explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
rpiaggio: Unfortunately, it does: we need it as a `val`, which forces us to declare a name.

toddburnside: Ah, yeah. Second parameter list is private by default.

PR #2107, opened 2022-08-17
Update highcharts and the accessibility module

cquiroz:
We have a pending highcharts update to version 10.
Among other changes on version it produced a big warning if you have the accessibility module disabled

This could be avoided but perhaps we can try using the module

This PR adds the code to add the accessibility module. Let's see how big it gets

### DIFF
--- a/explore/src/main/scala/explore/highcharts.scala
+++ b/explore/src/main/scala/explore/highcharts.scala
@@ -6,4 +6,11 @@ package explore.highcharts
 import gpp.highcharts.mod._
 import japgolly.scalajs.react._
 
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
 extension (c: Chart_) inline def showLoadingCB = CallbackTo(c.showLoading())
+
+@js.native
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/highcharts.scala
```diff
@@ -6,4 +6,11 @@ package explore.highcharts
...
 import japgolly.scalajs.react._
 
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
 extension (c: Chart_) inline def showLoadingCB = CallbackTo(c.showLoading())
+
+@js.native
+@JSImport("highcharts/es-modules/masters/modules/accessibility.src.js", JSImport.Default)
+object HighchartsAccesibility extends js.Object
```
rpiaggio, line 16: Can we put this in the facade in lucuma-react?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/highcharts.scala
cquiroz: Yes we could move these there

PR #2109, opened 2022-08-18
Fix a bug when restoring the layouts from the db

cquiroz:
This fixes a bug restoring the obs layout coming from the user preferences db

### DIFF
--- a/common/src/main/scala/explore/common/UserPreferencesQueries.scala
+++ b/common/src/main/scala/explore/common/UserPreferencesQueries.scala
@@ -5,6 +5,7 @@ package explore.common
 
 import cats.ApplicativeThrow
 import cats.MonadThrow
+import cats.Order._
 import cats.data.OptionT
 import cats.syntax.all._
 import clue.TransactionalClient
@@ -14,7 +15,7 @@ import explore.model.ResizableSection
 import explore.model.enums.PlotRange
 import explore.model.enums.TimeDisplay
 import explore.model.enums.Visible
-import explore.model.layout._
+import explore.model.layout.{_, given}
 import lucuma.core.enums.Site
 import lucuma.core.math.Angle
 import lucuma.core.math.Offset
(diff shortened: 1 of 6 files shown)


common/src/main/scala/explore/components/TileController.scala
```diff
@@ -104,9 +106,13 @@ object TileController {
         getBreakpointFromWidth(p.layoutMap.map { case (x, (w, _, _)) => x -> w }, p.gridWidth)
       }
       // Store the current layout
-      .useStateViewWithReuseBy((p, _, _) => updateResizableState(p.layoutMap))
+      .useStateViewBy((p, _, _) => updateResizableState(p.layoutMap))
+      // Update the current layout if it changes upstream
+      .useEffectWithDepsBy((p, _, _, _) => p.layoutMap)((_, _, _, current) =>
```
cquiroz, line 111: This is the core of the fix.

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -14,14 +15,16 @@ import explore.model.ResizableSection
 import explore.model.enums.PlotRange
 import explore.model.enums.TimeDisplay
 import explore.model.enums.Visible
-import explore.model.layout._
+import explore.model.layout.{_, given}
```
rpiaggio, line 18: I think Scala 2 style `import X._` imports givens already. Scala 3 style would be `import X.{*, given}`.

common/src/main/scala/explore/components/TileController.scala
```diff
@@ -16,7 +18,7 @@ import explore.implicits._
 import explore.model.Constants
 import explore.model.GridLayoutSection
 import explore.model.enums.TileSizeState
-import explore.model.layout._
```
rpiaggio, line 19: `import *`

common/src/main/scala/explore/components/TileController.scala
rpiaggio: Thanks. I wonder if we should implement a custom hook that implements the `useState`/`useEffectWithDeps` pattern.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/common/UserPreferencesQueries.scala
cquiroz: Yes, scalafmt for some reason converts them to `{_, given}` I guess we need two lines of imports

cquiroz: or maybe `OrganizeImports`

PR #2111, opened 2022-08-19
Remove duplicate lucuma-react-datepicker dependency

toddburnside:
There was also a triplicate is-regexp in package-lock.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5055,24 +5055,6 @@
         "node": ">=0.10.0"
       }
     },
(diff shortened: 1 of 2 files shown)


rpiaggio (approved):
LGTM. Should we sort alphabetically to make it harder for this to happen?

toddburnside:
> LGTM. Should we sort alphabetically to make it harder for this to happen?

I thought about that, but wondered if I was being too OCD. I'll do it!

PR #2114, opened 2022-08-19
Use exposure time for graphs

cquiroz:
This uses the actual exposure time and count for the selected mode

image

### DIFF
--- a/common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
+++ b/common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
@@ -36,6 +36,7 @@ import queries.schemas.implicits._
 
 import java.time.Instant
 import scala.collection.immutable.SortedMap
+import lucuma.core.model.ExposureTimeMode.FixedExposure
 
 object ObsQueries {
   type ObservationList = KeyedIndexedList[Observation.Id, ObsSummaryWithTitleConstraintsAndConf]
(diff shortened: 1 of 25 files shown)


common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
```diff
@@ -126,45 +105,66 @@ object ObsQueries {
...
-            mtch.id,
-            mtch.title,
-            mtch.subtitle,
-            mtch.constraintSet,
-            mtch.status,
-            mtch.activeStatus,
-            mtch.plannedTime.execution,
-            mtch.scienceMode,
-            mtch.visualizationTime
+  extension (self: ObsEditQuery.Data.type)
```
rpiaggio, line 108: Maybe we can define the extension over `ObsEditQuery.Data` and make `asObsEditData` return an `ObsEditData` directly. Or do we need the `Getter`?

common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
```diff
@@ -126,45 +105,66 @@ object ObsQueries {
...
+        }
 
-  implicit class ProgramObservationsQueryDataOps(val self: ProgramObservationsQuery.Data.type)
-      extends AnyVal {
-    def asObsSummariesWithConstraints = queryToObsSummariesWithConstraintsGetter
-  }
+  extension (self: ITCSuccess)
+    def asFixedExposureTime = FixedExposure(self.exposures, self.exposureTime)
+
+  extension (self: ProgramObservationsQuery.Data.type)
```
rpiaggio, line 132: Same here

rpiaggio (approved):
LGTM

common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
cquiroz: Good idea

PR #2116, opened 2022-08-20
Convert ags-test to enum

cquiroz:
Don't merge I want to see the effect of switching to enum instead of a trait on js size


cquiroz:
Actually this could be merged, we got a size decrease due to switching to enums

toddburnside (approved):
LGTM

PR #2125, opened 2022-08-21
Fix problem of exposure time override not updating ODB

toddburnside:


explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -377,7 +353,7 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
         (
           props,
           exposureModeEnum,
-          expTimeOverrideSecs,
+          // expTimeOverrideSecs,
```
cquiroz, line 356: Should we remove this?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
toddburnside: Yep. Missed that in my cleanup.

PR #2126, opened 2022-08-22
Update exposure mode override properly

toddburnside:
The exposure mode drop down was not being reset when config customizations were reverted. It also didn't change if the value was modified in a different session.


cquiroz (approved):
LGTM

PR #2128, opened 2022-08-23
Fix peak pixel display and a message if there is not enough data to g…

cquiroz:
…et the ITC graph


cquiroz:
I'll just merge this for testing

PR #2133, opened 2022-08-25
Update lucuma-core

cquiroz:


common/src/main/scala/explore/model/enums/Theme.scala
```diff
@@ -19,6 +19,5 @@ object Theme {
   }
 
   /** @group Typeclass Instances */
-  implicit val ThemeEnumerated: Enumerated[Theme] =
-    Enumerated.of(Dark, Light)
+  given Eq[Theme] = Eq.by(_.clazz)
```
rpiaggio, line 22: `derives Eq`? BTW, I'll be moving this to `lucuma-ui` soon to share it with `observe`.

cquiroz: Well I checked and in the current code base we don't use `Enumerated`

rpiaggio: Yup, I agree on removing it. I was just suggesting deriving Eq instead of defining it.

rpiaggio: With the enforcement of `tag`, I think `Enumerated` is more geared towards things we want to store in databases, so that they are not affected by renames.

That said, I agree that we never know when some project may want to store something in a DB. For example, in this case, we will store this in a user preferences DB. (So I'll make the `lucuma-ui` version have an `Enumerated`).

PR #2153, opened 2022-08-28
Fix redrawing of charts upon resizing

cquiroz:
For a while we've noticed that resizing the tiles with charts leads to artifacts on correctly drawing the plots often getting larger than the tile and thus showing scrollbars

This PR fixes this with 3 changes
* New `ResizingChart` component that calls `chart.reflow()` upon changing the size (to be moved to lucuma-ui)
* Css changes to better follow the tile size
* A bit unrelated but charts need cleanup or we start accumulating memory used on the charts

[sc-1367]


rpiaggio (approved):
LGTM, thanks!

PR #2154, opened 2022-08-28
Double initialization sync for workers.

rpiaggio:
There's a race condition between initializing workers and initializing their clients. By "initialize" I mean that the "onmessage" listener is executed.

In a previous PR we added a message for the server to notify the client when it's initialized and logic for the client to wait for that message before sending requests to the server.

However, the client may not be initialized when the server sends its ready message, so now we add another "client ready" message for the client to request the "server ready" message from the server.

The server therefore sends the "server ready" message twice: once as an answer to the "client ready" request, and once as soon the server is initialized. The first one may not happen if the server misses the "client ready" message. This way we make sure that the client gets at least one "server ready" message.


cquiroz (approved):
LGTM, thanks

PR #2159, opened 2022-08-31
Switch to Tooltips using floating-ui

cquiroz:
I was never too happy with SUI tooltips as they are kind of ugly and may not place themselves on the right place.

This PR includes a facade to use [floating-ui](https://floating-ui.com/) which is a low-level library to implement thing like tooltips. It doesn't include a Tooltip itsef thus a component is provided.

The tooltip looks better IMHO:
* before
Explore 2022-08-30 16-54-40(1)

* after:
Explore 2022-08-30 16-56-20

A nice feature is that it can adapt its location depending on the size available as can be seen here:

https://user-images.githubusercontent.com/3615303/187576984-eff7ff82-abd5-4d29-98f1-d9bc391a3340.mp4

### DIFF
--- a/common/src/main/scala/react/floatingui/Tooltip.scala
+++ b/common/src/main/scala/react/floatingui/Tooltip.scala
@@ -26,12 +26,12 @@ import js.annotation._
  * Tooltip base on floating ui see: https://floating-ui.com/docs/react-dom
  */
 case class Tooltip(trigger: VdomTag, tooltip: VdomNode, placement: Placement = Placement.Top)
-    extends ReactFnProps[Tooltip](Tooltip.component)
+    extends ReactFnProps(Tooltip.component)
 
 object Tooltip {
-  type Props = Tooltip
+  private type Props = Tooltip
 
-  val component =
+  private val component =
     ScalaFnComponent
       .withHooks[Props]
       .useState(false) // isOpen
--- a/common/src/main/scala/react/floatingui/Tooltip.scala
+++ b/common/src/main/scala/react/floatingui/Tooltip.scala
@@ -1,0 +1,135 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package react.floatingui
+
+import cats.syntax.all.*
+import japgolly.scalajs.react._
+import japgolly.scalajs.react.feature.ReactFragment
+import japgolly.scalajs.react.vdom.TagOf
+import japgolly.scalajs.react.vdom.html_<^._
+import lucuma.ui.syntax.all.*
+import lucuma.ui.syntax.all.given
+import org.scalajs.dom
+import org.scalajs.dom.html
+import react.common.ReactFnProps
+import react.common.Style
+import react.floatingui.hooks._
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSExportTopLevel
+import scala.scalajs.js.|
+
+import js.annotation._
+
+/**
+ * Tooltip base on floating ui see: https://floating-ui.com/docs/react-dom
+ */
+case class Tooltip(trigger: VdomTag, tooltip: VdomNode, placement: Placement = Placement.Top)
+    extends ReactFnProps[Tooltip](Tooltip.component)
+
+object Tooltip {
+  type Props = Tooltip
+
+  val component =
+    ScalaFnComponent
+      .withHooks[Props]
+      .useState(false) // isOpen
+      .useRefToVdom[dom.HTMLElement] // arrow
+      .useFloatingBy { (props, open, arrow) =>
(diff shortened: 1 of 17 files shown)


common/src/main/scala/react/floatingui/Tooltip.scala
```diff
@@ -0,0 +1,135 @@
...
+import js.annotation._
+
+/**
+ * Tooltip base on floating ui see: https://floating-ui.com/docs/react-dom
+ */
+case class Tooltip(trigger: VdomTag, tooltip: VdomNode, placement: Placement = Placement.Top)
+    extends ReactFnProps[Tooltip](Tooltip.component)
+
+object Tooltip {
+  type Props = Tooltip
```
rpiaggio, line 32: Can we make `Props` and `component` private and protected?

cquiroz: Are we doing that often?

common/src/main/scala/react/floatingui/Tooltip.scala
```diff
@@ -0,0 +1,135 @@
...
+import scala.scalajs.js.annotation.JSExportTopLevel
+import scala.scalajs.js.|
+
+import js.annotation._
+
+/**
+ * Tooltip base on floating ui see: https://floating-ui.com/docs/react-dom
+ */
+case class Tooltip(trigger: VdomTag, tooltip: VdomNode, placement: Placement = Placement.Top)
+    extends ReactFnProps[Tooltip](Tooltip.component)
```
rpiaggio, line 29: With Scala 3 we don't need the `React(Fn)Props` type parameter anymore. We can write eg: `extends ReactFnProps(Tooltip.component)`

common/src/main/scala/react/floatingui/Tooltip.scala
```diff
@@ -0,0 +1,135 @@
...
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSExportTopLevel
+import scala.scalajs.js.|
+
+import js.annotation._
+
+/**
+ * Tooltip base on floating ui see: https://floating-ui.com/docs/react-dom
+ */
+case class Tooltip(trigger: VdomTag, tooltip: VdomNode, placement: Placement = Placement.Top)
```
rpiaggio, line 28: Is the idea to move this to `lucuma-react` eventually?

common/src/main/scala/react/floatingui/hooks.scala
```diff
@@ -0,0 +1,85 @@
...
+import scala.scalajs.js.JSConverters._
+import scala.scalajs.js.|
+
+@js.native
+trait ElementProps extends js.Object
+
+type ElementPropsList = List[ElementProps]
+
+object HooksApiExt {
+  val floatingHook =
```
rpiaggio, line 18: Can we make the hook `val`s private or at least protected?

common/src/main/webapp/theme/semantic.less
```diff
@@ -130,9 +130,9 @@
 & {
   @import "fomantic-ui-less/definitions/modules/nag";
 }
-& {
-  @import "fomantic-ui-less/definitions/modules/popup";
-}
+// & {
```
rpiaggio, line 133: Commented code

rpiaggio: Maybe the idea is to leave it like this?

rpiaggio (approved):
LGTM

common/src/main/scala/react/floatingui/Tooltip.scala
rpiaggio: Some of our components have it already, but I think we should.

common/src/main/webapp/theme/semantic.less
cquiroz: Yes I'm going to leave it there, we could remove a few more and reduce css size

common/src/main/scala/react/floatingui/Tooltip.scala
cquiroz: Yes we should. I want to keep it here for a while until I'm sure it is working properly. The facade is very minimal and may need to be expanded

toddburnside, line 28: I wonder if we should have a `disabled` flag or something? I'm thinking of a situation where we have a element that will expect to always be in the DOM, but will only have a tooltip some of the time.

Another situation could be where we want the tooltip to always be displayed. For example, the way we currently display errors for inputs.

toddburnside (approved):
LGTM with a couple of comments.

primereact has their own tooltip component. It's functionality is built into many of their components which makes it easy to use, but less flexible than this implementation.

common/src/main/scala/react/floatingui/Tooltip.scala
cquiroz: Right, we could add those as needed

PR #2168, opened 2022-09-01
Switch to react style naming

cquiroz:
TIL the previous style naming is not compatible with react, instead we need to use `camelCase`

`In React, all DOM properties and attributes (including event handlers) should be camelCased.` from:
https://reactjs.org/docs/dom-elements.html

### DIFF
--- a/common/src/main/scala/react/floatingui/Tooltip.scala
+++ b/common/src/main/scala/react/floatingui/Tooltip.scala
@@ -94,24 +94,24 @@ object Tooltip {
         val placementStyle: Map[String, String | Int] =
           Placement.fromString(floating.placement) match {
             case Some(Placement.Top) | Some(Placement.TopStart) | Some(Placement.TopEnd)          =>
-              Map("bottom"              -> arrowShift,
-                  "border-right-width"  -> arrowBorder,
-                  "border-bottom-width" -> arrowBorder
+              Map("bottom"            -> arrowShift,
+                  "borderRightWidth"  -> arrowBorder,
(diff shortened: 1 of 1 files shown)


common/src/main/scala/react/floatingui/Tooltip.scala
```diff
@@ -94,24 +94,24 @@ object Tooltip {
         val placementStyle: Map[String, String | Int] =
           Placement.fromString(floating.placement) match {
             case Some(Placement.Top) | Some(Placement.TopStart) | Some(Placement.TopEnd)          =>
-              Map("bottom"              -> arrowShift,
-                  "border-right-width"  -> arrowBorder,
-                  "border-bottom-width" -> arrowBorder
+              Map("bottom"            -> arrowShift,
+                  "borderRightWidth"  -> arrowBorder,
```
toddburnside, line 98: Was this causing a problem? It seems that this is a style, not a dom property or attribute and they need to match the CSS name. Does react change the `borderRightWidth` to `border-right-width` automagically when it applies it as an inline style?

cquiroz: I was surprised too, and even worse the old method still works but you get a big warning on the log.
In the link it explains a bit how it works and it seems react does the conversion

toddburnside (approved):
LGTM

common/src/main/scala/react/floatingui/Tooltip.scala
cquiroz: Maybe it is easier to write in camel case in JSX/JS?

PR #2171, opened 2022-09-01
Move the target control and S/N info to the ITC tile title bar

cquiroz:
https://user-images.githubusercontent.com/3615303/188026236-b25fe389-b18d-488d-bf6a-a5470fec7816.mp4


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -132,7 +133,9 @@ object ObsTabTiles {
           )
           .reRunOnResourceSignals(ObservationEditSubscription.subscribe[IO](props.obsId))
       }
-      .render { (props, obsView) =>
+      // ITC selected target. Here to be shared by the ITC tile body and title
+      .useStateView(none[ItcTarget])
```
rpiaggio, line 137: Maybe we want to move this further up when we have a Configuration Panel?

rpiaggio (approved):
LGTM

toddburnside:
How does the dropdown interact with the other tiles when the itc panel is collapsed? In particular, how does it interact with the dropdown in the constraints tile? I think the current z-index solution to not having the contraints dropdown flow behind other panels may be limited in its ability to handle multiple dropdowns.

cquiroz:
@toddburnside That',s a good question. I didn't think about it but seems fine from the screenshots

image

image

cquiroz:
This seems fine too
image

toddburnside:
> This seems fine too
Excellent. I just looked at how the z-index for the dropdown works, and it's better than I thought. :)

toddburnside (approved):
LGTM

cquiroz:
Would prime react drop down handle these cases?

toddburnside:
> Would prime react drop down handle these cases?

Yes, it should work fine.

PR #2178, opened 2022-09-02
Setup local cache for AGS calculations

cquiroz:


workers/src/main/scala/workers/package.scala
```diff
@@ -40,3 +42,5 @@ object PromiseConverter:
 given PromiseConverter[IO] = new PromiseConverter[IO] {
   inline def convert[A](promise: => js.Promise[A]): IO[A] = IO.fromPromise(IO(promise))
 }
+
+extension (t: IO.type) def now(): IO[Instant] = IO(Instant.now)
```
rpiaggio, line 46: Nice. Maybe we can move this to lucuma-core in the future?

workers/src/main/scala/workers/AgsServer.scala
```diff
@@ -20,18 +25,34 @@ object AgsServer extends WorkerServer[IO, AgsMessage.Request] {
...
+    for
+      self  <- IO(dom.DedicatedWorkerGlobalScope.self)
+      cache <- Cache.withIDB[IO](self.indexedDB.toOption, "ags")
+      _     <- cache.evict(CacheRetention).start
+    yield invocation =>
+      invocation.data match {
+        case req @ AgsMessage.Request(_, _, _, _, _, _, _) =>
+          val cacheableRequest =
+            Cacheable(CacheName("ags"), CacheVersion(AgsCacheVersion), agsCalculation)
+          (IO.now() >>= (now => cache.evict(now.minus(CacheRetention)).start)) >>
```
rpiaggio, line 55: Do we want to run eviction on every request? We are alrady running it on startup.

rpiaggio (approved):
LGTM

workers/src/main/scala/workers/AgsServer.scala
cquiroz: ah you're right. I guess it will be kind of expensive if it gets longer

PR #2180, opened 2022-09-02
Add some tooltips to the program popup

cquiroz:


cquiroz:
image

explore/src/main/scala/explore/EditableLabel.scala
```diff
@@ -59,29 +69,72 @@ object EditableLabel {
...
+      editButtonTooltip,
       deleteButtonClass,
-      deleteButtonIcon
+      deleteButtonIcon,
+      deleteButtonTooltip,
+      okButtonTooltip,
+      discardButtonTooltip
     )
 
+  type Editing = Editing.Type
```
rpiaggio, line 80: Can this just be an `enum Editing: case NotEditing, InEdition` ?

rpiaggio (approved):
LGTM

toddburnside (approved):
LGTM

explore/src/main/scala/explore/EditableLabel.scala
cquiroz: Wouldn't be cheaper to use an opaque type?

rpiaggio: Generally speaking, yes. If we define `val`s for the values, probably yes too, but no so much ;) Ohh unless the vals are inline?

PR #2181, opened 2022-09-02
Accelerate semester plots

rpiaggio:
- Accelerate semester plot rendering (when cached) by batching responses from worker.
- Show message when target is not visible at all in the semester.


cquiroz (approved):
LGTM

PR #2185, opened 2022-09-04
Add syntax for making tooltips easier to use.

toddburnside:
If we go with this, I can go through and update the other uses of tooltips

### DIFF
--- a/common/src/main/scala/react/floatingui/syntax.scala
+++ b/common/src/main/scala/react/floatingui/syntax.scala
@@ -10,26 +10,9 @@ import react.floatingui.*
 
 import scala.scalajs.js
 
-// Unfortunately, we can't differentiate between the extension methods below and use
-// `withTooltip` for both because a `VdomTag` IS a `VdomNode`.
-
 extension (tag: VdomTag)
-  // This extension method can be used with html tags like `span` and `div`. These tags
-  // do not need to be wrapped in another tag, so this method should be used when possible.
   def withTooltip(
-    tooltip:   js.UndefOr[VdomNode],
+    tooltip:   VdomNode,
     placement: Placement = Placement.Top
   ): VdomNode =
-    tooltip.fold(tag)(tt => Tooltip(trigger = tag, tooltip = tt, placement = placement))
-
-extension (node: VdomNode)
-  // This extension method is to be used for react components, which need to be wrapped
-  // in a `span` in order to be the trigger for a tooltip.
-  // If possible, use `withTooltip`, above.
-  def componentWithTooltip(
-    tooltip:   js.UndefOr[VdomNode],
-    placement: Placement = Placement.Top
-  ): VdomNode =
-    tooltip.toOption.fold(node)(tt =>
-      Tooltip(trigger = <.span(node), tooltip = tt, placement = placement)
-    )
+    Tooltip(trigger = tag, tooltip = tooltip, placement = placement)
--- a/common/src/main/scala/react/floatingui/syntax.scala
+++ b/common/src/main/scala/react/floatingui/syntax.scala
@@ -1,32 +1,35 @@
 // Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 // For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-package react.floatingui
+package react.floatingui.syntax
 
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.ui.syntax.all.given
+import react.floatingui.*
 
 import scala.scalajs.js
 
-object syntax:
-  // extension (tag: VdomTag)
-  //   def withTooltip(tooltip: VdomNode, placement: Placement = Placement.Top): VdomNode =
-  //     Tooltip(trigger = tag, tooltip = tooltip, placement = placement)
-  //   def withOptionalTooltip(
-  //     tooltip:   js.UndefOr[VdomNode],
-  //     placement: Placement = Placement.Top
-  //   ): TagMod =
-  //     tooltip.fold(tag)(tt => Tooltip(trigger = tag, tooltip = tt, placement = placement))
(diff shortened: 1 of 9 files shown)


common/src/main/scala/react/floatingui/syntax.scala
```diff
@@ -0,0 +1,32 @@
...
+package react.floatingui
+
+import japgolly.scalajs.react.*
+import japgolly.scalajs.react.vdom.html_<^.*
+import lucuma.ui.syntax.all.given
+
+import scala.scalajs.js
+
+object syntax:
+  // extension (tag: VdomTag)
```
toddburnside, line 13: I want to add these extensions for when we want to add tooltips to dom elements that don't need to be wrapped in a `span`. However, I can't use the same names and still have default parameters because the compiler complains about `2 or more overloaded methods have default parameters`. Any suggestions?

rpiaggio: Low priority implicits?

common/src/main/scala/react/floatingui/syntax.scala
```diff
@@ -0,0 +1,32 @@
...
+  //   def withTooltip(tooltip: VdomNode, placement: Placement = Placement.Top): VdomNode =
+  //     Tooltip(trigger = tag, tooltip = tooltip, placement = placement)
+  //   def withOptionalTooltip(
+  //     tooltip:   js.UndefOr[VdomNode],
+  //     placement: Placement = Placement.Top
+  //   ): TagMod =
+  //     tooltip.fold(tag)(tt => Tooltip(trigger = tag, tooltip = tt, placement = placement))
+
+  extension (node: VdomNode)
+    def withTooltip(tooltip: VdomNode, placement: Placement = Placement.Top): VdomNode =
```
cquiroz, line 23: Could we do this with an extra `Tooltip` constructor?

cquiroz: Ah, nevermind this is to be applied to the trigger, nice

cquiroz (approved):
I like this syntax

common/src/main/scala/react/floatingui/syntax.scala
toddburnside: As I understand it, this isn't a problem of resolution, but a problem with how methods with default parameters are named. This results in two methods with the same name, and the other parameters don't disambiguate for some reason.

IIRC, this is one of the things the magnet pattern was designed for. I'll see how ugly that would be.

toddburnside: I did not recall correctly. The magnet pattern cannot handle default parameters.

common/src/main/scala/react/floatingui/syntax.scala
```diff
@@ -0,0 +1,32 @@
...
+      Tooltip(trigger = <.span(node), tooltip = tooltip, placement = placement)
+
+    def withOptionalTooltip(
```
toddburnside, line 26: `withOptionalTooltip` could be renamed `withTooltip` and would work for both cases. I started out with `Option[VdomNode` for the `tooltip` parameter, but the `js.UndefOr` allows either a `VdomNode` or a `Option[VdomNode]` to be used. More importantly, it allows the use of `Option[String]`. 

I kind of like that the name indicates in the code that the tooltip may not be present, but it's probably best to just have the one method. Thoughts?

cquiroz (approved):
LGTM

common/src/main/scala/react/floatingui/syntax.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import japgolly.scalajs.react.*
+import japgolly.scalajs.react.vdom.html_<^.*
+import lucuma.ui.syntax.all.given
+import react.floatingui.*
+
+import scala.scalajs.js
+
+// Unfortunately, we can't differentiate between the extension methods below and use
+// `withTooltip` for both because a `VdomTag` IS a `VdomNode`.
```
toddburnside, line 14: One last question: I dislike having extra crud in the DOM, but is it worth the extra extension method and name to avoid having an extra `span` in the DOM for tooltip triggers?

cquiroz: The thing with `span` is that we need a way to set a `Ref` and you can't do that on `VdomNode`, you need a `VdomTag`, having a `span` around is the easiest way.
The extension seems nice from an API point of view but I wonder if it may obscure the fact that we are wrapping and you may need to adjust your css

cquiroz: Maybe a bbetter option is to make `Toolitp` take a `ReactFnProps/GenericFNComponont/..` or something like that and set the `Ref` there

toddburnside: I understand the need for the `span` with the `VdomNode`. My question was about avoiding wrapping a `span` in another `span`. I'm doing that now, but it requires the extra method for use with `VdomNode`.

Good point about it affecting the CSS. I was more considering the idea that the `ToolTip` wrapper was sort of obscuring what the "main" elements in the html were. 🤔

toddburnside: > Maybe a bbetter option is to make `Toolitp` take a `ReactFnProps/GenericFNComponont/..` or something like that and set the `Ref` there

Hmmm. Let me see if I can come up with something like that.

toddburnside: I do not see any way forward with something like that. (Not that I couldn't be missing something.) So, maybe we could just keep the syntax for elements (VdomTag), and just require the user to explicitly wrap components in `span`s. This leaves `span`s when they aren't needed (for optional tooltips), but it does keep the dom consistent. I still think it is still a win wrt syntax.

toddburnside: I did the above, as well as remove the "optional" functionality. This should make it more apparent what you get in the DOM.

PR #2186, opened 2022-09-05
Fix PM and Parallax formatting.

rpiaggio:


model/shared/src/main/scala/lucuma/utils/package.scala
```diff
@@ -0,0 +1,32 @@
...
+    case (None, Some(someB))        => (Monoid[A].empty, someB).some
+    case (Some(someA), Some(someB)) => (someA, someB).some
+
+// We should move this to lucuma-ui or remove it if there's ever a coulomb-refined for Scala 3
+extension [V, U](q: Quantity[V, U]) {
+  @inline def toRefined[P](using Validate[V, P]): Either[String, Quantity[V Refined P, U]] =
+    refineQV[P](q)
+}
+
+abstract class NewType[Wrapped](using eq: Eq[Wrapped]):
```
cquiroz, line 27: Should we move this to `lucuma-core`?

cquiroz (approved):
LGTM

model/shared/src/main/scala/lucuma/utils/package.scala
rpiaggio: 👍

PR #2201, opened 2022-09-09
Initial integration of primereact

toddburnside:


common/src/main/scala/explore/components/primereact/EnumDropdownView.scala
```diff
@@ -0,0 +1,46 @@
...
+import lucuma.core.util.Enumerated
+import react.common.*
+import react.primereact.Dropdown
+import react.primereact.SelectItem
+import reactST.primereact.components.{Dropdown => CDropdown}
+
+import scalajs.js
+import scalajs.js.JSConverters.*
+
+object EnumDropdownView:
```
cquiroz, line 19: Would these be eventually moved to lucuma-ui?

common/src/main/scala/explore/components/primereact/EnumDropdownView.scala
```diff
@@ -0,0 +1,46 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.components.primereact
+
+import crystal.react.View
+import eu.timepit.refined.types.string.NonEmptyString
+import japgolly.scalajs.react.*
+import lucuma.core.util.Display
+import lucuma.core.util.Enumerated
+import react.common.*
```
cquiroz, line 11: Could we use only the minimal imports from `react.common`?

common/src/main/webapp/less/style.less
```diff
@@ -1072,6 +1072,10 @@ tfoot {
 
   .elevation-range-entry {
     width: 8em;
+
```
cquiroz, line 1075: Should we switch to sass step by step?

cquiroz (commented):
It is obviously hard to review 50k of code :D
Some questions:
* Are the theme css imported `as is`. Can we just replace them if they change upstream?
* Do the currently created css variables play a role?

As a general thing I'd like that we wait until we have the production  release stable before  merging this to avoid having two styles in production

rpiaggio:
Thank you for tackling this!

Would it be possible to reference the prime-react files instead of copying them over? That way we don't have to manually keep them up to date.

common/src/main/scala/explore/components/primereact/EnumDropdownView.scala
toddburnside: Yes. That is the idea. It was just easiest to start with them here.

toddburnside:
> * Are the theme css imported `as is`. Can we just replace them if they change upstream?
 
At the moment, they are imported `as is` from a modified instance of the theme designer (that I need to put in github as a private repository). We can just replace them if they change in that theme designer project, but I think we should find a better solution. Raul and I have discussed somehow pulling them out of a dependency jar or creating an npm package. I need to re-read the theme designer license (https://www.primefaces.org/designer/licenses) to see what we can do within the limits of the license.

toddburnside:
> * Do the currently created css variables play a role?

Do you mean the css variables in the currently deployed explore (non-primereact)? If so, I think most (or all) of the variables related to color should probably be superseded eventually by the primereact CSS variables. We can tweak the theme colors if needed.

toddburnside:
> As a general thing I'd like that we wait until we have the production release stable before merging this to avoid having two styles in production

Agreed. That's why this is a draft PR. :)

toddburnside:
> Would it be possible to reference the prime-react files instead of copying them over? That way we don't have to manually keep them up to date.

Reference in what way?

rpiaggio:
> Reference in what way?

Are they available in an npm package?

toddburnside:
> Are they available in an npm package?

They aren't currently. I'm pretty sure that creating an npm package with the sass code (designer) would violate the license. But, it may be OK to create a package with the processed CSS file. I need to read the license again.

common/src/main/scala/explore/components/primereact/FormEnumDropdownView.scala
```diff
@@ -0,0 +1,60 @@
...
+import lucuma.core.util.Enumerated
+import react.common.*
+import react.primereact.PrimeStyles
+import reactST.primereact.selectitemMod.SelectItem
+
+import scalajs.js
+import scalajs.js.JSConverters.*
+import lucuma.core.enums.Half.A
+
+final case class FormEnumDropdownView[A](
```
toddburnside, line 21: For this and the option drop down view for enums, the only reason to have a `Form` version and a `non-Form` version is that the `non-Form` version can return the builder, while the `Form` version can't because it also has a label. If we do a full facade or don't expose the builder, there is probably no real reason to have the `non-Form` variant.

common/src/main/scala/explore/components/primereact/FormInputText.scala
```diff
@@ -0,0 +1,76 @@
...
+import react.primereact.InputText
+import react.primereact.PrimeStyles
+import reactST.primereact.components.{Button => CButton}
+
+import scalajs.js
+
+final case class FormInputText(
+  id:               NonEmptyString,
+  value:            js.UndefOr[String] = js.undefined,
+  label:            js.UndefOr[TagMod] = js.undefined,
```
toddburnside, line 25: Having the label here can reduce some code while using, but more important in my opinion is that it ensures that the id of the input matches the `htmlFor` of the label for accessibility reasons.

common/src/main/webapp/sass/explore-primereact.scss
```diff
@@ -0,0 +1,164 @@
+body {
```
rpiaggio, line 1: We don't need to enclose the theme imports in `body`. It will save us some space since `body` is copied to every selector in the theme.

explore/src/main/webapp/main.jsx
```diff
@@ -1,4 +1,7 @@
 import "/common/theme/semantic.less";
+import "/common/sass/explore-primereact.scss";
+import "primereact/resources/primereact.min.css"
+import "primeicons/primeicons.css"
```
rpiaggio, line 4: Do you think we can avoid importing prime icons?

rpiaggio (approved):
LGTM

PR #2203, opened 2022-09-10
Dynamic loading help

cquiroz:
This is a bit of a test and would like input from @rpiaggio @toddburnside and @armanbilge 
This reverts to do dynamic loading of the help thus requiring us to use scala 3.2.1-RC1

It seems we may want to do more dynamic loading to reduce the time spent compiling. I'm roughly thinking we could dynamic load each tab. WDYT?

I think this would preclude us of using GCC for compression, right?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [3.2.0]
+        scala: [3.2.1-RC1]
         java: [temurin@17]
         project: [rootJS, rootJVM]
     runs-on: ${{ matrix.os }}
@@ -100,7 +100,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [3.2.0]
+        scala: [3.2.1-RC1]
         java: [temurin@17]
     runs-on: ${{ matrix.os }}
     steps:
@@ -187,7 +187,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [3.2.0]
+        scala: [3.2.1-RC1]
         java: [temurin@17]
     runs-on: ${{ matrix.os }}
     steps:
@@ -255,7 +255,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [3.2.0]
+        scala: [3.2.1-RC1]
         java: [temurin@17]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,8 +12,8 @@ pull_request_rules:
   - or:
     - body~=labels:.*early-semver-patch
     - body~=labels:.*early-semver-minor
-  - status-success=Build and Test (ubuntu-latest, 3.2.0, temurin@17, rootJS)
-  - status-success=Build and Test (ubuntu-latest, 3.2.0, temurin@17, rootJVM)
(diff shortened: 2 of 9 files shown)


armanbilge:
> It seems we may want to do more dynamic loading to reduce the time spent compiling.

Sorry, I need more context: what does dynamic loading have to do with compiling? Do you mean linking? How does it affect linking?

---

> I think this would preclude us of using GCC for compression, right?

Yes, if we do dynamic loading, this does preclude us from using GCC.

However, I spent this past week poking at GCC and to be honest I don't understand why this can't work. GCC supports chunking, dynamic imports, and ES modules. ES modules are obviously the future and it makes more sense to fix Scala.js instead of warping our pipeline to workaround this.

cquiroz:
We've noticed that in development fastLinkJS is getting rather slow. I conjecture (no data to back it up) that we could break it in dynamic chunks to make it faster

That will probably make small chunks but would be bigger overall, I think some dependencies will be in multiple chunks though I'm not really sure

armanbilge:
> I conjecture (no data to back it up) that we could break it in dynamic chunks to make it faster

If this actually works I'd be surprised :)

> That will probably make small chunks but would be bigger overall, I think some dependencies will be in multiple chunks though I'm not really sure

No, I don't think that's correct. I believe the linker will create additional internal chunks that can be used to share dependencies.

However, it is very possible that the overall size will be bigger, because adding the dynamic loading boundary prevents optimizations across that boundary.

rpiaggio:
Any insights as to why the new chunk has 141.45KB while the main chunk only decreases by 86.48KB ?

@armanbilge I think the rationale to investigate this a bit further is:

Linking the whole app takes between 1 and 2 minutes (with performance degrading over time as long as you don't reset the JVM). This becomes quite unbearable when developing.

If we split into dynamically loaded chunks, we will only need to link the chunks we modify, hopefully speeding up the development cycle.

rpiaggio (approved):
LGTM

armanbilge:
> Any insights as to why the new chunk has 141.45KB while the main chunk only decreases by 86.48KB ?

If I had to guess, it's for the reason I hypothesized in https://github.com/gemini-hlsw/explore/pull/2203#issuecomment-1242812361

> However, it is very possible that the overall size will be bigger, because adding the dynamic loading boundary prevents optimizations across that boundary.

---

> This becomes quite unbearable when developing.

> If we split into dynamically loaded chunks, we will only need to link the chunks we modify, hopefully speeding up the development cycle.

Are you using small-modules-for when developing? That should already be creating chunks, no?

https://github.com/gemini-hlsw/explore/blob/b6974a640f519d2fcb41a4363dd5dd8a83ee1441/build.sbt#L254-L257

Additionally, it's not obvious to me how dynamic loading (which is a runtime JS concept) relates to linking (which is a Scala.js SJSIR-level concept).

Finally, it seems like these changes which are motivated by dev/debug needs are impacting the production deploy.

I'd be happy to be wrong, but I'm struggling to see how it all fits together. Of course, if empirically this is solving your problems that's fantastic :)

cquiroz:
I guess we should experiment a bit on whether this is really helping or just an idea will little practical impart. We may want to switch one tab to see the if we feel there is a benefit

rpiaggio:
@armanbilge I was hoping the Scala.js linker detected dynamic loading boundaries so that when we modify something it only linked the affected chunk, thus speeding the linker. But that doesn't seem to be the case :( (at least with smallest-modules-for).

cquiroz:
I'm going to close this, using the FewestModules mode seems more appropriate

PR #2213, opened 2022-09-12
Use middle of the coverage range for itc

cquiroz:
We just learned that we should use the center of coverage wavelength for ITC requests rather than the requested wavelength.
They often match but not all the time.

This only fixes the issue for the table, we need another fix for the graph eventually

### DIFF
--- a/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
+++ b/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
@@ -78,7 +78,7 @@ import scala.concurrent.duration.*
 
 import scalajs.js.|
 
-final case class SpectroscopyModesTable(
+case class SpectroscopyModesTable(
   scienceMode:              View[Option[ScienceMode]],
   spectroscopyRequirements: SpectroscopyRequirementsData,
   constraints:              ConstraintSet,
@@ -87,7 +87,7 @@ final case class SpectroscopyModesTable(
   matrix:                   SpectroscopyModesMatrix,
   onSelect:                 Callback
 )(using val ctx:            AppContextIO)
-    extends ReactFnProps[SpectroscopyModesTable](SpectroscopyModesTable.component) {
+    extends ReactFnProps(SpectroscopyModesTable.component) {
   val brightestTarget: Option[ItcTarget] =
     for
       w  <- spectroscopyRequirements.wavelength
@@ -97,7 +97,7 @@ final case class SpectroscopyModesTable(
 }
 
 object SpectroscopyModesTable {
-  type Props = SpectroscopyModesTable
+  private type Props = SpectroscopyModesTable
 
   type ColId = NonEmptyString
 
@@ -190,11 +190,10 @@ object SpectroscopyModesTable {
     case (i @ Instrument.Gnirs, m) => s"${i.longName} $m"
     case (i, _)                    => i.longName
 
-  def formatFPU(r: FocalPlane): String = r match {
+  def formatFPU(r: FocalPlane): String = r match
     case FocalPlane.SingleSlit   => "Single"
     case FocalPlane.MultipleSlit => "Multi"
     case FocalPlane.IFU          => "IFU"
-  }
 
   private def itcCell(c: EitherNec[ItcQueryProblems, ItcResult]): VdomElement =
     val content: TagMod = c match {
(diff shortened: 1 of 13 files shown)


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -97,7 +97,7 @@ final case class SpectroscopyModesTable(
 }
 
 object SpectroscopyModesTable {
-  type Props = SpectroscopyModesTable
+  private type Props = SpectroscopyModesTable
 
   type ColId = NonEmptyString
```
rpiaggio, line 102: All members can probably be `private` too.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -338,53 +337,54 @@ object SpectroscopyModesTable {
...
+              basic = ScienceModeBasic.GmosSouthLongSlit(grating, filter, fpu),
+              advanced = ScienceModeAdvanced.GmosSouthLongSlit.Empty
+            )
+            .some
+        case _ => none
+      }
 
-  protected def equalsConf(row: SpectroscopyModeRow, conf: ScienceMode): Boolean =
-    rowToConf(row).exists(_ === conf)
+    def equalsConf(conf: ScienceMode): Boolean =
```
rpiaggio, line 362: Oh wait, are these invoked from somewhere else?

model/shared/src/main/scala/explore/model/boopickle/CommonPicklers.scala
```diff
@@ -80,6 +81,9 @@ trait CommonPicklers {
       Enumerated[A].tag(_)
     )
 
+  def picklerNewType[A: Pickler](nt: NewType[A]): Pickler[nt.Type] =
```
rpiaggio, line 84: Can this be `given` so we don't have to manually define instances?

model/shared/src/main/scala/explore/model/itc/ItcRequestParams.scala
```diff
@@ -11,17 +11,21 @@ import explore.modes.InstrumentRow
 import lucuma.core.math.Wavelength
 import lucuma.core.model.ConstraintSet
 import lucuma.core.model.NonNegDuration
+import lucuma.utils.NewType
+
+object CoverageCenterWavelength extends NewType[Wavelength]
+type CoverageCenterWavelength = CoverageCenterWavelength.Type
```
rpiaggio, line 17: Nice

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/model/boopickle/CommonPicklers.scala
cquiroz: We can't pass an explicit parameter to define a given (I think) maybe we'd need to make `NewType` a typeclass for that

rpiaggio: Oh you are right, we would need a type class, which defeats the purpose.

I'm guessing as an alternative we could add to `NewType`:
`given pickler(using p: Pickler[Wrapped]): Pickler[Type]        = p`

(I'm wondering if we should do the same with `Eq`)

cquiroz: Good idea, let's try that

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: WDYT about declaring the whole `SpectroscopyModesTable` private. It will achieve the same and apply to everything in one go
See my last commit

rpiaggio: I think it's a good idea to declare the object private, but we should also make members private. Otherwise, they are still accessible from within the same package, which in most cases we don't want to allow.

model/shared/src/main/scala/explore/model/boopickle/CommonPicklers.scala
rpiaggio: The downside is that we make `NewType` depend on `Pickle`, which may be an issue if we want to move `NewType` to `lucuma-core`.

PR #2229, opened 2022-09-16
Switch to fewest modules linking in dev.

rpiaggio:
Linking in dev with fewest modules will make HMR unfeasible, but it seems to speed up the fast linker considerably, with times of 7s when making a small change in code, vs 90s we had with smallest-modules-for.


cquiroz (approved):
Yes, let's test it

PR #2242, opened 2022-09-20
Overlay all science targets on aladin

cquiroz:
This is a change that in paper seemed simple but ended up being fairly large (thus the PR size)

When we have an asterism for Aladin and AGS they are treated as independent targets while instead you want to do both (vizualization and ags) on the base coordinates of the asterism (Understood as the calculated center)

image (2)

This is a fairly large change and it will likely lead to regression, thus testing will be needed


cquiroz:
https://user-images.githubusercontent.com/3615303/191371461-36043cb1-c108-466f-8bad-e111e5a44398.mp4

toddburnside (approved):
I won't pretend to understand all the changes but the code looks fine and I'll trust that any issues are uncovered by testing. :)

PR #2245, opened 2022-09-22
Fixes elevation plot regression

cquiroz:
This fixes a small regression where the elevation plot wouldn't get the appropriate coordinates of the current target/asterism

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -154,19 +154,24 @@ object ObsTabTiles {
             ).zoom(Asterism.fromTargetsListOn(props.focusedTarget).asLens)
           )
 
-        val vizTimeView: Pot[View[Option[Instant]]] =
-          obsViewPot.map(_.zoom(ObsEditData.visualizationTime))
-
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -154,19 +154,25 @@ object ObsTabTiles {
...
 
+        val vizTimeView: Pot[View[Option[Instant]]] =
+          obsViewPot.map(_.zoom(ObsEditData.visualizationTime))
+
         val vizTime = vizTimeView.toOption.flatMap(_.get)
 
-        // base coordinates corrected to vizTime
+        // asterism base coordinates at viz time or default to base coordinates
         val targetCoords: Option[CoordinatesAtVizTime] =
+          potAsterism.toOption.flatMap(_.get.map(_.baseTracking.baseCoordinates))
```
toddburnside, line 167: I'm not following this. It doesn't seem to be assigned to anything so does it actually do anything?

cquiroz: Uh, good catch, i restored this line by mistake

cquiroz: I removed the offending line

PR #2246, opened 2022-09-22
Compress data stored in indexed db via wasm-flate

cquiroz:
I've been learning a bit about wasm and thought we could benefit in some cases.
there is a compression wasm library we can use to compress the data stored in indexdb,[
We have seen an example using `pako` to do the same but `wasm-flate` claims to be seven time faster

image

This is also to test the waters about using `wasm` in places where it can be useful


cquiroz:
Some measurements on size reduction

ITC chart data from 190471 to 108419
AGS results from 2026 to 1107

cquiroz:
It seems this is blocked by
https://github.com/Menci/vite-plugin-wasm/issues/13

PR #2249, opened 2022-09-23
Better responsivness for the obs time editor

cquiroz:
These are just small improvements to the obs time editor, especially when we have reduced screen sizes

I also created a new `explore.scss` to start migrating the styles in less

Before
image

After
Explore 2022-09-23 10-53-55


toddburnside (approved):
LGTM

PR #2250, opened 2022-09-23
Update scala to 3.2.1-RC2, plus lucuma-react and other libs

toddburnside:


project/Settings.scala
```diff
@@ -217,6 +217,7 @@ object Settings {
 
     val LucumaReact = Def.setting(
       deps(
+        "edu.gemini" %%% "lucuma-react-table",
```
toddburnside, line 220: Reordered these dependencies to get rid of a compiler AssertionError. It is probably related to the ScalablyTyped generated facades having duplicates. A recent change to lucuma-react changed all of the generated facades to use `reactST` as the prefix. I think this changed the issues the scala 3 compiler is having.

cquiroz (approved):
LGTM

PR #2251, opened 2022-09-23
Fix tooltip at extreme zoom levels

cquiroz:
The target tooltip may get distorted at high magnification levels

This PR may help solve it

Before:
image

After:
image


toddburnside (approved):
I kind of like the `before` 😆 , but LGTM.

PR #2258, opened 2022-09-23
Vite plugin take 2

cquiroz:
This PR transform build.js and vite config to ES modules to let us upgrade vite-plugin-font and avoid having to do `npm install --force`


toddburnside (approved):
LGTM

PR #2264, opened 2022-09-26
Refactor common code and improve constraints on mobiles

cquiroz:
This is an internal improvement to share code related to the two panels in constraints and observations (target is pending)
It also improves the behavior of the constraints panel in small screens

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -338,6 +338,10 @@ tfoot {
   margin-top: 0;
   cursor: pointer;
   min-height: 36px;
+  flex-shrink: 2;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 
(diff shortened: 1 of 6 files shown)


common/src/main/webapp/less/style.less
```diff
@@ -338,6 +338,10 @@ tfoot {
   margin-top: 0;
   cursor: pointer;
   min-height: 36px;
+  flex-shrink: 2;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
```
rpiaggio, line 344: These 3 properties seem like something we want to use in a number of places. Do we want a mixin?

rpiaggio (approved):
LGTM

common/src/main/webapp/less/style.less
rpiaggio: Yes, I misread. I though we were on sass here.

PR #2265, opened 2022-09-26
Rewording of the add target dialog box

cquiroz:


cquiroz:
Explore 2022-09-26 16-46-13

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
```diff
@@ -53,6 +54,18 @@ case class TargetSelectionPopup(
 )(using val ctx: AppContextIO)
     extends ReactFnProps(TargetSelectionPopup.component)
 
+object SearchingState extends NewType[Boolean]:
+  val Searching: SearchingState = SearchingState(true)
+  val Idle: SearchingState      = SearchingState(false)
```
rpiaggio, line 59: I have a bit of a conflict with defining `val` constants for `NewType`s since they grow the compiled code. I know it's not 100% the same thing, but maybe we can replace with `inline def`s ?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
cquiroz: Do you think they will be smaller?

PR #2267, opened 2022-09-27
Allow disabling scroll to zoom in aladin

cquiroz:
Mouse interaction with aladin can now be disabled with a global (per user) setting.
interaction is also disabled on mobiles as it is very hard to scroll if you have aladin on the viewport

This also include settings for a separate production user prefs server and a little bit of documentation about user preferences in general


toddburnside (approved):
LGTM

rpiaggio (approved):
LGTM

PR #2282, opened 2022-09-28
Make some of the target prefs optional

cquiroz:
The target preferences table is getting a bit too big and having all the fields as required complicates the code a bit
This PR removes that restriction and it also setup the user prefs model to be properly handled via hasura migrations


common-graphql/src/main/resources/graphql/schemas/UserPreferencesDB.graphql
```diff
@@ -1570,19 +1571,18 @@ input lucuma_target_pk_columns_input {
 columns and relationships of "lucuma_target_preferences"
 """
 type lucuma_target_preferences {
-  agsCandidates: Boolean!
-  agsOverlay: Boolean!
-  fov: bigint
-  fovDec: bigint!
-  fovRA: bigint!
-  fullScreen: Boolean!
+  agsCandidates: Boolean
```
rpiaggio, line 1574: There's a `fetchUserPreferencesSchema.sh` so that we can import the schema from Hasura and avoid us the burden of maintaining it manually.

Maybe there's a way to declare the fields as optional in Hasura?

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -142,13 +142,14 @@ object AladinCell extends ModelOptics {
...
           import props.given
 
-          TargetPreferences
-            .queryWithDefault[IO](props.uid, props.tid, Constants.InitialFov)
-            .flatMap { (up, tp) =>
-              options
-                .set((up, tp).ready)
-                .to[IO]
-            }
+          IO.println(Constants.InitialFov.toMicroarcseconds) *>
```
rpiaggio, line 145: println

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -334,17 +325,15 @@ object AladinCell extends ModelOptics {
             overlay:    Visible => Visible,
             fullScreen: Boolean => Boolean
           ): Callback =
-            (fovView.get, agsCandidatesView.get, agsOverlayView.get, fullScreenView.get).mapN {
-              (f, a, o, s) =>
+            (agsCandidatesView.get, agsOverlayView.get, fullScreenView.get).mapN { (a, o, s) =>
+              Callback.log("Set prefs") *>
```
rpiaggio, line 329: log

rpiaggio (approved):
LGTM

PR #2285, opened 2022-09-29
Select the first available target for the elevation plot

cquiroz:
This is a small fix where the elevation plot is not show if the selection is focused on the obs though there are available targets


rpiaggio (approved):
LGTM

PR #2286, opened 2022-09-30
Better responsivness for the itc title bar

cquiroz:
Also the dropdown is switched to prime

Before:
Explore 2022-09-29 17-45-47

After:
image


rpiaggio (approved):
LGTM

PR #2287, opened 2022-09-30
Make elevation plot settings global

cquiroz:


common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -331,56 +329,41 @@ object UserPreferencesQueries {
 
   object ElevationPlotPreference:
     def updatePlotPreferences[F[_]: ApplicativeThrow](
-      uid:      User.Id,
-      targetId: Target.Id,
-      site:     Site,
-      range:    PlotRange,
-      time:     TimeDisplay
+      userId: User.Id,
+      site:   Site,
```
rpiaggio, line 333: Maybe we should keep the site bound to the target?

cquiroz: It was requested to make it global

rpiaggio (approved):
LGTM

cquiroz:
After some discussion, the site won't be stored as a persistent setting

PR #2293, opened 2022-10-03
Default to erg/s/cm2/A unit for emission lines

cquiroz:


common/src/main/scala/explore/model/enums/SEDType.scala
```diff
@@ -73,7 +73,7 @@ sealed abstract class SEDTypeEnum[T](implicit
         _ =>
           EmissionLines[T](
             SortedMap.empty,
-            enumFDCUnits.all.head.withValueTagged(BigDecimal(1).refined[Positive])
+            enumFDCUnits.all.last.withValueTagged(BigDecimal(1).refined[Positive])
```
rpiaggio, line 76: I suggest we explicitly declare the default value.

Otherwise this may break if we reshuffle the enum order, or add a new one at the end.

PR #2294, opened 2022-10-03
Itc improvements

cquiroz:
This PR contains two improvements:
* The title bar for ITC while now show when we are retrieving data or in an error
* Move the calculation of the wavelength at max to the backend
Pending on 
https://github.com/gemini-hlsw/lucuma-itc/pull/310

### DIFF
--- a/explore/src/main/scala/explore/itc/ItcPanelTitle.scala
+++ b/explore/src/main/scala/explore/itc/ItcPanelTitle.scala
@@ -6,11 +6,13 @@ package explore.itc
 import cats.effect.IO
 import cats.syntax.all.*
 import crystal.Pot
+import crystal.implicits.*
 import crystal.react.View
 import crystal.react.hooks.*
 import crystal.react.implicits.*
 import eu.timepit.refined.*
 import eu.timepit.refined.numeric.Positive
+import explore.Icons
 import explore.common.ObsQueries.*
 import explore.components.ui.ExploreStyles
 import explore.events.*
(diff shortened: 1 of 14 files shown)


explore/src/main/scala/explore/itc/ItcPanelTitle.scala
```diff
@@ -60,53 +69,77 @@ object ItcPanelTitle:
...
           ExploreStyles.ItcTileTitle,
           <.label(s"Target:"),
           Dropdown(
             clazz = ExploreStyles.ItcTileTargetSelector,
-            value = props.selectedTarget.get.orUndefined,
+            value = idx,
             onChange = {
-              case t @ ItcTarget(_, _, _) => props.selectedTarget.set(newSelected(t))
-              case _                      => Callback.empty
+              case t: Int => props.selectedTarget.set(newSelected(t))
```
rpiaggio, line 135: Why was it necessary to switch to value to an index? Just to know if this is something to consider in the future.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/itc/ItcPanelTitle.scala
cquiroz: Right, this was tricky.
The object set in value is compared (I think) internally as js objects which make it not match as you'd expect a scala object .
We may want to have a `Droopdown[A]` eventually

toddburnside (approved):
LGTM

explore/src/main/scala/explore/itc/ItcPanelTitle.scala
toddburnside: I have already changed it to `Dropdown[A]` in my recent work to make it safer. There will be a `Dropdown[A]` and a `DropdownOptional[A]`

cquiroz: Oh great, would it it take an `Eq[A]`?

toddburnside: When I upgrade explore to the new lucuma-react, I'll see if this works correctly.

toddburnside: > Oh great, would it it take an `Eq[A]`?

Hmm. It doesn't at the moment. So, the problem you are having is that the correct `SelectItem` isn't selected for the target?

PR #2299, opened 2022-10-04
Aladin image zooms when changing constraints

cquiroz:
There is a cycle with Aladin where you ask it to use a fov (x) and it sends you back what fov is actually using (y). They don't always match but we need to get the real value to react to zoom changes. The small differences add up to produce undesired zoom when e.g. resizing. Surprisingly this is more evident for Linux users.

This PR essentially skip recording very small fov changes.
Also it was noticed that the aladin window would flash in certain cases because of unnecessary re renderings.
To solve these `Reusability` was reintroduced.
I tested this quite a bit but there maybe some regression in certain cases

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
@@ -52,7 +52,7 @@ case class AladinContainer(
   // TODO Move the functionality of saving the FOV in ALadincell here
   updateFov:              Function1[Fov, Callback],
   updateViewOffset:       Function1[Offset, Callback],
-  centerOnTarget:         View[Boolean],
+  centerOnTarget:         View[CenterTargetTrigger],
   selectedGuideStar:      Option[AgsAnalysis],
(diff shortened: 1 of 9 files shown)


toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -239,16 +244,7 @@ object AladinContainer {
       }
       // Use fov from aladin
       .useState(none[Fov])
-      // full screen trigger reflow
-      .useEffectWithDepsBy((props, _, _, _, _, _, _, _) => props.options.fullScreen)(
```
rpiaggio, line 243: Why is this no longer needed?

cquiroz: Full screen triggers a redraw when the parent changes it. This only made the flashing effect worse by doing 2 extra repaints

rpiaggio:
This looks like a good use for reusability :+1:

PR #2300, opened 2022-10-04
useContext

rpiaggio:
This is a refactor so that components obtain the application context using `useContext` instead of passing it implicitly with properties everywhere.

This is a more React-idiomatic way of dealing with context.

There are also some changes towards Scala 3 syntax and some reorganization of code.


.github/workflows/ci.yml
```diff
@@ -67,8 +67,8 @@ jobs:
           node-version: 16
           cache: npm
 
-      - name: npm install
-        run: npm install
+      - name: npm clean-install
```
rpiaggio, line 70: Unrelated. Thought this would be more appropriate for CI.

common-queries/src/main/scala/queries/schemas/odb/implicits.scala
```diff
@@ -4,97 +4,91 @@
...
 object implicits {
-  implicit class ObsIdOps(val id: Observation.Id) extends AnyVal {
+  extension (id: Observation.Id)
     def toWhereObservation: WhereObservation =
       WhereObservation(id = WhereOrderObservationId(EQ = id.assign).assign)
-  }
 
-  implicit class ObsIdListOps(val ids: List[Observation.Id]) extends AnyVal {
+  extension (ids: List[Observation.Id])
+    @targetName("ObservationId_toWhereObservation")
```
cquiroz, line 40: Isn't this only relevant to java users?

common/src/main/scala/explore/HelpContext.scala
```diff
@@ -0,0 +1,50 @@
...
+case class HelpContext(
+  rawUrl:        Uri,
+  editUrl:       Uri,
+  user:          String,
+  project:       String,
+  displayedHelp: View[Option[Help.Id]]
+)
+
+object HelpContext:
+  val ctx: Context[HelpContext] = React.createContext(null) // No default value
```
cquiroz, line 28: So this becomes available for use e.g. in other hooks?

common/src/main/scala/explore/HelpContext.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  // given Reusability[Help.Id]     = Reusability.by(_.value)
```
cquiroz, line 30: Commented code

cquiroz (approved):
LGTM

common-queries/src/main/scala/queries/schemas/odb/implicits.scala
toddburnside: Well, yes. But, it also seems to be a workaround for having extension methods with the same name. `toWhereObservation` in this case. It doesn't seem like it should be necessary since the actual generated methods have the type which is being extended as a parameter. But, maybe that overload is not sufficient for java-land.

common/src/main/scala/explore/common/ConstraintGroupQueries.scala
```diff
@@ -34,13 +33,11 @@ object ConstraintGroupQueries {
...
 
-  object ConstraintSummaryWithObervations {
+  object ConstraintSummaryWithObervations:
     val constraintGroups = Focus[ConstraintSummaryWithObervations](_.constraintGroups)
     val observations     = Focus[ConstraintSummaryWithObervations](_.observations)
-  }
 
-  implicit val constraintsSummWithObsReuse: Reusability[ConstraintSummaryWithObervations] =
-    Reusability.derive
+  // given Reusability[ConstraintSummaryWithObervations] = Reusability.derive
```
toddburnside: commented code

common-queries/src/main/scala/queries/schemas/odb/implicits.scala
rpiaggio: It doesn't compile otherwise. I think it's because the name and type are the same after erasure.

toddburnside (approved):
LGTM

common/src/main/scala/explore/HelpContext.scala
rpiaggio: Yup, if you `.useContext(HelpContext.ctx)` then this will be available as another parameter in subsequent hooks (and render).

common/src/main/scala/explore/model/AppContext.scala
```diff
@@ -29,19 +33,29 @@ case class AppContext[F[_]](
...
+  given WebSocketClient[F, UserPreferencesDB] = clients.preferencesDB
+  given TransactionalClient[F, ITC]           = clients.itc
+
+  given itcWorker: WorkerClient[F, ItcMessage.Request]         = workerClients.itc
+  given catalogWorker: WorkerClient[F, CatalogMessage.Request] = workerClients.catalog
+  given agsWorker: WorkerClient[F, AgsMessage.Request]         = workerClients.ags
+  given plotWorker: WorkerClient[F, PlotMessage.Request]       = workerClients.plot
+
+object AppContext:
+  val ctx: Context[AppContext[IO]] = React.createContext(null) // No default value
```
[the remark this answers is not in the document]
rpiaggio: It's set in scope here: https://github.com/gemini-hlsw/explore/blob/contexts/explore/src/main/scala/explore/RootComponent.scala#L33

The default value (`null` in this case) means that if you attempt `.useContext` from a component that is not embedded within a `provide`, then it will get the default value.

If we wanted, we could run the whole app with another `AppContext` (say, a testing one) just by providing it instead of our current one.

PR #2301, opened 2022-10-04
Change user db to use camelCase

cquiroz:
Hasura tends to prefer `snake_case` but graphql tends to prefer `camelCase` and the odb does too
This PR is mechanical to rename all tables and fields to `camelCase`
There maybe a need to do more of these refactorings but this is the starting point


rpiaggio (approved):
LGTM

PR #2304, opened 2022-10-04
Separate toInput extensions again

rpiaggio:
Due to the way extension methods work, I joined all the `toInput` methods in the last PR since I couldn't get it to compile if kept separate.

See https://github.com/lampepfl/dotty/issues/15691#issuecomment-1186560928

However, it is desirable to have separately defined `toInput` methods for model classes (which we will move to `lucuma-schemas`) and for classes defined here. This PR separates them again using the suggestion from the above link to compose extension collections.


cquiroz (approved):
LGTM

PR #2305, opened 2022-10-05
Switch facade for hotkeys and setup some shortcuts

cquiroz:
`react-hotkeys` is not actively maintained so I'm switching to `react-hook-hotkeys` which is
I added some shortcuts to switch between tabs though I'm not going to promote them yet


toddburnside (approved):
LGTM

PR #2309, opened 2022-10-06
Global hotkeys

cquiroz:
Add new hooks to do global shortcuts and also to support the `Deps` mode which is needed as the hook memoizes the callbacks
Thanks @rpiaggio for the improved hooks


explore/src/main/scala/explore/HelpBody.scala
```diff
@@ -70,8 +70,10 @@ object HelpBody:
       .useEffectOnMountBy { (props, _, state) =>
         load(props.url).flatMap(v => state.set(Pot.fromTry(v)).to[IO])
       }
-      .useHotkeysBy((_, helpCtx, _) => UseHotkeysProps("esc", helpCtx.displayedHelp.set(none)))
-      .render { (props, helpCtx, state, _) =>
+      .useGlobalHotkeysBy((_, helpCtx, _) =>
```
rpiaggio, line 73: Just a note: the way `WithDeps` hook is implemented now implies the semantics here have changed: Now the callback is not being reused (before it was).

Maybe we want to change the default behavior when no deps are specified; or maybe we can make this explicit by having eg `useGlobalHotkeysAlwaysBy`.

BTW: Right now, usability can be achieved with `.useGobalHotkeysWithDepsBy( (_, _, _) => ()) ( ...)`

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/HelpBody.scala
cquiroz: Do you mean the callback is not reused even if the dependencies don't change?
I guess that's fine, I'm not totally sure about the benefit of reusing the callback in general

rpiaggio: >Do you mean the callback is not reused even if the dependencies don't change?

Exactly. 

I don't know what the impact is in the hook of not memoizing either.

PR #2311, opened 2022-10-06
Apply observation to a target

cquiroz:
Initial version of the ability to copy paste observations into other targets
It sets up some shortcuts to do the operation
There are a few things missing though:
* No undo support
* After the obs is cloned it takes a while for the UI to update
* not possible to copy/paste across browser tabs

Despite the shortcomings it is good for testing


toddburnside (approved):
LGTM

PR #2315, opened 2022-10-10
Fix center target feature

cquiroz:
This is a regression from a previous failure and also it brings several simplifications


rpiaggio (approved):
LGTM

PR #2324, opened 2022-10-11
New, prime, tables

rpiaggio:
- Migrate all tables to new version of `react-tabe` + `react-virtual`, rendering with `react-prime` styles.
- Column selector was extracted into a component in `lucuma-ui`.

A word about `SpectroscopyModesTable`: Behavior was modified slightly. Visible rows are no longer prioritized when querying ITC. This is because our way of prioritizing was cancelling current queries and querying again in different order upon scroll. This could cause a lot of cancelled requests if the user continually scrolled. We can reimplement priorization with a different approach, ideally without cancelling requests already in flight.


common/src/main/scala/explore/Icons.scala
```diff
@@ -130,18 +130,6 @@ object Icons {
   @JSImport("@fortawesome/pro-solid-svg-icons", "faCheck")
   val faCheck: FAIcon = js.native
 
-  @js.native
-  @JSImport("@fortawesome/pro-solid-svg-icons", "faSort")
```
cquiroz, line 134: Are these provided by prime?

rpiaggio: We are still using the same FA icons, but they are declared in `lucuma-ui` now.

common/src/main/webapp/sass/explore.scss
```diff
@@ -222,13 +224,11 @@ $search-preview-margin: 5px;
...
   align-items: center;
-  background: repeating-linear-gradient(
-    -45deg,
-    black,
-    black 10px,
-    var(--site-border-color) 10px,
-    var(--site-border-color) 11px
-  );
+  background: repeating-linear-gradient(-45deg,
+      black,
```
cquiroz, line 228: I guess this should be a theme variable?

cquiroz (approved):
LGTM. It is a bit hard to check everything but so far it seems fine
Thanks a lot for doing this

common/src/main/webapp/sass/explore.scss
rpiaggio: Agreed. This was not introduced in this PR though. It was autoformatted by prettier.

toddburnside (approved):
LGTM

PR #2326, opened 2022-10-12
Small touches to tables

cquiroz:
This makes the hover on tables work better (sometimes you can't see both insets) and adds borders for a few tables

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -340,3 +340,53 @@ $search-preview-margin: 5px;
 .explore-warning-icon {
   color: var(--warning-background-color);
 }
+
+// ------
+// Generic table
+// ------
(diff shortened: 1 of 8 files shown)


rpiaggio (approved):
LGTM

common/src/main/webapp/sass/explore.scss
```diff
@@ -339,4 +339,54 @@ $search-preview-margin: 5px;
...
+.explore-border-table,
+table.explore-border-table {
+  border: 1px solid var(--site-border-color);
+  border-radius: var(--border-radius);
+}
+
+// -------------
+// Move to lucuma-ui
+// -------------
+.pl-react-table {
```
cquiroz, line 386: This should go to the lucuma-ui css, how are we planning to handle these?

PR #2332, opened 2022-10-16
Fix grating sorting on the configurations table

cquiroz:


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -278,7 +278,8 @@ private object SpectroscopyModesTable:
           maxSize = 105
         ),
       column(GratingColumnId, SpectroscopyModeRow.grating.get)
-        .copy(cell = cell => formatGrating(cell.value), size = 96, minSize = 96, maxSize = 96),
+        .copy(cell = cell => formatGrating(cell.value), size = 96, minSize = 96, maxSize = 96)
+        .sortableBy(formatGrating),
```
cquiroz, line 282: This seems to fix but seems a bit odd, shouldn't it sort already by the cell string value?

rpiaggio: The value of the cells is what is returned by `SpectroscopyModeRow.grating.get` which returns a `Grating`. `react-table` doesn't know how to sort that since it's not a string.

The rendered value can't be used either since it's a `VdomNode` (which in this case it's implicitly obtained from a `String`).

We can pass an implicit `Ordering` by invoking `.sortable` (without parameters), which is what I think we should do whenever we have an `Ordering` defined (but doesn't seem to be the case here).

Alternatives here are: define an `Order` or `Ordering` and invoke `.sortable`. Or, we can set the cell value to the string value. Then we don't need to define a render function or a sorting function.

I still lean towards forcing us to explicitly define when a column is sortable and pass how it should be sorted. This would imply forcing `enableSorting = false` in our facade by default. WDYT?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: Sounds good, I'm   in the middle of another change to store the visibility preferences. I'll incorporate that change there, adding besides visibility properties per column, whether we allow sorting

cquiroz:
@rpiaggio I added `Ordering` for `Grating`, `Filter` and `FPU`
I also integrated the table into the preferences mini framework to remember the sorting criteria

rpiaggio (approved):
LGTM

PR #2334, opened 2022-10-16
Syntax changes

cquiroz:
Add a method to use pprint as Callback

### DIFF
--- a/common/src/main/scala/explore/syntax/ui/package.scala
+++ b/common/src/main/scala/explore/syntax/ui/package.scala
@@ -1,7 +1,7 @@
 // Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 // For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-package explore.syntax
+package explore.syntax.ui
 
 import cats.Eq
 import cats.syntax.all.*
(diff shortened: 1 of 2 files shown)


toddburnside (approved):
LGTM

common/src/main/scala/explore/syntax/ui/package.scala
```diff
@@ -1,80 +1,55 @@
...
-      val newClazz = input.clazz |+| ExploreStyles.ClearableInputPaddingReset
-      input.copy(icon = clearInputIcon[EV, A](input.value), clazz = newClazz)
-    }
-  }
-
-  extension [A](c: js.UndefOr[A => Callback])
-    def toJs: js.UndefOr[js.Function1[A, Unit]] = c.map(x => (a: A) => x(a).runNow())
-
-}
+extension (c: Callback.type) def pprintln[T](a: T): Callback = Callback(_root_.pprint.pprintln(a))
```
cquiroz, line 55: I think I'll move this to lucuma-ui

rpiaggio: Nice.

Can we have a `Logger.pdebug` too, please?

cquiroz: We have this too, is it in use?
https://github.com/gemini-hlsw/explore/blob/8f7d99556e53adc49f073e53184b258a7d26667a/common/src/main/scala/explore/utils/pprinter.scala

rpiaggio: I had completely forgotten about that.  It doesn't seem to be in use.

PR #2340, opened 2022-10-18
Table preferences

cquiroz:
This Pr adds storing and loading the table selected visible columns and sorting so on reload you recover your last setting.
It is only affecting the `ConstraintsTable` but the same logic should be applied to other tables.

I hope to generalize it for other tables as the behavior should be generic enough

There are also a couple of minor ui touches


explore/src/main/scala/ColumnSelector.scala
```diff
@@ -0,0 +1,62 @@
...
+import crystal.react.View
+import explore.model.TableColumnPref
+import japgolly.scalajs.react.*
+import japgolly.scalajs.react.vdom.html_<^.*
+import react.common.*
+import react.semanticui.modules.checkbox.Checkbox
+import react.semanticui.modules.dropdown.*
+import reactST.{tanstackTableCore => raw}
+
+case class NewColumnSelector[T](
```
rpiaggio, line 16: Can we keep using some version of the `ColumnSelector` we have in `lucuma-ui` so that we can share it with other projects and save preferences by using `withOnMod` on the `View`?

cquiroz: I was going to replace `ColumnSelector` with this one. I guess we can keep them both, Saving is now done using `withOnMod`

cquiroz: A big difference is that the old selector worked on a list of hidden columns, while this works on a list of all columns, you want to know all columns to save which ones are visible and which are not ideally in one go

rpiaggio: Ah, I see. Ideally we should keep only one version, and have it in lucuma-ui.

rpiaggio (approved):
LGTM

PR #2341, opened 2022-10-18
Properly set the fpu for ags calculations

cquiroz:
FPU wasn't properly being set for AGS calculations.
This PR also avoids some extra redraws when doing ags calculations and other changes


toddburnside (approved):
LGTM

PR #2343, opened 2022-10-18
Generalize table preferences

cquiroz:
This PR creates hooks to generalize the code to read/store table preferences and we add them to Target Summary and Asetrisms table


common/src/main/scala/explore/utils/TableHooks.scala
```diff
@@ -0,0 +1,99 @@
...
+trait TableHooks:
+  def toSortingRules(tuples: List[(String, Boolean)]): js.Array[raw.mod.ColumnSort] =
+    tuples.map { case (id, b) => raw.mod.ColumnSort(b, id) }.toJSArray
+
+  def fromTableState(state: raw.mod.TableState): List[(String, Boolean)] =
+    state.sorting.toList.map(sr => (sr.id.toString, sr.desc))
+
+  def useTablePreferencesLoad[T] =
+    CustomHook[
+      (Option[User.Id], AppContext[IO], TableId, List[ColumnDef[T, ?]], List[TableColumnPref])
```
rpiaggio, line 41: Can we use a `case class` for the options? That will name the parameters, and the context can be given.

cquiroz: Sure. I avoided it to skip creating an object on call but it is probably clearer to have it

rpiaggio (approved):
LGTM, neat

I'd like to build upon this and attempt a unique hook for tables with persistent state, but we can merge this and do that later.

PR #2351, opened 2022-10-19
Alternative table state persistence.

rpiaggio:
This PR implements a single hook to use tables with an external persistent store for the table state (`useReactTableWithStateStore`). The hook is general and should be moved to `lucuma-ui` or maybe even `lucuma-react`. It uses a trait `TableStateStore`. An implementation is provided that uses our user preferences DB.

The table's internal state is used directly, also in the `ColumnSelector`, thus avoiding the need to keep our own state for visibility or sorting.

Finally, the priority in which columns are sorted is now persisted.


common-graphql/src/main/resources/graphql/schemas/UserPreferencesDB.graphql
```diff
@@ -1140,6 +1140,7 @@ type LucumaTableColumnPreferences {
   columnId: String!
   createdAt: timestamptz
   sorting: LucumaSortDirectionEnum
+  sortingOrder: Int
```
cquiroz, line 1143: Neat

common/src/main/scala/explore/common/UserPreferencesQueries.scala
```diff
@@ -422,12 +450,26 @@ object UserPreferencesQueries:
...
 
-  extension (w: TableColumnPref)
-    def toInput(u: User.Id, t: TableId): LucumaTableColumnPreferencesInsertInput =
-      LucumaTableColumnPreferencesInsertInput(
-        userId = u.show.assign,
-        tableId = t.assign,
-        columnId = w.columnId.value.assign,
-        visible = w.visible.assign,
-        sorting = w.sorting.orIgnore
+  extension (tableColsPrefs: List[TableColumnPreferencesQuery.Data.LucumaTableColumnPreferences])
```
cquiroz, line 453: Is this more of a UI syntax rather than queries?

common/src/main/scala/explore/utils/TableHooks.scala
```diff
@@ -4,104 +4,90 @@
...
-        TableColumns
-          .queryColumns[IO](params.userId, params.tid)
-          .flatMap(r =>
-            prefs
-              .set(allColumns.withStored(r.lucumaTableColumnPreferences).sortBy(_.columnId.value))
-              .to[IO]
+  private def hook[T] =
+    CustomHook[TableOptionsWithStateStore[DefaultA, T]]
+      .useReactTableBy(_.tableOptions)
+      .useState(false) // prefsDone
```
cquiroz, line 40: I'm finding clearer to use NewTypes instead of boolean for these states, WDYT?

explore/src/main/scala/ColumnSelector.scala
```diff
@@ -14,10 +13,9 @@ import react.semanticui.modules.dropdown.*
 import reactST.{tanstackTableCore => raw}
 
 case class NewColumnSelector[T](
-  table:         raw.mod.Table[T],
-  columnNames:   Map[String, String],
-  hiddenColumns: View[List[TableColumnPref]],
-  clazz:         Css = Css.Empty
+  table:       raw.mod.Table[T],
+  columnNames: Map[String, String],
```
cquiroz, line 17: is it a map of `name` to ...?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -454,36 +455,26 @@ private object SpectroscopyModesTable extends TableHooks:
...
-      .useReactTableBy((_, _, _, rows, _, _, cols, _, prefs) =>
-        TableOptions(
-          cols,
-          rows,
-          getRowId = (row, _, _) => row.id.toString,
-          enableSorting = true,
-          initialState = raw.mod
-            .InitialTableState()
-            .setSorting(toSortingRules(prefs.get.sortingColumns))
+      .useReactTableWithStateStoreBy((props, ctx, _, rows, _, _, cols, _) =>
```
cquiroz, line 459: Nice

common/src/main/scala/explore/common/UserPreferencesQueries.scala
rpiaggio: Not sure, these are extensions to the result type of `TableColumnPreferencesQuery`.

explore/src/main/scala/ColumnSelector.scala
rpiaggio: id -> label

There was a `ColumnId` newtype which is lost here. I do think we should use it though but it wasn't straightforward to use it consistently everywhere. I'm thinking of adding it to the facade.

common/src/main/scala/explore/common/UserPreferencesQueries.scala
cquiroz: yes but mixed with UI code for the table

rpiaggio: Yup, it translates the query result into table state. I'm open to moving it elsewhere. Do you have a suggestion? Its own file maybe?

explore/src/main/scala/ColumnSelector.scala
rpiaggio: I changed this to a function, it doesn't have to be a `Map`.

common/src/main/scala/explore/common/UserPreferencesQueries.scala
rpiaggio: I started extracting this and realized the whole implementation of `TableStateStore` (lines 392-445) has this characteristic of addressing both the user preferences DB and the table state. I suggest we leave everything together, either here, or moving the whole `TableStateStore` + extensions into a `TableColumnUserPreferences`.

After all, interacting with table state is as much UI as interacting with the grid layout or charts.

Maybe it will look better if we evolve the table facade to use friendlier types other than the `raw` ones, which I'd like to do at some point.

cquiroz: 👍

cquiroz (approved):
LGTM, remember to migrate the other user dbs

toddburnside (approved):
Very cool. LGTM

PR #2352, opened 2022-10-19
Update to Cats Effect 3.4.0-RC2

armanbilge:


explore/src/main/scala/explore/Explore.scala
```diff
@@ -189,7 +189,7 @@ object ExploreMain extends IOApp.Simple {
       }
 
     (for {
-      dispatcher       <- Dispatcher[IO]
+      dispatcher       <- Dispatcher.parallel[IO]
```
armanbilge, line 192: Cats Effect 3.4.0 introduces two new `Dispatcher` modes.
1. `Dispatcher.parallel`: creates a new fiber for each task and runs them in parallel
2. `Dispatcher.sequential`: runs tasks one-by-one on a single fiber

The "sequential" `Dispatcher` is far better optimized, because it does not have the overhead of a a fiber-per-task. But it is not safe to use it globally: while the `Dispatcher` is busy with one task (even if it is waiting for an async callback) it will be unable to run any other tasks.

Since these `Dispatcher`s are globally shared across many components, we can use the "parallel" `Dispatcher` for now. But in the long-term creating a "sequential" `Dispatcher` per-component can be more efficient.

rpiaggio (approved):
LGTM, this poses an interesting challenge of how to manage dispatchers.

Also, there's another dispatcher inside scalajs-react.

armanbilge:
> LGTM, this poses an interesting challenge of how to manage dispatchers.

Yes it does. The new sequential `Dispatcher` can be counter-intuitive: when I did the upgrade in FS2, in some places I realized that actually you want _two_ sequential `Dispatcher`s to replace a single parallel `Dispatcher` e.g. one for reading and one for writing.

- https://github.com/typelevel/fs2/pull/3000#discussion_r984074090

The new sequential mode heavily favors for creating highly-localized `Dispatcher`s that are effectively single-purpose.

toddburnside (approved):
LGTM

cquiroz (approved):
Yes, let's test this

PR #2353, opened 2022-10-19
Use primereact for View Sequence dialog

toddburnside:


cquiroz:
Screenshots?

toddburnside:
It doesn't look a lot different. Except for the cheerful button, of course.
image

common/src/main/scala/explore/components/ui/ExploreStyles.scala
```diff
@@ -371,6 +371,9 @@ object ExploreStyles {
   val TargetSummarySubRowCell: Css = Css("explore-target-summary-subrow-cell")
   val ConstraintsSummaryEdit: Css  = Css("explore-constraints-summary-edit")
 
+  object Dialog: // SUI has Mini, Tiny, Small, Large and Fullscreen.
+    val Small: Css = Css("pl-dialog-small")
```
cquiroz, line 375: Should these go to `PrimeStyle` eventually?

explore/src/main/scala/explore/config/SequenceEditorPopup.scala
```diff
@@ -36,41 +35,38 @@ object SequenceEditorPopup:
   private val component =
     ScalaFnComponent
       .withHooks[Props]
-      .useState(false)        // isOpen
+      .useStateView(false)    // isOpen
```
cquiroz, line 38: I've been trying to switch these to use `NewTypes` to make them more readable

cquiroz (approved):
LGTM

common/src/main/scala/explore/components/ui/ExploreStyles.scala
toddburnside: I don't know. The CSS depends on the breakpoints we set in explore. It doesn't seem like the `Css` should go there if it isn't backed by actual styles.

explore/src/main/scala/explore/config/SequenceEditorPopup.scala
toddburnside: Sure, I can switch those in the morning.

common/src/main/scala/explore/components/ui/ExploreStyles.scala
toddburnside: It's possible we could move it all to Lucuma UI to be used across lucuma projects.

PR #2357, opened 2022-10-20
Multiple selection

cquiroz:
This is to setup multiple selection on the target table. It doesn't do anything useful yet:

https://user-images.githubusercontent.com/3615303/196839006-0030cbca-5252-4da8-aeab-4c201849356a.mp4


common/src/main/scala/explore/components/ui/ExploreStyles.scala
```diff
@@ -328,6 +328,7 @@ object ExploreStyles {
   val ConfigurationFilter: Css          = Css("explore-configuration-filter") |+| Css("field")
   val ConfigurationFilterItem: Css      = Css("explore-configuration-filter-item")
   val ExploreTable: Css                 = Css("explore-table")
+  val ExploreSelectableTable: Css       = Css("explore-selectable-table")
```
rpiaggio, line 331: Can we put this in `lucuma-ui`?

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
```diff
@@ -141,7 +148,20 @@ object TargetSummaryTable extends TableHooks:
         <.div(
           props.renderInTitle(
             React.Fragment(
-              <.span, // Push column selector to right
+              <.div(
```
rpiaggio, line 151: Maybe we can put this in `lucuma-ui` too as a component.

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
```diff
@@ -155,11 +175,56 @@ object TargetSummaryTable extends TableHooks:
...
+                    else {
+                      val currentId      = row.id
+                      // selectedRow is not empty, these won't fail
+                      val first          = selectedRows.head.id
+                      val last           = selectedRows.last.id
+                      val indexOfCurrent = allRows.indexWhere(_.id === currentId)
+                      val indexOfFirst   = allRows.indexWhere(_.id === first)
+                      val indexOfLast    = allRows.indexWhere(_.id === last)
+                      if (indexOfCurrent === -1 || indexOfFirst === -1 || indexOfLast === -1) {
+                        println("Illegal selection state")
```
rpiaggio, line 203: println

rpiaggio (approved):
LGTM, looks great. 

It'd be nice if we could generalize this logic to make any table multi-selectable.

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
```diff
@@ -155,11 +175,56 @@ object TargetSummaryTable extends TableHooks:
...
+            rowMod = row =>
+              TagMod(
+                ExploreStyles.TableRowSelected.when_(row.getIsSelected()),
+                ^.onClick --> Callback.log(s"cmd ${isHotkeyPressed("cmd")}") *> Callback {
+                  // If cmd is pressed add to the selection
+                  if (!isCmdCtrlPressed) table.toggleAllRowsSelected(false)
+                  if (isShiftPressed) {
+                    // If shift is pressed extend
+                    val selectedRows = table.getSelectedRowModel().rows.toList.sortBy(_.id)
+                    val allRows      = table.getRowModel().rows.toList.sortBy(_.id)
```
toddburnside, line 192: Does this work if they have sorted the table by something other than ID?

cquiroz: Good question, let me review. It is also a question for generalization. can we assume rows have an id?

toddburnside:
Does the interaction between `Cmd click` and `Shift click` click work as expected? Don't you have to track the last selected row to make that work correctly?

cquiroz:
Do you mean if you press both cmd and shift? I guess in that case shift should win but I didn't test that either

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
toddburnside: You'd have to check the current table implementation, but I seem to remember the old table having an id. You could assign it, otherwise it would assign one itself. But, I may be thinking of a different table.

rpiaggio: All rows have an id. From the doc for `getRowId()`:

`This optional function is used to derive a unique ID for any given row. If not provided the rows index is used (nested rows join together with . using their grandparents' index eg. index.index.index)`

toddburnside:
> Do you mean if you press both cmd and shift? I guess in that case shift should win but I didn't test that either

I meant if you use cmd to select 2 non-contiguous rows, then use shift on a third row. I believe the idiomatic behavior is to select all the rows between the new click and the most recent previous click.

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
toddburnside: It seems like that would allow us to generalize the selection process.

cquiroz: Ok, I added a fix that relies on the position of the selected rows rather than the id and it seems to work

rpiaggio: Does it work with virtualized tables?

cquiroz: 🙉

PR #2359, opened 2022-10-21
Improvements to manually changed coordinates

cquiroz:
This fixes a regression where the aladin view doesn't update when coordinates change manually.
View offsets will be reset when the offset changes thus recentering the view


toddburnside (approved):
LGTM

PR #2363, opened 2022-10-21
Color code the guide speeds

cquiroz:
image


swalker2m:
Wow, cool.  Is the selected one blue?  How can you tell its guide speed?  Are out-of-range options always red regardless of guide speed?

cquiroz:
Yes, cyan is selected, I'm not sure maybe we just want to use the same color as corresponding to fast.
The speed can be seen on the overlay on the bottom light.
Out of rang will be red

swalker2m:
> Out of rang will be red

In the OT at least we were told to color code even the out of range ones so that they could tell whether there were good candidates just out of reach if they moved the PA or base or whatever.

common/src/main/webapp/sass/visualization.scss
```diff
@@ -161,22 +161,50 @@
...
       stroke-width: 0.5px;
-      fill: hsl(180deg 100% 50%);
+      stroke: hsl(120deg 60.7% 33.9%);
+      fill: var(--ags-fast-guide-speed-color);
+      fill-opacity: 70%;
     }
 
-    &.guide-star-candidate-target-visible {
-      display: inline;
+    // fast -> yellow
```
rpiaggio, line 176: Should the comment say `medium` ?

explore/src/main/scala/explore/targeteditor/AgsOverlay.scala
```diff
@@ -65,10 +66,14 @@ object AgsOverlay {
             ),
             <.div(
               ExploreStyles.AgsDescription,
-              analysis.quality match {
-                case AgsGuideQuality.DeliversRequestedIq =>
-                  Icons.CircleSmall.clazz(ExploreStyles.AgsGoodIQ)
-                case _                                   => ""
+              analysis.match {
+                case AgsAnalysis.Usable(_, _, Some(GuideSpeed.Fast), _, _)   =>
```
rpiaggio, line 70: Can we use the `case` clauses just for the class?

rpiaggio (approved):
LGTM

toddburnside (approved):
LGTM

cquiroz:
For performance reasons we are calculating whether a target falls into thte patrol field and only then we calculate quality.
To color the stars outside we'd need to calculate it for each.
The most expensive part is calculating vignetting maybe we could calculate speed but not vignetting?

swalker2m:
> The most expensive part is calculating vignetting maybe we could calculate speed but not vignetting?

Seems reasonable to me. The vignetting isn't really relevant if the probe can't reach the guide star anyway.  That is, it would likely change if the observation were edited such that the probe _could_ reach the candidate.

swalker2m:
Just an idea, but maybe out of reach candidates could be more transparent or something to distinguish them a bit.

cquiroz:
I guess you can have unusable guidestars inside the patrol field, right?

andrewwstephens:
> In the OT at least we were told to color code even the out of range ones so that they could tell whether there were good candidates just out of reach if they moved the PA or base or whatever.

I think we want to do the same thing here assuming it won't slow things down too much.

cquiroz:
I'll investigate the changes to the AGS algorithm in a separate ticket

PR #2368, opened 2022-10-23
Select the focused target on the target table

cquiroz:
Fixes a small bug on the target tab with asterisms

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
@@ -68,7 +68,7 @@ object AladinContainer {
   private given Reusability[Props]               =
     Reusability.by(x => (x.asterism, x.obsConf, x.allowMouseScroll, x.options))
   private given Reusability[Fov]                 = Reusability.by(x => (x.y, x.y))
-  private given Reusability[Asterism]            = Reusability.by(_.toSiderealTracking)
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -68,7 +68,7 @@ object AladinContainer {
   private given Reusability[Props]               =
     Reusability.by(x => (x.asterism, x.obsConf, x.allowMouseScroll, x.options))
   private given Reusability[Fov]                 = Reusability.by(x => (x.y, x.y))
-  private given Reusability[Asterism]            = Reusability.by(_.toSiderealTracking)
+  private given Reusability[Asterism]            = Reusability.by(x => (x.toSiderealTracking, x.focus.id))
```
toddburnside, line 71: Since it is used in both places, should `Reusability[Asterism]` be moved someplace common? Or, if it is specific enough to Aladin to keep it isolated but it needs to be the same in these 2 places, maybe it could be passed as a given from `AladinCell` to `AladinContainer`?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
cquiroz: Yes, it is aladin specific, normally you'd you use `byEq` but I don't want a refresh if e.g. you change the brightness

PR #2380, opened 2022-10-26
Support target import of csv files

cquiroz:
This is the first PR to support target import. This lets you open a csv file with name/ra/dec columns and import the targets.

In future updates we'll support importing brightnesses, etc

The current API only accepts adding targets one by one thus this gets very slow when adding, e.g. 1000 targets but it is very usable for a few hundreds.

https://user-images.githubusercontent.com/3615303/198096928-9b57cd51-d37e-4e3c-8729-9ea914e78b0b.mp4

### DIFF
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -61,9 +61,9 @@ import react.draggable.Axis
 import react.gridlayout.*
 import react.hotkeys.*
 import react.hotkeys.hooks.*
+import react.primereact.PrimeStyles
 import react.primereact.Toast
 import react.primereact.hooks.all.*
-import react.primereact.PrimeStyles
 import react.resizable.*
 import react.resizeDetector.*
 import react.resizeDetector.hooks.*
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -63,6 +63,7 @@ import react.hotkeys.*
 import react.hotkeys.hooks.*
 import react.primereact.Toast
 import react.primereact.hooks.all.*
+import react.primereact.PrimeStyles
 import react.resizable.*
 import react.resizeDetector.*
 import react.resizeDetector.hooks.*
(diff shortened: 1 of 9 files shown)


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -61,6 +61,7 @@ import react.draggable.Axis
 import react.gridlayout.*
 import react.hotkeys.*
 import react.hotkeys.hooks.*
+import react.primereact.PrimeStyles
```
toddburnside, line 64: Is this being used here?

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
```diff
@@ -174,34 +184,55 @@ object TargetSummaryTable extends TableHooks:
...
+                        ^.cls     := "pl-compact p-component p-button p-fileupload",
+                        ^.htmlFor := "target-import",
+                        Icons.FileImport
+                ),
+                <.input(^.tpe     := "file",
+                        ^.onChange ==> onTextChange,
+                        ^.id      := "target-import",
+                        ^.name    := "file",
+                        ^.accept  := ".csv"
+                ),
```
toddburnside, line 217: Clever to hide the button. 

It's strange to me that clicking on the label would open the dialog, but useful.

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
```diff
@@ -174,34 +184,55 @@ object TargetSummaryTable extends TableHooks:
...
+                .runAsyncAndForget,
             acceptClass = PrimeStyles.ButtonSmall,
             rejectClass = PrimeStyles.ButtonSmall,
             icon = Icons.SkullCrossBones.color("red")
           )
 
+        def onTextChange(e: ReactEventFromInput): Callback =
+          val files = e.target.files.toList
+          // set value to null so we can reuse the import button
+          (Callback(e.target.value = null) *> filesToImport.set(files)).when_(files.nonEmpty)
```
toddburnside, line 200: What does the setting to `null` do?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/tabs/TargetTabContents.scala
cquiroz: Maybe at some point in time 😄

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
cquiroz: It is a bit odd but it seems the standard way to customize the button
Note that prime react has a `FileUpload` component but it assumes the file is uploaded to a backend where we are here processing it locally

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
cquiroz: Once you select a file value is set, thus you cannot click again to open another one. The solution is to set it to null thus allowing it to be set again

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
toddburnside: Interesting. Thanks.

I was just looking at FileUpload. It seems like you could use it if you want and set `customUpload` to true and provide a `uploadHandler`. Not sure what extra it would give you or how hard it would be to use.  I could do a facade for it if you think it would be useful.

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
toddburnside: Thanks for the information. I never would have guessed.

cquiroz: Me neither, it took me a while and somo stack overflow searches to find out

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
cquiroz: I don't think we'll need it very soon

rpiaggio (approved):
LGTM, nice

PR #2396, opened 2022-10-29
Summary virtual

cquiroz:


cquiroz:
Fixes [sc-1760]

cquiroz:
This includes a fix we'd need for testing. I'll just merge it as it is fairly small

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
```diff
@@ -241,11 +245,13 @@ object TargetSummaryTable extends TableHooks:
...
             )
           ),
-          PrimeTable(
+          PrimeVirtualizedTable(
             table,
+            _ => 32.toPx,
             striped = true,
             compact = Compact.Very,
             tableMod = ExploreStyles.ExploreTable |+| ExploreStyles.ExploreSelectableTable,
+            containerMod = ExploreStyles.ExploreFullHeightTable,
```
rpiaggio, line 254: Did you try `AutoHeightPrimeVirtualizedTable`? Maybe this class is not necessary.

cquiroz: Ah I could try that, I thought it was auto row height

cquiroz: Yeah that works, thanks

PR #2398, opened 2022-10-29
Allow sorting by mag regardless of units

cquiroz:


explore/src/main/scala/explore/targets/TargetColumns.scala
```diff
@@ -149,7 +149,8 @@ object TargetColumns:
             t => targetBrightnesses.get(t).flatMap(_.get(band))
           ).setCell(_.value.map(displayWithoutError(_)(displayBrightness)).orEmpty)
             .setSize(80.toPx)
-            .setEnableSorting(false) // We cannot sort since there may be different units.
+            // By user request we allow sorting by value though there maybe a mix of units
+            .sortableBy(_.map(_.value))
```
rpiaggio, line 153: I think Andy expected sorting to group rows with the same unit and then sort by value within them.

Maybe we can define an `Ordering[Measure[N]]` given an `Ordering[N]` and `Ordering[Units]` ?

cquiroz: I discussed this and it seems just the name is enough as by their range they'd naturally group by units.
@andrewwstephens WDYT?

andrewwstephens: I am hopeful that simply sorting by value will be good enough.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targets/TargetColumns.scala
```diff
@@ -132,6 +133,9 @@ object TargetColumns:
         <.span(ExploreStyles.UnitsTableLabel, measure.units.shortName.replace(" mag", ""))
       )
 
+    // Order first by unit alphabetically and then value
+    private given Order[Measure[BigDecimal]] = Order.by(x => (x.units.abbv, x.value))
```
cquiroz, line 137: @rpiaggio Does this seem right?

rpiaggio: That should work, yes :+1:

PR #2402, opened 2022-10-31
Show errors per row imported

cquiroz:
Now the import target box will show a table with an error per row that has failed

image

### DIFF
--- a/explore/src/main/scala/explore/config/TargetImportPopup.scala
+++ b/explore/src/main/scala/explore/config/TargetImportPopup.scala
@@ -21,9 +21,12 @@ import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.catalog.csv.TargetImport
 import lucuma.core.model.Program
 import lucuma.core.model.Target
+import lucuma.react.syntax.*
+import lucuma.react.table.*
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/config/TargetImportPopup.scala
```diff
@@ -128,14 +162,27 @@ object TargetImportPopup:
...
+            ).unless(state.value.genericError.isEmpty),
+            <.div(
+              ExploreStyles.TargetImportErrors |+| ExploreStyles.ExploreBorderTable,
+              PrimeAutoHeightVirtualizedTable(
+                table,
+                estimateRowHeight = _ => 34.toPx,
+                tableMod = ExploreStyles.ExploreTable,
+                striped = true,
+                compact = Compact.Very,
+                emptyMessage = "No lines defined"
```
rpiaggio, line 183: Is this the right emptyMessage for this table?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/TargetImportPopup.scala
cquiroz: Good catch, will fix it

PR #2408, opened 2022-11-01
Fix spectroscopy matrix rows ids, optimize table rendering

rpiaggio:
Some rows had duplicated ids, which broke sorting.

Also, precompute ITC result summary, storing it in the table row, so it doesn't have to be computed on every render.


cquiroz (approved):
LGTM

PR #2417, opened 2022-11-03
lib updates

cquiroz:


cquiroz:
It seems atto was 12k of compressed js 😃

toddburnside (approved):
LGTM

PR #2428, opened 2022-11-04
Fix stylelint errors

toddburnside:


cquiroz:
Weird this produced a +1.1k change in bundle size

toddburnside:
> Weird this produced a +1.1k change in bundle size

I think BundleMon just randomly picks an increase or decrease. It doesn't actually keep track of previous sizes.

PR #2429, opened 2022-11-06
Switch to prime react sidebar

cquiroz:


explore/src/main/scala/explore/HelpBody.scala
```diff
@@ -71,9 +69,6 @@ object HelpBody:
       .useEffectOnMountBy { (props, _, state) =>
         load(props.url).flatMap(v => state.set(Pot.fromTry(v)).to[IO])
       }
-      .useGlobalHotkeysBy((_, helpCtx, _) =>
-        UseHotkeysProps(CloseHelp.value, helpCtx.displayedHelp.set(none))
```
toddburnside, line 75: Is the `CloseHelp` hotkey being used anywhere else? If not, can it be deleted?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/HelpBody.scala
cquiroz: You're right. I'll delete it. prime handles `closeOnEsc` by itself

rpiaggio: You can do `state.setAsync` (requires `import crystal.react.implicits.*`).

PR #2433, opened 2022-11-08
Update lucuma-react, scalac

rpiaggio:


explore/src/main/scala/explore/highcharts.scala
```diff
@@ -28,6 +27,6 @@ def commonOptions: ChartOptions =
   ChartOptions()
     .setStyledMode(true)
     .setAlignTicks(false)
-    .setZooming(ChartZoomingOptions().setType(OptionsTypeValue.xy))
+    .setZooming(ChartZoomingOptions().setType(highchartsStrings.xy.toString))
     .setPanning(ChartPanningOptions().setEnabled(true))
-    .setPanKey(OptionsPanKeyValue.shift)
+    .setPanKey("shift")
```
rpiaggio, line 32: There's no constant anymore for this.

cquiroz: WDYM, on the facade? Maybe we should have our own

cquiroz (approved):
LGTM

explore/src/main/scala/explore/highcharts.scala
rpiaggio: Yup, ST is not generating it anymore.

We can add our own to the facade, yes.

toddburnside (approved):
LGTM

PR #2436, opened 2022-11-09
Replace TargetSelectionPopup with primereact Dialog and related changes

toddburnside:
image

### DIFF
--- a/common/src/main/webapp/sass/explore-primereact.scss
+++ b/common/src/main/webapp/sass/explore-primereact.scss
@@ -26,4 +26,26 @@
       width: 760px;
     }
   }
+
+  &.pl-dialog-large {
+    width: 95%;
+
+    @media (min-width: 1200px) {
+      width: 1080px;
+    }
+  }
 }
+ 
+// TODO: Move to lucuma-ui
+// From the SUI Header component
+.pl-header {
+  font-size: 1.2857em;
+  font-weight: bold;
+  line-height: 1.2857em;
+  margin: calc(2rem - 0.1429em) 0 1rem;
+
+  &.pl-small {
+    font-size: 1.0714em;
+  }
+}
+
--- a/explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
+++ b/explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
@@ -31,18 +31,15 @@ import lucuma.core.model.Target
 import lucuma.core.util.NewType
 import lucuma.refined.*
 import lucuma.ui.forms.FormInputEV
+import lucuma.ui.primereact.*
 import lucuma.ui.reusability.*
 import lucuma.ui.syntax.all.*
 import lucuma.ui.syntax.all.given
 import react.aladin.*
 import react.common.ReactFnProps
-import react.semanticui.elements.button.Button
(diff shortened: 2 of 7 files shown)


explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
```diff
@@ -301,16 +287,7 @@ object TargetSelectionPopup:
                     )
                   )
                 )
-              }.toTagMod
-            )
-          )(
-            ^.autoComplete.off,
-            ^.onSubmit ==> (e =>
```
toddburnside, line 308: It is no longer a form that can be submitted, but that shouldn't matter since search is done when the text changes. I'm not sure what actually submitted this, anyway, since there was no `submit` button.

cquiroz:
Neat, can you verify it works on small screens. There is some extra logic to make it work so

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
rpiaggio: This allowed search to be triggered by pressing <ENTER> (text change search waits a few milliseconds after you stop typing).

toddburnside:
> Neat, can you verify it works on small screens. There is some extra logic to make it work so

image

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
toddburnside: Ah. Do you think that is important? I think I could use a standard form and do that.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
rpiaggio: I personally like it, but I don't think it's absolutely necessary.

cquiroz:
is the tooltip chrome's?

common/src/main/webapp/sass/explore-primereact.scss
```diff
@@ -26,4 +26,26 @@
...
+    @media (min-width: 1200px) {
+      width: 1080px;
+    }
+  }
 }
+ 
+// TODO: Move to lucuma-ui
+// From the SUI Header component
+.pl-header {
+  font-size: 1.2857em;
```
cquiroz, line 42: I'm guessing these could be (eventually) converted to variables

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
cquiroz: Such that pressing enter does a submit?

cquiroz (approved):
LGTM

toddburnside:
> is the tooltip chrome's?

Yeah. I forgot to get rid of that before the screenshot.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targets/TargetSelectionPopup.scala
toddburnside: I just put a form around the input and it worked.

PR #2439, opened 2022-11-10
Timing windows

cquiroz:
This is a first stab at supporting timing windows. This implements the UI and adds backend support with a very simple table in hasura
The table is global thus every user will see the same and it is not associated to any observation or program
This will eventually be replaced by the actual odb api

Here is a small video on how it works:

https://user-images.githubusercontent.com/3615303/200969393-ae357300-97d8-49bb-8dd8-1b69579b0337.mp4


cquiroz:
Add support for timing windows [sc-1793]

common/src/main/webapp/less/vendor/react-grid-layout.less
```diff
@@ -64,10 +64,10 @@
 }
 
 .rgl-tile-overlay {
-  display: none; // Enable for grid debugging
+  // display: none; // Enable for grid debugging
   position: absolute;
```
toddburnside, line 68: DId you want this commented out?

common/src/main/webapp/sass/explore.scss
```diff
@@ -991,3 +994,104 @@ table.explore-border-table {
...
+
+  .timing-window-editor {
+    @include react-datepicker-input;
+    @include react-datepicker-wrapper;
+
+    grid-area: editor;
+    padding: 0.5em;
+    font-size: smaller;
+
+    & span {
```
toddburnside, line 1051: Are all of these `&`s needed? The `&` is needed when you want to add something like another class to the outer selector. But here, it seems like these are children.

explore/src/main/scala/explore/constraints/TimingWindowsPanel.scala
```diff
@@ -0,0 +1,398 @@
...
+
+        val current = table.getSelectedRowModel().rows.headOption.map(_.original)
+        val pos     = rows.indexWhere((x: TimingWindow) => current.exists(_.id === x.id))
+
+        val selectedTW =
+          props.windows
+            .zoom(Index.index[List[TimingWindow], Int, TimingWindow](pos))
+            .withOnMod {
+              case Some(tw) =>
+                Callback.pprintln(tw) *> TimingQueries.updateTimingWindow[IO](tw).runAsyncAndForget
```
toddburnside, line 203: Do you want to keep the `pprintln`?

explore/src/main/scala/explore/constraints/TimingWindowsPanel.scala
```diff
@@ -0,0 +1,398 @@
...
+                    id = "remain-open",
+                    checked = e.remainOpenFor,
+                    onChange = (_, checked) =>
+                      selectedTW
+                        .mod(_.toRemainOpen(NonNegDuration.unsafeFrom(Duration.ofDays(2))))
+                        .when_(checked)
+                  ),
+                  <.label("Remain open for", ^.htmlFor := "remain-open"),
+                  selectedRemainOpenFor.asView.map { remainOpen =>
+                    FormInputEV(
```
toddburnside, line 312: It seems like you have a mix of SUI and primereact. Could you use FormInputTextView here?

explore/src/main/scala/explore/tabs/ConstraintsTabContents.scala
```diff
@@ -40,53 +50,105 @@ import lucuma.core.model.ConstraintSet
...
+        x = 0,
+        y = ConstraintsHeight.value,
+        w = DefaultWidth.value,
+        h = TimingWindowsHeight.value,
+        isResizable = false
+      )
+    )
+  )
+
+  private val defaultTargetLayouts = defineStdLayouts(
```
toddburnside, line 124: Is this really a Targets layout?

toddburnside (approved):
LGTM

toddburnside:
I don't think I understand how the timing windows work. SInce the timing windows are in the constraints editor, does that mean that the constraint grouping includes (or will include in the future) the timing windows as part of the grouping criteria? In other words if the observations have the same Cloud Extinction, Water Vapor, etc., but different timing windows, will there be (in the future) separate groups in the tree panel for them?

cquiroz:
That's an excellent question and I can only share your confusion. I think we want to group them as well but the grouping will be different than conditions. In fact grouping will be hard because there are so many degrees of freedom in windows that the combinations will likely explode
My recommendation is that they belong to an observation and be a new tile in the observation tab instead but there are still some discussions about it
This PR establishes the basis for the windows but there are this issues to discuss in the near future.

cquiroz:
I fear too that trying to group by timing windows will be fairly expensive

explore/src/main/scala/explore/constraints/TimingWindowsPanel.scala
cquiroz: I noticed, I wasn't sure if we had a prime react version

common/src/main/webapp/sass/explore.scss
cquiroz: Hmm, I guess I missundrestood, will review

common/src/main/webapp/less/vendor/react-grid-layout.less
cquiroz: nope, that's a leftover

PR #2445, opened 2022-11-11
Migrate enums to native graphql enums

cquiroz:


toddburnside (commented):
LGTM

common-graphql/src/main/scala/queries/common/UserPreferencesQueriesGQL.scala
```diff
@@ -123,7 +123,7 @@ object UserPreferencesQueriesGQL {
   @GraphQL
   trait UserPreferencesElevPlotUpdate extends GraphQLOperation[UserPreferencesDB] {
     val document = """
-      mutation userPreferencesUpsert($userId: String = "", $elevationPlotRange: elevation_plot_range = "", $elevationPlotTime: elevation_plot_time = "") {
+      mutation userPreferencesUpsert($userId: String = "", $elevationPlotRange: ExplorePlotRangeEnum = "", $elevationPlotTime: ExplorePlotTimeEnum = "") {
```
rpiaggio, line 126: The default values don't seem like valid enum instances.

cquiroz: Good point should the parser detect that?

rpiaggio: That would be nice. For parsing in clue we use grackle and we are "a few" versions behind on it (0.1.9 vs 0.8.2). We should try updating and if that doesn't fix it then see how we can add it to grackle.

cquiroz: Maybe I should remove the default for now?

rpiaggio (approved):
LGTM

PR #2456, opened 2022-11-13
Quick fix for brightness units pulldown menu hidden

toddburnside:
I did a quick fix for the problem of the brightness units dropdown dropping down behind the table rows. But it's a bit of an ugly hack, so I don't know if we want to use. I'm just thinking it would be nice if it worked in the demo this week...

I set out to convert the whole target editor to primereact, but I wasn't going to have time to finish it. So, I rigged some temporary CSS to make it work. The rows are still slightly taller than the old table.
image

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -1347,69 +1347,94 @@ html {
   padding: 1em;
 }
 
+// This selector is never matched after the new table transistion. But, the table doesn't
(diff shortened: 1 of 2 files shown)


cquiroz:
Are you going to redo this in another PR?

toddburnside:
> Are you going to redo this in another PR?

Yes. I'll do a more complete fix that includes the entire sidereal target editor.

PR #2469, opened 2022-11-23
Convert SiderealTargetEditor to primereact

toddburnside:
Some images of the converted sidereal target editor.  See comments in the PR.

Note: The `Epoch` field and the ones below it used to be indented further than the `Name`, `RA` and `Dec` fields. Carlos said he thought that just fell out of the structure of the HTML and wasn't intentional or necessary. This indentation could be restored if desired.

With bands:
image

With emission lines:
image

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -233,16 +233,15 @@ $tree-section-width: 300px;
   }
 }
 
-@mixin expanded-aladin {
-  grid-template-areas: 'aladin';
-  grid-template-columns: 1fr;
+@mixin hide-hidden-target-cells($all) {
+  @if $all {
+    .explore-title-undo-buttons {
+      display: none;
+    }
 
-  .explore-title-undo-buttons {
-    display: none;
-  }
-
-  .target-form {
-    display: none;
+    .target-form {
+      display: none;
+    }
   }
 
   .target-proper-motion-form {
@@ -259,13 +258,12 @@ $tree-section-width: 300px;
 .tile-md-width,
 .tile-sm-width {
   .target-grid {
-    grid-template-columns: 1fr 2fr;
-    grid-template-rows: min-content min-content min-content 1fr;
-    grid-template-areas:
-      'undo           aladin'
-      'coordinates    aladin'
-      'proper-motion  aladin'
-      'source-profile aladin';
+    grid-template:
+      'undo           aladin' min-content
+      'coordinates    aladin' min-content
+      'proper-motion  aladin' min-content
+      'source-profile aladin' 1fr
+      / 1fr 2fr;
   }
 
   .target-aladin-cell {
@@ -274,38 +272,15 @@ $tree-section-width: 300px;
 }
 
 .tile-xs-width {
-  
-  // undo buttons are not show in xs tiles, only aladin
-  // For very small tables make the undo buttons remove the text
-  // .explore-buttons-undo-label {
-  //   display: none;
-  // }
-  
   .target-grid {
-    @include expanded-aladin;
+    @include hide-hidden-target-cells(false);
+
+    grid-template:
+      'undo' min-content
+      'coordinates' min-content
+      'aladin' 1fr
+      / 1fr
   }
-  
-  // brightness are not even displayed in xs tiles.
-  // For very small tables hide the headers and add fake "header" to the left of the row
-  // .explore-brightnesses-container {
-
-  //   .ui.table,
-  //   thead,
-  //   tbody,
-  //   th,
-  //   td,
-  //   tr {
-  //     // SUI already has an !important
-  //     display: block !important;
-  //   }
-
-  //   // Hide table headers (but not display: none;, for accessibility)
-  //   thead tr {
-  //     position: absolute;
-  //     top: -9999px;
-  //     left: -9999px;
-  //   }
-  // }
 }
 
 .target-tile-body {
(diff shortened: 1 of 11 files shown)


common/src/main/webapp/sass/explore.scss
```diff
@@ -157,7 +161,212 @@ $tree-section-width: 300px;
...
+.tile-md-width,
+.tile-sm-width {
+  .target-grid {
+    grid-template-columns: 1fr 2fr;
+    grid-template-rows: min-content min-content min-content 1fr;
+    grid-template-areas:
+      'undo           aladin'
+      'coordinates    aladin'
+      'proper-motion  aladin'
+      'source-profile aladin';
```
toddburnside, line 268: It seems the layout was different for the different sizes at one point? This grid layout provides more opportunities for changing the layout. However, currently, either everything is shown, or only `aladin` is shown (small screens, full-screen aladin). If we want to keep this behavior, I could simplify this grid layout and remove some of the commented out CSS below.

common/src/main/webapp/sass/explore.scss
```diff
@@ -157,7 +161,212 @@ $tree-section-width: 300px;
...
+  }
+
+  .target-aladin-cell {
+    grid-row: span 4;
+  }
+}
+
+.tile-xs-width {
+  
+  // undo buttons are not show in xs tiles, only aladin
```
toddburnside, line 278: Will delete this and the stuff below before merging if we are going to keep the current layouts mentioned above.

explore/src/main/scala/explore/targeteditor/SearchForm.scala
```diff
@@ -70,6 +70,7 @@ object SearchForm {
         _ => // Auto-select name field on new targets.
           // Accessing dom elements by id is not ideal in React, but passing a ref to the <input> element of a
           // Form.Input is a bit convoluted. We should reevaluate when and if we switch to another component library.
+          // TODO: I don't think this even works.
```
toddburnside, line 73: Unless I misunderstand what this `useEffectWithDepsBy` is supposed to do, it does not seem to be working in the current staging. I can try to fix it, or delete it, depending on whether we think it is necessary.

common/src/main/webapp/sass/explore.scss
cquiroz: Right. I think we wanted to show aladin and hide the lesser parts of the form on small screen. Essentially show the coordinates only and below aladin in full width.
This is worth testing in several configurations and also on full screen mode

toddburnside: OK. Currently we do NOT show the coordinates on small screens. (The `expanded-aladin` mixin above is used for both full screen aladin and small screens.) I can change it so that the coordinates are displayed above (or below) aladin for small screens if that is what we want.

explore/src/main/scala/explore/targeteditor/SearchForm.scala
cquiroz: Do we know if the element is even found? Seems like doing it via Ref would be better though the main question is:
Why are we doing this? to make it simpler to replace the currently selected text?
I may recall that detail and thinking it could be useful in many cases

common/src/main/webapp/sass/explore.scss
cquiroz: Did we only show the name before?
I guess coordinates is often usefil

toddburnside: At this time (in staging and production) we don't show the name. We only show aladin. At some point maybe we showed the name and/or coordinates? If we show the coordinates, we probably want to show the undo/redo buttons, too, I guess?

cquiroz: Ok, and we are showing the table right? In that case the coords are there

explore/src/main/scala/explore/targeteditor/SearchForm.scala
cquiroz: I don't think it is critical, though it maybe nice to have, if simple we could try to restore it but remove it if it is complicated

common/src/main/webapp/sass/explore.scss
toddburnside: Ah. OK. Do you want to be able to edit the name and coords? I pushed a change that show them above aladin. But I could remove them again. They can still go to full screen aladin in the small screen, so 🤷‍♂️

cquiroz: I'd keep as i twas before

toddburnside: Which is to not allow editing in small screen and only show aladin?

explore/src/main/scala/explore/targeteditor/SearchForm.scala
toddburnside: I'm trying to debug it right now.

toddburnside: Ah. Silly me. I didn't notice that it is only supposed to happen when you add a "New Empty Target". It is, indeed, functioning as intended.

cquiroz: Yes that's what I recall, so you can type the new name right away

cquiroz:
Looks great, keeps the style with the new controls 👍

cquiroz (approved):
LGTM

explore/src/main/scala/explore/targeteditor/SearchForm.scala
```diff
@@ -99,48 +99,33 @@ object SearchForm {
...
             )
 
-        def iconKeyPress(e: ReactKeyboardEvent): Callback =
-          search >> e.stopPropagationCB *> e.preventDefaultCB
-
-        def submitForm: Form.OnSubmitE =
-          (e: Form.ReactFormEvent, _: FormProps) =>
-            e.preventDefaultCB >> search.when(enabled.value).void
+        def onKeyPress = (e: ReactKeyboardEvent) =>
+          Callback.log(s"Key Code: ${e.key}") >>
```
rpiaggio, line 103: Debug messages.

PR #2474, opened 2022-11-25
Update lucuma-ui and make changes for ViewLike and convert some controls

toddburnside:
In addition to the changes for ViewLike, the PR:

- Changes the inputs in the timing window edit form to use primereact. This required the viewlike changes because `ViewOpt` is involved.
- On the shared target edit warning, switches out the checkboxes for radiobuttons since only one can be selected.

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -13,6 +13,12 @@ $tree-section-width: 300px;
   }
 }
 
+// There is still an analog to this in style.less
+@mixin warning-label-mixin {
+  color: var(--warning-text-color);
+  background-color: var(--warning-background-color);
+}
+
 .p-inputgroup.explore-warning-input {
   box-shadow: -0 0 2px 1px var(--warning-background-color);
 }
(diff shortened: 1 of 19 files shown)


common/src/main/webapp/sass/explore.scss
```diff
@@ -1258,15 +1287,20 @@ table.explore-border-table {
...
-      .field {
+      .pl-form-field {
         margin-right: 0.5em;
+        width: 9ch;
 
-        .ui.input {
-          width: 9ch;
-        }
+        // .ui.input {
+        //   width: 9ch;
```
cquiroz, line 1295: Commented code

explore/src/main/scala/explore/constraints/TimingWindowsPanel.scala
```diff
@@ -60,6 +60,8 @@ import java.time.Duration
 import java.time.ZonedDateTime
 import scala.concurrent.duration.FiniteDuration
 
+import scalajs.js.timers
```
cquiroz, line 63: Is this being used?

cquiroz (approved):
LGTM thanks

explore/src/main/scala/explore/constraints/TimingWindowsPanel.scala
toddburnside: It is being used when the code in the `n times` textbox is selected.

rpiaggio (approved):
LGTM

PR #2475, opened 2022-11-25
Remove Target Library on Targets Tab

toddburnside:
I moved the `+ Sidereal` button to the top of the tree.
image
However, any target they create will not even show up in the tree because it won't have observations. So, maybe the button should go on with the `Target Summary` buttons. But, the Target Summary doesn't have undo/redo. I guess we could have the Target Summary use the same undo/redo as the tree? Or don't have undo for target add? It all gets very confusing...

Which makes me realize, there is no longer any way to delete targets. There used to be a delete button on the targets in the target library. So, maybe we need to put a delete button in the rows of the Target Summary Table and add undo/redo?

Update: `+ Sidereal` has been moved to the Target Summary. And the undo context is shared between the tree and the summary. Delete targets now also supports undo.
image


cquiroz:
You can multi select and delete

toddburnside:
> You can multi select and delete

Ah. I see. When you have something selected, the trash can appears. 

And there is a confirmation dialog about it being undoable when you click it. 👍 

Should I move the `+ Sidereal` button there, too, and also make it undoable? They can always just delete it.

toddburnside:
> Should I move the `+ Sidereal` button there, too, and also make it undoable? They can always just delete it.

Alternatively, I could wire up the delete to the undo/redo for the tree, or have a separate undo/redo context for the table. The undo/redo logic is already available.

rpiaggio:
Seems like the tree and the table are now more connected. They should share an undo context and a drag'n'drop context. Maybe we could move both of these to TargetTabContents?

cquiroz:
Seems good, we may want to have more undoable actions on the big table

toddburnside:
Yeah. I think that's probably the best idea. I'll move the `+` button to the table, too. 

I knew this wouldn't be as simple as just deleting some code. 😄

toddburnside:
> Can we remove react-reflex?

I just did a search, and it seems it is no longer used. I'll remove it, too. I think we could use a primereact splitter if we need something similar in the future.

common/src/main/webapp/sass/explore.scss
```diff
@@ -37,8 +37,6 @@ $tree-section-width: 300px;
   }
 
   .p-button.p-button-sm {
-    max-width: 4rem;
```
toddburnside, line 40: @cquiroz Was there a reason you set the max-width for the buttons in the summary table toolbar to 4rem? I removed it so the whole text of `+ Sidereal` would display.

cquiroz: I don't recall

toddburnside:
> They should share an undo context and a drag'n'drop context. Maybe we could move both of these to TargetTabContents?

@rpiaggio I am sharing the undo context, but not the drag'n'drop context. Would it be possible to drag rows from the table?

rpiaggio:
> Would it be possible to drag rows from the table?

I don't think you can drag'n'drop from one context into another, but it might work.

rpiaggio:
Ah, yes, that should be possible. There's an example of dragging from one list into another: https://codesandbox.io/s/vertical-list-with-multiple-drop-targets-forked-7mt55r.

rpiaggio (approved):
LGTM

PR #2477, opened 2022-11-26
Fix for ExploreModelValidators test

cquiroz:
It seems BigDecimal can parse digits in other locales, e.g. the number 1 in tibetan: `༡`

To fix it this is verifying that the string matches a regex for US numbers


toddburnside (commented):
Good catch! LGTM.

model/shared/src/main/scala/explore/model/ExploreModelValidators.scala
```diff
@@ -71,13 +72,18 @@ object ExploreModelValidators:
           .toErrorsValidWedge
       )
 
+  private val bdPattern = """"-?(?:\\d+(?:\\.\\d+)?|\\.\\d+)""".r
```
rpiaggio, line 75: Can we add a comment explaining the regex?

rpiaggio (approved):
LGTM, thanks for fixing this

PR #2490, opened 2022-12-03
Improvements to overlay calculations

cquiroz:
Fixes an edge case where the visualization is not executed and some small improvements for safari


toddburnside (approved):
LGTM

PR #2495, opened 2022-12-07
Add undo/redo for pasting observations into targets

toddburnside:
This also makes the expanded nodes behave in a reasonable manner.


cquiroz:
This is so great. Thanks

PR #2509, opened 2022-12-09
Update lucuma-core and make SED optional

cquiroz:
Needs some validation against the latest odb and it servers before merging


model/shared/src/main/scala/explore/model/ExploreModelValidators.scala
```diff
@@ -43,14 +44,14 @@ object ExploreModelValidators:
...
-      .toNel()
+      .toNel(".".refined)
       .withErrorMessage("Invalid wavelength dither values".refined)
       .optional
 
   val offsetQNELValidWedge: InputValidWedge[Option[NonEmptyList[Offset.Q]]] =
     MathValidators.truncatedAngleSignedDegrees
       .andThen(Offset.Component.angle[Axis.Q].reverse)
-      .toNel()
+      .toNel(".".refined)
```
toddburnside, line 54: It seems this might be breaking the ValidWedge Laws.

cquiroz: Thanks. it was a "," before, reverting

cquiroz:
This seems to be working fine with the lastest tmp-api

PR #2510, opened 2022-12-09
Allow observation creation on target tab

toddburnside:
Adds a `+ Obs` button next to the `+ Tgt` button on the target tab. This is part of the workflow @andrewwstephens requested when we discussed target copy/paste. If there are one or more observations selected in the target tree, the same asterism as for that (those) observation(s) is used for the new observation. Otherwise, the targets that are selected in the summary (if any) are used.
image

### DIFF
--- a/explore/src/main/scala/explore/targets/ObservationInsertAction.scala
+++ b/explore/src/main/scala/explore/targets/ObservationInsertAction.scala
@@ -1,0 +1,91 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.targets
+
+import cats.Order.*
+import cats.effect.IO
+import cats.syntax.all.*
+import clue.TransactionalClient
+import crystal.react.View
+import crystal.react.implicits.*
+import explore.common.AsterismQueries.*
+import explore.model.ObsIdSet
+import explore.model.ObsSummaryWithConstraintsAndConf
+import explore.model.syntax.all.*
+import explore.undo.*
+import lucuma.core.model.Observation
+import lucuma.core.model.Program
+import lucuma.core.model.Target
+import lucuma.schemas.ObservationDB
+import lucuma.schemas.ObservationDB.Types.*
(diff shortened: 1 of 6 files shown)


rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targets/ObservationInsertAction.scala
```diff
@@ -0,0 +1,91 @@
...
+
+object ObservationInsertAction {
+  private def getter(
+    obsId: Observation.Id
+  ): AsterismGroupsWithObs => Option[ObsSummaryWithConstraintsAndConf] = _.observations.get(obsId)
+
+  private def setter(obsId: Observation.Id)(
+    optObs:                 Option[ObsSummaryWithConstraintsAndConf]
+  ): AsterismGroupsWithObs => AsterismGroupsWithObs = agwo =>
+    optObs.fold {
```
cquiroz, line 35: Neat

explore/src/main/scala/explore/targets/ObservationInsertAction.scala
```diff
@@ -0,0 +1,91 @@
...
+  ): Action[AsterismGroupsWithObs, Option[ObsSummaryWithConstraintsAndConf]] =
+    Action(getter = getter(obsId), setter = setter(obsId))(
+      onSet = (agwo, optObs) =>
+        expandedIds.mod(updateExpandedIds(obsId, agwo, optObs)).to[IO] >> setPage(obsId.some),
+      onRestore = (agwo, optObs) =>
+        expandedIds.mod(updateExpandedIds(obsId, agwo, optObs)).to[IO] >>
+          optObs.fold(
+            ObsQueries.deleteObservation[IO](obsId) >>
+              setPage(none) >>
+              postMessage(s"Deleted observation $obsId")
```
cquiroz, line 84: Is this raising a notification?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/targets/ObservationInsertAction.scala
toddburnside: Yes. It's posting a toast.

PR #2514, opened 2022-12-11
Don't draw probe outline

cquiroz:
Oddly I don't see it in chrome or safari but at least one user sees it on safari


toddburnside (approved):
LGTM

PR #2515, opened 2022-12-12
Use new css device units

cquiroz:
Starting in chrome 108 new dynamic device css units are supported
https://web.dev/viewport-units/

This PR uses them instead of the old non dynamic ones. This produces a better result putting the side tabs in position and should allow us to target more devices

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -2,6 +2,49 @@
 
 $tree-section-width: 300px;
 
+// Scrollbar
+*::-webkit-scrollbar {
+  width: 0.8em !important;
+}
+
+*::-webkit-scrollbar-track {
+  background: var(--scrollbar-background) !important;
+}
+
+*::-webkit-scrollbar-thumb {
+  background: var(--scrollbar-background-thumb) !important;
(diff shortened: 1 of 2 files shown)


common/src/main/webapp/sass/explore.scss
```diff
@@ -2,6 +2,49 @@
...
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  overflow: hidden;
+  position: fixed;
+  min-height: 100svh;
```
toddburnside, line 40: I've only looked over the web page you provided in the link, so I'm not clear on all of this. Why did you choose small viewport dimensions rather than the dynamic ones?

cquiroz: Good question. I tested this on a tablet and the responsive mode but wasn't able to do so on my phone where the toolbars re in place.
I'll give it a go too

cquiroz: Ok, after some fiddling with my local network I managed to test on the phone and in practice it doesn't matter because we set the height anyway. Regardless using the dynamic units seems more sensible

toddburnside (approved):
LGTM

PR #2518, opened 2022-12-13
Copy/paste of targets into observations

toddburnside:


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -717,16 +731,31 @@ object TargetTabContents extends TwoPanels:
                         .getOrElse(IO.unit)
                     else IO.unit
 
+                  case LocalClipboard.CopiedTargets(tids) =>
+                    (props.focused.obsSet, optViewAgwo).tupled
+                      .foldMap((obsIds, agwov) =>
+                        val undoContext    = UndoContext(props.listUndoStacks, agwov)
+                        // don't want to do anything if the target astersism already contains the targets
```
cquiroz, line 738: Do you mean we don't put the same target twice in the obs?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/tabs/TargetTabContents.scala
toddburnside: I know, crazy, right?

toddburnside: Your comment made me realize this bit needed improvement. 😄

PR #2519, opened 2022-12-13
Show executed visits in sequence popup

rpiaggio:
Kapture 2022-12-13 at 16 04 26


rpiaggio:
All the components related to sequences now live in the `explore.config.sequence` package.

common/src/main/scala/explore/givens.scala
```diff
@@ -88,13 +88,21 @@ extension [F[_], A](f: F[A])
   )(using Temporal[F]): fs2.Stream[F, A] =
     reRunOnSignal(signals.reduceLeft(_ merge _), debounce)
 
+  /**
+   * Given an effect producing an A and a bunch of `Resource`s providing signal streams, runs the
+   * effect and then re-runs it whenver a signal is received, producing a Stream[A].
+   */
```
toddburnside, line 94: Nice. Thanks for the documentation!

explore/src/main/scala/explore/config/sequence/GmosSequenceTable.scala
```diff
@@ -0,0 +1,105 @@
...
+          atom.steps.tail.map(step => GmosSequenceRow.FutureStep.fromStep(step, atom.id, none))
+      )
+      .flatten
+
+  val bracketDef =
+    svg(^.width := "0", ^.height := "0")(
+      defs(
+        path(
+          transform   := "scale(1, 0.28)",
+          fill        := "white", // FIXME Use CSS
```
toddburnside, line 68: A FIXME comment

toddburnside (approved):
Looks great.

explore/src/main/scala/explore/config/sequence/GmosSequenceTable.scala
rpiaggio: This was here before, but we should fix it and use Prime colors. I'll address it in another PR.

PR #2526, opened 2022-12-14
Avoid scrollbars on target tab.

rpiaggio:


cquiroz:
What if we have 100 targets?

cquiroz:
Ah I see, thanks

cquiroz (approved):
LGTM

PR #2527, opened 2022-12-15
Limit asterism table height (with scrollbars).

rpiaggio:


common/src/main/webapp/sass/explore.scss
```diff
@@ -370,6 +370,10 @@ body {
   flex-direction: column;
 }
 
+.target-asterism-table {
+  max-height: 250px;
```
cquiroz, line 374: This is inside a flex grid, maybe we can specify it using flex grow/shrink/basis?

toddburnside:
This for the asterism editor? I think I (we) had scrollbars there once, and Andy explicitly wanted all the targets to show so I removed them.  But, maybe if enough of them show, he'd be happy. I think before, I needed to specify an exact height for some reason, so it needed to be small so that it didn't waste space for one or 2 targets. So, if it can shrink as needed, I think it would be OK.

rpiaggio:
> This for the asterism editor? I think I (we) had scrollbars there once, and Andy explicitly wanted all the targets to show so I removed them. But, maybe if enough of them show, he'd be happy. I think before, I needed to specify an exact height for some reason, so it needed to be small so that it didn't waste space for one or 2 targets. So, if it can shrink as needed, I think it would be OK.

Yes, with this setting it shrinks, and grows to show up to 7 rows. Then the scrollbars kick in.

rpiaggio:
@andrewwstephens Is this OK or should we grow the table indefinitely?

common/src/main/webapp/sass/explore.scss
cquiroz: I guess a fixed max-height is fine. I just thought that we may get something similar if we make the aladin part grow faster than the table. Not sure when the overflow will be activated

cquiroz:
IMHO if we have n targets the table should scroll or we'd hide the aladin part

andrewwstephens:
I prefer to avoid scrolling if we can avoid it, but 7 targets should be enough for anybody, right?  ;-)

rpiaggio:
This is to put some bound on the UI, since nothing stops you from adding 100 targets to an observation.

cquiroz (approved):
LGTM

PR #2535, opened 2022-12-19
Port the pos angle editor to prime-react

cquiroz:
image

### DIFF
--- a/explore/src/main/scala/explore/config/ObsConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/ObsConfigurationPanel.scala
@@ -21,8 +21,12 @@ import lucuma.core.math.validation.MathValidators
 import lucuma.core.model.Observation
 import lucuma.core.model.PosAngleConstraint
 import lucuma.refined.*
-import lucuma.ui.forms.EnumViewSelect
 import lucuma.ui.input.ChangeAuditor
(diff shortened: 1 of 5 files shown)


rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/ObsConfigurationPanel.scala
```diff
@@ -94,25 +96,28 @@ object ObsConfigurationPanel:
...
+            "Position Angle",
+            HelpIcon("configuration/positionangle.md".refined)
+          ),
+          EnumDropdownView(
             clazz = ExploreStyles.ObsConfigurationObsPA,
-            id = "pos-angle-alternative",
+            id = "pos-angle-alternative".refined,
             value = posAngleOptionsView,
             disabled = !props.agsState.get.canRecalculate
           ),
```
toddburnside, line 123: FYI. You could use a FormEnumDropdownView here to combine the label and the dropdown. IIRC, it does require using a fragment for the label when you have the help icon, but you don't have to repeat the id/htmlFor.

cquiroz: I tried this but I'm not sure how to add an (i) button there, Any ideas?

PR #2538, opened 2022-12-20
Convert ObsBadge and UndoButtons to PrimeReact

toddburnside:
The observation badge looks very much the same as it did before. I used an outline style for the undo/redo buttons because I thought it looked better in the button set, but I did increase the contrast from the standard `Secondary` button color.

image


explore/src/main/scala/explore/tabs/TwoPanels.scala
```diff
@@ -47,16 +46,13 @@ trait TwoPanels {
...
-          pv.set(SelectedPanel.Tree)
-      )
-    )(^.href := ctx.pageUrl(appTab, programId, Focused.None), Icons.ChevronLeft)
+      icon = Icons.ChevronLeft,
+      severity = Button.Severity.Secondary,
+      text = true,
+      clazz = ExploreStyles.TileBackButton,
+      onClick = ctx.pushPage(appTab, programId, Focused.None) >>
+        pv.set(SelectedPanel.Tree)
+    ).mini.compact
```
toddburnside, line 55: This no longer displays the link on hover, but since it usually will only be displayed on smaller devices, they wouldn't usually see the hover link, anyway. It could be switched to a `span` or a tooltip added if we need it.

common/src/main/webapp/sass/explore.scss
```diff
@@ -654,42 +698,146 @@ $search-preview-margin: 5px;
...
+  width: 100%;
   margin: 0;
-  padding: 0 !important;
+  padding: 1em;
+  background-color: var(--obs-badge-background);
+  border: solid 1px var(--site-border-color);
+  border-radius: 5px;
+
+  // defined in lucuma-ui.scss. Linting doesn't capture that.
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
```
toddburnside, line 710: I'm not sure how we can capture these variables in the linting since the sass file doesn't exist until after linking is done and the sass files are extracted from the lucuma-ui jar.

rpiaggio (approved):
LGTM

PR #2539, opened 2022-12-20
Remove SUI Loader

toddburnside:


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -765,15 +764,6 @@ object TargetTabContents extends TwoPanels:
...
             Toast(Toast.Position.BottomRight).withRef(toastRef.ref),
-            if (active.get.value) {
-              <.div(
-                ^.cls := "ui active dimmer",
-                Loader(
-                  active = true,
-                  "Cloning observation..."
-                )
-              )
-            } else EmptyVdom,
```
toddburnside, line 776: I was never able to actually observe this `Loader`. I think maybe it got overridden by the `asterismGroupsWithObs.render`? The `+Obs` button is disabled and `loading`, anyway.

explore/src/main/scala/explore/HelpBody.scala
```diff
@@ -83,12 +83,17 @@ object HelpBody:
             ExploreStyles.HelpTitle,
             <.h4(ExploreStyles.HelpTitleLabel, "Help"),
             <.div(
-              Button(as = <.a, size = Mini, compact = true, onClick = helpView.set(None))(
-                Icons.Edit
-              )(^.href := editUrl.toString(), ^.target := "_blank"),
-              Button(size = Mini, compact = true, onClick = helpView.set(None))(
-                Icons.Close
-              )
+              <.a(Button(icon = Icons.Edit,
```
cquiroz, line 86: Cool, I was woing to change these :D
Is the `<.a>` here so you get the link showing?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/HelpBody.scala
toddburnside: Yes, exactly. We can't do `as` anymore, so this allows the link to show on hover.

PR #2541, opened 2022-12-20
Convert some SUI code on the itc plots

cquiroz:


common/src/main/webapp/sass/explore.scss
```diff
@@ -1648,6 +1648,17 @@ table.explore-border-table {
...
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
+  font-size: var(--pl-tiny-font-size);
+}
+
+.p-component.p-button.pl-compact {
+  row-gap: 0.5em;
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
+  padding: var(--compact-button-padding);
+}
+
```
toddburnside, line 1661: You should be getting the font size and padding from lucuma-ui.css. Was it not working? And `row-gap` seems to be more specific than would apply to every compact button, isn't it?

cquiroz: on lucuma-ui they are linked to just `button` and not `p-button`. I thought that when we move these to lucuma-ui we may want to replicate all the cases

cquiroz: We could delete row-gap, that's true

toddburnside: Hmmm. Yeah. In lucuma-ui.scss it is `button.p-component.p-button`. I think I did that to increase specificity because all the primereact css is getting layered beneath the `.dark-theme` or `.light-theme`. Maybe just the `.pl-tiny` and `.pl-compact` is enough to increase the priority and we could remove `button` in lucuma-ui.css?

Is this targeting the `SelectButton` and it isn't actually a `button`?

toddburnside: Maybe in lucuma-ui.scss we just need to move the `button` to where it is needed. I `think` that the selectors like `&.p-button-sm,`  may need more specificity. So, maybe we just move those into their own "outer" selector instead of trying to include them with the `&.pl-small` etc.

cquiroz: Yes, it is for `SelectButton`

toddburnside: Or, maybe `button&.p-button-small` would work? I'm not sure if `&` can be used that way or not. I can try some things and see if it maintains your sizing. What UI element would I look at to be sure?

cquiroz:
The button on the left is a `Button` while the ones on the right are a `SelectButton` and I want them too look the same
image

PR #2542, opened 2022-12-21
Keep target summary visible when editing target

rpiaggio:
Also, if a target is being edited within an observation, it remains selected when switching to the Target Summary.

Kapture 2022-12-21 at 11 53 20

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -33,7 +33,6 @@ case class Tile(
   canMinimize:       Boolean = false,
   canMaximize:       Boolean = false,
   state:             TileSizeState = TileSizeState.Normal,
-  fadeIn:            Boolean = true,
   sizeStateCallback: TileSizeState => Callback = _ => Callback.empty,
   controllerClass:   Option[Css] = None, // applied to wrapping div when in a TileController.
   bodyClass:         Option[Css] = None, // applied to tile body
@@ -105,7 +104,7 @@ object Tile {
             .runNow()
 
         <.div(
-          ExploreStyles.Tile |+| ExploreStyles.FadeIn.when_(p.fadeIn) |+| p.tileClass.orEmpty,
+          ExploreStyles.Tile |+| ExploreStyles.FadeIn |+| p.tileClass.orEmpty,
           ^.key := p.id.value
         )(
           // Tile title, set classes based on size
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -33,8 +33,8 @@ case class Tile(
   canMinimize:       Boolean = false,
   canMaximize:       Boolean = false,
   state:             TileSizeState = TileSizeState.Normal,
+  fadeIn:            Boolean = true,
   sizeStateCallback: TileSizeState => Callback = _ => Callback.empty,
-  key:               js.UndefOr[Key] = js.undefined,
   controllerClass:   Option[Css] = None, // applied to wrapping div when in a TileController.
   bodyClass:         Option[Css] = None, // applied to tile body
   tileClass:         Option[Css] = None, // applied to the tile
@@ -105,7 +105,8 @@ object Tile {
             .runNow()
 
         <.div(
-          ExploreStyles.Tile |+| ExploreStyles.FadeIn |+| p.tileClass.orEmpty
+          ExploreStyles.Tile |+| ExploreStyles.FadeIn.when_(p.fadeIn) |+| p.tileClass.orEmpty,
+          ^.key := p.id.value
         )(
           // Tile title, set classes based on size
           ResponsiveComponent(
--- a/common/src/main/scala/explore/components/TileController.scala
+++ b/common/src/main/scala/explore/components/TileController.scala
@@ -46,7 +46,8 @@ case class TileController(
   layoutMap:     LayoutsMap,
   tiles:         List[Tile],
   section:       GridLayoutSection,
-  clazz:         Option[Css] = None
+  clazz:         Option[Css] = None,
+  storeLayout:   Boolean = true
 ) extends ReactFnProps(TileController.component)
 
 object TileController:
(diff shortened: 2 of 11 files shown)


common/src/main/scala/explore/components/TileController.scala
```diff
@@ -109,10 +110,10 @@ object TileController:
       // Make a local copy of the layout fixing the state of minimized layouts
       .useStateViewBy((p, _, _, _) => updateResizableState(p.layoutMap))
       // Update the current layout if it changes upstream
-      .useEffectWithDepsBy((p, _, _, _, _) => p.layoutMap)((_, _, _, _, current) =>
-        layout => current.set(updateResizableState(layout))
+      .useEffectWithDepsBy((p, _, _, _, _) => p.layoutMap)((_, _, _, _, currentLayout) =>
```
rpiaggio, line 113: I did some renaming here

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -605,37 +637,34 @@ object TargetTabContents extends TwoPanels:
       .useResizeDetector()
       // Initial target layout
       .useStateView(Pot.pending[LayoutsMap])
-      // Keep a record of the initial target layout
-      .useMemo(())(_ => defaultTargetLayouts)
```
rpiaggio, line 609: We can just use the `val` here.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -690,27 +719,29 @@ object TargetTabContents extends TwoPanels:
...
                       optViewAgwo
                         .map(agwov =>
-                          applyObs(id.idSet.toList,
-                                   treeTargets,
-                                   loading,
-                                   agwov,
-                                   ctx,
-                                   props.listUndoStacks,
-                                   props.expandedIds
+                          applyObs(
```
toddburnside, line 722: In `val treeTargets` above I think you want to get rid of the `getOrElse` because `selectedIds` is more accurate than `props.focused.target` isn't it? Also, I think if you replaced that `getOrElse` with `selectedIds` you could get rid of this `if/else`.

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
```diff
@@ -232,7 +231,7 @@ object AsterismGroupObsList:
         props.focused.obsSet
           .flatMap(ids => props.asterismsWithObs.get.asterismGroups.get(ids).map(_.targetIds))
           .orElse(props.focused.target.map(SortedSet(_)))
-          .getOrElse(SortedSet.from(props.selectedSummaryTargets))
```
toddburnside, line 235: I think you also want to get rid of the `orElse` above, right?

toddburnside:
Does it behave well with multiple selection in the summary table?

rpiaggio:
> Does it behave well with multiple selection in the summary table?

If multiple targets are selected, then the editor is not shown. The layout is the same as when nothing is selected.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
rpiaggio: Let's try that.

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -105,7 +104,8 @@ object Tile {
             .runNow()
 
         <.div(
-          ExploreStyles.Tile |+| ExploreStyles.FadeIn |+| p.tileClass.orEmpty
+          ExploreStyles.Tile |+| ExploreStyles.FadeIn |+| p.tileClass.orEmpty,
+          ^.key := p.id.value
```
cquiroz, line 108: Good

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -648,11 +677,11 @@ object TargetTabContents extends TwoPanels:
       }
       .useToastRef
       // Selected targets on the summary table
-      .useStateView(List.empty[Target.Id])
```
cquiroz, line 651: Con we still multi select?

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
```diff
@@ -184,7 +198,16 @@ object TargetSummaryTable extends TableHooks:
...
       .useEffectWithDepsBy((_, _, _, _, table, _, _) =>
         table.getSelectedRowModel().rows.toList.map(_.original.id)
-      )((props, _, _, _, _, _, _) => ids => props.selectedTargetIds.set(ids))
+      )((props, ctx, _, _, _, _, _) =>
+        ids =>
+          props.selectedTargetIds.set(ids) >>
+            ids.headOption
+              .filter(_ => ids.length === 1) // Only if there's just 1 target selected
+              .map(targetId =>
+                ctx.pushPage(AppTab.Targets, props.programId, Focused.target(targetId))
```
cquiroz, line 207: Does this mean we always see the target summary, even when selecting a target on the left side?

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -130,6 +141,30 @@ object TargetTabContents extends TwoPanels:
...
   )
 
+  private val singleLayoutMedium: Layout = Layout(
+    List(
+      LayoutItem(
+        i = ObsTabTilesIds.TargetSummaryId.id.value,
+        x = 0,
+        y = 0,
+        w = DefaultWidth.value,
+        h = 0, // This doesn't matter, we are forcing 100%.
```
cquiroz, line 151: How are we doing the 100% forcing?

toddburnside (approved):
LGTM

cquiroz (approved):
LGTM

explore/src/main/scala/explore/tabs/TargetTabContents.scala
rpiaggio: > Con we still multi select?

Yes, but the editor is not shown in that case.

explore/src/main/scala/explore/targets/TargetSummaryTable.scala
rpiaggio: > Does this mean we always see the target summary, even when selecting a target on the left side?

No, the target summary is only shown when no asterism or observation is selected.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
rpiaggio: > How are we doing the 100% forcing?

Via `ExploreStyles.SingleTileMaximized`, which I see you mentioned in another comment. It justs sets `height: 100% !important` (same for `width`) in the `TileController` and the `Tile`.

PR #2545, opened 2022-12-22
Target magnifying glass opens catalog search popup

rpiaggio:
Kapture 2022-12-21 at 23 19 59

Also, propagate `disabled` flag to all target editor components.

### DIFF
--- a/explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
+++ b/explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
@@ -25,7 +25,7 @@ import java.time.Instant
 
 object SiderealTargetEditorTile {
 
-  def siderealTargetEditorTile(
+  def sideralTargetEditorTile(
     userId:     Option[User.Id],
     targetId:   Target.Id,
     target:     View[Target.Sidereal],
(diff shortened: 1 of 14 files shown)


rpiaggio:
For the moment, only a catalog search is performed. When a target is selected, its information overwrites the current target's.

Matches within the target library are not shown. It's not clear to me what should happen if an existing target is selected. One option is to replace the target in the selected asterism/observation, but we may be editing the target directly in the library, where there's no asterism or observation. Furthermore, this would be an operation in the asterism, which doesn't seem right that it could be done from within one of its targets.

toddburnside:
> When a target is selected, its information overwrites the current target's.

Does this work with the shared targets "only these observations" bit? In other words, if only some of the observations for a target are being edited, does it do the cloning be overwriting?

>  It's not clear to me what should happen...

Yeah, I thought this approach would have some behavior questions.

explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
```diff
@@ -25,7 +25,7 @@ import java.time.Instant
 
 object SiderealTargetEditorTile {
 
-  def siderealTargetEditorTile(
```
toddburnside, line 28: Why the renaming? `sidereal` is correct, isn't it?

rpiaggio:
> Does this work with the shared targets "only these observations" bit? In other words, if only some of the observations for a target are being edited, does it do the cloning be overwriting?

That's a good question and the answer is: it does the same as it did before (the magnifying glass button), which I'm not sure what it was.

explore/src/main/scala/explore/tabs/SiderealTargetEditorTile.scala
rpiaggio: Good catch. That's a mistake, I must have renamed it something else and then renamed it back.

cquiroz (approved):
LGTM

PR #2546, opened 2022-12-22
Improve the AGS flip mode with a new algorithm and allow manual

cquiroz:
In this version we are globally looking at all the PAs (just two in flip mode) and find the best possible including brightness which wasn't included before.
We can also now do manual override and will change the PA if needed:

https://user-images.githubusercontent.com/3615303/209148348-0e92a620-309e-4fe8-9185-93d3a7d06dbd.mp4


toddburnside (approved):
Looks good as far as I can tell. 😄

PR #2547, opened 2022-12-22
Convert TopBar and main menu to primereact

toddburnside:
It looks like what I demoed in the UI meeting.


common/src/main/webapp/sass/explore.scss
```diff
@@ -1648,17 +1682,6 @@ table.explore-border-table {
   }
 }
 
-.p-component.p-button.pl-tiny {
```
cquiroz, line 1651: I guess these are now on lucuma-ui

explore/src/main/scala/explore/TopBar.scala
```diff
@@ -69,10 +67,11 @@ object TopBar:
       .useState(false) // isProgramsOpen
       .useStateView(IsAboutOpen(false))
       .useState(false) // just to force rerenders
-      .useEffectResultWithDepsBy((_, _, _, _, toggle) => toggle.value)((_, _, _, _, _) =>
+      .usePopupMenuRef
```
cquiroz, line 70: I guess this comes from lucuma-ui too?

cquiroz (approved):
LGTM

common/src/main/webapp/sass/explore.scss
toddburnside: Yes. The lucuma-ui classes work for this now.

explore/src/main/scala/explore/TopBar.scala
toddburnside: That comes from lucuma-react

toddburnside: It's like the ToastRef

PR #2550, opened 2022-12-23
Enforce SED always present when importing

rpiaggio:
When the target data was replaced by using the magnifying glass, the DB update was failing because of the missing SED.


cquiroz (approved):
LGTM

PR #2555, opened 2022-12-24
Target save error handling and global toastRef

rpiaggio:
- There is a single `Toast` instance in the app and its `toastRef` is stored in the `AppContext`.
- User is informed now if there's an error when saving a target to the DB.

We should consider notifying users of errors in all interactions with the DB.

### DIFF
--- a/common/src/main/scala/explore/utils/package.scala
+++ b/common/src/main/scala/explore/utils/package.scala
@@ -3,15 +3,20 @@
 
 package explore.utils
 
+import cats.Applicative
 import cats.Endo
+import cats.Monad
 import cats.Monoid
+import cats.effect.Deferred
 import cats.effect.Sync
+import cats.effect.kernel.Sync
 import cats.syntax.all.*
 import clue.data.*
 import clue.data.syntax.*
 import crystal.Pot
 import crystal.PotOption
 import crystal.react.View
+import crystal.react.implicits.*
 import crystal.react.reuse.*
 import eu.timepit.refined.*
 import eu.timepit.refined.types.string.NonEmptyString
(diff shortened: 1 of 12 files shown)


explore/src/main/scala/explore/targeteditor/SearchForm.scala
```diff
@@ -106,8 +107,14 @@ object SearchForm:
               ).tiny.compact,
               onSelected = targetWithId =>
                 (targetWithId.target match
-                  case t @ Target.Sidereal(_, _, _, _) => props.targetSet(t)
-                  case _                               => Callback.empty
+                  case t @ Target.Sidereal(_, _, _, _) =>
+                    props.targetSet(t)
+                  // .onError { case t: Throwable =>
```
cquiroz, line 112: Commented code

cquiroz (approved):
LGTM

rpiaggio:
Although errors are now informed to the users, I'm not entirely happy with this resolution. It leaves the UI and DB in different states.

What do you all think we should do in this case? Undo the UI change? Requery the DB in order to sync the UI? Retry the mutation? Should we differentiate recoverable and unrecoverable errors? Let's discuss in https://app.shortcut.com/lucuma/story/1393/handle-errors-when-calling-the-api-also-in-subscriptions

Whatever we decide to do, let's adopt it for all DB mutation invocations.

common/src/main/scala/explore/utils/package.scala
```diff
@@ -150,16 +155,18 @@ def clearInputIcon[EV[_], A](
...
     .orUndefined
 
+// TODO Move these to lucuma-ui
 extension (toastRef: ToastRef)
-  def info(text: String) =
+  def show(text: String, severity: Message.Severity = Message.Severity.Info): Callback =
     toastRef.show(
       MessageItem(
         content = <.span(Icons.InfoLight.withSize(IconSize.LG), text),
+        severity = severity,
```
toddburnside, line 164: I know this is old news now, but it looks to me as though all toasts using this method use the `InfoLight` icon regardless of severity?

rpiaggio: Yup, good catch. We should probable make the whole `content: VdomNode` a parameter.

toddburnside: Are the `built in` prime icons so different from the FA ones that they are not usable for the toasts?

toddburnside: You can set the icon in the `MessageItem`, but it only seems to work for `Messages`, not `Toasts`. I'll look deeper and see if there is a way to make it work. But, if that doesn't work, maybe rather than passing the whole `content` as a parameter, we could select an FA icon based on the severity.

PR #2561, opened 2022-12-26
Port basic spectroscopy and imaging configuration panels to prime react

cquiroz:
There is a customized filter drop down view not yet ported. It will be tackled in another PR


rpiaggio (approved):
LGTM

PR #2563, opened 2022-12-27
Port the side tabs to prime

cquiroz:


common/src/main/scala/explore/components/ui/ExploreStyles.scala
```diff
@@ -69,6 +69,9 @@ object ExploreStyles:
   val MainMenuDropdown: Css      = Css("main-menu-dropdown")
   val SideTabs: Css              = Css("sidetabs")
   val SideButton: Css            = Css("side-button")
+  val SideTabGroup1: Css         = Css("side-tabs-group-1")
+  val SideTabGroup2: Css         = Css("side-tabs-group-2")
+  val SideTabGroup3: Css         = Css("side-tabs-group-3")
```
rpiaggio, line 74: It seems that group-3 is not used.

rpiaggio: And group-1 and -2 are treated the same way, can we unify?

cquiroz: You're right we could simplify

rpiaggio (approved):
LGTM

PR #2565, opened 2022-12-28
Better handling of ags flip mode

cquiroz:
Another attempt at handling all the cases where we  do ags flip be automatically or by user request

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinCell.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinCell.scala
@@ -95,14 +95,13 @@ case class AladinCell(
 }
 
 trait AladinCommon:
-  val asReu: Reusability[Asterism] = Reusability.by(x => (x.toSiderealTracking, x.focus.id))
-  given Reusability[Asterism]      = asReu.logNonReusable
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -94,18 +95,23 @@ case class AladinCell(
 }
 
 trait AladinCommon:
-  given Reusability[Asterism] = Reusability.by(x => (x.toSiderealTracking, x.focus.id))
-  given Reusability[AgsState] = Reusability.byEq
+  val asReu: Reusability[Asterism] = Reusability.by(x => (x.toSiderealTracking, x.focus.id))
+  given Reusability[Asterism]      = asReu.logNonReusable
```
rpiaggio, line 99: Do we want to keep the log in production?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinCell.scala
cquiroz: Nope, I'll remove it

PR #2566, opened 2022-12-28
More prime migration, plus other fixes

rpiaggio:
- I was using an Accordion for brightness table when a Panel sufficed.
- ITC errors in the table tooltips didn't show a newline.

### DIFF
--- a/explore/src/main/scala/explore/config/sequence/GeneratedSequenceTables.scala
+++ b/explore/src/main/scala/explore/config/sequence/GeneratedSequenceTables.scala
@@ -10,8 +10,7 @@ import lucuma.core.model.Observation
 import lucuma.core.model.sequence.*
 import lucuma.ui.syntax.all.given
 import react.common.ReactFnProps
-import react.semanticui.elements.header.Header
-import react.semanticui.elements.segment.Segment
+import react.primereact.Panel
 
 case class GeneratedSequenceTables(obsId: Observation.Id, config: FutureExecutionConfig)
     extends ReactFnProps(GeneratedSequenceTables.component)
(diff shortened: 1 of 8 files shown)


explore/src/main/scala/explore/config/sequence/GeneratedSequenceTables.scala
```diff
@@ -22,20 +21,18 @@ object GeneratedSequenceTables:
   private val component =
     ScalaFnComponent
       .withHooks[Props]
-      .render { props =>
-        <.div(^.height := "100%", ^.overflow.auto)(
```
cquiroz, line 26: Nice there are done now in css?

project/Settings.scala
```diff
@@ -323,6 +324,12 @@ object Settings {
       Seq("edu.gemini" %%% "refined-algebra" % refinedAlgebra)
     )
 
+    val ScalaCollectionContrib = Def.setting(
+      deps(
+        "org.scala-lang.modules" %%% "scala-collection-contrib"
```
cquiroz, line 329: What are we using this for? I thought this mostly needed for cross built libraries

cquiroz (approved):
LGTM

explore/src/main/scala/explore/config/sequence/GeneratedSequenceTables.scala
rpiaggio: Turns out it's not necessary anymore. I assume it was in the migration to Prime.

project/Settings.scala
rpiaggio: For `list.intersperse(...)`.

PR #2575, opened 2022-12-30
Migrate proposal editor to prime.

rpiaggio:
This was a bit challenging since the grid structure was based on the fact the a `form`'s `label`+`input` were contained in a `div` before, and now they aren't.


cquiroz (approved):
LGTM, thanks

PR #2585, opened 2023-01-04
Support unconstrained and parallactic override modes

cquiroz:
This required a bit more work than I expected as we needed to support the undefined constraint case. Took the opportunity to do a bit of a cleanup

https://user-images.githubusercontent.com/3615303/210599697-21d4178d-12c9-4bfd-9322-359814db8eb9.mp4

### DIFF
--- a/common/src/main/scala/explore/model/PAProperties.scala
+++ b/common/src/main/scala/explore/model/PAProperties.scala
@@ -6,7 +6,6 @@ package explore.model
 import cats.syntax.all.*
 import crystal.react.View
 import crystal.react.implicits.*
-import crystal.react.reuse.*
 import explore.model.enums.AgsState
 import explore.model.reusability.given
 import japgolly.scalajs.react.Reusability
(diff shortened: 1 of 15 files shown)


common/src/main/scala/explore/model/PAProperties.scala
```diff
@@ -0,0 +1,44 @@
...
+object PAProperties:
+  // Move to lucuma-catalog
+  given Reusability[AgsAnalysis] = Reusability {
+    case (AgsAnalysis.Usable(a, b, c, d, e), AgsAnalysis.Usable(f, g, h, i, j)) =>
+      (a === f) && (b === g) && (c === h) && (d === i) && (e === j)
+    case _                                                                      => false
+  }
+
+  given Reusability[PAProperties] = Reusability.by(x =>
+    (x.oid, x.selectedGS.reuseByValue, x.agsState.reuseByValue, x.constraint.reuseByValue)
```
rpiaggio, line 43: Might be simpler just to use `.get`.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/model/PAProperties.scala
cquiroz: oh good to know

PR #2600, opened 2023-01-09
Replace last SUI component and remove lucuma-react-semantic-ui

toddburnside:
I was able to remove `lucuma-react-semantic-ui` from the build and clean up our CSS somewhat, but am not yet able to remove the `fomantic-ui-less` npm package. It seems some of our styles still rely on it. I will continue to look into that in parallel with the work on getting the new backend working in Explore.

image


cquiroz (approved):
Looking Great!

PR #2617, opened 2023-01-11
Launcher improvemnts and better detection of ctrl key

cquiroz:
Attempts to fix ctrl key detection (isHotkeyPressed seems unreliable) and adds some improvements to the application launch logic

### DIFF
--- a/explore/src/main/webapp/main.jsx
+++ b/explore/src/main/webapp/main.jsx
@@ -39,6 +39,5 @@ if (import.meta.hot) {
   import.meta.hot.accept();
   import.meta.hot.dispose((_) => {
     // Reset the IO runtime
-    Explore.resetIOApp();
   });
 }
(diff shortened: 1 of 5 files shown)


explore/src/main/webapp/main.jsx
```diff
@@ -39,6 +39,5 @@ if (import.meta.hot) {
   import.meta.hot.accept();
   import.meta.hot.dispose((_) => {
```
toddburnside, line 40: Do we need this at all if it doesn't do anything?

toddburnside (approved):
LGTM

explore/src/main/webapp/main.jsx
cquiroz: We don't really need it I thought to keep it as a reminder but maybe I should rather delete it

PR #2618, opened 2023-01-12
Rework the way shortcuts are built

cquiroz:
With the new version of hotkeys some shortcuts are not working. This should re enable cmd/ctrl+v/c

### DIFF
--- a/explore/src/main/scala/explore/shortcuts.scala
+++ b/explore/src/main/scala/explore/shortcuts.scala
@@ -58,13 +58,11 @@ val Down = Shortcut("j")
 val Up   = Shortcut("k")
 
 val CopyAlt1 = Shortcut("y")
-val CopyAlt2 = Shortcut("ctrl+c")
-val CopyAlt3 = Shortcut("meta+c")
+val CopyAlt2 = Shortcut("meta+c")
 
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/shortcuts.scala
```diff
@@ -23,29 +34,37 @@ type ShortcutCallbacks = PartialFunction[Shortcut, Callback]
...
 
 val Esc           = Shortcut("ESC")
-val ShortcutsHelp = Shortcut("shift+/")
+val ShortcutsHelp = Shortcut("F1")
 
 val Down = Shortcut("j")
 val Up   = Shortcut("k")
 
 val CopyAlt1 = Shortcut("y")
 val CopyAlt2 = Shortcut("ctrl+c")
```
toddburnside, line 61: I'm not sure how this all works, but there is no `ctrl` in the `withModifiers` method. Is that OK?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/shortcuts.scala
cquiroz: Right, I should delete these, `meta` should cover `cmd` and `ctrl`

PR #2636, opened 2023-01-20
Remove leftover sui files

cquiroz:


common/src/main/webapp/less/legacyvariables.less
```diff
@@ -1,6 +1,11 @@
-// vite with SUI has troubles loading fonts
+// ----------------------------------------
+// Legacy variables from the SUI dark theme
+// ----------------------------------------
 @importGoogleFonts: false;
 
+/* This is the single variable that controls them all */
+@emSize   : 14px;
+
```
toddburnside, line 8: This is great! I wasn't sure what variables I need to pull from the old SUI files to keep this working for now. 
I guess the next step is to figure out how to switch to the primereact variables. I started working on that a couple of times while waiting for Rob to make changes, but didn't get very far.

toddburnside (commented):
LGTM. Thanks!

common/src/main/webapp/less/legacyvariables.less
cquiroz: Right, that's the next step

PR #2645, opened 2023-01-25
First try at calculating the average parallactic angle for AGS

cquiroz:
This is a first attempt at getting the average parallactic angle for AGS

The code here needs to be moved to lucuma-core/lucuma-catalogg but it is here for user validation


toddburnside (approved):
LGTM

PR #2647, opened 2023-01-26
Detect avg paralactic angle changes

cquiroz:
Reusability is setup to only recalculate if the vizTime changes a month but we also need to detect if the avg parallactic angle change. This PR uses the angle as a factor for reusability


rpiaggio (approved):
LGTM

PR #2664, opened 2023-02-02
Add debug info for moon brightness, fix (c) year

rpiaggio:
Andy pointed out that the moon brightness value is not exactly the expected one. I'm adding debug info so we can track the issue.


model-testkit/shared/src/main/scala/explore/boopickle/package.scala
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
```
toddburnside, line 1: Heh. I wondered how adding debug information for the moon phase could require modifying 19 files.

toddburnside (approved):
LGTM

PR #2665, opened 2023-02-03
Sort the target tree by name

toddburnside:
I'm not sure this is the best approach. But, since the target names aren't part of the data that is creating the tree, the best I came up with was sorting it at the usage site.

Originally, it was essentially being sorted by the first obsId in the the group, but somewhere along the line, that sorting got lost. So, the initial sort order was by the number of observations in the group. Which is really not a reasonable sort order.


rpiaggio (approved):
LGTM

PR #2682, opened 2023-02-09
Improve dithers and offsets editing

toddburnside:
Instead of just having a placeholder of the `default` value for dithers or offsets when an `explicit` value has not been set, this puts the `default` value in the text edit and allows the user to edit it. But, it also allows the user to revert to `none` via either a `x` icon or editing it to the original value.


rpiaggio (approved):
LGTM

PR #2688, opened 2023-02-10
More configuration editing improvements

toddburnside:
All of the controls now have the appearance of editing the configuration directly - hiding the fact that there are still essentially `overrides`

Also, something @rpiaggio said yesterday made me realize that the clear button was not really the proper means of reverting a config value. So, now the icon that indicates a field has been `customized` is clickable and will revert the value. The only control that still potentially has a clear button is the filter, which really can be set to `None`.


rpiaggio (approved):
LGTM

PR #2699, opened 2023-02-14
Subqueries

rpiaggio:


common-graphql/src/main/scala/queries/common/AsterismQueriesGQL.scala
```diff
@@ -7,18 +7,18 @@ import clue.GraphQLOperation
...
 
 object AsterismQueriesGQL {
 
   @GraphQL
   trait AsterismGroupObsQuery extends GraphQLOperation[ObservationDB] {
-    val document: String = """
-      query($programId: ProgramId!) {
-        asterismGroup(programId: $programId) {
+    val document: String = s"""
+      query($$programId: ProgramId!) {
```
toddburnside, line 20: Double `$` now? I guess a single `$` is for fragments?

PR #2706, opened 2023-02-15
More subqueries and renames

rpiaggio:


common-graphql/src/main/scala/queries/common/AsterismQueriesGQL.scala
```diff
@@ -63,20 +63,7 @@ object AsterismQueriesGQL {
...
-                centralWavelength $WavelengthSubquery
-              }
-              gmosSouthLongSlit {
-                grating
-                filter
-                fpu
-                centralWavelength $WavelengthSubquery
-              }
-            }
+            observingMode $BasicConfigurationSubquery
```
toddburnside, line 66: It is very cool that the fragments are typed!

toddburnside (approved):
LGTM. Thanks for taking on all of these changes!

PR #2711, opened 2023-02-16
Fix problem of ITC not being called for spectroscopy modes table.

toddburnside:
The problem was that the rows that needed itc calls for were checked before the data for the table was loaded. It wasn't checked again unless an itc parameter was changed. So, now I also check to see if the table data has changed and recalculate the itc if it has. 

I'm guessing that until recently, the data had always loaded by the time this point was hit. But, something changed that...

### DIFF
--- a/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
+++ b/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
@@ -557,7 +557,6 @@ private object SpectroscopyModesTable extends TableHooks:
                 (for
                   _       <- Resource.eval(itcProgress.setStateAsync(progressZero))
                   request <-
-                    println("Calling ItcClient.request(ItcMessage.Query)")
                     ItcClient[IO]
                       .request(
                         ItcMessage.Query(w, sn, constraints, t, modes.map(_.entry), signalToNoiseAt)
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -505,21 +505,23 @@ private object SpectroscopyModesTable extends TableHooks:
       // atTop
       .useState(false)
       // Recalculate ITC values if the wv or sn change or if the rows get modified
-      .useStreamResourceBy((props, _, _, _, _, _, _, _, _, visibleRows, _) =>
```
toddburnside, line 508: I think at one point the visible rows were sent to the itc first, but that is no longer used. I guess it was probably replaced by getting the rows via the table sort order.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -555,6 +557,7 @@ private object SpectroscopyModesTable extends TableHooks:
                 (for
                   _       <- Resource.eval(itcProgress.setStateAsync(progressZero))
                   request <-
+                    println("Calling ItcClient.request(ItcMessage.Query)")
```
rpiaggio, line 560: Debug message?

rpiaggio (approved):
LGTM. Good catch!

PR #2717, opened 2023-02-19
Fix scrolling issues with spectroscopy table when sorted.

toddburnside:


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -111,12 +112,22 @@ case class SpectroscopyModesTable(
...
   private given Reusability[EitherNec[ItcQueryProblems, ItcResult]]                        = Reusability.byEq
   private given Reusability[SpectroscopyModesMatrix]                                       = Reusability.by(_.matrix.length)
   private given Reusability[ItcResultsCache]                                               = Reusability.by(_.cache.size)
   private given Reusability[Map[ItcRequestParams, EitherNec[ItcQueryProblems, ItcResult]]] =
     Reusability.never
 
+  private given Reusability[ColumnId]      = Reusability.by(_.value)
+  private given Reusability[SortDirection] = Reusability.by(_.toDescending)
+  private given Reusability[ColumnSort]    = Reusability.derive
+  private given Reusability[Sorting]       = Reusability.by(_.value)
```
rpiaggio, line 129: Are these general enough that they can be moved to `explore.model.reusability` (or `lucuma.ui.reusability`) ?

toddburnside: I was wondering about that. Currently they are private in the table, but I can see we might want to use them again, so I could move them to lucuma-ui.

rpiaggio (approved):
LGTM

PR #2725, opened 2023-02-21
PM correct the base target of asterisms

cquiroz:
We should calculate the base position with the average of each target position corrected by proper motion. At the moment the base is calculated with the uncorrected values

before:
Explore 2023-02-21 13-06-55

After
Explore 2023-02-21 13-10-59


rpiaggio (approved):
LGTM

toddburnside (approved):
LGTM

PR #2726, opened 2023-02-21
Paint visualization time in night plot

rpiaggio:
For the moment, duration is fixed at 1 hour since we are not computing observation time.

image

Showing only elevation and parallactic angle:

image


cquiroz:
Looks very good, I wonder if we should only plot parallactic angle, @andrewwstephens

cquiroz (approved):
LGTM

rpiaggio:
> Looks very good, I wonder if we should only plot parallactic angle, @andrewwstephens

I'll merge for now and we can "unpaint" the other series later.

andrewwstephens:
I think that looks great!  :+1:

PR #2727, opened 2023-02-21
Fix scrollTo of spectroscopy modes table.

toddburnside:
Having the `auto scrollTo` depend on the data resulted in it scrolling back to the selected row with each time update from ITC. It seems just `scrolling to` initially and then when the sort order changes is sufficient for our needs.


cquiroz (approved):
LGTM

PR #2728, opened 2023-02-21
Display catalog object info in target editor

rpiaggio:
image

image


explore/src/main/scala/explore/targeteditor/SpectralDefinitionEditor.scala
```diff
@@ -195,6 +194,14 @@ sealed abstract class SpectralDefinitionEditorBuilder[
         )
 
       React.Fragment(
+        props.catalogInfo.flatMap(ci =>
+          ci.objectType.map(ot =>
+            React.Fragment(
+              FormLabel(htmlFor = "catalogInfo".refined)(ci.catalog.shortName),
```
cquiroz, line 200: What if we get this from an imported target?

cquiroz: I see it comes from the `Display` instance

rpiaggio: Right. But I'll add a label for the imported case.

PR #2729, opened 2023-02-22
Use custom zoom buttons on the search aladin control

cquiroz:
The aladin zoom controls on the search popup are the original ones and don't match what we use on the main editor. This fixes it

Before:
image

After:
image


rpiaggio (approved):
LGTM

PR #2731, opened 2023-02-22
In modes table, handle case of no brightnesses

toddburnside:
In the case where the target for an observation didn't have any brightnesses defined, all the rows in the spectroscopy modes table would have spinners for the itc results an would never resolve. This is because the `brightestAt` extension I modified would return a target, which would result in the SpectroscopyModesTable attempting to make an `ItcClient.request`. However, `ItcRequests.queryItc` calls the `selectedBand` method and maps on the result for the requests. Since `selectedBand` returns `None`, no itc call is made, but the SpectroscopyModesTable doesn't know that.

The easiest solution for me to understand was to filter the targets in `brightestAt` so that if there are no targets with brightnesses,  the table rows will have a `Missing target info` itc error message. It is possible, even likely, this is not the best solution, but I created this PR as a starting point.

UPDATE:
 I added an `ItcQueryProblems.MissingBrightness` to differentiate the message from the `MissingTargetInfo`. 

Also, whenever the wavelength was missing, the `MissingTargetInfo` error would come along for the ride because even if the target(s) are OK, there is no `brightestAt` without an `At` wavelength. So, I suppressed the `MissingTargetInfo` message if there isn't a wavelength. This does mean they may only see a `MissingWavelength` error, but when they fix it a `MissingBrightness` will appear. That seemed better than having the spurious error.

### DIFF
--- a/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
+++ b/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
@@ -211,7 +211,10 @@ private object SpectroscopyModesTable extends TableHooks:
     case FocalPlane.MultipleSlit => "Multi"
     case FocalPlane.IFU          => "IFU"
 
-  private def itcCell(c: EitherNec[ItcQueryProblems, ItcResult]): VdomElement = {
+  private def itcCell(
+    c: EitherNec[ItcQueryProblems, ItcResult],
+    w: Option[Wavelength]
+  ): VdomElement = {
     val content: TagMod = c match
       case Left(nel)                        =>
         if (nel.exists(_ == ItcQueryProblems.UnsupportedMode))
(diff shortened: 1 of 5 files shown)


cquiroz:
Could the `Missing Target Info` message be something like `No brightness defined` instead? (For this specific case)

toddburnside:
> Could the `Missing Target Info` message be something like `No brightness defined` instead? (For this specific case)

I guess that would be better. I just have to figure out a different way of surfacing that error (and add another ItcQueryProblem).  I think I can move the `selectedBand` check to the `ItcResultsCache`. The brightest target will already have been selected at that point, but I can check to see if `selectedBand` returns a Some for it.

rpiaggio (approved):
LGTM

cquiroz (approved):
LGTM

PR #2732, opened 2023-02-22
 Add help icon for catalog info, label for imported targets

rpiaggio:


explore/src/main/scala/explore/targeteditor/SpectralDefinitionEditor.scala
```diff
@@ -197,7 +198,13 @@ sealed abstract class SpectralDefinitionEditorBuilder[
         props.catalogInfo.flatMap(ci =>
           ci.objectType.map(ot =>
             React.Fragment(
-              FormLabel(htmlFor = "catalogInfo".refined)(ci.catalog.shortName),
+              FormLabel(htmlFor = "catalogInfo".refined)(
```
cquiroz, line 201: Would it be better to change the `Display` instance?

rpiaggio: It didn't seem right that the display instance for `CatalogName.Import` was "Object Type".

The thing is, we are collapsing two fields (catalog name and object info) into one. One is the label and the other the value. We could instead show 2 fields and then use the `Display` instance for the catalog name.

toddburnside, line 201: Just a note. You can pass the label with icon directly to FormInputText by passing it as a Fragment instead of wrapping the label and input in a Fragment.

cquiroz: 👍

cquiroz (approved):
LGTM

explore/src/main/scala/explore/targeteditor/SpectralDefinitionEditor.scala
rpiaggio: I considered that, but does that render the same structure? This is in the context of a `grid` and we must keep the html structure. For some reason, we are using `FormLabel` for the other fields here, so I went with that.

In fact, I broke the brightness table, which is now pushed to the bottom. I will fix in a future PR.

toddburnside: I think it keeps the same structure. But, it's not a big deal to do it either way.

rpiaggio: Ah ok, the case where we have an explicit `FormLabel` is an `EnumDropdown`, which doesn't have a `label`. I'll change it in the next PR.

PR #2738, opened 2023-02-23
Use localStorage for copy/paste of targets/observations

toddburnside:
This implements Andy's request. We just need to decide if we want this implementation. It uses localStorage, so the "clipboard" is shared between all tabs within the same browser of the same "instance" of explore (staging, production, etc.). It also persists between runs of the browser, which could be seen as good or bad, I guess.

### DIFF
--- a/common/src/main/scala/explore/model/ExploreClipboard.scala
+++ b/common/src/main/scala/explore/model/ExploreClipboard.scala
@@ -1,0 +1,32 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats.effect.IO
+import explore.model.LocalClipboard.*
+import org.scalajs.dom
+
+object ExploreClipboard:
+  private val storageKey = "clipboard"
+
+  def get: IO[LocalClipboard] = IO {
+    // getItem returns null when storage isn't set.
+    Option(dom.window.localStorage.getItem(storageKey)).fold(LocalClipboard.Empty)(value =>
+      ObsIdSet.fromString
+        .getOption(value)
+        .map(LocalClipboard.CopiedObservations.apply)
+        .orElse(TargetIdSet.fromString.getOption(value).map(LocalClipboard.CopiedTargets.apply))
+        .getOrElse(LocalClipboard.Empty)
+    )
+  }
+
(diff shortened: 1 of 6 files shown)


common/src/main/scala/explore/model/ExploreClipboard.scala
```diff
@@ -0,0 +1,32 @@
...
+package explore.model
+
+import cats.effect.IO
+import explore.model.LocalClipboard.*
+import org.scalajs.dom
+
+object ExploreClipboard:
+  private val storageKey = "clipboard"
+
+  def get: IO[LocalClipboard] = IO {
```
toddburnside, line 13: This doesn't really have to be `IO`. I think everywhere we use it eventually wants a `Callback`. I just more or less recreated the interface of the previous `exploreClipboard`.

common/src/main/scala/explore/model/ExploreClipboard.scala
```diff
@@ -0,0 +1,32 @@
...
+    // getItem returns null when storage isn't set.
+    Option(dom.window.localStorage.getItem(storageKey)).fold(LocalClipboard.Empty)(value =>
```
cquiroz, line 15: Do we want to also verify if `localStorage` is present. in principle some browser don't have it or the user has disallowed it

cquiroz (approved):
LGTM

common/src/main/scala/explore/model/ExploreClipboard.scala
toddburnside: Hmmm. I can check to see how we can determine that. Then we'd have to figure out what to do if it isn't present or is disabled.

cquiroz: I think it is enough to check if `localStorage` is null

toddburnside: I guess it will throw an exception if it is disabled. I can try disabling it and see. Maybe we just log a message to console and in the the help for the shortcut keys say copy/paste requires that localStorage be enabled?

toddburnside: Or log a message to the logger, I guess.

toddburnside: OK. It seems like all of the browsers require turning off both cookies and local storage at the same time. Which would break explore, anyway. But, I suppose I should still deal with the case where local storage could be full.

PR #2746, opened 2023-02-25
Timing windows in plot

rpiaggio:
Not sure how to display these. Currently displaying as yellow-ish bands in the chart. We can adjust this.

image

image

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
+++ b/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
@@ -13,6 +13,7 @@ import crystal.react.hooks.*
 import crystal.react.implicits.*
 import eu.timepit.refined.auto.*
 import explore.*
+import explore.common.TimingWindowQueries.*
 import explore.common.UserPreferencesQueries.*
 import explore.components.HelpIcon
 import explore.components.ui.ExploreStyles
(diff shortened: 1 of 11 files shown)


explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
```diff
@@ -112,7 +117,17 @@ object ElevationPlotSection:
...
       }
-      .render { (props, ctx, options) =>
+      // FIXME This will be provided by the caller when timing windows are in the ODB
+      .useStreamResourceOnMountBy { (props, ctx, _) =>
+        import ctx.given
+
+        TimingWindowsQuery
+          .query()
+          .map(data => EntryToTimingWindows.get(data.tmpTimingWindows))
+          // _.zoom(TimingWindowsList))
```
cquiroz, line 127: Commented code

model/shared/src/main/scala/explore/model/TimingWindow.scala
```diff
@@ -106,6 +110,48 @@ case class TimingWindow private (
 
   def repeatForever: Boolean =
     repetition.exists(_.toOption.exists(_.repeatPeriod.exists(_.repeatFrequency.isEmpty)))
+
+  def toIntervalSeq(within: BoundedInterval[Instant]): IntervalSeq[Instant] = {
```
cquiroz, line 114: I wonder if this expansion logic should go to core, we probably will need it for the scheduler

cquiroz (approved):
Looks fantastic

model/shared/src/main/scala/explore/model/TimingWindow.scala
rpiaggio: Do we want the model classes in core? In that case the method will go with them.

PR #2754, opened 2023-02-28
Set the viz time from the db

cquiroz:


common/src/main/scala/explore/common/AsterismQueries.scala
```diff
@@ -117,7 +117,7 @@ object AsterismQueries:
       obsR.plannedTime.execution.toDuration,
       obsR.targetEnvironment.asterism.map(_.id).toSet,
       obsR.observingMode,
-      none, // obsR.visualizationTime,
+      obsR.visualizationTime.map(_.toInstant),
```
toddburnside, line 120: Good catch. I guess this didn't get restored after the new odb supported it.

toddburnside (approved):
LGTM

PR #2757, opened 2023-02-28
Quick fix for the ODB change to the itc query

toddburnside:
This is a temporary fix for the change to the ODB itc query that was deployed with this PR https://github.com/gemini-hlsw/lucuma-odb/pull/268

I will update lucuma-schemas, and then we can decide which way we want to go. We can 

1. call the itc directly
2. have the odb reinstate a non-failing itc query
3. use the odb itc query, but make it so that it won't fail our entire ObsEditQuery. Which, I think it will do now. I will double check that after I update lucuma-schemas.

Update:
Yes, an invalid ITC breaks the entire ObsEditQuery.
BTW, we are not currently displaying the ITC results, anyway.


cquiroz (approved):
LGTM

PR #2760, opened 2023-03-01
Update explore for the itc/sequence changes in the odb

toddburnside:
This sort of restores the sequence viewing. However:

1. Visits are not currently in the schema, so I commented that stuff out.
2. Not everything in the schema is currently mapped, so actually trying to view the sequence will result in an error. When the server has everything mapped, we shouldn't need any changes for it to work.


cquiroz (approved):
LGTM

PR #2763, opened 2023-03-02
Better handling of repeated timing windows

cquiroz:
We got a bug report when the timing windows repeat forever and the obs time is far in the future.
At the moment in that particular case we calculate many windows between the start of the window and the obs time which makes the browser hang.
This PR trims the calculation to only consider the windows that are going to be displayed on the elevation plot.


toddburnside (approved):
LGTM

rpiaggio (approved):
LGTM, thank you

PR #2772, opened 2023-03-06
Add UI to create and delete api keys

cquiroz:
Adds a user preferences UI where we can see some user properties and manage api keys

image

### DIFF
--- a/common-graphql/src/main/scala/queries/schema/SSO.scala
+++ b/common-graphql/src/main/scala/queries/schema/SSO.scala
@@ -8,8 +8,8 @@ import clue.annotation.GraphQLSchema
 @GraphQLSchema
 trait SSO:
   object Scalars:
-    type UserId   = String
-    type RoleId   = String
+    type UserId   = lucuma.core.model.User.Id
+    type RoleId   = lucuma.core.model.StandardRole.Id
     type ApiKeyId = String
 
   object Enums:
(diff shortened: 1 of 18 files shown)


common-graphql/src/main/scala/queries/schema/SSO.scala
```diff
@@ -0,0 +1,16 @@
...
+
+package queries.schemas
+
+import clue.annotation.GraphQLSchema
+
+@GraphQLSchema
+trait SSO:
+  object Scalars:
+    type UserId   = String
+    type RoleId   = String
```
rpiaggio, line 12: Can use use `User.Id` and `StandardRole.Id` here?

common/src/main/scala/explore/model/GraphQLClients.scala
```diff
@@ -9,31 +9,39 @@ import cats.syntax.all.*
...
+import org.http4s.syntax.all.*
 import org.typelevel.log4cats.Logger
 import queries.schemas.*
 
 case class GraphQLClients[F[_]: Async: Parallel] protected (
   odb:           WebSocketClient[F, ObservationDB],
   preferencesDB: WebSocketClient[F, UserPreferencesDB],
-  itc:           TransactionalClient[F, ITC]
+  itc:           TransactionalClient[F, ITC],
+  sso:           WebSocketClient[F, SSO]
```
rpiaggio, line 23: Is it worth it to keep a web socket connection open to SSO in every instance of explore? Maybe a `TransactionalClient` is enough?

(Or are you using it temporarily while we cannot refresh the token in a TransactionalClient?)

cquiroz: Yes I want to use `TransactionalClient` but we need to be able to set the auth token first

common-graphql/src/main/scala/queries/schema/SSO.scala
cquiroz: Let me try that

rpiaggio (approved):
LGTM

PR #2774, opened 2023-03-07
Link offsets to the db

cquiroz:
Currently the offsets are hardcoded but they are now available in the db on the sequence.
This PR adds a query to get them and will use the values there to drive the offset display and ags

I took the opportunity to do some refactoring as some of the obs configuration info was in several places

### DIFF
--- a/common-graphql/src/main/scala/queries/common/GeneratedSequenceSQL.scala
+++ b/common-graphql/src/main/scala/queries/common/GeneratedSequenceSQL.scala
@@ -14,18 +14,6 @@ import lucuma.schemas.decoders.given
 import lucuma.schemas.odb.*
 
 object GeneratedSequenceSQL {
-  given offsetComponentDecoder[T]: Decoder[Offset.Component[T]] = Decoder.instance(c =>
-    c.downField("microarcseconds")
-      .as[Long]
-      .map(Angle.signedMicroarcseconds.reverse.andThen(Offset.Component.angle[T].reverse).get)
-  )
-
-  given Decoder[Offset] = Decoder.instance(c =>
-    for {
-      p <- c.downField("p").as[Offset.P]
-      q <- c.downField("q").as[Offset.Q]
-    } yield Offset(p, q)
-  )
 
   @GraphQL
   trait SequenceSteps extends GraphQLOperation[ObservationDB] {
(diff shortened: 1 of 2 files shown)


common-graphql/src/main/scala/queries/common/GeneratedSequenceSQL.scala
```diff
@@ -14,6 +14,18 @@ import lucuma.schemas.decoders.given
 import lucuma.schemas.odb.*
 
 object GeneratedSequenceSQL {
+  given offsetComponentDecoder[T]: Decoder[Offset.Component[T]] = Decoder.instance(c =>
```
rpiaggio, line 17: There's a decoder already in `lucuma-schemas`.

rpiaggio: https://github.com/gemini-hlsw/lucuma-schemas/blob/main/model/src/main/scala/lucuma/schemas/decoders/CoreModelDecoders.scala#L143

common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
```diff
@@ -125,10 +130,13 @@ object ObsQueries:
             constraints = obs.constraintSet,
             targets = obs.targetEnvironment,
             posAngle = obs.posAngleConstraint,
+            offsets =
+              NonEmptyList.fromList(data.sequence.foldMap(_.executionConfig.allOffsets).distinct),
```
rpiaggio, line 134: Maybe we want a `NonEmptySet` ?

common-graphql/src/main/scala/queries/common/GeneratedSequenceSQL.scala
cquiroz: Hmm, I just moved this around. I guess we should remove it

common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
cquiroz: Could be, order is important, would Set preserve it?

rpiaggio: No, I thought order was not important.

PR #2775, opened 2023-03-08
Show acquisition offsets

cquiroz:
image


explore/src/main/scala/explore/visualization/OffsetSVG.scala
```diff
@@ -16,6 +16,9 @@ import react.common.Style
 import react.floatingui.*
 import react.floatingui.hooks.*
 
+enum OffsetType:
+  case Science, Acquisition
```
rpiaggio, line 20: Can we use `lucuma.core.enums.SequenceType` here?

cquiroz: Hmm, it is not exactly the same but the options match, I guess it can be argued that they are offsets for those sequence types. I'll make the change

PR #2776, opened 2023-03-08
Clue migration

rpiaggio:
Clue has undergone a lot of changes.

The new changes allow:
- Pass transport-layer parameters (eg: HTTP headers) when executing queries.
- Handling cases where both `data` and `errors` are included in the response.

`Transactional{Client, Backend}` has been renamed to `Fetch{Client, Backend}`.


common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
```diff
@@ -202,29 +204,29 @@ object ObsQueries:
         elevationRange = createER.assign
       ).assign
     )
-    UpdateObservationMutation
-      .execute[F](
+    val xxx: F[UpdateObservationMutation.Data] = UpdateObservationMutation[F]
```
cquiroz, line 207: `xxx` 😄

common/src/main/scala/explore/model/AppContext.scala
```diff
@@ -50,20 +51,22 @@ case class AppContext[F[_]](
...
     setPageVia(appTab, programId, focused, SetRouteVia.HistoryReplace)
 
-  given WebSocketClient[F, ObservationDB]     = clients.odb
-  given WebSocketClient[F, UserPreferencesDB] = clients.preferencesDB
-  given TransactionalClient[F, ITC]           = clients.itc
-  given TransactionalClient[F, SSO]           = clients.sso
+  given WebSocketJSClient[F, ObservationDB]     = clients.odb
+  given WebSocketJSClient[F, UserPreferencesDB] = clients.preferencesDB
+  given FetchJSClient[F, ITC]                   = clients.itc
+  given WebSocketJSClient[F, SSO]               = clients.sso
```
cquiroz, line 57: We may want to switch this one to `FetchJSClient` too but can we reset the header already?

workers/src/main/scala/workers/ItcServer.scala
```diff
@@ -57,14 +58,14 @@ object ItcServer extends WorkerServer[IO, ItcMessage.Request] with ItcPicklers {
...
-        given TransactionalBackend[IO] = FetchJSBackend[IO](FetchMethod.GET)
-        TransactionalClient.of[IO, ITC](config.itcURI, "ITC")
+      self                          <- IO(dom.DedicatedWorkerGlobalScope.self)
+      cache                         <- Cache.withIDB[IO](self.indexedDB.toOption, "explore-itc")
+      _                             <- cache.evict(CacheRetention).start
+      matrix                        <- Deferred[IO, SpectroscopyModesMatrix]
+      config                        <- fetchConfig[IO]
+      given FetchClient[IO, ?, ITC] <- {
+        given FetchJSBackend[IO] = FetchJSBackend[IO](FetchMethod.GET)
+        FetchJSClient.of[IO, ITC](config.itcURI.toString, "ITC")
```
cquiroz, line 68: Why no plain urls in clue clients?

cquiroz (approved):
LGTM, thanks

common-queries/src/main/scala/queries/schemas/odb/ObsQueries.scala
rpiaggio: LOL, that was an experiment

common/src/main/scala/explore/model/AppContext.scala
rpiaggio: Yes, we should be able to by using an additional `modParams` parameter present in all queries. Let's attempt to do that in the next PR.

workers/src/main/scala/workers/ItcServer.scala
rpiaggio: `Uri` is an `http4s` class. I removed the `http4s` dependency from the `clue-scalajs` module, since it was used just to import this class. `FetchJSBackend` now exposes what the underlying `fetch` API uses for URLs, which is a plain `String`.

There are `http4s` clients in `clue` which do make use of `Uri`. We should evaluate whether to use these clients.

Finally, I guess we *could* use `java.util.Uri` here, I'm not sure.

PR #2780, opened 2023-03-08
Fix for partially working ObsEditQuery

cquiroz:
This is  fix for queries not working when we dont' have enough data to calculate a sequence, e.g. if an obs has no target yet

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -6,6 +6,7 @@ package explore.tabs
 import cats.data.NonEmptyList
 import cats.effect.IO
 import cats.syntax.all.*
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -136,9 +137,9 @@ object ObsTabTiles:
         import ctx.given
 
         ObsEditQuery[IO]
-          .query(props.programId, props.obsId)
+          .query(props.programId, props.obsId)(ErrorPolicy.RaiseOnNoData)
           .map(
-            _.asObsEditData
+            _.data.asObsEditData
```
rpiaggio, line 142: Are errors ignored when there's data present then?

rpiaggio (approved):
LGTM

PR #2782, opened 2023-03-08
Switch SSO client to FetchJS

rpiaggio:


explore/src/main/scala/explore/users/UserPreferencesPopup.scala
```diff
@@ -97,8 +99,18 @@ object UserPreferencesContent:
   private val IdColumnId: ColumnId      = ColumnId("id")
   private val RoleColumnId: ColumnId    = ColumnId("role")
 
-  private def deleteKey(key: String, active: View[IsActive], newKey: View[NewKey])(using
-    FetchClient[IO, ?, SSO],
+  private def addToken(vault: UserVault)(request: FetchJSRequest): FetchJSRequest =
```
cquiroz, line 102: I guess this is going to be a somewhat common patternn, maybe add it to `UserVault` or make it a common syntax?

PR #2783, opened 2023-03-10
Avoid double scrollbars

cquiroz:


rpiaggio (approved):
LGTM

common/src/main/webapp/sass/explore.scss
```diff
@@ -2141,11 +2141,9 @@ div.spectroscopy-table-emptymessage {
 
 .explore-single-tile-maximized.react-grid-layout {
   height: 100% !important; // !important overrides element attributes
-  width: 100% !important;
```
rpiaggio, line 2144: I wonder why we added these.

cquiroz: Me too

PR #2788, opened 2023-03-13
Fix hidden brightnesses table height

hugo-vrijswijk:
image
image

Scrolling list:

image

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -1,4 +1,4 @@
-@use 'common.scss';
+@use "common.scss";
 
 $tree-section-width: 300px;
 
@@ -55,7 +55,7 @@ body {
 
 // Reusability overlays.
 body.hide-reusability {
-  & > div:not(#root)[title^='Last update reason'] {
+  & > div:not(#root)[title^="Last update reason"] {
     display: none;
   }
 }
@@ -350,7 +350,7 @@ a:hover {
     }
 
     label[data-abbrv]::after {
-      content: '';
+      content: "";
     }
 
     label[data-abbrv] > span {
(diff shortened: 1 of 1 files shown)


common/src/main/webapp/sass/explore.scss
```diff
@@ -65,7 +65,7 @@ body.hide-reusability {
   line-height: var(--pl-input-line-height);
   border: 1px dashed transparent; // the border makes the sizing work.
   /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
-  padding: var(--pl-input-padding)
+  padding: var(--pl-input-padding);
```
cquiroz, line 68: Did you run this through a formatter to get these?
We may want to agree if so, I remember using prettier at some moment but we haven't enforced it

hugo-vrijswijk: Yeah my editor auto-formatted it with Prettier. I can revert it if you want, but since it also added semicolons (which _should_ be in valid css anyway) I decided to commit it. I can revert it if you want

explore/src/main/scala/explore/targeteditor/BrightnessesEditor.scala
```diff
@@ -165,36 +165,41 @@ sealed abstract class BrightnessesEditorBuilder[T, Props <: BrightnessesEditor[T
...
-                  Button(
-                    icon = Icons.New,
-                    onClick = addBrightness,
-                    disabled = props.disabled,
-                    severity = Button.Severity.Secondary
-                  ).mini.compact
-                )
-              }
-              .whenDefined
+          <.tr(
```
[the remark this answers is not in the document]
hugo-vrijswijk: Sort of, the footer is now in the actual `<footer>` element

common/src/main/webapp/sass/explore.scss
cquiroz: I think it is fine, maybe others can chime in but we may want to have the same settings or we could endup changing the format back and forth 😄

cquiroz (approved):
LGTM

PR #2790, opened 2023-03-14
Fix missing magnitudes in case of more columns

hugo-vrijswijk:
In #2788, the table height was fixed only if a certain amount of rows are in the grid. With this change, the table is always the 'last' row, so it is always shown. I tried making the row height dynamic, but it didn't work out. I think this is a good compromise.


rpiaggio (approved):
LGTM

PR #2791, opened 2023-03-14
Improvements when inserting new target

rpiaggio:
- Show a toast notifying the new target was created.
- If the target summary table was in view, scroll it to the new target (This also has the effect of scrolling a manually selected target to the center of the view, which seems unnecessary, but seems like it would be a bit of work to avoid it).
- Introduce a `ToastCtx[F[_]]` typeclass, which makes it easier to invoke toasts by passing it implicitly.

### DIFF
--- a/common/src/main/scala/explore/common/TargetQueries.scala
+++ b/common/src/main/scala/explore/common/TargetQueries.scala
@@ -1,0 +1,31 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.common
+
+import queries.common.TargetQueriesGQL
+import lucuma.core.model.Target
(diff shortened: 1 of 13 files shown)


common/src/main/scala/explore/common/TargetQueries.scala
```diff
@@ -0,0 +1,31 @@
...
+    target:    Target.Sidereal
+  )(using
+    FetchClient[F, ?, ObservationDB],
+    ToastCtx[F]
+  ): F[Target.Id] =
+    TargetQueriesGQL
+      .CreateTargetMutation[F]
+      .execute(target.toCreateTargetInput(programId))
+      .map(_.createTarget.target.id)
+      .flatTap(id => ToastCtx[F].showToast(s"Created new target [$id]"))
```
cquiroz, line 31: Cool I could use this

cquiroz (approved):
Nice, I especially wanted the scroll to new feature

LGTM

PR #2793, opened 2023-03-15
Display FPU in short configuration

hugo-vrijswijk:
image

### DIFF
--- a/model-tests/shared/src/test/scala/explore/model/ObsSummarySuite.scala
+++ b/model-tests/shared/src/test/scala/explore/model/ObsSummarySuite.scala
@@ -5,7 +5,9 @@ package explore.model
 
 import cats.kernel.laws.discipline.EqTests
 import explore.model.arb.all.*
+import lucuma.schemas.model.BasicConfiguration
 import munit.DisciplineSuite
+import org.scalacheck.Prop.*
 
 class ObsSummarySuite extends DisciplineSuite {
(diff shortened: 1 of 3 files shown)


cquiroz:
Do we want to show it also as a column on the obs summary table?

hugo-vrijswijk:
> Do we want to show it also as a column on the obs summary table?

It's also included there, I probably should've added the screenshot:

image

cquiroz:
I see, I expected it to generate a new column but now I'm not sure if that's what we really want, maybe we need to confirm with @andrewwstephens, e.g. we may want to sort by fpu

toddburnside (commented):
LGTM

toddburnside (approved):

model-tests/shared/src/test/scala/explore/model/ObsSummarySuite.scala
```diff
@@ -14,4 +16,18 @@ class ObsSummarySuite extends DisciplineSuite {
            EqTests[ObsSummaryWithTitleConstraintsAndConf].eqv
   )
   checkAll("Eq[ObsSummaryWithTitleAndConf]", EqTests[ObsSummaryWithTitleAndConf].eqv)
+
+  property("ObsWithConf should include grating and fpu shortName") {
```
cquiroz, line 20: Great to have more testts

cquiroz (approved):
LGTM

PR #2798, opened 2023-03-17
Alternative loading indicator

cquiroz:
https://codepen.io/cquiroz/pen/rNZKYOP


toddburnside (approved):
Very Cool! I wonder if we should replace the progress spinner everywhere (or maybe most places) with something similar?

cquiroz:
That is an option make all indicators the same, I'll wait to get feedback on this new one and we can propagate

PR #2807, opened 2023-03-21
Implement RA, Dec, SED and multi-target rows for obs summary table

hugo-vrijswijk:
image
image

### DIFF
--- a/common-graphql/src/main/scala/queries/common/ObsQueriesGQL.scala
+++ b/common-graphql/src/main/scala/queries/common/ObsQueriesGQL.scala
@@ -54,13 +54,7 @@ object ObsQueriesGQL:
                 id
               }
             }
-            target {
-              id
-              sidereal {
-                ra $AngleSubquery
-                dec $AngleSubquery
-              }
-            }
+            target $TargetWithIdSubquery
           }
         }
       }
--- a/explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
+++ b/explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
@@ -16,10 +16,13 @@ import explore.common.UserPreferencesQueries.TableStore
 import explore.components.Tile
 import explore.components.ui.ExploreStyles
 import explore.model.AppContext
+import explore.model.Asterism
 import explore.model.Focused
 import explore.model.ObsIdSet
 import explore.model.ObsSummaryWithTitleConstraintsAndConf
 import explore.model.ObsWithConstraints
+import explore.model.TargetSummary
+import explore.model.display.given
 import explore.model.enums.AppTab
 import explore.model.enums.TableId
 import explore.model.reusability.given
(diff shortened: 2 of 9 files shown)


cquiroz:
Do you have a screenshot?

common-graphql/src/main/scala/queries/common/ObsQueriesGQL.scala
```diff
@@ -54,13 +54,7 @@ object ObsQueriesGQL:
                 id
               }
             }
-            target {
-              id
-              sidereal {
-                ra $AngleSubquery
-                dec $AngleSubquery
```
toddburnside, line 61: Were we even using the ra/dec from here before?

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -134,44 +159,134 @@ object ObsSummaryTable extends TableHooks:
...
+                  ^.onClick ==> (_.preventDefaultCB *> goToConstraint(id)),
+                  constraintsSummary
+              )
             )
           ),
           // TODO: FindingChartColumnId
-          column(ConfigurationColumnId, _.conf),
-          column(DurationColumnId, _.executionTime.toHoursMinutes)
+          column(ConfigurationColumnId, _.obs.conf),
+          column(DurationColumnId, _.obs.executionTime.toHoursMinutes),
```
cquiroz, line 215: Is the executionTime already in the model?

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -134,44 +159,134 @@ object ObsSummaryTable extends TableHooks:
...
 
           // TODO: PriorityColumnId
-          // TODO: RAColumnId
-          // TODO: DecColumnId
+          column(
+            RAColumnId,
+            r =>
+              r.asterism
+                .map(_.baseTracking.baseCoordinates.ra) // TODO: baseTrackingAt
```
cquiroz, line 222: So, in case of an asterism we'll show the base coordinates on each target row?
It seems we'd like to show the target's coordinates (maybe we need a synthetic additional base target)

common-graphql/src/main/scala/queries/common/ObsQueriesGQL.scala
hugo-vrijswijk: No it wasn't used

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
hugo-vrijswijk: It's returned from the backend query, yes

hugo-vrijswijk:
> Do you have a screenshot?

Added them

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
hugo-vrijswijk: The target coordinates are shown in the expanded target row (if an observation has more than 1 target). If an observation has exactly 1 target, the asterism and target coordinates are the same (I assume)

common-graphql/src/main/scala/queries/common/ObsQueriesGQL.scala
```diff
@@ -54,13 +54,7 @@ object ObsQueriesGQL:
...
-              }
-            }
+            target $TargetWithIdSubquery
```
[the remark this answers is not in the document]
hugo-vrijswijk: The data we need from a target is:

- Name
- Id
- RA
- Dec
- sourceProfile (to calculate SED)

I agree it would be cleaner to only pull this data and put it in the `TargetSummary` (that's actually how I started out). The reason for this solution is:
- I'm using the `Asterism` class to calculate the base coordinates. `Asterism` needs `TargetWithId`'s. I _could_ instead use a list of `Coordinates` from `TargetSummary` targets to calculate  the base coordinates
- Getting the SED needs some (partial optional) data I don't fully understand. Suggestions are welcome if we can make the query smaller

PR #2808, opened 2023-03-22
Show different column set for Simbad results & rearrange column order

rpiaggio:
Also: simplify sticky cols styles by using previous sibling selector (no more `with-`).

### DIFF
--- a/common/src/main/webapp/less/style.less
+++ b/common/src/main/webapp/less/style.less
@@ -527,6 +527,19 @@ html {
 
 .explore-target-summary-delete {
   left: 0;
+  overflow: hidden;
+
+  ~ .explore-target-summary-type {
+      left: 35px;
+
+    ~ .explore-target-summary-name {
+      left: 70px;
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


common/src/main/webapp/less/style.less
```diff
@@ -527,6 +527,19 @@ html {
 
 .explore-target-summary-delete {
   left: 0;
+  overflow: hidden;
```
cquiroz, line 530: Should we move these to scss?

cquiroz (changes requested):
Looks fine to me but I'd like that when we need to change the less file we rather convert it to scss

rpiaggio:
Good idea, I'll add another commit with that

cquiroz:
Do you have a screenshot?

rpiaggio:
> Do you have a screenshot?

image

image

Simbad doesn't show catalog name/cz/morphology/sed.

cquiroz:
Thanks

PR #2809, opened 2023-03-22
Target summary table: scroll to selected on resize

rpiaggio:
This solves the issue of bringing into view the selected target when the editor appears on screen.

It has the added advantage that it will keep it visible if the table tile is resized.


toddburnside (commented):
LGTM

PR #2826, opened 2023-03-29
Program cache for targets and observation summaries

rpiaggio:
I'm opening this a draft so it can start being reviewed.

The idea here is to query once all the targets and observation summaries of a program as soon as the user logs into a program. Then the cache is kept up to date via subscriptions that inform only of changed targets or observations.

This way we avoid refetching the whole list of targets and observations each time the user changes the view tab.

For the moment, and maybe indefinitely, we rebuild the asterism groups and target observations manually, since there's no subscription that provides updates for this information.

Up to the point of opening this PR, the cache is only feeding the target tab. I'd like to make it feed the observations and constraints tab before merging.


common/src/main/scala/explore/common/AsterismQueries.scala
```diff
@@ -109,20 +145,6 @@ object AsterismQueries:
     def findWithTargetIds(targetIds: SortedSet[Target.Id]): Option[AsterismGroup] =
       self.find { case (_, ag) => ag.targetIds === targetIds }.map(_._2)
 
-  private def obsResultToSummary(obsR: ObservationResult): ObsSummaryWithConstraintsAndConf =
```
rpiaggio, line 112: We are now decoding directly into `ObsSummaryWithConstraintsAndConf`.

common/src/main/scala/explore/common/AsterismQueries.scala
```diff
@@ -93,13 +94,45 @@ object AsterismQueries:
...
+          .mapValues(obsIds => SortedSet.from(obsIds))
+          .toMap
+
+      val updatedTargets: TargetWithObsList =
+        targetsWithObs.map((targetId, targetWithObs) =>
+          targetId ->
+            TargetWithObs.obsIds.replace(targetObservations.get(targetId).orEmpty)(targetWithObs)
+        )
+
+      val observationTargets: AsterismGroupList =
```
toddburnside, line 115: If we have this code to calculate the asterismGroups, should we even include them in the initial query?

Alternatively, I think the asterismGroups are only required on the target tab? If so, could they could just be created as needed there? Although I guess they only need to be recalculated when the observations change, not the targets. 🤔

explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -148,14 +151,20 @@ object ExploreLayout:
...
-                ExploreStyles.SideTabs,
-                SideTabs(routingInfo)
-              ),
-              <.div(
-                ExploreStyles.MainBody,
-                props.resolution.renderP(props.view)
-              )
+              routingInfo.optProgramId
+                .map(programId =>
+                  ProgramCache.Provider(ProgramCache(programId))(
```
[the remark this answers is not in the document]
toddburnside: I don't have an opinion on this at the moment, but I'll try and think about it.

toddburnside (commented):
It's looking good. I'll pull down the code sometime soon so I can look at it in more detail.

common/src/main/scala/explore/common/AsterismQueries.scala
rpiaggio: Yup, we could definitely compute this instead of querying the groups. Maybe we can even delete the maintenance methods (`insertObs`, `removeObsWithTargets`). I'll attempt that.

rpiaggio:
I'm marking this as ready, so that we can merge it.

It's mostly working, although some functionality is missing. Namely: auto expansion and focus. In some situations, undo may not work.

Actually, there's a `TODO.md` documenting all the things that need to be addressed. These will be the focus of future PRs.

I managed to unify the undo stacks in some cases, in some others it will require a bit more work. And we will need a way to navigate back to the change that's being undone.

explore/src/main/scala/explore/observationtree/ObsBadge.scala
```diff
@@ -45,9 +45,24 @@ case class ObsBadge(
...
 ) extends ReactFnProps(ObsBadge.component)
 
-object ObsBadge {
+object ObsBadge:
   private type Props = ObsBadge
 
+  enum Section derives Eq:
+    case None, Header, Detail
+
+  case class Layout(
```
toddburnside, line 54: No more matching on subclasses? What a shame. 😆

model/shared/src/main/scala/explore/optics/all.scala
```diff
@@ -176,11 +178,27 @@ object all extends ModelOptics {
...
+  given treeSeqMapAt[K, V]: At[TreeSeqMap[K, V], K, Option[V]] =
     At(i =>
       Lens((_: TreeSeqMap[K, V]).get(i))(optV => map => optV.fold(map - i)(v => map + (i -> v)))
     )
 
-  implicit def indexTreeSeqMap[K, V]: Index[TreeSeqMap[K, V], K, V] =
-    fromAt(atTreeSeqMap)
+  given treeSeqMapIndex[K, V]: Index[TreeSeqMap[K, V], K, V] = fromAt
+
+  // Not used after all, is it worth keeping?
```
toddburnside, line 188: Probably not, unless you foresee a future usage. On the other hand, I wouldn't find it easy to reproduce...

toddburnside (approved):
It's hard to fully understand the changes. It'll probably take some time using it to have a clearer picture. However, I don't see any glaring issues, so I think we can merge it and move forward.

Thanks for all the work on this. I think it will be a good step forward.

common/src/main/scala/explore/utils/CloneListView.scala
```diff
@@ -0,0 +1,17 @@
...
+
+package explore.utils
+
+import cats.Monad
+import crystal.ViewF
+import crystal.ViewListF
+import japgolly.scalajs.react.util.DefaultEffects.{Sync => DefaultS}
+
+// Turn a ViewList into a View, assuming all returned objects are clones.
+// and therefore just returning the head upon get.
```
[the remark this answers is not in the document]
rpiaggio: A `ViewList` can be seen as a `View`+`Traversal`, where the `get` returns a `List[A]` but the `mod`/`set` operates on `A` (not on `List[A]`), thus affecting all the elements the same way.

`CloneListView` assumes all the elements in the `List` are the same (clones of each other), and therefore exposes regular `View[A]` semantics by returning the `head` upon `get` (but it could return any element from the list, since they are all clones).

I found this very convenient when working with groupings. For example, when grouping observations by constraint set, we can create a `ViewList` into the `ConstraintSet` of the observations in the same group. The `ConstraintSet` of all these observations will be clones of each other (since that was actually the criterion for grouping them). We can pass a `CloneListView[ConstraintSet]` into a component that expects a `View[ConstraintSet]` and all the changes will be applied to all the observations in the group.

model/shared/src/main/scala/explore/optics/all.scala
rpiaggio: It is actually used, the comment is outdated! :D

rpiaggio: For the moment we are not using `TreeSeqMap`s directly, only through `KeyedIndexedList`, which has its own `FilterIndex` instance. This was copied and adapted from `monocle`'s implementation for `Map`s, it should be trivial to do it again if need be.

common/src/main/scala/explore/utils/CloneListView.scala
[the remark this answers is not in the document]
toddburnside: Cool!

PR #2841, opened 2023-04-11
Revert "Switch to fewest modules linking in dev."

armanbilge:
Reverts gemini-hlsw/explore#2229. Just to try! The changes in Scala.js 1.13.1 may have significantly improved the situation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -259,8 +259,8 @@ lazy val esModule = Seq(
   Compile / fastLinkJS / scalaJSLinkerConfig ~= { _.withSourceMap(false) },
   Compile / fullLinkJS / scalaJSLinkerConfig ~= { _.withSourceMap(false) },
   Compile / fastLinkJS / scalaJSLinkerConfig ~= (_.withModuleSplitStyle(
-    // Linking with smaller modules seems to take way longer.
-    ModuleSplitStyle.FewestModules
+    // If the browser is too slow for the SmallModulesFor switch to ModuleSplitStyle.FewestModules
+    ModuleSplitStyle.SmallModulesFor(List("explore"))
   )),
   Compile / fullLinkJS / scalaJSLinkerConfig ~= (_.withModuleSplitStyle(
     ModuleSplitStyle.FewestModules


hugo-vrijswijk:
I'll give it a try tomorrow, I think I played with it before, but only in fullLinkJS (once) where it didn't make much of a difference. Faster dev cycles would be nice though 😃

toddburnside (approved):
It's definitely worth testing!

rpiaggio:
I'll take it out for a spin

hugo-vrijswijk:
I feel like it is a little slower on initial compile, but for incremental it is faster (which makes sense if it doesn't have to overwrite one giant fastLink file). I think combined with Vite this works great 🥳

hugo-vrijswijk (dismissed):

armanbilge:
If that error is legit — i.e. Scala.js is emitting different / broken code under a different split style, that's a pretty serious bug that we should report.

hugo-vrijswijk:
It might be a case sensitivity issue? Looking at the code in VS Code gives this hint: 
image

Fixing it removes the error, though another similar one pops up after where the browser cannot find a function on the imported module

hugo-vrijswijk:
I was testing out `SmallestModulesFor` a bit more with #3049, and it actually works! But it didn't feel much faster. It turns out that the more modules you have, the slower Vite gets. We have around 750. Most of the time taken is the browser loading all the different modules. Which actually makes it feel a little slower (I have no real numbers, just based on poorly counting seconds in my head). 

This article also mentions that for a very high number of modules, Vite can become slower: https://betterprogramming.pub/is-vite-really-faster-than-webpack-b414f6cc751c

So, I think for now it is best to stay on `FewestModules` and take advantage of Scala.js bundling.

It might help to make 1 bundle for all dependencies, and 1 for explore itself, but I don't know if that is an option. That way, dependencies could stay cached while the app bundle changes during development (similar to how webpack outputs `main` and `vendor` bundles)

cquiroz:
THat's my experience too. Having FewestModules should make it possible to build less js files on `scala.js` but vite and browsers start to suffer with too many modules
There is an intermediate mode too that could make sense
https://github.com/scala-js/scala-js/issues/4327

PR #2842, opened 2023-04-11
Small fixes for the empty messages on tables

cquiroz:


toddburnside (approved):
LGTM

explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
```diff
@@ -257,6 +257,7 @@ object ConstraintsSummaryTable extends TableHooks:
             striped = true,
             compact = Compact.Very,
             tableMod = ExploreStyles.ExploreTable,
+            emptyMessage = <.div("No constraints present"),
```
rpiaggio, line 260: Since constraints are actually groupings of observations, would it be more clear here to display `No observations defined` ?

PR #2846, opened 2023-04-12
Update ITC and ODB schemas

cquiroz:


toddburnside (approved):
LGTM

model/shared/src/main/scala/explore/model/ExploreModelValidators.scala
```diff
@@ -52,6 +56,19 @@ object ExploreModelValidators:
...
     )
 
+  val snPosBigDecimal: Prism[PosBigDecimal, SignalToNoise] =
+    Prism[PosBigDecimal, SignalToNoise](bd =>
+      SignalToNoise.FromBigDecimalExact.getOption(bd.value)
+    )(
+      _.toPosBigDecimal
+    )
+
+  val signalToNoiseValidSplitEpi =
```
rpiaggio, line 66: Can we test the optics?

cquiroz: I will do that when I move this to core

PR #2856, opened 2023-04-13
Fixes to previous cache refactor

rpiaggio:
- `ObsSummaryTable` doesn't need a `View[ObservationList]` (just the `ObservationList`).
- Fix some copy & paste naming errors.
- Simplify logic for removing observations: it's not necessary to pass its targets anymore since there's no maintenance of a target's observations (it is recomputed).

PR #2857, opened 2023-04-13
Move cache to state instead of contest

rpiaggio:
Happens to also fix the issue of blank explore when navigating to the root.


rpiaggio:
Subsumes #2856 

Also, reinitializes cache update fibers upon program change.

explore/src/main/scala/explore/programs/ProgramsPopup.scala
```diff
@@ -53,6 +53,7 @@ object ProgramsPopup {
     .useContext(AppContext.ctx)
     .useStateView(IsOpen(true))
     .render { (props, ctx, isOpen) =>
+      println("hello")
```
hugo-vrijswijk, line 56: Accidentally added?

common/src/main/scala/explore/cache/CacheComponent.scala
```diff
@@ -6,37 +6,36 @@ package explore.cache
...
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.feature.Context
 import japgolly.scalajs.react.util.DefaultEffects.{Async => DefaultA}
+import japgolly.scalajs.react.util.DefaultEffects.{Sync => DefaultS}
 import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.ui.syntax.pot.*
 
-trait CacheComponent[Props: Reusability, A]:
-  private type F[T] = DefaultA[T]
+import scala.tools.nsc.doc.model.Def
```
toddburnside, line 20: Is this being used? Or was metals being helpful?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/programs/ProgramsPopup.scala
rpiaggio: Good catch :D

common/src/main/scala/explore/cache/CacheComponent.scala
rpiaggio: Metals, definitely

PR #2862, opened 2023-04-16
Auto-expanding of groups reinstated

rpiaggio:
I'm not sure this is working when cloning observations in the target tab since attempting it is raising an exception in the ODB.

Also added logic to allow developers to have a local configuration, in case we use a local instance of the ODB or other services.

### DIFF
--- a/explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
+++ b/explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
@@ -225,7 +225,7 @@ object ConstraintsSummaryTable extends TableHooks:
       )
       // Memo rows
       .useMemoBy((props, _, _) => props.constraintList)((_, _, _) =>
-        _.map(ConstraintGroup.fromTuple).toList
+        _.map(ConstraintGroup.fromTuple).toList.sortBy(_.constraintSet.summaryString)
       )
       .useReactTableWithStateStoreBy((props, ctx, cols, rows) =>
         import ctx.given
--- a/explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
+++ b/explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
@@ -98,26 +98,33 @@ object ConstraintGroupObsList:
       draggedIds  <- getDraggedIds(result.draggableId, focusedObsSet)
       if !destIds.intersects(draggedIds)
       newCs       <- constraintGroups.get(destIds)
-    } yield (newCs, draggedIds)
+      srcIds      <- constraintGroups.findContainingObsIds(draggedIds)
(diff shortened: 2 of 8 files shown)


explore/src/main/scala/explore/constraints/ConstraintsSummaryTable.scala
```diff
@@ -225,7 +225,7 @@ object ConstraintsSummaryTable extends TableHooks:
       )
       // Memo rows
       .useMemoBy((props, _, _) => props.constraintList)((_, _, _) =>
-        _.map(ConstraintGroup.fromTuple).toList
+        _.map(ConstraintGroup.fromTuple).toList.sortBy(_.constraintSet.summaryString)
```
rpiaggio, line 228: Keeps table sorting stable (as long as the user didn't select another sort criterion).

explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
```diff
@@ -98,26 +98,33 @@ object ConstraintGroupObsList:
...
-        )
-        .set(newCs)
+    oData.foldMap { case (newCs, destIds, draggedIds, srcIds) =>
+      expandedIds.mod(ids =>
+        val base = ids - draggedIds - destIds + (destIds ++ draggedIds)
+        (srcIds -- draggedIds).fold(base)(base + _)
+      ) >>
+        ConstraintsQueries
+          .UndoView(programId, draggedIds, csUndoCtx)(
+            // There should be a better way to do this.
```
cquiroz, line 122: In what sense? about using `identity`?

vite.config.mjs
```diff
@@ -92,8 +92,11 @@ export default defineConfig(({ command, mode }) => {
   const publicDirDev = path.resolve(common, 'src/main/publicdev');
   const lucumaCss = path.resolve(__dirname, 'explore/target/lucuma-css')
   fs.mkdir(publicDirDev, (err) => {
+    const localConf = path.resolve(publicDirProd, 'local.conf.json');
+    const devConf = path.resolve(publicDirProd, 'development.conf.json');
+
     fs.copyFileSync(
-      path.resolve(publicDirProd, 'development.conf.json'),
+      fs.existsSync(localConf) ? localConf : devConf,
```
cquiroz, line 99: Great idea

cquiroz (approved):
LGTM

explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
rpiaggio: Yes, maybe we can have a method that doesn't require `zoom`ing functions. I'll look into it in the future, but for the time being this works.

PR #2864, opened 2023-04-16
Show the intersection of patrols fields at each offset

cquiroz:
We want to show the intersection of each guided offset patrol fields.
This took a while to fix as this operation uncovered a bug on the JTS library we are using for geometry

Here's how it looks like:
image


toddburnside (approved):
I have no idea what a `Patrol Field` is, but LGTM.

cquiroz:
Who knows what a `Patrol Field` is 😄

PR #2867, opened 2023-04-17
Fix sorting by exposure time on the modes table

cquiroz:


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -131,8 +131,8 @@ private object SpectroscopyModesTable extends TableHooks:
     entry:  SpectroscopyModeRow,
     result: EitherNec[ItcQueryProblems, ItcResult]
   ):
-    lazy val exposureTime: Option[Int] =
-      result.toOption.collect { case ItcResult.Result(e, t) => e.toMillis.toInt * t }
+    lazy val totalItcTime: Option[Double] =
```
toddburnside, line 134: Is there a reason to use Double vs Long? I think `e.toMillis` is a Long, right?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: It is but the number is very large and it produces wrapping, I was a bit surprised too

PR #2868, opened 2023-04-17
Remove human-format

cquiroz:
I'm not sure where this npm dependency comes from but I'm not sure it is needed, is it?


rpiaggio:
It's used by the script that runs terser on the output files, to inform of the new file size.

We talked about removing the terser script altogether at some point, but I don't remember what was the conclusion.

cquiroz:
I see I'll remove this

PR #2871, opened 2023-04-18
Small touches to the patrol field display

cquiroz:
Change the color of the patrol field and only show the intersection


rpiaggio (approved):
LGTM

PR #2875, opened 2023-04-19
Share versions with meta-build & update clue

rpiaggio:
Old `Settings.scala` is split into `Versions.scala` and `Dependencies.scala`.

Plugin versions can now be declared in `Versions.scala`. The meta-build sees this through a symlink (which is added to the project).


project/Dependencies.scala
```diff
@@ -0,0 +1,328 @@
...
+
+object Dependencies {
+
+  private def deps(modules: PlatformDepsGroupArtifactID*)(version: String): Seq[ModuleID] =
+    modules.map(_ % version)
+
+  def In(configuration: Configuration)(dependencies: Seq[ModuleID]): Seq[ModuleID] =
+    dependencies.map(_ % configuration)
+
+    val Boopickle = Def.setting(
```
toddburnside, line 15: Is this supposed to be indented this far?

toddburnside (approved):
As long as lucuma-steward is happy with this, LGTM.

project/Dependencies.scala
rpiaggio: Nope, good catch.

PR #2877, opened 2023-04-21
Update ags logic and add a default 0 offset for ags calculations

cquiroz:
Fixes an issue where the selected guidestar could be out of the patrol field and another issue when using flip or unconstrained modes


rpiaggio (approved):
LGTM

PR #2886, opened 2023-04-25
Allow using SmallestModulesFor

cquiroz:
I'm not convinced SmallesModules would be usable but these changes fix the issue we saw with `UserVault`


toddburnside (approved):
LGTM

PR #2887, opened 2023-04-26
Use the new ITC api

cquiroz:
This PR updates explore to use the new simpler itc api. It will also make queries when using s/n@wavelength a lot faster.
It contains a  small UI improvement too


workers/src/main/scala/workers/itc/ITCRequests.scala
```diff
@@ -60,70 +62,41 @@ object ITCRequests:
...
+          .map(r => Map(params -> itcResults(r)))
+          .handleError { e =>
+            val msg = e match
+              case ResponseException(errors) =>
+                errors.map(_.message).mkString_("\n")
+              case e                         =>
+                e.getMessage
+            Map(params -> (ItcQueryProblems.GenericError(msg): ItcQueryProblems).leftNec[ItcResult])
+          }
+      }.sequence
```
hugo-vrijswijk, line 99: Minor thing, but can't `.map(...).sequence` be replaced by `.traverse`?

toddburnside: The answer is always traverse.

toddburnside (approved):
LGTM

PR #2899, opened 2023-05-01
Use itc client

cquiroz:
On this PR we start using the lucuma-itc client instead of doing graphql calls directly

Though we loose a bit of flexibility as we need to change the queries on a dependency, it gives us a smaller codebase which will save compilation time and also in my tests seem to reduce the final artifact size


common/src/main/public/development.conf.json
```diff
@@ -2,7 +2,7 @@
   "environment": "DEVELOPMENT",
   "odbURI": "wss://lucuma-postgres-odb-dev.herokuapp.com/ws",
   "preferencesDBURI": "wss://user-prefs-development.herokuapp.com/v1/graphql",
-  "itcURI": "https://itc-development.herokuapp.com/itc",
+  "itcURI": "http://itc-development.herokuapp.com/itc",
```
hugo-vrijswijk, line 5: Why is this http now?

hugo-vrijswijk (commented):
Nice!

workers/src/main/scala/workers/ItcServer.scala
```diff
@@ -25,10 +25,12 @@ import japgolly.scalajs.react.callback.*
 import log4cats.loglevel.LogLevelLogger
 import lucuma.ags.Ags
 import lucuma.ags.AgsAnalysis
+import lucuma.itc.client.ItcClient
```
rpiaggio, line 28: I think you don't need the `clue` imports anymore (lines 11-14). However, I suggest changing the `lucuma-itc` API so that it receives the `clue` client directly instead of building it from a `http4s` client. This will make it the API independent from `http4s` and clients are free to use it or not.

common-queries/src/main/scala/queries/schemas/itc/ITCConversions.scala
```diff
@@ -25,184 +25,192 @@ import lucuma.core.math.dimensional.Measure
...
-      GmosNITCInput(r.grating, r.fpu.toInput, filter = r.filter.orIgnore).assign
-
-  extension (r: GmosSouthSpectroscopyRow)
-    def toGmosSITCInput: Input[GmosSITCInput] =
-      GmosSITCInput(r.grating, r.fpu.toInput, filter = r.filter.orIgnore).assign
-
-  extension (r: RadialVelocity)
-    def toITCInput: RadialVelocityInput =
-      RadialVelocityInput(metersPerSecond = r.rv.value.assign)
+  // type SpectroscopyIntegrationTimeInput = ITC.Types.SpectroscopyIntegrationTimeInput
```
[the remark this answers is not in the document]
cquiroz: I'll remove it

common/src/main/public/development.conf.json
cquiroz: By mistake, it has to be `https`

workers/src/main/scala/workers/ItcServer.scala
cquiroz: I guess it would mean changing it here right?
https://github.com/gemini-hlsw/lucuma-itc/blob/main/modules/client/shared/src/main/scala/lucuma/itc/client/ItcClient.scala#L50

`lucuma-itc` is in use by the odb as well. Would it be fine to do this on the jvm side as well?

rpiaggio: Yes, it should be OK.

cquiroz: I'll do that in a separate PR as it needs an update on the odb side

rpiaggio (approved):
LGTM

PR #2900, opened 2023-05-02
Use graphQL enums without creating them in scala.

toddburnside:
The goal is to have the database be the source of truth for enumerations, rather than having them specifically created in scala code and keeping them in sync with the database. Currently, the ODB creates several graphQL enums from database tables, and provides `Meta` queries to get the additional information the scala Enumerated would normally have. For example, there is the `ObsAttachmentType` graphQL enum and the associated `ObsAttachmentMetaQuery`. 

This PR uses the `ObsAttachmentType` to try this out. It loads the `ObsAttachmentMetaQuery` data into a React context and provides Enumerated and Display values for both the `ObsAttachmentTypeMeta` and the `ObsAttachmentType`.

Usage will be:
- Use `useContext(EnumsContext.ctx)`
- In `render` you will have an `enums` parameter on which you can `import enums.given`
- Then you can use either `ObsAttachmentType` or `ObsAttachmentTypeMeta` in our `*EnumDropDown*`s.

I have tried this out and it works. I did not include an example in this PR because there was no good place to put it.

Once we agree on an implementation, I will add the `ProposalAttachmentType`.

Note: This implementation is based on the related implementation in the ODB by Shane.


cquiroz (approved):
Very nice, I remember discussing about this approach and it is great being closer to reality

common/src/main/scala/explore/Enums.scala
```diff
@@ -0,0 +1,58 @@
...
+    Display.by(t => enumMeta.obsAttachmentType(t).shortName,
+               t => enumMeta.obsAttachmentType(t).longName
+    )
+}
+
+object Enums {
+  case class Meta(
+    obsAttachmentType: Map[ObsAttachmentType, ObsAttachmentTypeMeta]
+  )
+  def load()(using StreamingClient[IO, ObservationDB]): IO[Enums] =
```
rpiaggio, line 51: It's enough to require a `FetchClient` here.

rpiaggio:
Thank you for tackling this!

It's looking good but I'm a bit confused as to how this will work. I see the ODB provides the enum instances in the schema, and then a meta query which will return the same collection of instances as defined in the schema (adding the display names).

I'm guessing that if a new instance of an enum is added, the ODB will adjust both the schema (and the query result) to reflect this, and our goal is to be able to dynamically read these new instance without the need to deploy again, is this right? Or is it OK for us to regenerate the schema class in `lucuma-schemas` and deploy explore again?

common/src/main/scala/explore/Enums.scala
```diff
@@ -0,0 +1,58 @@
...
+      enumMeta.obsAttachmentType.tail.map(_._2).toList*
+    )
+    .withTag(_.tag.show)
+
+  given Enumerated[ObsAttachmentType] = Enumerated
+    .from(
+      enumMeta.obsAttachmentType.head._1,
+      enumMeta.obsAttachmentType.tail.map(_._1).toList*
+    )
+    .withTag(_.show)
```
swalker2m, line 37: What would you do for a case that requires distinguishing the enum values for logic in the code?  For example, if you need to treat attachment type `Foo` differently than `Bar` in some context?

toddburnside: We can use the clue generated enums. See the `ObsAttachmentType` in `ObsAttachmentTypeMeta.scala`

rpiaggio: Or, we could do that with additional properties.

toddburnside:
Better implementation in #3223

PR #2903, opened 2023-05-03
Alternate approach without depending on schema enum

rpiaggio:
In this approach, there's only one class defining the enum instances, but 2 `Decoder`s are used:
- One for creating the instances from the meta query, only to be used in initialization.
- Another one for referencing instances, which is actually the one from the `Enumerated` instance (and will also provide an `Encoder` for inputs).

WDYT?


toddburnside:
> * Another one for referencing instances, which is actually the one from the `Enumerated` instance (and will also provide an `Encoder` for inputs).

I don't understand what you mean here: `Another one for referencing instances.`

rpiaggio:
> I don't understand what you mean here: `Another one for referencing instances.`

I mean that if at some point we want to decode `ObsAttachment(tpe: ObsAttachmentType, ...)`, we will use the `Decoder` from the `Enumerated` (which maps a tag to an existing instance), and not the meta decoder.

toddburnside:
> > * Another one for referencing instances, which is actually the one from the `Enumerated` instance (and will also provide an `Encoder` for inputs).
> 
> I don't understand what you mean here: `Another one for referencing instances.`

Ah. `Enumerated`s have their own decoder instance. 🤔

toddburnside:
Final implementation in #3223

PR #2905, opened 2023-05-06
PWA fixes

cquiroz:
This PR introduces a few small but important fixes around PWA. This started because locally I didn't get updates and it was at last linked to running out of drive space which made Chrome block the service worker installation.

It is an edge case which cannot be generally solved but can be alleviated a bit. The setup of the service worker is now more robust and includes a few other fixes:

* Update the interface of the PWA plugin
* More robust checking in case the user is offline or the server unreachable
* Purge caches when they are full
* In some cases the upgrade toast would get not displayed and instead would go to the background. This is now fixed
* The upgrade toast includes an experimental link to the commits since the last upgrade
* Update to broadcast_channel

### DIFF
--- a/explore/src/main/scala/explore/ExploreLayout.scala
+++ b/explore/src/main/scala/explore/ExploreLayout.scala
@@ -102,14 +102,27 @@ object ExploreLayout:
               x.event match {
                 // TODO: Handle logout events
                 case ExploreEvent.PWAUpdateId =>
-                  toastRef
-                    .prompt(
-                      "A new version of explore is available!",
-                      IO(
-                        ctx.broadcastChannel.postMessage(ExploreEvent.PWAReload)
-                      ).runAsyncAndForget
-                    )
-                    .to[IO]
+                  // Clear other toasts first
+                  toastRef.clear().to[IO] *>
+                    toastRef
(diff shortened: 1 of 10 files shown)


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -102,14 +102,27 @@ object ExploreLayout:
...
+                  // Clear other toasts first
+                  toastRef.clear().to[IO] *>
+                    toastRef
+                      .upgradePrompt(
+                        <.span(
+                          "A new version of ",
+                          <.a(
+                            ExploreStyles.UpgradeLink,
+                            "Explore",
+                            ^.href   := s"https://github.com/gemini-hlsw/explore/compare/${utils.gitHash.orEmpty}...HEAD",
```
rpiaggio, line 114: nice

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/ExploreLayout.scala
cquiroz: Let's experiment with this, it may not be the best for all users

PR #2916, opened 2023-05-09
Timing windows in ODB

rpiaggio:
- Switch to ODB timing windows, with the model now in lucuma-core.
- Add timing windows tile to observation view.
- Address inverting how we plot timing windows ([sc-2189]).
- Add a `.toastErrors` extension method to `F[_]`. This is a minimal approach to notify users in case there is an invocation error. I only added it here to a few places, feel free to propagate.


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -336,6 +339,18 @@ object ObsTabTiles:
             )
           )
 
+        val timingWindows: View[List[TimingWindow]] =
```
cquiroz, line 342: So timing windows go to their own tile? and removed from constraints?

cquiroz (approved):
Thanks a lot, LGTM

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
rpiaggio: It already had its own tile, which was only shown in the constraints view. Now it is shown there and in the observation tabs.

If we create a new timing windows tab, we can show it there instead of the constraints tab.

PR #2917, opened 2023-05-09
Do only one query for the itc chart data

cquiroz:
The ITC tile graph ad title do the same query, we can instead do it only once and share the results between the graph and the title.
It also reduces the need to pass some parameters along


explore/src/main/scala/explore/itc/ItcPanelProps.scala
```diff
@@ -64,26 +68,6 @@ trait ItcPanelProps(
   // TODO: Revisit when we have exposure mode in spectroscopy requirements
   val signalToNoise: Option[SignalToNoise] = spectroscopyRequirements.flatMap(_.signalToNoise)
 
-  // val signalToNoise: Option[PosBigDecimal] = observingMode match
```
hugo-vrijswijk, line 67: Are these still relevant? I usually prefer to just remove commented code

explore/src/main/scala/explore/itc/ItcPlotControl.scala
```diff
@@ -27,15 +30,30 @@ case class ItcPlotControl(
...
+  } {
+    case AllowedChartType.S2N    => ChartType.S2NChart
+    case AllowedChartType.Signal => ChartType.SignalChart
+  }
+
   private val component = ScalaFnComponent[Props] { props =>
-    val descText             = if (props.showDetails.get.exists(_.value)) "Hide details" else "Show details"
-    def label(ch: ChartType) = ch match
-      case ChartType.S2NChart         => "S/N"
-      case ChartType.SignalPixelChart => "Pixel"
```
toddburnside, line 37: No Pixel chart anymore?

toddburnside (approved):
LGTM

cquiroz:
I expected this to reduce code size 🤷

PR #2918, opened 2023-05-10
Request the itc data to use it when the mode has been selected

cquiroz:
Once a mode is selected we loose track of the actual results of the itc and cannot do the graph.
This PR fixes it requesting the itc info from the odb and it uses it to do the plot as needed

### DIFF
--- a/common-queries/src/clue/scala/queries/common/ObsQueriesGQL.scala
+++ b/common-queries/src/clue/scala/queries/common/ObsQueriesGQL.scala
@@ -145,14 +145,10 @@ object ObsQueriesGQL:
       }
     """
 
-    object Scalars:
-      type SignalToNoise = lucuma.core.math.SignalToNoise
-
     object Data:
       object Itc:
         object Result:
           type ExposureTime = lucuma.core.util.TimeSpan
-          type Exposures    = NonNegInt
 
       object Sequence:
         type ExecutionConfig = explore.model.ExecutionOffsets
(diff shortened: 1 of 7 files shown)


common-queries/src/clue/scala/queries/common/ObsQueriesGQL.scala
```diff
@@ -134,7 +145,15 @@ object ObsQueriesGQL:
       }
     """
 
+    object Scalars:
+      type SignalToNoise = lucuma.core.math.SignalToNoise
```
rpiaggio, line 149: This shouldn't be necessary, since it's declared in the schema (in `lucuma-schemas`).

common-queries/src/clue/scala/queries/common/ObsQueriesGQL.scala
```diff
@@ -134,7 +145,15 @@ object ObsQueriesGQL:
...
+
     object Data:
+      object Itc:
+        object Result:
+          type ExposureTime = lucuma.core.util.TimeSpan
+          type Exposures    = NonNegInt
```
rpiaggio, line 155: Same here, I think you can omit both of these types.

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -544,7 +543,7 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
         }
 
         val invalidateITC: Callback =
```
rpiaggio, line 545: Do we want to keep this around?

rpiaggio (approved):
LGTM, with a couple of comments

common-queries/src/clue/scala/queries/common/ObsQueriesGQL.scala
cquiroz: I added this because ExposureTime was failing to give me a usable type but I'll try again

PR #2919, opened 2023-05-10
Populate cache beyond query limit

rpiaggio:
Keep querying the ODB if there are more than 1000 observations or targets in the program.

### DIFF
--- a/common/src/main/scala/explore/cache/ProgramCache.scala
+++ b/common/src/main/scala/explore/cache/ProgramCache.scala
@@ -6,30 +6,30 @@ package explore.cache
 import cats.Order.given
 import cats.effect.IO
 import cats.effect.kernel.Resource
-import cats.syntax.all.given
+import cats.syntax.all.*
 import clue.StreamingClient
+import clue.data.syntax.*
 import explore.DefaultErrorPolicy
 import explore.common.AsterismQueries.*
+import explore.model.ObsSummary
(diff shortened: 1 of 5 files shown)


common/src/main/scala/explore/cache/ProgramCache.scala
```diff
@@ -43,13 +43,47 @@ case class ProgramCache(
 object ProgramCache extends CacheComponent[ProgramSummaries, ProgramCache]:
   given Reusability[ProgramCache] = Reusability.by(_.programId)
 
+  private def drain[A, Id, R](
+    fetch:      Option[Id] => IO[R],
+    getList:    R => List[A],
+    getHasMore: R => Boolean,
+    getId:      A => Id
+  ): IO[List[A]] = {
+    def go(id: Option[Id], accum: List[A]): IO[List[A]] =
```
cquiroz, line 52: tailrec?

cquiroz: Is this fetching results in batches? cool

cquiroz (approved):
LGTM

common/src/main/scala/explore/cache/ProgramCache.scala
rpiaggio: > tailrec?

I tried that initially but it doesn't work because of the flatmap, which IIRC is stack safe for IO.

PR #2922, opened 2023-05-11
Release ITC resource as soon as we have results

rpiaggio:
This fixes a bug where after a few calls to ITC, it became stuck and no further responses were processed.

### DIFF
--- a/common/src/main/scala/workers/WorkerClient.scala
+++ b/common/src/main/scala/workers/WorkerClient.scala
@@ -29,10 +29,8 @@ class WorkerClient[F[_]: Concurrent: UUIDGen: Logger, R: Pickler] private (
     worker.streamResource
       .use(
         _.map(decodeFromTransferableEither[FromServer]).rethrow
-          .collectFirst { case FromServer.ServerReady =>
-            initLatch.complete(())
-          }
-          .evalMap(identity)
+          .collectFirst { case FromServer.ServerReady => initLatch.complete(()) }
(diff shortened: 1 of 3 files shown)


common/src/main/scala/workers/WorkerClient.scala
```diff
@@ -70,7 +68,7 @@ class WorkerClient[F[_]: Concurrent: UUIDGen: Logger, R: Pickler] private (
         case FromServer.Complete(mid) if mid === id      =>
           none
         case FromServer.Error(mid, error) if mid === id  =>
-          throw error
```
rpiaggio, line 73: This is an unrelated fix.

explore/src/main/scala/explore/itc/ItcPanelProps.scala
```diff
@@ -125,10 +125,12 @@ case class ItcPanelProps(
         mode        <- instrumentRow
       yield beforeStart *>
         ItcClient[IO]
-          .request(
-            ItcMessage.GraphQuery(w, ex.time, exposures, constraints, t, mode)
+          .requestSingle(ItcMessage.GraphQuery(w, ex.time, exposures, constraints, t, mode))
```
rpiaggio, line 128: This is a syntax improvement.

toddburnside (approved):
Awesome. I think I was running into the bug and couldn't figure out why ITC wasn't responding sometimes. Thanks!

cquiroz (approved):
This is great, thanks a lot

PR #2937, opened 2023-05-15
Display and/or groups in obsList

hugo-vrijswijk:
Sorting is based on groups sorting, so might slightly change from the original order of 'just' observations. Group view works for both AND and OR groups (based on minimumRequired field of a group).

TODO (other PR's):

- 'Edit' groups view
- Create new groups
- Delete groups
- Move observations between groups
- Keep tree state when switching between observations (now collapses all)

Lucuma-schemas release is still processing, so build will fail until then.

### DIFF
--- a/explore/src/main/scala/explore/observationtree/ObsListTree.scala
+++ b/explore/src/main/scala/explore/observationtree/ObsListTree.scala
@@ -34,50 +34,15 @@ object ObsListTree:
   private type Props[A] = ObsListTree[A]
 
   private val ColDef = ColumnDef[Expandable[ObsNode]]
-  // private def treeToData(
-  //   tree:             KeyedIndexedTree[ObsNode.Id, ObsNode],
-  //   collapsedItemIds: Set[AtlasTree.ItemId]
-  // ): AtlasTree.Data[ObsNode] =
-  //   def valueChildrenToData(
-  //     keyValue: Option[(ObsNode.Id, ObsNode)],
-  //     children: List[Node[(ObsNode.Id, ObsNode)]],
-  //     accum:    List[(String, AtlasTree.Item[ObsNode])] = List.empty
-  //   ): List[(String, AtlasTree.Item[ObsNode])] = {
-  //     val getKey: ((ObsNode.Id, ObsNode)) => String = (id, _) => id.fold(_.toString, _.toString)
-  //     val key                                       = keyValue.map(getKey).getOrElse("")
-  //     children.foldLeft(
-  //       accum :+
-  //         (key -> AtlasTree.Item[ObsNode](
-  //           key,
-  //           children.map(node => getKey(node.value)),
-  //           hasChildren = children.nonEmpty,
-  //           isExpanded = !collapsedItemIds.contains(key),
-  //           isChildrenLoading = false,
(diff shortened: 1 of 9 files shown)


explore/src/main/scala/explore/observationtree/obsTree.scala
```diff
@@ -0,0 +1,87 @@
...
+import react.primereact.Tree.Node
+
+import java.util.UUID
+import scala.scalajs.js.JSConverters._
+
+sealed trait ObsNode
+
+object ObsNode {
+
+  // given Reusability[ObsNode] = Reusability.byEq
```
rpiaggio, line 29: commented code

explore/src/main/scala/explore/observationtree/obsTree.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  case class Obs(value: ObsSummary) extends ObsNode
```
rpiaggio, line 31: Can we use an `enum` for this?

explore/src/main/scala/explore/observationtree/ObsListTree.scala
```diff
@@ -0,0 +1,48 @@
...
+import lucuma.ui.reusability.given
+import lucuma.react.table.*
+import lucuma.react.table.ColumnDef
+import org.scalajs.dom.HTMLDivElement
+import explore.observationtree.ObsNode.given
+import explore.components.ui.ExploreStyles
+import react.primereact.Tree
+import lucuma.typed.primereact.treeTreeMod.TreeNodeTemplateOptions
+
+final case class ObsListTree[A](
```
rpiaggio, line 28: `final` is not necessary anymore for `case class`es in Scala 3.

explore/src/main/scala/explore/tabs/ObsTabContents.scala
```diff
@@ -389,6 +395,16 @@ object ObsTabContents extends TwoPanels:
                         callbacks
         )
       }
-      .render((props, ctx, twoPanelState, resize, layouts, defaultLayout) =>
-        renderFn(props, twoPanelState, defaultLayout, layouts, resize, ctx)
+      .useStreamResourceViewBy((props, _, _, _, _, _) => props.programId) {
```
rpiaggio, line 398: Can we put this in `ProgramSummaries` and have it cached?

rpiaggio (approved):
LGTM.

We can move the groups to the cache in a future PR.

explore/src/main/scala/explore/observationtree/ObsListTree.scala
hugo-vrijswijk: I blame Metals for adding it :)

explore/src/main/scala/explore/observationtree/obsTree.scala
hugo-vrijswijk: Good catch

explore/src/main/scala/explore/observationtree/obsTree.scala
hugo-vrijswijk: Yes, refactored it

PR #2939, opened 2023-05-15
Actually show timing windows, log other twilights

rpiaggio:
All the logic was in place but for some reason in the merged code we were not passing the timing windows to the plot. Looks like a rebase error.

Unrelated, we now show in the console the nautical twilights instead of the official nows, by request from Andy.


cquiroz (approved):
LGTM

PR #2941, opened 2023-05-16
Support accepting multiple target's graphs

cquiroz:
In the case of multiple targets we were only plotting one, An improvement for larger asterisms could be to request each target  in parallel

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -174,7 +174,7 @@ object ObsTabTiles:
         itcQueryProps(obsView, selectedConfig.get, props.allTargets)
       )
       // Chart results
-      .useState(Pot.pending[Map[ItcTarget, ItcChartResult]])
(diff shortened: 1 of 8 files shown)


toddburnside (approved):
LGTM

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -185,21 +185,23 @@ object ObsTabTiles:
...
+                    k -> (Pot.Ready(e): Pot[ItcChartResult])
+                }.toMap
+                charts.setStateAsync(r) *> loading.setState(LoadingState.Done).to[IO]
+              },
               (charts.setState(
-                Pot.error(new RuntimeException("Not enough information to calculate the ITC graph"))
+                itcProps.targets
+                  .map(t =>
+                    t -> Pot.error(
+                      new RuntimeException("Not enough information to calculate the ITC graph")
```
rpiaggio, line 201: Why `RuntimeException` instead of a regular `Exception`?

rpiaggio (approved):
LGTM

PR #2946, opened 2023-05-16
Handle a particular case when the geometry is calculated and there is…

cquiroz:
… no mode

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/GmosGeometry.scala
+++ b/explore/src/main/scala/explore/targeteditor/GmosGeometry.scala
@@ -64,12 +64,14 @@ object GmosGeometry:
     configuration: Option[BasicConfiguration],
     port:          PortDisposition,
     extraCss:      Css = Css.Empty
-  ): (Css, ShapeExpression) =
-    (Css("patrol-field-intersection") |+| extraCss) ->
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/targeteditor/GmosGeometry.scala
```diff
@@ -64,12 +64,14 @@ object GmosGeometry:
...
-    (Css("patrol-field-intersection") |+| extraCss) ->
-      offsets
-        .map(patrolField(posAngle, _, configuration, port))
-        .collect { case Some(s) => s }
-        .reduce(_ ∩ _)
+  ): Option[(Css, ShapeExpression)] =
+    configuration.map(_ =>
+      (Css("patrol-field-intersection") |+| extraCss) ->
+        offsets
+          .map(patrolField(posAngle, _, configuration, port))
```
hugo-vrijswijk, line 71: `patrolField` is not used anywhere else, I think. So it could be updated to accept a `BasicConfiguration` instead of `Option[BasicConfiguration]`

cquiroz: Right, we could do that

PR #2948, opened 2023-05-17
Add groups to programCache and fix observations updates in obsList

hugo-vrijswijk:
- Move program groups query to programCache
- Fix new observations not appearing due to groupElements not being updated

Moving items between groups is not properly updated yet, this needs odb support


common/src/main/scala/explore/cache/ProgramCache.scala
```diff
@@ -20,13 +23,19 @@ import lucuma.core.model.Observation
 import lucuma.core.model.Program
 import lucuma.core.model.Target
 import lucuma.schemas.ObservationDB
+import lucuma.schemas.ObservationDB.Enums.EditType
+import lucuma.schemas.ObservationDB.Enums.EditType.Created
+import lucuma.schemas.ObservationDB.Enums.EditType.Updated
```
rpiaggio, line 28: Code seems to use `EditType.Created`and `EditType.Updated`. Maybe these 2 imports can be omitted?

rpiaggio (approved):
LGTM, with a couple of very minor comments.

common/src/main/scala/explore/cache/ProgramCache.scala
hugo-vrijswijk: It was used in a pattern match, but that's now matching on EditType.Created/Updated too

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -52,14 +54,14 @@ case class ObsList(
   focusedObs:      Option[Observation.Id],
   focusedTarget:   Option[Target.Id],
   setSummaryPanel: Callback,
-  groups:          Pot[View[List[AllGroupElements]]]
+  groups:          GroupList
```
cquiroz, line 57: Much nicer :D

cquiroz (approved):
LGTM

PR #2951, opened 2023-05-17
Add Observation Attachments Table

toddburnside:
This creates an `Observation Attachments` tile on the overview tab, with a table of observations and allows uploading, updating and deleting of attachments. It also adds placeholder tabs for the Proposal Attachments and the Warnings And Errors. There are still things that need to be done that will be handled in other PRs.

- Make the editing of descriptions and the checked column actually update the attachment via graphql. (See comment in ObsAttachmentsTable.scala.) This could be handled in this PR if a solution is found.
- Save/restore the tile layout for the overview tab to/from the user preferences.
- Add the ability to download an attachment.
- The table should probably be a virtualized table.

### DIFF
--- a/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
+++ b/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
@@ -1,0 +1,407 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.attachments
+
+import cats.Order.*
+import cats.effect.IO
+import cats.syntax.all.*
+import crystal.react.*
+import crystal.react.hooks.*
+import crystal.react.implicits.*
+import crystal.react.reuse.*
+import eu.timepit.refined.types.numeric.NonNegLong
+import eu.timepit.refined.types.string.NonEmptyString
+import explore.EditableLabel
+import explore.Icons
+import explore.common.ProgramQueries
+import explore.components.ui.ExploreStyles
+import explore.model.AppContext
+import explore.model.Constants
+import explore.model.ObsAttachment
+import explore.model.reusability.given
+import explore.syntax.ui.*
+import explore.utils.OdbRestClient
+import explore.utils.*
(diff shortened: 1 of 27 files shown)


explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
```diff
@@ -0,0 +1,407 @@
...
+                  .format(cell.value.toInstant)
+              ),
+            ColDef(
+              DescriptionColumnId,
+              _.withOnMod(oa =>
+                ProgramQueries
+                  .updateObsAttachmentDescription[IO](props.pid, oa.id, oa.description)
+                  .runAsync
+              )
+                .zoom(ObsAttachment.description),
```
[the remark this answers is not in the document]
rpiaggio: I can take a look at this. AFAIC feel free to merge anyway in the meantime.

explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
```diff
@@ -0,0 +1,407 @@
...
+              )
+            )
+          )
+
+        <.div(
+          PrimeTable(
+            table,
+            striped = true,
+            compact = Compact.Very,
+            footerMod = footer,
```
rpiaggio, line 390: I just noticed we've been approaching other table footers in this same way. However, the `*mod` parameters of `PrimeTable` are designed so that you define the footers for columns and let the library do the rendering. You can add the  `^.colSpan := 7` via `footerCellMod`.

This obviously works, but just to bear in mind for the future: 
- This doesn't add the standard styling classes the facade inserts.
- I think a footer `tr` is supposed to have `th`s, not `td`s.

rpiaggio (approved):
LGTM

cquiroz (approved):
I read quite a bit of it but it is a lot to take in
LGTM

explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
toddburnside: Yeah. I just basically copied this from the brightnesses table. 😄 I'll look at this for the proposal attachments table and when I switch to virtualized table.

PR #2953, opened 2023-05-17
Move observation configuration to program cache

rpiaggio:
You can consider this PR as the final part of the program cache refactor. There was an intermediate state where we were doubly querying parts of the observation. Now everything's in the cache, including the configuration, and we query only once.

Switching observations in the observations tab should be faster than before.

Also, as a "side effect", the configuration undo stacks become unified with the general observation stacks.


toddburnside:
> Also, as a "side effect", the configuration undo stacks become unified with the general observation stacks.

Inching towards universal undo. 🤣

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -230,39 +249,36 @@ object ObsTabTiles:
...
-          val (obsViewPot, _, observingMode, scienceReqs, chartExposureTime) =
-            obsProperties(obsView)
+          // PREVIOUS CODE WAS: (BUT WE ALWAYS HAD "none" IN itcExposureTime
+          //   val chartExposureTime                        = obsView.toOption
+          //     .flatMap(
+          //       _.get.itcExposureTime
+          //         .map(r => ItcChartExposureTime(OverridenExposureTime.FromItc, r.time, r.count))
+          //     )
+          //     .orElse(obsView.toOption.flatMap(_.get.itc.map(_.toItcChartExposureTime)))
+          //   (obsViewPot, scienceData, observingMode, scienceReqs, chartExposureTime)
```
toddburnside, line 259: Do you want to keep this commented code?

toddburnside (approved):
LGTM - It's nice to have this all wrapped up! :)

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
rpiaggio: I think yes, for the moment. This is a pending fix. WDYT @cquiroz ?

PR #2957, opened 2023-05-19
Keep group state when selecting observations (and between moving tabs)

hugo-vrijswijk:
https://github.com/gemini-hlsw/explore/assets/10114577/41ab7a3c-310f-477e-b6dc-66c80f3a8ce1

### DIFF
--- a/common/src/main/scala/explore/model/RootModel.scala
+++ b/common/src/main/scala/explore/model/RootModel.scala
@@ -19,7 +19,6 @@ import lucuma.core.model.User
 import monocle.Focus
 import monocle.Lens
 import monocle.Optional
-import react.primereact.Tree
 
 import scala.collection.immutable.HashSet
 case class RootModel(
--- a/common/src/main/scala/explore/model/RootModel.scala
+++ b/common/src/main/scala/explore/model/RootModel.scala
@@ -19,26 +19,29 @@ import lucuma.core.model.User
 import monocle.Focus
 import monocle.Lens
 import monocle.Optional
+import react.primereact.Tree
 
 import scala.collection.immutable.HashSet
 case class RootModel(
-  vault:                Option[UserVault],
-  localPreferences:     ExploreLocalPreferences,
-  expandedIds:          ExpandedIds = ExpandedIds(),
-  searchingTarget:      Set[Target.Id] = HashSet.empty,
-  userSelectionMessage: Option[NonEmptyString] = none,
-  programSummaries:     Option[ProgramSummaries] = none,
-  undoStacks:           ModelUndoStacks[IO] = ModelUndoStacks[IO]()
+  vault:                 Option[UserVault],
+  localPreferences:      ExploreLocalPreferences,
+  expandedIds:           ExpandedIds = ExpandedIds(),
+  searchingTarget:       Set[Target.Id] = HashSet.empty,
+  userSelectionMessage:  Option[NonEmptyString] = none,
+  programSummaries:      Option[ProgramSummaries] = none,
+  undoStacks:            ModelUndoStacks[IO] = ModelUndoStacks[IO](),
+  expandedObsListGroups: Map[Tree.Id, Boolean] = Map.empty
 ) derives Eq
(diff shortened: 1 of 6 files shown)


common/src/main/scala/explore/model/RootModel.scala
```diff
@@ -19,26 +19,29 @@ import lucuma.core.model.User
...
-  programSummaries:     Option[ProgramSummaries] = none,
-  undoStacks:           ModelUndoStacks[IO] = ModelUndoStacks[IO]()
+  vault:                 Option[UserVault],
+  localPreferences:      ExploreLocalPreferences,
+  expandedIds:           ExpandedIds = ExpandedIds(),
+  searchingTarget:       Set[Target.Id] = HashSet.empty,
+  userSelectionMessage:  Option[NonEmptyString] = none,
+  programSummaries:      Option[ProgramSummaries] = none,
+  undoStacks:            ModelUndoStacks[IO] = ModelUndoStacks[IO](),
+  expandedObsListGroups: Map[Tree.Id, Boolean] = Map.empty
```
toddburnside, line 33: Do you need a `Map` here? Or would a `Set` or `SortedSet` work to just keep track of the expanded ones?

Also, I think this should probably go in `ExpandedIds` with the other expanded states.

hugo-vrijswijk: The Primereact type is a Map (with the values always `true`), but I agree a `Set`is probably cleaner and we can keep the map an implementation detail for the primereact component

hugo-vrijswijk: Should I change it here or in lucuma-react? I don't know how close to the primereact API we want to stay in lucuma-react

toddburnside: Ie. The closer we stay to the API, the easier it is to use the primereact documentation for using our components. But, if we can clean it up a bit...

cquiroz:
Would there be any value on keeping this state on the user prefs db?

cquiroz:
> We aren't currently doing that for any of the other `Expanded States`.

Ok, that's fine, we are consistent. we can consider it if there is user demand

common/src/main/scala/explore/model/RootModel.scala
```diff
@@ -19,6 +19,7 @@ import lucuma.core.model.User
 import monocle.Focus
 import monocle.Lens
 import monocle.Optional
+import react.primereact.Tree
```
toddburnside, line 22: Not needed?

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -119,12 +123,53 @@ object ObsList:
...
+      )
       .render { (props, ctx, _, adding, treeNodes) =>
 
         import ctx.given
 
         val observations = props.observations.toList
 
+        // TODO: remove after https://github.com/gemini-hlsw/lucuma-react/pull/465 is merged and released
+        val expandedGroups =
+          props.expandedGroups.zoom(_.map(k => Tree.Id(k.toString) -> true).toMap)(f =>
```
toddburnside, line 164: We'll still need to convert from Set[Group.Id] to Set[Tree.Id] somewhere, right? Maybe we could memoize that or something?

hugo-vrijswijk: We could also keep a `Set[Tree.Id]`, but that feels slightly less clean (since it is always for a group)

common/src/main/scala/explore/model/RootModel.scala
toddburnside: It's going to be chaos when that arrives. I hope I'm on vacation. 🤔

explore/src/main/scala/explore/observationtree/ObsList.scala
toddburnside: Agreed

toddburnside (approved):

PR #2974, opened 2023-05-24
Ability to download attachments

toddburnside:
Provides a presigned link to the attachment on S3, styled as a button. 

Unfortunately, the user is not able to view images via this link, only download the files. If we want view capabilities (which I'm sure we will), we'll need to provide a different method. But, this is a start.

The URL is good for 2 hours. We could go through and replace links on a timer if we feel it is necessary,

### DIFF
--- a/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
+++ b/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
@@ -247,9 +247,7 @@ object ObsAttachmentsTable extends TableHooks:
               }
               .to[IO]
           val getUrls      =
-            newOas
-              .traverse(oa => getAttachmentUrl(props, oa, urlMap))
-              .void
+            newOas.traverse_(oa => getAttachmentUrl(props, oa, urlMap))
 
           updateUrlMap *> getUrls
       )
--- a/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
+++ b/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
@@ -6,6 +6,7 @@ package explore.attachments
 import cats.Order.*
 import cats.effect.IO
 import cats.syntax.all.*
+import crystal.Pot
 import crystal.react.*
 import crystal.react.hooks.*
 import crystal.react.implicits.*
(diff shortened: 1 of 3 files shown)


cquiroz (approved):
LGTM

explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
```diff
@@ -209,9 +233,29 @@ object ObsAttachmentsTable extends TableHooks:
...
+          val updateUrlMap =
+            urlMap
+              .mod { umap =>
+                val filteredMap = umap.filter((k, v) => allCurrentKeys.contains(k))
+                newOas.foldRight(filteredMap)((oa, m) => m.updated(oa.toMapKey, Pot.pending))
+              }
+              .to[IO]
+          val getUrls      =
+            newOas
+              .traverse(oa => getAttachmentUrl(props, oa, urlMap))
```
hugo-vrijswijk, line 251: Can you use `.traverse_` here?

toddburnside: Yep. Good catch.

hugo-vrijswijk: The answer is always `traverse_`

explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
```diff
@@ -209,9 +233,27 @@ object ObsAttachmentsTable extends TableHooks:
...
+              .to[IO]
+          val getUrls      =
+            newOas.traverse_(oa => getAttachmentUrl(props, oa, urlMap))
+
+          updateUrlMap *> getUrls
+      )
       // Columns
-      .useMemoBy((_, _, _) => ())((props, ctx, action) =>
-        _ =>
+      .useMemoBy((_, _, _, urlMap) => urlMap.reuseByValue)((props, ctx, action, _) =>
```
rpiaggio, line 255: You can use `urlMap.get` in the dependencies, since only the value seems to be used.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
toddburnside: Good point

toddburnside: Good point

PR #2980, opened 2023-05-26
Allow filtering of cacheable items, e.g. to not cache errors

cquiroz:
This adds a feature to the cache to make it optional no a per request basis
On ITC we are storing error cases but we may not want to do that in all cases. In particular if the server is unreachable we want to retry.

### DIFF
--- a/workers/src/main/scala/workers/itc/ITCGraphRequests.scala
+++ b/workers/src/main/scala/workers/itc/ITCGraphRequests.scala
@@ -121,7 +121,20 @@ object ITCGraphRequests:
         .map(_.toList.flattenOption.toMap)
 
     // We cache unexpanded results, exactly as received from server.
-    val cacheableRequest = Cacheable(CacheName("itcGraphQuery"), CacheVersion(5), doRequest)
+    val cacheableRequest =
+      Cacheable(
+        CacheName("itcGraphQuery"),
+        CacheVersion(5),
+        doRequest,
+        (r, g) =>
+          r.target.forall(t =>
+            g.get(t).forall {
+              case Right(_)                               => true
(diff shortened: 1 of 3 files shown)


workers/src/main/scala/workers/itc/ITCGraphRequests.scala
```diff
@@ -121,7 +121,20 @@ object ITCGraphRequests:
...
+      Cacheable(
+        CacheName("itcGraphQuery"),
+        CacheVersion(5),
+        doRequest,
+        (r, g) =>
+          r.target.forall(t =>
+            g.get(t).forall {
+              case Right(_)                               => true
+              case Left(ItcQueryProblems.GenericError(_)) => false
+              case Left(_)                                => false
```
toddburnside, line 134: The last case would also handle the middle case. Is this what you meant?

workers/src/main/scala/workers/itc/ITCRequests.scala
```diff
@@ -110,7 +110,18 @@ object ITCRequests:
...
+    val cacheableRequest =
+      Cacheable(
+        CacheName("itcQuery"),
+        cacheVersion,
+        doRequest,
+        (r, g) =>
+          g.exists(_.get(r).forall {
+            case Right(_)                               => true
+            case Left(ItcQueryProblems.GenericError(_)) => false
+            case Left(_)                                => false
```
toddburnside, line 122: Same as above.

toddburnside (approved):
LGTM with comments.

workers/src/main/scala/workers/itc/ITCGraphRequests.scala
cquiroz: oh right, those last ones should be true, thanks

PR #2988, opened 2023-05-29
update highcharts

cquiroz:


explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
```diff
@@ -73,116 +73,111 @@ object ElevationPlotSemester:
...
+                      .XAxisPlotLinesOptions()
+                      .setId("date")
+                      .setValue(zonedDateTime.toInstant.toEpochMilli.toDouble)
+                      .setZIndex(1000)
+                      .setClassName("plot-plot-line-date")
+                  )
                 }
-              ).orEmpty)
-      )
-      .useEffectWithDepsBy((_, _, resize, _, _) => resize)((_, _, _, chartOpt, _) =>
```
toddburnside, line 179: We don't need this resize here anymore? Does the chart handle resizing better?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
cquiroz: I should have commented this. This code was producing an exception when you try to resize. [
Now it won't but the graph won't be updated. I'll make a separate PR to fix it

PR #2990, opened 2023-05-29
Support unsetting SED

cquiroz:


common/src/main/scala/explore/model/enums/SEDType.scala
```diff
@@ -47,29 +47,28 @@ sealed abstract class SEDTypeEnum[T](
   import SpectralDefinition.*
 
   private def toBandNormalized[T](
-    sed: UnnormalizedSED
+    sed: Option[UnnormalizedSED]
```
rpiaggio, line 50: I think this change is not necessary after all. It's always called with `.some` and it's more compact to have it here than in all the callers.

The `None` case is handled somewhere else.

common/src/main/scala/explore/model/enums/SEDType.scala
```diff
@@ -80,32 +79,33 @@ sealed abstract class SEDTypeEnum[T](
...
-      case BandNormalized(Some(Quasar(_)), _)          => QuasarType
-      case BandNormalized(Some(HIIRegion(_)), _)       => HIIRegionType
-      case BandNormalized(Some(PlanetaryNebula(_)), _) => PlanetaryNebulaType
-      case EmissionLines(_, _)                         => EmissionLineType
-      case BandNormalized(Some(PowerLaw(_)), _)        => PowerLawType
-      case BandNormalized(Some(BlackBody(_)), _)       => BlackBodyType
-      case BandNormalized(Some(UserDefined(_)), _)     => UserDefinedType
-      case BandNormalized(_, _)                        => UserDefinedType
-    }
+  def fromSpectralDefinition(spectralDefinition: SpectralDefinition[T]): Option[SEDType[T]] =
```
rpiaggio, line 95: Same here. Unless I'm missing something, this can be left as before.

explore/src/main/scala/explore/targets/TargetSearchResult.scala
```diff
@@ -30,7 +30,7 @@ object TargetSearchResult {
         none,
         Target.Sidereal.sourceProfile
           .andThen(SourceProfile.unnormalizedSED)
-          .modify(_.orElse(Constants.DefaultSED.some))(r.target) // Ensure SED is always present.
+          .modify(identity)(r.target)
```
rpiaggio, line 33: Can all this be replaced with just `r.target`?

cquiroz: Right

common/src/main/scala/explore/model/enums/SEDType.scala
rpiaggio: Actually this one is needed.

rpiaggio (approved):
LGTM

PR #2991, opened 2023-05-30
Don't query ITC if SED is not defined

cquiroz:


toddburnside (approved):
LGTM, other than the odd build error. I can't see anything in this PR that should cause that.

PR #2995, opened 2023-06-01
Small graph ui improvements

cquiroz:
Some small updates
* a bit thicker line
* border on the tooltips
* less faded inactive series

These are a side effect of the last highcharts update


rpiaggio (approved):
LGTM

PR #2998, opened 2023-06-01
Setup a finder chart tile

cquiroz:
https://github.com/gemini-hlsw/explore/assets/3615303/a8eb4414-86d7-4dc2-a203-0909e31242f4


common/src/main/scala/explore/Icons.scala
```diff
@@ -286,6 +286,38 @@ object Icons {
...
+  @JSImport("@fortawesome/pro-light-svg-icons", "faMagnifyingGlassPlus")
+  val faMagnifyingGlassPlus: FAIcon = js.native
+
+  @js.native
+  @JSImport("@fortawesome/pro-light-svg-icons", "faMagnifyingGlassMinus")
+  val faMagnifyingGlassMinus: FAIcon = js.native
+
+  @js.native
+  @JSImport("@fortawesome/pro-light-svg-icons", "faArrowsRepeat")
+  val faArrowsRetweet: FAIcon = js.native
```
toddburnside, line 315: `Retweet`? Although maybe this is better since Raul has the same icon as `faArrowsRepeat` as a `pro-regular-svg-icons`.

cquiroz: Ups. I was using the retweet icon but then I switched. maybe it should be fArrowsRepeatLight

toddburnside (approved):
Very cool! LGTM

PR #3011, opened 2023-06-12
Add assigned obs ids to obs attachments table

toddburnside:
This also includes a number of other changes, including, but not limited to:

- Add the assigned `ObsAttachmentId`s to the `ObsSummary`
- Remove the Proposal Attachments tile from the overview tab. I discussed it with Andy and it isn't necessary.
- Move the attachment upload button and dropdown to the tile title from the table footer. It is now more like the target summary table.
- Fix issue where there would be a stale `OdbRestClient` when the auth token expired.

### DIFF
--- a/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
+++ b/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
@@ -222,16 +222,19 @@ object ObsAttachmentsTable extends TableHooks:
       }
       .flatMap(p => urlMap.mod(_.updated(oa.toMapKey, p)).to[IO])
 
-  def deletePrompt(props: Props, aid: ObsAtt.Id)(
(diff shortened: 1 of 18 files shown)


explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
```diff
@@ -104,16 +114,44 @@ object ObsAttachmentsTable extends TableHooks:
...
+    val tag:        String,
+    val name:       String,
+    val gql:        ObsAttachmentType,
+    val extensions: List[String]
+  ) derives Enumerated {
+    case Finder
+        extends AttachmentType("FINDER",
+                               "Finder Chart",
+                               ObsAttachmentType.Finder,
+                               List("jpg", "png")
```
cquiroz, line 132: gif maybe?

PR #3017, opened 2023-06-13
Fix a subtle bug that makes the flip angles not to converge

cquiroz:
When doing allow flip mode we could endup in a loop where the selected angle jumps between the searched angles forever.
This fixes it by making sure we write to the db only once and by sorting the angles in flip mode

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinCell.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinCell.scala
@@ -91,12 +91,10 @@ case class AladinCell(
       paConstraint  <- conf.posAngleConstraint
       angles        <-
         paConstraint.anglesToTestAt(configuration.siteFor, asterism.baseTracking, vizTime)
-    } yield {
-      // We sort the angles or we could end up in a loop where the angles are tested back and forth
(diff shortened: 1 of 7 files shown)


explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -91,12 +91,18 @@ case class AladinCell(
       paConstraint  <- conf.posAngleConstraint
       angles        <-
         paConstraint.anglesToTestAt(configuration.siteFor, asterism.baseTracking, vizTime)
-    } yield angles
+    } yield {
+      // We sort the angles or we could end up in a loop where the angles are tested back and forth
+      // This is rare but can happen if each angle finds an equivalent guide star
+      given Order[Angle] = Angle.AngleOrder
+      angles.sorted
```
rpiaggio, line 98: `angles.sorted(using Angle.AngleOrder)` ?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AladinCell.scala
cquiroz: Oh didn't know that

PR #3019, opened 2023-06-13
Support scrolling on the obs summary table

cquiroz:


explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -298,10 +299,12 @@ object ObsSummaryTable extends TableHooks:
         TableStore(props.userId, TableId.ObservationsSummary, cols)
       )
     )
-    .render { (props, ctx, _, _, table) =>
+    .useRef(none[HTMLTableVirtualizer])
```
rpiaggio, line 302: It seems like we don't need the ref to the virtualizer since it's not being used. I think we can safely omit this `useRef` and the `virtualizerRef` parameter below.

rpiaggio (approved):
LGTM

PR #3022, opened 2023-06-14
Put program list in the cache and validate the program id in the URL.

toddburnside:


common/src/main/scala/explore/cache/ProgramCache.scala
```diff
@@ -205,7 +217,17 @@ object ProgramCache extends CacheComponent[ProgramSummaries, ProgramCache]:
...
+            ProgramSummaries.programs
+              .modify(_.updated(data.programEdit.value.id, data.programEdit.value))
+          )
+        )
+
     // TODO Handle errors, disable transparent resubscription upon connection loss.
-    (updateTargets, updateObservations, updateGroups, updateAttachments).mapN(
-      _.merge(_).merge(_).merge(_)
+    (updateTargets, updateObservations, updateGroups, updateAttachments, updatePrograms).mapN(
+      _.merge(_).merge(_).merge(_).merge(_)
```
cquiroz, line 232: lots of merging

explore/src/main/scala/explore/Routing.scala
```diff
@@ -102,20 +116,25 @@ object Routing:
...
+      SequenceEditor(routingInfo.programId)
+    )
 
   private def showProgramSelectionPopup(model: View[RootModel]): VdomElement =
-    ProgramsPopup(
-      currentProgramId = none,
-      undoStacks = model.zoom(RootModel.undoStacks)
-    )
+    // Because we are not supplying a program id, the ProgramsPopup will be displayed
+    withProgramSummaries(none, model)(_ => <.div("Programmer error!"))
```
cquiroz, line 137: Cool, I'm looking forward to the first faul report with "Programmer error"

cquiroz:
So, if the program id doesn't exist (or not allowed) the program popup gets shown?

common/src/main/scala/explore/cache/ProgramCache.scala
toddburnside: Is there a better way?

explore/src/main/scala/explore/Routing.scala
toddburnside: Maybe I should add my name? 😆

common/src/main/scala/explore/cache/ProgramCache.scala
cquiroz: not sure what merge does, maybe `foldLeft`?

toddburnside:
> So, if the program id doesn't exist (or not allowed) the program popup gets shown?

Yep. And if there was an id in the url, a message is displayed at the bottom of the popup. 
image
 Because of the way the program cache works, if the program is deleted the user will be immediately taken to the popup. As long as the programEdit subscription works for it, the same thing will happen if authorization is revoked.

cquiroz (approved):
LGTM, this is good

toddburnside:
> As long as the programEdit subscription works for it, the same thing will happen if authorization is revoked.

Actually, we'll probably need to do a bit more to get authorization revoking to work right. We'll need some more information in the subscription, I think.

common/src/main/scala/explore/cache/ProgramCache.scala
toddburnside: > not sure what merge does, maybe `foldLeft`?

Raul thumbed up this comment. I'm not sure what that means.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/cache/ProgramCache.scala
rpiaggio: That we can use `List(...).reduceLeft(_ merge _)`.

toddburnside: Ah. I change that.

toddburnside: Actually, I think it was simpler with `sequence.map` in this case.

rpiaggio (approved):
LGTM

PR #3023, opened 2023-06-15
Better sorting of the modes table

cquiroz:
With the update of react table of 8.9.2 the logic of sorting undefined changed a bit.
After some experiments I thought it maybe be better to take full control defining different orderings for each direction in code.
Now the times are sorted such that all the non available values go to the end regardless if we are ascending or descending.
There are also some small ui touches for the configuration tile

### DIFF
--- a/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
+++ b/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
@@ -54,15 +54,9 @@ import japgolly.scalajs.react.*
 import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.enums.FocalPlane
 import lucuma.core.enums.*
-import lucuma.core.math.BoundedInterval
-import lucuma.core.math.Coordinates
-import lucuma.core.math.SignalToNoise
-import lucuma.core.math.Wavelength
+import lucuma.core.math.*
 import lucuma.core.math.units.Micrometer
-import lucuma.core.model.ConstraintSet
-import lucuma.core.model.CoordinatesAtVizTime
-import lucuma.core.model.SiderealTracking
-import lucuma.core.model.User
+import lucuma.core.model.*
 import lucuma.core.syntax.all.*
 import lucuma.core.util.Display
 import lucuma.core.util.NewType
@@ -208,7 +202,17 @@ private object SpectroscopyModesTable extends TableHooks:
   private given Order[InstrumentRow#FPU]     = Order.by(_.toString)
   private given Order[BasicConfigAndItc]     = Order.by(_.configuration.configurationSummary)
 
-  private given Order[TimeSpan | Unit] = Order.by(_.toOption)
+  private def timeOrder(sort: Option[ColumnSort]): Order[Option[TimeSpan]] =
+    Order.from {
+      case (Some(a), Some(b)) =>
+        a.toMicroseconds.compareTo(b.toMicroseconds)
+      case (Some(a), None)    =>
+        if (sort.exists(_.direction == SortDirection.Descending)) 1 else -1
+      case (None, Some(a))    =>
+        if (sort.exists(_.direction == SortDirection.Descending)) -1 else 1
+      case (None, None)       =>
+        -1
+    }
 
   private def formatInstrument(r: (Instrument, NonEmptyString)): String = r match
     case (i @ Instrument.Gnirs, m) => s"${i.longName} $m"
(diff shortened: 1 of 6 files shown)


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -499,37 +508,41 @@ private object SpectroscopyModesTable extends TableHooks:
...
           (configuration, requirements, sortedRows) =>
             selectedIndex.setState(
               selectedRowIndex(configuration, requirements.wavelength, sortedRows)
             )
       }
       // Set the selected config if the rows change because it might have different itc data.
       // Note, we use rows for the dependency, not sorted rows, because sorted rows also changes with sort.
-      .useEffectWithDepsBy((_, _, _, rows, _, _, _, _, _, _, _) => rows) {
-        (props, _, _, _, _, _, _, _, _, sortedRows, selectedIndex) => _ =>
+      .useEffectWithDepsBy((_, _, _, rows, _, _, timeSort, _, _, _, _, _) => rows) {
```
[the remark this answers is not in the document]
cquiroz: Oh right, I had to do some experiments trying to find a bug on saving table prefrences

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -499,37 +508,41 @@ private object SpectroscopyModesTable extends TableHooks:
...
+          table.setSorting(s) *>
+            sorting.setState(s.value.headOption) *> scrollTo
+              .setState(ScrollTo.Scroll)
       }
-      .useMemoBy((_, _, _, rows, _, _, _, table, _) => (rows, table.getState().sorting)) {
-        (_, _, _, _, _, _, _, table, _) => (_, _) =>
-          table.getSortedRowModel().rows.map(_.original).toList
+      .useMemoBy((_, _, _, rows, _, _, timeSort, _, table, _) =>
+        (rows, timeSort.value, table.getState().sorting)
+      ) { (_, _, _, _, _, _, sorting, _, table, _) => (_, s, _) =>
```
rpiaggio, line 519: It seems named `s` is not used in the hook body.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -499,37 +508,41 @@ private object SpectroscopyModesTable extends TableHooks:
       // By having it as state with the following `useEffectWithDepsBy`, the scrollTo effect will run
       // in the following "hook cycle" and get the proper index.
       .useState(ScrollTo.Scroll)
-      .useEffectWithDepsBy((_, _, _, _, _, _, _, table, _) => table.getState().sorting) {
-        (_, _, _, _, _, _, _, _, scrollTo) => _ => scrollTo.setState(ScrollTo.Scroll)
+      .useEffectWithDepsBy((_, _, _, _, _, _, _, _, table, _) => table.getState().sorting) {
+        (_, _, _, _, _, _, sorting, _, table, scrollTo) => s =>
+          table.setSorting(s) *>
+            sorting.setState(s.value.headOption) *> scrollTo
```
rpiaggio, line 514: I'm not sure the whole mechanism will work if time is not the first column in the sort. Eg: we can sort first by instrument and then by time.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -208,7 +202,17 @@ private object SpectroscopyModesTable extends TableHooks:
   private given Order[InstrumentRow#FPU]     = Order.by(_.toString)
   private given Order[BasicConfigAndItc]     = Order.by(_.configuration.configurationSummary)
 
-  private given Order[TimeSpan | Unit] = Order.by(_.toOption)
+  private def timeOrder(sort: Option[ColumnSort]): Order[Option[TimeSpan]] =
```
rpiaggio, line 205: Does this mechanism push undefined values to the bottom when the column is not sorted?

cquiroz: Nope, the order is defined by the other columns

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: II don't think multi sort is enabled in this table. AFAIK only one column is sorted

rpiaggio (approved):
LGTM

PR #3024, opened 2023-06-15
Migrate to new sequence model

rpiaggio:


explore/src/main/scala/explore/config/sequence/GmosSequenceRow.scala
```diff
@@ -35,14 +37,14 @@ sealed trait GmosSequenceRow:
   // TODO Not in model yet, we are just simulating
   lazy val guided: Boolean                                  =
```
cquiroz, line 38: Is this whether offsets are guided?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/config/sequence/GmosSequenceRow.scala
rpiaggio: I'm assuming so. I took the columns a while ago from the inception document.

rpiaggio: Great, I saw that, it's actually in the query now and in memory. We just have to switch to showing it.

PR #3028, opened 2023-06-16
Alternate sorting strategy

rpiaggio:
This PR leverages the strategy of capturing the sorting state introduced in the base PR with using again `react-table`'s functionality for treating undefined values instead of reimplementing it.

It also contemplates multisort, and fixes the sorting by instrument column.


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -294,7 +287,13 @@ private object SpectroscopyModesTable extends TableHooks:
         .setCell(cell => itcCell(cell.row.original.result, cw))
         .setColumnSize(FixedSize(80.toPx))
         .setEnableSorting(progress.isEmpty)
-        .sortable(using timeOrder(timeSort).toOrdering),
+        .setEnableSorting(true)
+        .setSortUndefined {
+          timeSortDirection match
```
cquiroz, line 292: I see, I'll merge it and test it

PR #3031, opened 2023-06-16
Test firebase

toddburnside:


toddburnside:
Firebase deployment will be done in conjunction with a move to Cirrus CI.

PR #3044, opened 2023-06-20
Switch to single config file

toddburnside:


rpiaggio (approved):
LGTM

common/src/main/public/README.txt
```diff
@@ -0,0 +1,10 @@
+The "conf.json" and "staging.conf.json" files are deprecated. They
```
cquiroz, line 1: I guess this file will be made public right?

model/shared/src/main/scala/explore/model/AppConfig.scala
```diff
@@ -42,34 +41,36 @@ case class AppConfig(
...
+          .orElse(confs.find(_.hostName === "*"))
+          .fold(
+            Async[F].raiseError(new Exception("Host not found in configuration."))
+          )(_.pure)
+      )
 
   def fetchConfig[F[_]: Async](env: ExecutionEnvironment, client: Client[F]): F[AppConfig] =
     client
-      .get(confName(env))(_.decodeJson[AppConfig])
+      .get(configFile)(_.decodeJson[List[AppConfig]])
```
cquiroz, line 66: Nice this makes it simpler than what I thought it would be

PR #3045, opened 2023-06-20
Remove queries subproject

rpiaggio:
With `sbt-clue`, it is no longer necessary to have a project for the queries.

This PR redistributes the files previously in the `queries` project in order to simplify the build.


cquiroz:
What's the role of sbt-clue?

rpiaggio:
It basically invokes scalafix to generate code, but it doesn't require the code templates to be in a separate project. It can be in a `clue` subfolder in the same project and it will create a project behind the scenes.

See https://github.com/gemini-hlsw/clue/pull/483

cquiroz (approved):
LGTM

PR #3046, opened 2023-06-20
Support local.conf.json again.

toddburnside:
The `local.conf.json` file will need to use the new config file format with an array of configurations. It essentially needs to be the `environments.conf.json` file with an updated development entry. The production and staging entries can be deleted.


rpiaggio (approved):
LGTM

PR #3047, opened 2023-06-20
Firebase promotion scripts

rpiaggio:


toddburnside (approved):
LGTM - except we probably don't need `size.sh`?

rpiaggio:
> LGTM - except we probably don't need `size.sh`?

That file is already in the repo, I just changed its permissions to `+x`, probably that's why it appears as empty on the diff.

toddburnside:
> That file is already in the repo, I just changed its permissions to `+x`, probably that's why it appears as empty on the diff.

Ah. OK.

PR #3049, opened 2023-06-21
Vite build improvements

hugo-vrijswijk:
Move Vite config to use async API's instead of sync (which is not recommended as it is blocking). File copying  from the public dir is now also done in parallel, and copy everything in the dir, instead of a list of files.


hugo-vrijswijk:
I'm unsure why this change makes the bundle bigger 🤔

cquiroz (approved):
LGTM

PR #3057, opened 2023-06-22
Fix selecting new observation resetting scroll position and scroll to new observations

hugo-vrijswijk:
* Fix selecting new observation resetting scroll position
* Scroll to new observations when added

### DIFF
--- a/explore/src/main/scala/explore/observationtree/ObsList.scala
+++ b/explore/src/main/scala/explore/observationtree/ObsList.scala
@@ -43,6 +43,8 @@ import lucuma.ui.syntax.all.*
 import lucuma.ui.syntax.all.given
 import lucuma.ui.utils.*
 import monocle.Lens
+import org.scalajs.dom
+import org.scalajs.dom.Element
 import org.typelevel.log4cats.Logger
 import queries.schemas.odb.ObsQueries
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -267,11 +283,14 @@ object ObsList:
                   clazz = ExploreStyles.ButtonSummary
                 )
               ),
-              Tree(
-                treeNodes,
-                renderItem,
-                expandedKeys = expandedGroups.get,
-                onToggle = expandedGroups.set
+              <.div(
+                ^.overflow := "auto",
```
cquiroz, line 287: Why do we need the overflow?

hugo-vrijswijk: To actually make the list scrollable. Before it came from a parent, but the wrapping in the div is needed to prevent scroll-resetting. I spent a week on this, don't ask me why 🪄

cquiroz: I can certainly believe that

cquiroz (approved):
LGTM

PR #3058, opened 2023-06-22
Support multiple finder charts

cquiroz:
https://github.com/gemini-hlsw/explore/assets/3615303/13b21afb-edef-4ef4-8bf0-5ef9ad7c1dfb


common/src/main/scala/explore/components/state/SolarProgress.scala
```diff
@@ -0,0 +1,36 @@
...
+package explore.components
+
+import japgolly.scalajs.react.*
+import japgolly.scalajs.react.vdom.html_<^.*
+import react.common.Css
+import react.common.ReactFnProps
+
+case class SolarProgress(css: Css = Css.Empty) extends ReactFnProps(SolarProgress.component)
+
+object SolarProgress {
```
rpiaggio, line 13: Maybe we can move this to lucuma-ui ?

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/state/SolarProgress.scala
cquiroz: Yes. I'll do it next

PR #3060, opened 2023-06-22
Add `switching` View bracket utility extension

hugo-vrijswijk:
I saw this pattern used a lot in the codebase, so I thought it would be good to generalize it a little bit.

~_Very_ open to naming suggestions. Earlier thought was `useAdding`, but I also don't really like that~ `switching`, for now

### DIFF
--- a/common/src/main/scala/explore/syntax/ui/package.scala
+++ b/common/src/main/scala/explore/syntax/ui/package.scala
@@ -4,7 +4,7 @@
 package explore.syntax.ui
 
 import cats.*
-import cats.effect.MonadCancel
+import cats.effect.MonadCancelThrow
 import cats.syntax.all.*
 import clue.ResponseException
 import clue.js.FetchJSRequest
@@ -54,7 +54,7 @@ extension (vault: UserVault)
     request
 
 extension [F[_]: ApplicativeThrow: ToastCtx, A](f: F[A])
-  def toastErrors: F[A]                                                                          =
+  def toastErrors: F[A] =
     f.onError {
       case ResponseException(errors, _) =>
         errors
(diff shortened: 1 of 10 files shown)


cquiroz:
Seems a good idea. Not sure about the name `useXXX` suggested me it was a hook. Maybe `withBoolSwitch`?

common/src/main/scala/explore/syntax/ui/package.scala
```diff
@@ -66,3 +68,21 @@ extension [F[_]: ApplicativeThrow: ToastCtx, A](f: F[A])
         ToastCtx[F]
           .showToast(throwable.getMessage, Message.Severity.Error, sticky = true)
     }
+
+/**
+ * Switch the value of a ViewF to true-ish (by the function) while executing the given effect, then
+ * switch it back to false when the effect is finished
+ */
+extension [F[_]: Monad, A](adding: ViewF[F, A])
+  def useBoolSwitchBy[B](makeBool: Boolean => A)(f: F[B])(using MonadCancel[F, Throwable]): F[B] =
```
rpiaggio, line 77: This seems to require both `Monad` and `MonadCancel`, which is redundant. Maybe you can just do `[F[_]: MonadThrow]` ?

rpiaggio:
This is a neat idea, thank you!

Other name suggestions: `switchAround`/`switchAroundBy`.

Actually, have you considered the inverse pattern? That is, an extension to `F[A]` such that if we have `effect: F[A]` and a `switch: Boolean => F[Unit]` we can do `effect.switching(switch)` (or `effect.switching(switch, Loading(_))`).

hugo-vrijswijk:
> This is a neat idea, thank you!
> 
> Other name suggestions: `switchAround`/`switchAroundBy`.
> 
> Actually, have you considered the inverse pattern? That is, an extension to `F[A]` such that if we have `effect: F[A]` and a `switch: Boolean => F[Unit]` we can do `effect.switching(switch)` (or `effect.switching(switch, Loading(_))`).
> 
> I think this would keep the focus on `effect` instead of the switch.

Yeah, I like this. I've refactored it with your suggestion.

There's also a pattern of switching between two enum values, but I decided against using this pattern in those cases as the boolean switching seemed a bit arbitrary (It would look like: `f.switching(action.async, if (_) Action.Replace else Action.None`), thoughts?

rpiaggio:
> There's also a pattern of switching between two enum values, but I decided against using this pattern in those cases as the boolean switching seemed a bit arbitrary (It would look like: f.switching(action.async, if (_) Action.Replace else Action.None), thoughts?

We could provide values for before and after. Something like `f.switching(setter, Status.Active, Status.Inactive)`.

rpiaggio (approved):
LGTM

PR #3077, opened 2023-06-27
Small user prefs migration

cquiroz:
This replaces some leftover columns defined using custom postgres types rather than the recommended table enums

### DIFF
--- a/fetchUserPreferencesSchema.sh
+++ b/fetchUserPreferencesSchema.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-gq https://user-prefs-development.herokuapp.com/v1/graphql --introspect >common-queries/src/clue/resources/UserPreferencesDB.graphql
+gq https://user-prefs-master.herokuapp.com/v1/graphql --introspect >common/src/clue/resources/UserPreferencesDB.graphql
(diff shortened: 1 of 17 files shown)


fetchUserPreferencesSchema.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-gq https://user-prefs-development.herokuapp.com/v1/graphql --introspect >common-queries/src/clue/resources/UserPreferencesDB.graphql
```
rpiaggio, line 3: Maybe we want to keep referencing `development` here?

rpiaggio (approved):
LGTM

fetchUserPreferencesSchema.sh
cquiroz: Not sure, do we normally get the schema from dev, I'd think we only update it when we change it on master?

rpiaggio: I was under the impression that development is for unmerged changes and master is for merged ones.

In other words, when we make a change, we change it on development first, and fetch it so as to be able to work on explore.

When the PR is merged we promote the development hasura to master. If we promote earlier we might break it for other developers using master.

That said, I see that the development environment in `environments.conf.json` points to development instead of master. Should we change this? Otherwise, the names are a bit confusing.

cquiroz: I thought it was the other way around 😄

cquiroz: I guess we can revert in another PR if needed

PR #3079, opened 2023-06-27
Store finder chart transformations on the user prefs

cquiroz:


hugo-vrijswijk (commented):
Do we need to do any migration for current data?

PR #3080, opened 2023-06-28
A step closer to universal undo

rpiaggio:
This PR merges the undo stacks for all the data we have in the program cache.

The only thing left out for the moment is proposal data, but we should put that into the cache and finish unifying undo in a future PR.

Also pending is some mechanism to auto-navigate to the site that is being undone or redone. I think it shouldn't be too complex to store the URL in the stack.

Finally, redo is not working for targets for some reason (but undo does work). I'll look into this in a next PR.


common/src/main/scala/explore/undo/UndoContext.scala
```diff
@@ -76,7 +76,7 @@ case class UndoContext[M](
       _ <- push(undoStack)(Restorer[DefaultA, M, A](model.get, getter, setter, onRestore))
       _ <- reset(redoStack)
       _ <- model.mod.compose(setter)(v)
-      f <- onSet(model.get, v).runAsync
+      f <- onSet(model.get, v).runAsyncAndForget // TODO: Log errors
```
hugo-vrijswijk, line 79: Does `runAsync` not have the same behaviour, but log errors?

common/src/main/scala/explore/undo/Undoer.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import japgolly.scalajs.react.util.DefaultEffects.{Sync => DefaultS}
+
+/** Undo controls */
+trait Undoer:
+  def undo: DefaultS[Unit]
+  def redo: DefaultS[Unit]
+  def isUndoEmpty: Boolean
+  def isRedoEmpty: Boolean
+  def working: Boolean
```
cquiroz, line 14: What does `working` mean?

cquiroz (approved):
LGTM

common/src/main/scala/explore/undo/UndoContext.scala
rpiaggio: Yes, but we need a `Logger` for that and we are now creating a global `UndoContext` from the router, which doesn't have a `Logger` instance on hand.

I'll have to look into how to get a `Logger` there, or else modify the `UndoContext` so that the `Logger` is required by the methods instead of the constructor.

Future PR, for now I put this note in the code.

PR #3093, opened 2023-06-30
Move the attachment selector to the toolbar

cquiroz:
image (14)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -19,18 +19,19 @@ import react.common.style.*
...
+  control:            TileSizeState => Option[VdomNode] = _ => None,
+  canMinimize:        Boolean = false,
+  canMaximize:        Boolean = false,
+  state:              TileSizeState = TileSizeState.Normal,
+  sizeStateCallback:  TileSizeState => Callback = _ => Callback.empty,
+  controllerClass:    Option[Css] = None, // applied to wrapping div when in a TileController.
+  bodyClass:          Option[Css] = None, // applied to tile body
+  tileClass:          Option[Css] = None, // applied to the tile
+  tileTitleClass:     Option[Css] = None, // applied to the title
+  renderInTitleClass: Option[Css] = None  // applied to the portal in the title
```
rpiaggio, line 34: This comment goes beyond the scope of this PR, but I wonder if we need `Option[Css]` here when we have `Css.Empty`.

rpiaggio (approved):
LGTM

common/src/main/scala/explore/components/Tile.scala
cquiroz: That's a good suggestion, I'll implement it

PR #3108, opened 2023-07-05
Put the back button on the first tile always

cquiroz:
In small screens we have a back  button but it is always on the first tile as defined statically
This PR changes the logic to dynamically select the first tile in case the user changes the order


swalker2m (approved):
Certainly seems credible, as far as I know.  :-)

PR #3109, opened 2023-07-06
Require signal to noise at to call ITC

cquiroz:
Signal to Noise At will be required on the client
Depending on user feedback we may force it  on the server too


swalker2m (approved):
This is excellent, it will prevent really slow configurations from being submitted.  :+1:

PR #3110, opened 2023-07-07
Implement cache for user preferences

cquiroz:
We added a better client side cache for the odb and we can do something similar for the user preferences.
This lays out the ground work and uses it for the grid layout state.
The code downstream gets nicely simplified IMHO


swalker2m (approved):
I don't know what any of this is / looks great to me! :+1:

PR #3112, opened 2023-07-08
Replace react-moon

cquiroz:
React-moon has been flagged as obsolete lately.
It could be fully removed just using css. This also lets us remove `styled-components`


explore/src/main/scala/explore/targeteditor/ElevationPlotNight.scala
```diff
@@ -428,15 +428,17 @@ object ElevationPlotNight {
...
-              MoonPhase(
-                phase = moonPhase,
-                size = 20,
-                border = "1px solid black",
-                darkColor = "#303030"
-              ),
-              <.small("%1.0f%%".format(moonIllum * 100))
-            )
+            // Adapted from https://dev.to/thormeier/use-your-i-moon-gination-lets-build-a-moon-phase-visualizer-with-css-and-js-aih
+            <.div(ExploreStyles.MoonSphere)(
```
hugo-vrijswijk, line 432: Should we extract this into a separate component similar to the `MoonPhase` component? It doesn't really fit in the responsibility of this component now, I think

common/src/main/webapp/sass/solar-system.scss
```diff
@@ -1,98 +0,0 @@
...
-  }
-
-  to{
-    transform: rotate(360deg);
-  }
-}
-
-/* GRID STYLING */
-
-:root {
```
hugo-vrijswijk, line 15: Have we lost the solar system spinner? 😢

hugo-vrijswijk (commented):
Just 1 comment about the component

common/src/main/webapp/sass/solar-system.scss
cquiroz: Yes, it is now on lucuma-ui

explore/src/main/scala/explore/targeteditor/ElevationPlotNight.scala
cquiroz: Right, we could make it a component and move it to lucuma-ui

cquiroz: I made it a component, I'll move it lucuma-ui later on

PR #3119, opened 2023-07-10
Fix configuration input selection not sticking during input changes

hugo-vrijswijk:
Now uses a `selectedRow` instead of `selectedIndex` (which is still there, but a memo). Sorting changing during selection changes will follow the selected row

Video:

https://github.com/gemini-hlsw/explore/assets/10114577/3049ba93-5224-4dbe-bf4b-967f7ce683c4

### DIFF
--- a/common/src/main/scala/explore/syntax/ui/package.scala
+++ b/common/src/main/scala/explore/syntax/ui/package.scala
@@ -66,17 +66,25 @@ extension [F[_]: ApplicativeThrow: ToastCtx, A](f: F[A])
     }
 
 extension [F[_]: MonadCancelThrow, A](f: F[A])
+
   /**
    * Switch the value of a ViewF to true while executing the given effect, then switch it back to
    * false when the effect is finished
    */
   def switching(
     view: ViewF[F, Boolean]
-  ): F[A] = switching(view, identity)
+  ): F[A] = switching(view, true, false)
 
   /**
    * Switch the value of a ViewF to true-ish (by the function) while executing the given effect,
    * then switch it back to false when the effect is finished
    */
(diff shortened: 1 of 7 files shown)


cquiroz:
The video looks great

common/src/main/scala/explore/syntax/ui/package.scala
```diff
@@ -66,17 +66,25 @@ extension [F[_]: ApplicativeThrow: ToastCtx, A](f: F[A])
...
   def switching[B](view: ViewF[F, B], boolToB: Boolean => B): F[A] =
-    MonadCancelThrow[F].bracket(view.set(boolToB(true)))(_ => f)(_ => view.set(boolToB(false)))
+    switching(view, boolToB(true), boolToB(false))
+
+  /**
+   * Switch the value of a ViewF to the @param acquire value while executing the given effect, then
+   * switch it back to @param release when the effect is finished
+   */
+  def switching[B](view: ViewF[F, B], acquire: B, release: B): F[A] =
+    MonadCancelThrow[F].bracket(view.set(acquire))(_ => f)(_ => view.set(release))
```
cquiroz, line 90: Interesting, this could have some uses elsewhere

cquiroz (approved):
Looks great, thanks

common/src/main/scala/explore/syntax/ui/package.scala
hugo-vrijswijk: Like where? It's sort of a specialized `Resource` almost

cquiroz: Just the concept of acquire/use/release

PR #3120, opened 2023-07-10
Use effect for scrolling to newly created obs instead of callback

hugo-vrijswijk:
This seems slightly more reliable than the callback. It still doesn't _always_ scroll to the newly created observation, but I think because it is 1 tick later it's more reliable


cquiroz (approved):
LGTM

PR #3124, opened 2023-07-12
Correctly handle if getElementById returns null

hugo-vrijswijk:
I'm not sure in what cases it returns null or undefined (or if me seeing the undefined was a fever dream), but this handles both cases

### DIFF
--- a/explore/src/main/scala/explore/observationtree/ObsList.scala
+++ b/explore/src/main/scala/explore/observationtree/ObsList.scala
@@ -72,17 +72,14 @@ object ObsList:
     )
 
   private def scrollIfNeeded(targetObs: Observation.Id) =
-    Callback.suspend {
+    Callback {
       Option(dom.document.getElementById(s"obs-list-${targetObs.toString}"))
         .filterNot(js.isUndefined)
-        .map { target =>
-          Callback {
-            val rect = target.getBoundingClientRect()
-            if (rect.top < 0) target.scrollIntoView()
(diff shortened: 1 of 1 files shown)


toddburnside (approved):
LGTM

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -73,14 +73,13 @@ object ObsList:
...
     Callback {
-      // 'getElementById' does not return any optional type, so coerce to js.UndefOr
-      val target: js.UndefOr[Element] =
-        dom.document.getElementById(s"obs-list-${targetObs.toString}")
-      target.foreach { t =>
-        val rect = t.getBoundingClientRect()
-        if (rect.top < 0) t.scrollIntoView()
-        if (rect.bottom > dom.window.innerHeight) t.scrollIntoView(false)
-      }
+      Option(dom.document.getElementById(s"obs-list-${targetObs.toString}"))
```
cquiroz, line 76: This got me thinking, are refs the proper react way to do this?

hugo-vrijswijk: Potentially, but with a list I don't think that would work. Or you'd have to have a single ref that you set on the last added observation, but that seems a lot more complex than a query selector

cquiroz: Right, now I recall I tried that

cquiroz (approved):
LGTM

PR #3125, opened 2023-07-12
Switch to the moon component in lucuma-ui

cquiroz:


toddburnside (approved):
LGTM, but I think there are styles in `ExploreStyles` that can be removed.

PR #3131, opened 2023-07-14
Improve saving/restoring of tile layouts in user preferences

toddburnside:
The min/max values width/height for the tiles were being lost because we were directly using the data from User Preferences, where only x, y, width and height are saved. This PR restores merging of the user preferences data back onto the current values.


cquiroz (approved):
This looks great, and thanks for the tests

PR #3136, opened 2023-07-17
Add empty programs tab

hugo-vrijswijk:
The tab is now always shown. I'm not sure if this is what is needed, or based on what we should hide/show the tab. Input is welcome 😅.

Also adds a migration for the tab layout in the userPreferences db.

image


cquiroz:
Can you test on phone sized screens, the tabs go to the bottom and may not fit now

hugo-vrijswijk:
@cquiroz It does become a bit cramped. Should we have the buttons overflow? 

image

hugo-vrijswijk:
I've updated it to have an overflow, so when the viewport is too small you can just scroll sideways. 

I've also fixed the 1px slide that the buttons do (on desktop too) when hovering over them. That's been bothering me since I started 😅

https://github.com/gemini-hlsw/explore/assets/10114577/275ff08f-d429-4424-851c-18e446cd274e

PR #3140, opened 2023-07-19
Setup linting with Prettier

hugo-vrijswijk:
Also sets up formatting on save in VS Code settings.

- Setup lint checking with prettier
- Format with prettier
- Add 'Format with prettier' to .git-blame-ignore-revs


hugo-vrijswijk:
> Seems like a good idea. The plugin to check the variables is maybe old now, right?

What plugin do you mean? `stylelint-value-no-unknown-custom-properties` is still active. The oldest plugin is `stylelint-color-format` which was last published 3 years ago, but still works fine

cquiroz (approved):
LGTM

PR #3142, opened 2023-07-19
Move SSO and state classes to lucuma-ui

rpiaggio:
This implied a few changes:
- I had to tweak the logic a bit in order to seprate things specific to explore from things we can reuse in other projects. As a result, the `IfLogged` component now has more parameters.
- `model` cannot see `UserVault` and `SSOConfig` anymore, since those classes were moved to `lucuma-ui`, which is Scala.js-only and `model` is cross-compiled. This meant I had to move `RootModel` and `AppConfig` to `common`. The `ItcServer` worker used `AppConfig` but now it cannot see it anymore since it doesn't depend on `common`. In any case, it only needed the ITC server URL from the `AppConfig` which is now passed by the client in a new initialization message.


cquiroz:
Should we just merge model and common and not do cross compilation of model?

rpiaggio:
> Should we just merge model and common and not do cross compilation of model?

We shouldn't merge `model` and `common` (although we may want to rename). `model` is shared by the UI and `workers`, but `common` is only used by the UI.

However, if we do drop the JVM compilation, we could then simplify things by merging `common` and `explore`

cquiroz (approved):
LGTM

PR #3148, opened 2023-07-21
Move global user preferences to the cache

cquiroz:
There are user preferences that are global and can be moved to the user prefs cache
This simplifies several parts of the code


common/src/clue/scala/queries/common/UserPreferencesQueriesGQL.scala
```diff
@@ -125,28 +128,61 @@ object UserPreferencesQueriesGQL {
           saturation
           brightness
         }
+      }
+    """
+  }
+
+  case class LucumaUserPreferences(
+    val aladinMouseScroll:  Option[Boolean],
```
rpiaggio, line 136: Why the `val`s?

rpiaggio: Also, do you think it's possible to decode directly into `GlobalPreferences`? I guess it would take a custom `Decoder`.

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -123,37 +121,39 @@ object AladinCell extends ModelOptics with AladinCommon:
   }
 
   private given Reusability[Props] =
-    Reusability.by(x => (x.uid, x.tid, x.obsConf, x.asterism, x.fullScreen.get))
+    Reusability.by(x =>
+      (x.uid,
+       x.tid,
+       x.obsConf,
+       x.asterism,
+       x.fullScreen.reuseByValue,
```
rpiaggio, line 129: `.get` has the same effect upon reusability here than `.reuseByValue` and avoids creating `Reuse` instances.

rpiaggio (approved):
LGTM with some comments and suggestions

common/src/clue/scala/queries/common/UserPreferencesQueriesGQL.scala
cquiroz: Could be, though in that case I'd change the db to make the columns no nullable, that's probably the best

rpiaggio (approved):
LGTM

PR #3174, opened 2023-08-03
Partial fix for changed to plannedTime API

cquiroz:
The latest change on the API related to plannedtime broke explore
This makes it work but there is still work pending to adapt to the new API, I think @hugo-vrijswijk is working on planned time in general


rpiaggio (approved):
LGTM

toddburnside (approved):
Looks good for a short term fix to get explore working. 👍

swalker2m:
How are things set up?  When this is merged, does it go directly to staging?

toddburnside:
> How are things set up? When this is merged, does it go directly to staging?

That is correct. We'll need to promote the odb when this is merged.

PR #3176, opened 2023-08-03
Update vite

cquiroz:


vite.config.mjs
```diff
@@ -97,25 +99,29 @@ const pathExists = async (path) => {
 
 // https://vitejs.dev/config/
 export default defineConfig(async ({ mode }) => {
-  const scalaClassesDir = path.resolve(__dirname, `explore/target/scala-3.3.0`);
+
+  const _dirname = typeof __dirname !== 'undefined'
+    ? __dirname
+    : dirname(fileURLToPath(import.meta.url))
```
rpiaggio, line 105: Why is this necessary?

rpiaggio (approved):
LGTM

vite.config.mjs
cquiroz: It is a change on vite 4.4.0 which requires to pass not a function but a value on the externaal build method.
this breaks where `_dirname` is available

See:
https://github.com/vitejs/vite/commit/998512b
https://github.com/vitejs/vite/issues/6946

PR #3187, opened 2023-08-05
Fix a bug when calculating the link on the targets tab

cquiroz:
Under some circumstances the target list may not redirect to the correct address. This fixes it and also shows where we'd jump when selecting a target


explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
```diff
@@ -322,14 +322,18 @@ object AsterismGroupObsList:
               <.span(ExploreStyles.ObsCount, s"${obsIds.size} Obs")
             )
 
+            val clickFocus = props.focused.target.fold(Focused.fromAsterismGroup(asterismGroup))(
```
cquiroz, line 325: This is the core changee

toddburnside (approved):
LGTM

PR #3193, opened 2023-08-08
Sequence tile

cquiroz:
The sequence is sometimes hard to find. This PR moves it from the popup in the configuration to its own tile
Note that it can only be seen in approved observations

This PR also introduces the concept of hidden tiles
image


swalker2m:
One of the goals of GPP was to keep the PI from having to worry about the sequence.  Maybe it _should_ be hard to find?  You might argue that seeing the sequence is going to encourage making manual edits to it.

toddburnside:
> One of the goals of GPP was to keep the PI from having to worry about the sequence. Maybe it _should_ be hard to find? You might argue that seeing the sequence is going to encourage making manual edits to it.

Hmm. That's a good point. There was all that effort made to discourage users from making edits to parts of the configuration. This seems a bit at odds with that.

rpiaggio:
How does this behave? The `View Sequence` button now unhides the tile instead of showing a popup?

Is this a requirement?

andrewwstephens:
There were a few motivations to try putting the sequence in its own panel:
- make it easier to see how the sequence changes as conditions change (currently one must switch back and forth, while a panel would allow seeing both at the same time).
- avoid duplicating the buttons to change offsets and wavelength dithers.
- provide a place for the sequence editor.
- provide a place to show executed step details (file names, QA states, comments).

toddburnside (approved):
LGTM

PR #3194, opened 2023-08-10
Update lucuma-core to 0.84.0 for ObjectTracking changes.

toddburnside:


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -291,16 +291,12 @@ object ObsTabTiles:
           // asterism base coordinates at viz time or default to base coordinates
```
toddburnside, line 291: Do we want to default this to the base coordinates or to the current time?

toddburnside: For one thing, when we add non-sidereals, we'll lose the concept of base coordinates.

cquiroz: So we'd likke to have reference coordinates that are not time dependent
That probably make no sense for a non-sidereals though we could have mixed asetrisms right?

model/shared/src/main/scala/explore/model/ObsSummary.scala
```diff
@@ -121,6 +121,8 @@ case class ObsSummary(
   lazy val constraintsSummary: String =
     s"${constraints.imageQuality.label} ${constraints.cloudExtinction.label} ${constraints.skyBackground.label} ${constraints.waterVapor.label}"
 
+  def visualizationTimeOrNow: Instant = visualizationTime.getOrElse(Instant.now())
```
cquiroz, line 124: Should this be a `F[Instant]`?

cquiroz: In general we always have a viz time, in the obs is defined as optional but on the UI we try to get it from the db but if missing we default to `now`
That is done at a higher level so that models below assume there is one

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
toddburnside: Right. But it works for now. Non-sidereals are going to require a lot of changes...

model/shared/src/main/scala/explore/model/ObsSummary.scala
cquiroz: See
https://github.com/gemini-hlsw/explore/blob/master/explore/src/main/scala/explore/targeteditor/TargetTable.scala#L118

model/shared/src/main/scala/explore/model/boopickle/CatalogPickler.scala
```diff
@@ -105,14 +105,14 @@ trait CatalogPicklers extends CommonPicklers {
...
-
   given Pickler[ObjectTracking.SiderealObjectTracking] = generatePickler
 
+  given Pickler[ObjectTracking.SiderealAsterismTracking] = generatePickler
+
   given Pickler[ObjectTracking] =
     compositePickler[ObjectTracking]
-      .addConcreteType[ObjectTracking.ConstantTracking]
       .addConcreteType[ObjectTracking.SiderealObjectTracking]
+      .addConcreteType[ObjectTracking.SiderealAsterismTracking]
```
cquiroz, line 115: We need to invalidate the catalog cache with this change

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
cquiroz: Yeah it will be a big task

model/shared/src/main/scala/explore/model/ObsSummary.scala
toddburnside: Hmmm. Yeah, it probably should. I know we're doing this elsewhere, too. I'll see if I can find and fix those, too.

model/shared/src/main/scala/explore/model/boopickle/CatalogPickler.scala
toddburnside: Does the cache rely on the `ObjectTracking` or on the coordinates based on the `ObjectTracking`?

model/shared/src/main/scala/explore/model/ObsSummary.scala
toddburnside: I'm moving this to an extension method on Option[Instant]. I can see where it is possible to pass it as a "parameter", but sometimes we need it for every row in a table.

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -185,21 +187,15 @@ object ObsSummaryTable extends TableHooks:
           .setCell(_.value),
         column(
           RAColumnId,
-          r =>
-            r.asterism
-              .map(_.baseTracking.baseCoordinates.ra) // TODO: baseTrackingAt
-              .orElse(r.targetWithId.map(_.target).flatMap(Target.baseRA.getOption)),
-          r => Target.baseRA.getOption(r.targetWithId.target)
```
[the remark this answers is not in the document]
toddburnside: I don't know if it is possible to default these to the current time since getting the current time is effectful. I can default to base coordinates for now. This will need to change when we support non-sidereals, but at that point we may need to get coordinates from the ODB or somewhere, anyway.

model/shared/src/main/scala/explore/model/ObsSummary.scala
toddburnside: This is no longer necessary at all.

explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
```diff
@@ -115,19 +115,17 @@ object ObsAttachmentsTable extends TableHooks with ObsAttachmentUtils:
...
                                  dom.readReadableStream(IO(f.stream()))
             ) *>
-            props.obsAttachments
-              .mod(
-                _.updatedWith(oa.id)(
-                  _.map(
-                    _.copy(fileName = name,
-                           updatedAt = Timestamp.unsafeFromInstantTruncated(Instant.now()),
-                           checked = false
+            IO(Timestamp.unsafeFromInstantTruncated(Instant.now())).flatMap { now =>
```
cquiroz, line 118: We have this
https://github.com/gemini-hlsw/explore/blob/master/workers/src/main/scala/workers/package.scala#L46

Though it may need to be moved to be reused

model/shared/src/main/scala/explore/model/boopickle/CatalogPickler.scala
cquiroz: As discussed we don't really need to invalidate

cquiroz (approved):
LGTM

explore/src/main/scala/explore/attachments/package.scala
```diff
@@ -121,21 +121,23 @@ trait ObsAttachmentUtils:
...
-                                Timestamp.unsafeFromInstantTruncated(Instant.now())
-                              )
-                    )
-                  )).toAsync
+              IO(Timestamp.unsafeFromInstantTruncated(Instant.now())).flatMap { now =>
+                (onSuccess(id) *>
+                  obsAttachments
+                    .mod(
+                      _.updated(id,
+                                ObsAttachment(
```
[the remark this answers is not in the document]
toddburnside: Because we need to proactively update the local times. There is a `lastUpdated` time. It won't be exact with the server, but it's close enough because we only display the date.

explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
toddburnside: I moved `IO.now()` into the model project (explore.model.syntax.all). It could be used by the workers, but doesn't seem to be at this point.

cquiroz: 👍

PR #3209, opened 2023-08-17
Check itc version and cleanup if data version changes

cquiroz:
This PR adds a task to check  every 1h if the data on ITC has changed and if so the local itc cache is cleared

### DIFF
--- a/workers/src/main/scala/workers/ItcServer.scala
+++ b/workers/src/main/scala/workers/ItcServer.scala
@@ -10,6 +10,7 @@ import cats.syntax.all.*
 import explore.events.*
 import explore.itc.ITCGraphRequests
 import explore.itc.ITCRequests
+import explore.itc.ITCVersionsRequests
 import explore.model.StaticData
 import explore.model.boopickle.ItcPicklers
 import explore.modes.SpectroscopyModesMatrix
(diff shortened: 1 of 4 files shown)


workers/src/main/scala/workers/ItcServer.scala
```diff
@@ -50,7 +51,11 @@ object ItcServer extends WorkerServer[IO, ItcMessage.Request] with ItcPicklers {
     } yield { invocation =>
       invocation.data match
         case ItcMessage.Initialize(itcURI) =>
-          ItcClient.create[IO](itcURI, createClient) >>= (client => itcClient.complete(client).void)
+          for {
+            client <- ItcClient.create[IO](itcURI, createClient)
+            _      <- itcClient.complete(client).void
+            _      <- ITCVersionsRequests.queryItc[IO](cache, client).andWait(1.hour).foreverM.start
```
toddburnside, line 57: So we could have up to an hour of using old itc version data, but the itc changes infrequently so it's not a big deal?

toddburnside (approved):
LGTM

workers/src/main/scala/workers/ItcServer.scala
cquiroz: Exactly this only hapeens when we change the itc data which doesn't change often. The server version changes on every lib update

PR #3215, opened 2023-08-17
Add control to adjust a finder chart to the parallactic angle

cquiroz:
It also fixes a bug when storing the transformation on the user database

image


toddburnside (approved):
LGTM

PR #3219, opened 2023-08-20
Provide sensible roi and ccd to ITC queries

cquiroz:
There is a mismatch between the exposure time shown in the modes table and what we get from the odb.
The reason is that we are not providing roi nor ccd and the odb sets them depending on the mode an image quality

this PR replicates this logic in explore

### DIFF
--- a/model/shared/src/main/scala/queries/schemas/itc/syntax.scala
+++ b/model/shared/src/main/scala/queries/schemas/itc/syntax.scala
@@ -5,6 +5,7 @@ package queries.schemas.itc
 
 import cats.Hash
 import cats.syntax.all.*
+import eu.timepit.refined.types.numeric.PosDouble
 import explore.model.AsterismIds
 import explore.model.TargetList
 import explore.model.itc.ItcTarget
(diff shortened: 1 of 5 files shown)


model/shared/src/main/scala/queries/schemas/itc/syntax.scala
```diff
@@ -5,31 +5,65 @@ package queries.schemas.itc
...
 
 trait syntax:
 
   extension (row: InstrumentRow)
-    def toItcClientMode: Option[InstrumentMode] = row match {
+    def toItcClientMode(p: SourceProfile, iq: ImageQuality): Option[InstrumentMode] = row match {
       case g: GmosNorthSpectroscopyRow =>
-        val roi = g.modeOverrides.flatMap(_.roi)
-        val ccd = g.modeOverrides.flatMap(_.ccdMode)
+        val isFPU =
```
rpiaggio, line 33: Is this the same logic as in the ODB? If so, can we move it to a common lib? core?

rpiaggio (approved):
LGTM. Not sure if we are duplicating logic. If we are, I suggest we unify.

model/shared/src/main/scala/queries/schemas/itc/syntax.scala
cquiroz: Yes I added this as a method to core

rpiaggio: Nice, I was wondering if the could unify the override logic, but I see now that the overrides are defined in explore.

PR #3223, opened 2023-08-23
Dynamic enums

rpiaggio:
This PR implements "dynamic enums" by importing the file produced by the ODB for this purpose.

`main.jsx` now resolves the ODB REST URL from `environments.conf.json`, queries it for the dynamic enum metadata and sets it in a global variable so that Scala code can access it and build the `Enumerated` instance from static code.

Only `ObsAttachmentType` is dealt with in this PR.

Implementing this implied reorganizing the code somewhat since we need to reference `DynamicEnums` from the queries and `DynamicEnums` is JS-only code. Therefore, the queries were moved to the `explore` project, as well as the code in `common` which depended on them.


explore/src/main/webapp/main.jsx
```diff
@@ -29,15 +29,35 @@ if (import.meta.env.DEV) {
   process.env = { CATS_EFFECT_TRACING_MODE: 'none' };
 }
 
-Explore.runIOApp();
+// Get dynamic enums static data from ODB rest point and into a global variable.
+// ODB rest URL is resolved from environments.conf.json.
+fetch('/environments.conf.json').then((response) => {
```
hugo-vrijswijk, line 34: This file is available at build-time, right? We should be able to do `import environment from "/environments.conf.json"`, I think (or the similar in scala code), and Vite will resolve it, saving us a network hop.

vite.config.mjs
```diff
@@ -168,6 +166,10 @@ export default defineConfig(async ({ mode }) => {
           find: '@workers',
           replacement: workersSjs,
         },
+        {
+          find: '@odbRestURL',
+          replacement: 'hello',
```
hugo-vrijswijk, line 171: I assume this needs to be removed before merging?

toddburnside (approved):
Cool! LGTM after Hugo's comments are addressed. Thanks for doing this, it's great to have it out of the way.

explore/src/main/webapp/main.jsx
rpiaggio: Good idea! That should work, I'll implement it.

vite.config.mjs
rpiaggio: Yeah, I was testing things here but ended up not going that way.

explore/src/main/webapp/main.jsx
```diff
@@ -29,15 +29,35 @@ if (import.meta.env.DEV) {
...
-  ExplorePWA.runServiceWorker();
-}
+    const specificHostURL = getODBRestURLForHost(window.location.host);
+    const url = specificHostURL ? specificHostURL : getODBRestURLForHost('*');
 
-// Setting this here shouldn't be necessary if we get `vite-plugin-environment` to work.
+    // Suppress vite warning about dynamic imports it can't handle.
+    import(/* @vite-ignore */ `${url}/export/enumMetadata`).then((enumMetadataModule) => {
+      // Set it globally so it can be read in Scala code.
+      window.enumMetadataString = enumMetadataModule.enumMetadata;
```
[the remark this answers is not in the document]
rpiaggio: I considered that, but then we are back again in the problem we are trying to solve with this approach. We want to derive dependent decoders at compile time, for which the dynamic decoders must be statically reachable, we can't construct them with parameters once in Scala code. Maybe there's another way to bootstrap this that I'm missing.

PR #3233, opened 2023-08-30
Fix pointer events on FF

cquiroz:


common/src/main/webapp/sass/visualization.scss
```diff
@@ -138,12 +138,14 @@ $brightness: var(--aladin-image-brightness);
 
   .circle-target {
     display: none;
+    pointer-events: fill;
```
hugo-vrijswijk, line 141: What is the difference in this case between the default (`auto`/`visiblePainted`) and `fill`?

common/src/main/webapp/sass/tooltips.scss
```diff
@@ -42,7 +42,7 @@
 }
 
 .svg-target-tooltip-area {
-  pointer-events: bounding-box;
```
hugo-vrijswijk, line 45: I'm wondering if there's some sort of s?css linting we can set up that warns for non-standard usage

cquiroz: Yeah that would be useful

common/src/main/webapp/sass/visualization.scss
cquiroz: I tried several combinations that would work on FF and Chrome and `auto/visiblePainted` didn't. Couldn't find any documentation that would indicate why they are different

rpiaggio (approved):
LGTM

PR #3241, opened 2023-09-04
Recalculate the sequence when the target changes

cquiroz:
Changes in the target, e.g. SED will now trigger a request to recalculate the sequence

### DIFF
--- a/explore/src/clue/scala/queries/common/TargetQueriesGQL.scala
+++ b/explore/src/clue/scala/queries/common/TargetQueriesGQL.scala
@@ -92,7 +92,6 @@ object TargetQueriesGQL {
 
   @GraphQL
   trait TargetEditSubscription extends GraphQLOperation[ObservationDB] {
-    // We need to include the `value {id}` to avoid a bug in grackle.
     val document = """
       subscription($targetId: TargetId!) {
         targetEdit(input: {targetId: $targetId}) {
--- a/explore/src/clue/scala/queries/common/TargetQueriesGQL.scala
+++ b/explore/src/clue/scala/queries/common/TargetQueriesGQL.scala
@@ -91,15 +91,12 @@ object TargetQueriesGQL {
   }
 
   @GraphQL
-  trait ProgramTargetEditSubscription extends GraphQLOperation[ObservationDB] {
+  trait TargetEditSubscription extends GraphQLOperation[ObservationDB] {
     // We need to include the `value {id}` to avoid a bug in grackle.
     val document = """
-      subscription($programId: ProgramId!) {
-        targetEdit(input: {programId: $programId}) {
+      subscription($targetId: TargetId!) {
+        targetEdit(input: {targetId: $targetId}) {
           id
-          value {
-            id
-          }
         }
       }
     """
--- a/explore/src/main/scala/explore/config/sequence/GeneratedSequenceViewer.scala
+++ b/explore/src/main/scala/explore/config/sequence/GeneratedSequenceViewer.scala
@@ -4,6 +4,7 @@
 package explore.config.sequence
 
 import cats.effect.IO
+import cats.syntax.all.*
 import crystal.Pot
 import crystal.react.*
 import crystal.react.given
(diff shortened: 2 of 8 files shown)


explore/src/clue/scala/queries/common/TargetQueriesGQL.scala
```diff
@@ -91,15 +91,12 @@ object TargetQueriesGQL {
   }
 
   @GraphQL
-  trait ProgramTargetEditSubscription extends GraphQLOperation[ObservationDB] {
+  trait TargetEditSubscription extends GraphQLOperation[ObservationDB] {
     // We need to include the `value {id}` to avoid a bug in grackle.
```
rpiaggio, line 95: The comment that was in the code seems no longer to apply.

explore/src/main/scala/explore/config/sequence/GeneratedSequenceViewer.scala
```diff
@@ -38,21 +42,26 @@ object GeneratedSequenceViewer:
...
+
         SequenceQuery[IO]
           .query(props.obsId)
           .map(_.observation.map(_.execution.config))
           .attemptPot
           .resetOnResourceSignals(
-            ObsQueriesGQL.ObservationEditSubscription.subscribe[IO](props.obsId)
+            for {
+              o <- ObsQueriesGQL.ObservationEditSubscription.subscribe[IO](props.obsId)
+              t <- targetChanges
```
rpiaggio, line 59: It seems that we subscribe to all the changes in the observation and all the targets and rerequest the sequence in those cases.

Can we limit the sequence requests to only when relevant fields change?

explore/src/main/scala/explore/config/sequence/GeneratedSequenceViewer.scala
```diff
@@ -62,8 +71,14 @@ object GeneratedSequenceViewer:
               case InstrumentExecutionConfig.GmosSouth(config) =>
                 GmosSouthGeneratedSequenceTables(props.obsId, config)
             },
-            errorRender = _ => {
-              val msg = "No sequence avaialble, you may need to setup a configuration"
+            errorRender = m => {
+              val msg = m match {
+                case clue.ResponseException(errors, _)
+                    if errors.exists(_.message.startsWith("ITC returned")) =>
```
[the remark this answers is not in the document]
cquiroz: That would be a lot better but for that we need structured errors, see:
https://github.com/gemini-hlsw/gsp-graphql/issues/467

explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
```diff
@@ -68,7 +68,8 @@ case class SiderealTargetEditor(
   onClone:           TargetWithId => Callback = _ => Callback.empty,
   renderInTitle:     Option[Tile.RenderInTitle] = None,
   fullScreen:        View[AladinFullScreen],
-  globalPreferences: View[GlobalPreferences]
+  globalPreferences: View[GlobalPreferences],
+  sequenceChanged:   Callback = Callback.empty
```
rpiaggio, line 72: May I suggest changing this parameter name to `invalidateSequence`? It took me a bit to understand what it did.

rpiaggio (approved):
LGTM with a couple of comments

explore/src/main/scala/explore/config/sequence/GeneratedSequenceViewer.scala
rpiaggio: I was thinking something along the lines of having the relevant fields in the the stream of changes and analyzing if any of them changes before emitting a signal in the final stream.

This may be overkill though. I think the ODB caches the sequence, which wouldn't make it that costly to overfetch it.

explore/src/main/scala/explore/config/sequence/GeneratedSequenceViewer.scala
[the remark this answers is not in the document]
rpiaggio: Sounds good, we can revise once structured errors are available.

explore/src/main/scala/explore/config/sequence/GeneratedSequenceViewer.scala
cquiroz: right not every change will actually change the sequence, it would basically be the target brightness and profile. Not sure how doable is that with the front end only

PR #3248, opened 2023-09-06
Add control to add a new observation if none is present

cquiroz:
image


common/src/main/webapp/sass/explore.scss
```diff
@@ -2469,3 +2469,15 @@ a.explore-upgrade-link {
     min-width: 20ch;
   }
 }
+
+.observations-summary-table {
+  td.p-datatable-emptymessage {
+    background-color: var(--table-striped-background);
+
+    span {
```
rpiaggio, line 2477: Maybe we want a specific class for the button?

cquiroz: could be

PR #3250, opened 2023-09-06
Fix to handle pasting into multiple targets on the summary table

cquiroz:
Pasting an obs needs a different logic when the targets are selected on the sidebar or on the summary table


rpiaggio (approved):
LGTM

PR #3252, opened 2023-09-07
Fix plannedTime decoding error

hugo-vrijswijk:
Should hopefully be slightly easier to spot the problem with https://github.com/gemini-hlsw/lucuma-odb/pull/690


cquiroz (approved):
Thanks, LGTM

PR #3270, opened 2023-09-10
Fix scale calculation bug on tooltips

cquiroz:
This fixes a bug that could make some tooltips too small
It also removes some duplicated code making it common across different svg entities


hugo-vrijswijk (approved):
LGTM

PR #3271, opened 2023-09-11
Update node version to 20 in CI

hugo-vrijswijk:
Node 16 is [EOL](https://nodejs.dev/en/about/releases/).
Node 20 is the latest LTS version.

I swear I randomly checked the EOL date and it happened to be today 😅


rpiaggio (approved):
LGTM

PR #3273, opened 2023-09-11
Setup firebase deploy

armanbilge:
Sorry, I forgot exactly what we are going for 😅 currently this enables deploying to `explore-gemini-dev` whenever a PR merges to `master`. But let's adjust :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,16 @@ jobs:
           BUNDLEMON_PROJECT_APIKEY: ${{ secrets.BUNDLEMON_PROJECT_APIKEY }}
         run: yarn bundlemon
 
+      - name: Deploy staging app to Firebase
+        if: github.ref == 'refs/heads/master'
+        uses: FirebaseExtended/action-hosting-deploy@v0
+        with:
+          firebaseServiceAccount: ${{ secrets.FIREBASE_SERVICE_ACCOUNT_EXPLORE_GEMINI }}
+          projectId: explore-gemini
+          repoToken: ${{ secrets.GITHUB_TOKEN }}
+          target: dev
+          channelId: live
+
   heroku:
     name: Deploy to Heroku
     if: (github.event_name == 'push' && github.ref == 'refs/heads/master' && startsWith(github.repository, 'gemini'))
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -320,7 +320,7 @@ lazy val firebaseDeployReview = firebaseDeploy(
 
 lazy val firebaseDeployStaging = firebaseDeploy(
```
toddburnside, line 321: Should we rename this `firebaseDeployDev`?

toddburnside:
> Sorry, I forgot exactly what we are going for 😅 currently this enables deploying to `explore-gemini-dev` whenever a PR merges to `master`. But let's adjust :)

I think that is exactly what we want. We then have scripts to promote to staging and dev

toddburnside (approved):
LGTM

toddburnside:
Notes: 
- Before we promote to staging on firebase and have users start using the new url, we'll need to merge #3048.
- Before we can use production on firebase, we'll need to get IT to point `explore.gemini.edu` to firebase.

.github/workflows/ci.yml
```diff
@@ -138,6 +138,16 @@ jobs:
           BUNDLEMON_PROJECT_APIKEY: ${{ secrets.BUNDLEMON_PROJECT_APIKEY }}
         run: yarn bundlemon
 
+      - name: Deploy staging app to Firebase
+        if: github.ref == 'refs/heads/master'
```
toddburnside, line 142: Is this condition already checked on line 90? Also, why is the condition different than for the heroku deploy, which uses `github.event_name == 'push'` instead of `'pull_request'`?

toddburnside: "> On line 90 we are checking a more general form of this condition

Oops. I missed the `||` 😞

> I'm not sure, but semantically the difference is that `github.event_name == 'push'` will run for _any_ push to a branch, whether it's `master` or something else, and whether or not there is a PR open for that branch.

:D So, since we would `never` push directly to `master`, there is no difference. 🙈

armanbilge: > So, since we would `never` push directly to `master`, there is no difference. 🙈

Merging a PR to `master` counts as push event I believe.

PR #3274, opened 2023-09-11
Extract common code and styles

rpiaggio:


cquiroz (approved):
LGTM

rpiaggio:
I'm having trouble with `stylelint` since variables are now defined in scss externally.

The main issue seems to be that variables are defined only within the theme classes but we use them freely from outside. I attempted moving sites where we use them within a `.dark-theme, .light-theme` selector, but couldn't get it to work.

Defeated, I'm removing the check for undefined variables. I accept suggestions on how to reinstate them.

THEMING.md
```diff
@@ -90,7 +90,9 @@ As mentioned the theme is mostly based on `slate` but it has been further refine
 
 # Linting
 
-[Stylelint](https://stylelint.io/) has been setup to check for common errors on the css/less files. Note it is not run against the SUI files
+[Stylelint](https://stylelint.io/) has been setup to check for common errors on the css/less/sass file
+
+s. Note it is not run against the SUI files
 
 A tricky part of using `stylelint` is the way variables are checked.
 A small sed script can take the variables from either the dark or light theme and export them as a `vars.css` file.
```
toddburnside, line 98: Is this section true anymore? Hadn't you removed it earlier?

.stylelintrc.json
```diff
@@ -1,19 +1,10 @@
 {
   "extends": ["stylelint-config-standard-scss"],
-  "plugins": [
-    "@ronilaukkarinen/stylelint-value-no-unknown-custom-properties",
```
toddburnside, line 4: This could be removed from package.json, too.

common/src/main/webapp/sass/explore-primereact.scss
```diff
@@ -12,21 +12,24 @@
 @import '/lucuma-css/lucuma-ui-prime';
 /* stylelint-enable no-invalid-position-at-import-rule */
 
-// specific to Explore - may be able to remove this when we're fully primereact - may have to apply to tiles or something
-.constraints-tile .explore-tile-body,
-.explore-configuration-grid {
-  background-color: var(--surface-card);
-}
+.dark-theme,
+.light-theme {
```
toddburnside, line 16: So these had no effect?

toddburnside (approved):
I seem to remember when we first moved some of the css to lucuma-ui linting could still work because the CSS had already been unpackaged from the lucuma-ui jar.  I think I just had to change some paths somewhere. Is it possible the light and dark variable files are somewhere where you could reach them to still create the vars.css file? It not, I have no ideas.

rpiaggio:
> I seem to remember when we first moved some of the css to lucuma-ui linting could still work in CI because the CSS had already been unpackaged from the lucuma-ui jar. I think I just had to change some paths somewhere. Is it possible the light and dark variable files are somewhere where you could reach them to still create the vars.css file? It not, I have no ideas.

Yes the files are reachable in the `lucuma-css` directory, but the issue seems to be the switch to `sass`. I guess we could revert them to `less`.

common/src/main/webapp/sass/explore-primereact.scss
rpiaggio: Unfortunately, not. I still left those on purpose which seemed correct.

.stylelintrc.json
rpiaggio: Noted.

THEMING.md
rpiaggio: Ah yes, but when I attempted to make linting work again I undeleted it. I will edit.

rpiaggio:
> > I seem to remember when we first moved some of the css to lucuma-ui linting could still work in CI because the CSS had already been unpackaged from the lucuma-ui jar. I think I just had to change some paths somewhere. Is it possible the light and dark variable files are somewhere where you could reach them to still create the vars.css file? It not, I have no ideas.
> 
> Yes the files are reachable in the `lucuma-css` directory, but the issue seems to be the switch to `sass`. I guess we could revert them to `less`.

We could also try to reproduce the previous behaviour, where, if I understood correctly, we used `sed` to remove all lines that didn't declare a variable and import that as CSS. I will try that.

rpiaggio:
The check is reinstated, in the same way as before.

PR #3275, opened 2023-09-11
Extend the staging banner to all non-production instances

toddburnside:
Because we now have a dev instance on firebase, I thought we should have a banner to differentiate it, too.


cquiroz (approved):
LGTM

PR #3285, opened 2023-09-12
Show S/N on the sequence table per observation class

cquiroz:
S/N is obtained from the top level query and not part of each step model. Thus we need to pass it through to get it to the sequence table

Explore 🔊 2023-09-12 15-26-25


explore/src/main/scala/explore/config/sequence/GmosSequenceTable.scala
```diff
@@ -48,19 +59,31 @@ private sealed trait GmosSequenceTableBuilder[D: Eq]:
...
-    ) +: SequenceColumns.gmosColumns(ColDef, _.step.some, _.index.some)
+    ) +: SequenceColumns.gmosColumns(
+      ColDef,
+      _.step.some,
+      _.index.some,
+      r => {
+        val showScience = obsClass === ObserveClass.Science && r.flatMap(_.science).isDefined
+        val showAcq     = obsClass === ObserveClass.Acquisition && r.flatMap(_.fpuName).isEmpty
+        sn.filter(_ => showScience || showAcq)
+      }
```
rpiaggio, line 80: We need this logic in observe too. Can we place it in `lucuma-ui`?

cquiroz: I'm not sure we need to show the S/N column in observe, do we?

rpiaggio: It appears in the inception document: 

image

rpiaggio (approved):
LGTM

PR #3289, opened 2023-09-14
Fix the way AllowFlip behaves and some cleanups

cquiroz:


explore/src/main/scala/explore/targeteditor/package.scala
```diff
@@ -16,3 +20,15 @@ extension (fov: Fov)
 
 object AddDisabled extends NewType[Boolean]
 type AddDisabled = AddDisabled.Type
+
+def domRoot: Option[HTMLElement] =
```
rpiaggio, line 24: I think this can be a `val`.

rpiaggio (approved):
LGTM

PR #3301, opened 2023-09-21
Use common exposure time for all targets when calculating the itc chart

cquiroz:
This is relevant for asterisms such that graphs use the same exposure time even though it will produce different S/N results


toddburnside (approved):
LGTM

PR #3302, opened 2023-09-23
Fix negative times

cquiroz:


explore/src/main/scala/explore/targeteditor/ElevationPlotNight.scala
```diff
@@ -210,7 +210,10 @@ object ElevationPlotNight {
             case TimeDisplay.UT       => timezoneInstantFormat(instant, ZoneOffset.UTC)
             case TimeDisplay.Sidereal =>
               val skycalc = ImprovedSkyCalc(props.site.place)
-              val sid     = skycalc.getSiderealTime(instant)
+              val sid     = {
+                val sid = skycalc.getSiderealTime(instant)
+                if (sid < 0) sid + 24 else sid
```
toddburnside, line 215: Is sid constrained to +/- 24?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/ElevationPlotNight.scala
cquiroz: Yes, it is not very clearly stated on the type :D

PR #3311, opened 2023-09-25
Update browserslist configuration

hugo-vrijswijk:
Browserslist is used to generate browser compatibility polyfills for JS and CSS.
The previous configuration was very wide and [included browsers like IE10](https://browsersl.ist/#q=supports+es6-module%0Alast+2+versions%0AFirefox+ESR). This splits the configuration into 2 parts:
- Production: [defaults](https://browsersl.ist/#q=defaults) splits into `> 0.5%, last 2 versions, Firefox ESR, not dead`. We include `supports es6-module` into it. This uses the sane default that doesn't try to support very old browsers.
- Development: We want working code, with as little transpilation as possible for speed. So this just targets the last version of each development browser.

Also fixes the stylelint plugin not working in VS Code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,9 +67,14 @@
     "build": "node --unhandled-rejections=strict build.mjs",
     "serve": "vite preview"
   },
-  "browserslist": [
-    "supports es6-module",
-    "last 2 versions",
-    "Firefox ESR"
-  ]
+  "browserslist": {
+    "production": [
+      "defaults and supports es6-module"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -67,9 +67,14 @@
...
     "serve": "vite preview"
   },
-  "browserslist": [
-    "supports es6-module",
-    "last 2 versions",
-    "Firefox ESR"
-  ]
+  "browserslist": {
+    "production": [
+      "defaults and supports es6-module"
```
cquiroz, line 72: We may want to keep Firefox ESR. That is used on the operational consoles at the telescope

hugo-vrijswijk: Yes, `defaults` includes Firefox ESR

cquiroz (approved):
LGTM

PR #3318, opened 2023-09-28
Use `fs2.dom.BroadcastChannel`

armanbilge:


build.sbt
```diff
@@ -111,7 +112,6 @@ lazy val common = project
   .settings(
     libraryDependencies ++=
       LucumaSSO.value ++
-        LucumaBC.value ++
```
armanbilge, line 114: We also need to remove broadcast-channel from `package.json` but I currently couldn't do anything with `npm` while FA is borked. Maybe someone with it cached can run `npm uninstall broadcast-channel`.

explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -127,7 +127,7 @@ object ExploreLayout:
                       .toAsync
                 case _                        => IO.unit
               }
-          ): (ExploreEvent => IO[Unit])
+          ).compile.drain.unsafeRunAndForget()(IORuntime.global)
```
armanbilge, line 130: This is not great, but apparently lucuma-broadcast-channel was also just using `IORuntime.global` for its callback.

In any case, I am fairly certain there should be no memory leak on unmount, because the only reference to this fiber would be via the callback, so long as the callback holder is GCed the this fiber can be as well.

If there is some way to attach the lifecycle of a `Resource[IO, Unit]` to the lifecycle of a component then that's ideally what we should do.

toddburnside: `crystal` has some useful hooks for `useStream*` and 'useResource*`

toddburnside: It seems like here we probably want `useStreamOnMountBy`.

rpiaggio: Yup, that should do the trick. You will have an extra useless parameter in the next hooks/render though. I wonder if we want a `useEvalStream` (or something like that) for this kind of situations.

toddburnside: Sure. We do the same think in `CacheComponent`. I'll work with what I have for now.

explore/src/main/scala/explore/ExplorePWA.scala
toddburnside, line 1: This file should probably be thoroughly reviewed since it was changed quite a bit. I tested it and the service worker update works, but that doesn't mean it is entirely correct.

explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -96,45 +95,40 @@ object ExploreLayout:
...
+                        ExploreStyles.UpgradeLink,
+                        "Explore",
+                        ^.href   := s"https://github.com/gemini-hlsw/explore/compare/${utils.gitHash.orEmpty}...HEAD",
+                        ^.target := "_blank"
+                      ),
+                      " is available!"
+                    ),
+                    ctx.broadcastChannel
+                      .postMessage(ExploreEvent.PWAReload)
+                      .handleErrorWith(e => IO(e.printStackTrace()))
```
armanbilge, line 120: I think the `IORuntime` will already be printing any unhandled errors to the console. (I think this was previously bugged). Not a big deal tho :)

explore/src/main/scala/explore/ExplorePWA.scala
```diff
@@ -75,88 +82,83 @@ object ExplorePWA {
...
-      .setIntervalMs(intervalMS.toDouble)
-      .void
+  val updateInteval = FiniteDuration(10, TimeUnit.MINUTES)
+
+  def scheduleUpdateCheck(u: String, r: ServiceWorkerRegistration): IO[Unit] =
+    Stream
+      .awakeEvery[IO](updateInteval)
+      .evalTap { _ =>
+        IO.println("Check for new explore version") >>
+          pingSW(u).flatMap(go => if (go) IO.blocking(r.update()) else IO.unit)
```
armanbilge, line 92: `blocking` doesn't do anything on JS, because there is no blocking threadpool to move the task to. Anyway, almost nothing is blocking in JS.

```suggestion
          pingSW(u).flatMap(go => if (go) IO(r.update()) else IO.unit)
```

armanbilge:
```suggestion
          pingSW(u).ifM(IO(r.update()), IO.unit)
```

explore/src/main/scala/explore/ExplorePWA.scala
```diff
@@ -75,88 +82,83 @@ object ExplorePWA {
...
+            (IO.println(
+              s"Service worker registered, setup self update task, offline: $isOffline"
+            ) *>
+              pingSW(u)
+                .flatMap(go =>
+                  // initial check
+                  if (go) IO.sleep(Duration(2, TimeUnit.SECONDS)) >> IO.blocking(r.update())
+                  else IO.unit
+                )
+                .attempt *>
```
armanbilge, line 137:
```suggestion
                .voidError *>
```

explore/src/main/scala/explore/ExplorePWA.scala
```diff
@@ -75,88 +82,83 @@ object ExplorePWA {
...
-                  )
-                  .toCallback *>          // Inital check
-                scheduleUpdateCheck(u, r) // Periodic checks
-          )
-        )
-    }
+  def setupSW: IO[Unit] =
+    (for {
+      bc <- BroadcastChannel[IO, ExploreEvent]("explore")
+      _  <- Resource.eval(handleEvents(bc).start)
```
armanbilge, line 157:
```suggestion
      _  <- handleEvents(bc).background
```

explore/src/main/scala/explore/ExplorePWA.scala
```diff
@@ -75,88 +82,83 @@ object ExplorePWA {
...
+    } yield ()).useForever
 
   @JSExport
   def runServiceWorker(): Unit =
-    setupSW().handleError(t => Callback(t.printStackTrace())).runNow()
+    setupSW.handleError(t => IO(t.printStackTrace())).unsafeRunAndForget()
```
armanbilge, line 162: Same here, errors should be logged automatically.

```suggestion
    setupSW.unsafeRunAndForget()
```

explore/src/main/scala/explore/ExplorePWA.scala
```diff
@@ -75,88 +82,83 @@ object ExplorePWA {
...
+            val isOffline =
+              !(!(r.installing && !js.isUndefined(org.scalajs.dom.window.navigator))) ||
+                (dom.window.navigator.hasOwnProperty("connection") && dom.window.navigator.onLine)
+            (IO.println(
+              s"Service worker registered, setup self update task, offline: $isOffline"
+            ) *>
+              pingSW(u)
+                .flatMap(go =>
+                  // initial check
+                  if (go) IO.sleep(Duration(2, TimeUnit.SECONDS)) >> IO.blocking(r.update())
```
armanbilge, line 134:
```suggestion
                  if (go) IO.sleep(Duration(2, TimeUnit.SECONDS)) >> IO(r.update())
```

explore/src/main/scala/explore/ExplorePWA.scala
cquiroz, line 92: What does it mean `blocking` in a browser?

toddburnside: Nothing. I wasn't thinking. 😆

toddburnside: Good point. I clearly wasn't paying attention.

explore/src/main/scala/explore/ExplorePWA.scala
toddburnside: Cool. I didn't know that was a thing. 😄

cquiroz (approved):
LGTM

PR #3319, opened 2023-09-29
Remove deprecated field on ObservationEditSubscription

hugo-vrijswijk:
```
  id: Long! @deprecated(reason: "id is no longer computed; a constant value is returned")
```


rpiaggio (approved):
LGTM

PR #3328, opened 2023-10-01
Switch to Unfonts

cquiroz:
We are using a deprecated vite plugin to get fonts links
This PR uses the plugin that replaced the former one and also will now provide the font locally. This will make the app more resilient


rpiaggio (approved):
LGTM, thanks

PR #3336, opened 2023-10-03
Update lucuma-core 0.86.1 with new GMOS Binning

toddburnside:


vite.config.mjs
```diff
@@ -106,7 +106,6 @@ export default defineConfig(async ({ mode }) => {
   const workersSjs = isProduction
     ? path.resolve(workersScalaClassesDir, 'workers-opt')
     : path.resolve(workersScalaClassesDir, 'workers-fastopt');
-  const rollupPlugins = isProduction ? [] : [visualizer()];
```
toddburnside, line 109: unrelated, but visualizer was removed recently, and `rollupPlugins` isn't used anywhere.

model/shared/src/main/scala/queries/schemas/itc/syntax.scala
```diff
@@ -38,8 +39,8 @@ trait syntax:
           .orElse(
             GmosCcdMode(
               if (isFPU) GmosXBinning.One
-              else xbinNorth(g.fpu, p, iq, PosDouble.unsafeFrom(2.0)),
-              if (isFPU) GmosYBinning.One else DefaultYBinning,
```
toddburnside, line 42: `DefaultBinning` is gone, but it used to be an alias for `GMosYBinning.Two`. I don't know if this is still correct.

cquiroz: hmm this is tricky, we may need to replicate the calculation done for each mode:
https://github.com/gemini-hlsw/lucuma-core/pull/810

vite.config.mjs
cquiroz: 👍

model/shared/src/main/scala/queries/schemas/itc/syntax.scala
toddburnside: Would that mean calling one of the `*spatialBinning` functions? We would need a `detector` (or `pixelSize`) or default to `Hamamatsu`.

cquiroz: we default to `Hamamatsu`

cquiroz: I'm not fully aware of the details of the calculation, perhaps @swalker2m could enlighten us

swalker2m: Each will provide its own binning calculation implemented to call the more generic `gmos.binning` method as appropriate.  So we wouldn't need to call any `*spatialBinning` directly.  In this case, we'd use `lucuma.core.model.sequence.gmos.longslit.northBinning`.  It defaults to `Hamamatsu` and defaults the sampling to the new preferred value of `2.5`.

So something like `northBinning(g.fpu, p, iq, g.grating)`.

Does that make sense?

toddburnside: Yep. And that will give us both the X and Y binning. Thanks!

toddburnside: Oh, one other thing: Do I not need the `isFPU` anymore? Or is that still necessary?

toddburnside: Thanks. I'll push an update in a couple of minutes.

swalker2m: One other thing.  The `isIFU` check is already defined in `GmosNorthFpu` and `GmosSouthFpu` so that definition can be deleted and then you can call `g.isIFU` directly.

toddburnside: Cool. Pushed that change, too. 😄

swalker2m (approved):
Looks great!  :+1:

PR #3337, opened 2023-10-04
Program tab details tile

hugo-vrijswijk:


toddburnside:
@hugo-vrijswijk Did you discuss the inclusion of the minimum calculated execution time on the proposal with someone like Andy? If not, I think there might be a misunderstanding of the minimum time for the proposal. This is one of the many unfortunate cases where the naming is confusing.  It has nothing to do with the minimum execution time calculated by the ODB, which is the minimum execution time based on which and/or groups are actually observed. On the proposal tab, the user enters the minimum percent of the calculated execution time that is acceptable for them. The ITC doesn’t always grant as much time as the user would ideally like. So, on the proposal tab, the minimum time displayed is the calculated execution time times the minimum percent.

For the execution time on the proposal time, I am currently using the max calculated time. I am trying to get a discussion with Andy about how he wants to handle the fact that, with and/or groups, we no longer have a definitive execution time for the program to put on the proposal tab. It seems to me that the minimum acceptable time is something that could be handled by OR groups instead of entering a percent on the proposal. But, then we'd probably want to show the ranges on the proposal, not just the maximum. Or, maybe just show the minimum calculated execution time where you have it (but without multiplying by a percent).

toddburnside (approved):
LGTM

PR #3346, opened 2023-10-06
Handle group update subscriptions

hugo-vrijswijk:
Moved all modify functions to CacheModifierUpdaters.scala (naming suggestions are welcome). If the groups subscription does end up triggering events for old/new parents part of this can be removed.

### DIFF
--- a/explore/src/main/scala/explore/cache/CacheModifierUpdaters.scala
+++ b/explore/src/main/scala/explore/cache/CacheModifierUpdaters.scala
@@ -194,8 +194,6 @@ trait CacheModifierUpdaters {
         }
         // Moved to root group or different group
         else if (groupId.isEmpty || groupId.isDefined && !groupId.contains(grouping.id)) {
-          pprint.pprintln(groupId)
-          pprint.pprintln(grouping.id)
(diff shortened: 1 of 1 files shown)


cquiroz:
Do we plan to get this merged or will the backend get fixed?

hugo-vrijswijk:
> Do we plan to get this merged or will the backend get fixed?

The backend is planning to get fixed to give events for parent nodes. That means we won't have to do most of the mapping. 

If we want to do optimistic updates, like when drag'n'dropping nodes and updating the list before the backend event comes through, we'll still need most of this code for updating the elements list (which I think we do want)

toddburnside:
> If we want to do optimistic updates, like when drag'n'dropping nodes and updating the list before the backend event comes through, we'll still need most of this code for updating the elements list (which I think we do want)

Yeah, we definitely want optimisitic updates. Otherwise the UI experience is a bit painful. Your plan is to move some of this code around once the backend is fixed to use it for optimistic updates?

toddburnside (approved):
If there is a plan for the backend fix and optimistic updates, LGTM.

PR #3349, opened 2023-10-09
Setup custom cache step for `node_modules`

armanbilge:
Based on https://github.com/actions/setup-node/issues/835#issuecomment-1753052021. Seems to be working for them to stay under the FA bandwidth limit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,13 +52,22 @@ jobs:
         if: matrix.java == 'temurin@17' && steps.setup-java-temurin-17.outputs.cache-hit == 'false'
         run: sbt -v -J-Xmx6g +update
 
-      - name: Use Node.js
+      - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
           node-version: 20
           cache: npm
 
+      - name: Cache node_modules
+        id: cache-node_modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: ${{ hashFiles('package.json') }}
+          restore-keys: ${{ hashFiles('package.json') }}
+
       - name: npm clean-install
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -52,13 +52,22 @@ jobs:
...
         with:
           node-version: 20
           cache: npm
 
+      - name: Cache node_modules
+        id: cache-node_modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: ${{ hashFiles('package.json') }}
```
rpiaggio, line 66: Shall we switch the key to `package-lock.json`? The actual version of a package may change even `package.json` doesn't, because of the `~version` notation.

Also the original solution seems to include node version in the key, but I'm not sure that's actually necessary.

armanbilge: Ack, good catch, this was supposed to be `package-lock.json`, I was being hasty 😢 

The official action is using `${{ runner.os }}-npm-${{ hashFiles('<package-lock-path>') }}`, but that's not caching `node_modules/`. I feel like the Node version is not important, but the npm version is 🤔 
https://github.com/actions/setup-node/blob/2a017f350dbf6c4b6bb4508cc83809719115162e/docs/adrs/0000-caching-dependencies.md?plain=1#L29

toddburnside: So, we'll still probably download FA anytime any npm package changes, but not on every build?

rpiaggio (approved):
LGTM

.github/workflows/ci.yml
armanbilge: Hmm, actually with some adjustments to the `restore-key` I think we can even use an existing cache as a fallback when `package-lock.json` changes, hopefully to avoid downloading FA.

PR #3350, opened 2023-10-10
update libs: Adds duration to ave parallactic angle calculation

toddburnside:
For now, a duration time of 1 hour is being used. This is what was hard-coded in lucuma-core before. When we have the observation duration available, we probably want to switch to that - maybe with a maximum duration because not all observations can actually be scheduled in one night.


cquiroz (approved):
LGTM

PR #3361, opened 2023-10-12
Implement drag'n'drop for observation list

hugo-vrijswijk:
Depends on https://github.com/gemini-hlsw/lucuma-react/pull/628 and https://github.com/gemini-hlsw/explore/pull/3346


explore/src/main/scala/explore/common/GroupQueries.scala
```diff
@@ -0,0 +1,39 @@
...
+
+object GroupQueries:
+
+  /**
+   * @param groupId
+   *   Group to move
+   * @param parentGroup
+   *   New parent group, `None` to move to top level
+   * @param parentGroupIndex
+   *   New index in parent group, `None` to leave position unchanged
```
toddburnside, line 25: What does it mean to leave the index `unchanged` if it is a new parent group?

toddburnside (approved):
LGTM. Hopefully Prime React will release a new version soon. Andy will be very excited. 😄

explore/src/main/scala/explore/common/GroupQueries.scala
hugo-vrijswijk: 'Unchanged' is definitely the wrong word. It means "let the odb figure out the index". Though in reality this is always filled (primereact-tree gives always gives the index)

hugo-vrijswijk:
> LGTM. Hopefully Prime React will release a new version soon. Andy will be very excited. 😄

There is a new release. But the CI disagrees 🙃

toddburnside:
> There is a new release. But the CI disagrees 🙃

Dang computers.

explore/src/main/scala/explore/observationtree/obsTree.scala
```diff
@@ -8,40 +8,59 @@ import cats.Order.*
...
+import explore.data.tree.{Tree => IndexTree}
+import explore.model.*
+import explore.model.syntax.all.*
+import lucuma.core.model.Group
 import lucuma.core.model.Observation
 import lucuma.react.primereact.Tree
 import lucuma.react.primereact.Tree.Node
+import monocle.Iso
 
 enum ObsNode derives Eq:
```
cquiroz, line 24: Why an `enum`? It seems this would be more an `ADT`

cquiroz (approved):
Thanks, LGTM

explore/src/main/scala/explore/observationtree/obsTree.scala
hugo-vrijswijk: Aren't enums with params basically ADT's?

toddburnside: Yes

PR #3362, opened 2023-10-13
Allow logout from the Programs popup.

toddburnside:
This PR adds a `Logout` button the the Programs popup if you reached the popup due to a missing or invalid program id in the url. 
image
It also moves the popup code for missing or invalid programs out of the routing code and into ExploreLayout.


cquiroz (approved):
LGTM

PR #3366, opened 2023-10-19
Handle errors in ProgramCache and DynamicEnums initialization

rpiaggio:
Also, remove `downField("data")` when parsing dynamic enums.


toddburnside (approved):
LGTM - Thanks for finding and fixing this!

PR #3379, opened 2023-10-26
Downgrade remark-gfm

cquiroz:
there are some incompatibilities gfm 4.0 is not compatible with react-markdown 8, you'd need react-markdown 9 but that in turns requires react 8

The combination in this PR seems to work just fine


toddburnside (approved):
LGTM

PR #3386, opened 2023-10-30
Add reduceWithin stream combinator and group subscription updates

hugo-vrijswijk:


toddburnside:
@armanbilge, as our fs2 expert, would you be willing to review this. 😃

common/src/main/scala/explore/utils/package.scala
```diff
@@ -170,3 +176,66 @@ extension (bytes: NonNegLong)
...
+   *
+   * When an item is received in the stream, a timer is started. Within that time, items are
+   * combined until the timer expires, at which point the combined item is emitted.
+   *
+   * If no items are received within the given duration, no item is emitted.
+   */
+  def reduceWithin(d: FiniteDuration, f: (O, O) => O)(using F: Temporal[F]): Stream[F, O] =
+    given Semigroup[O] with
+      def combine(x: O, y: O): O = f(x, y)
+    reduceSemigroupWithin(d)
```
armanbilge, line 192: Nit, but I don't think it makes sense to express the implementations in this direction. Not every "reducer" `f` is a lawful `Semigroup`.

armanbilge (commented):
Implementation seems reasonable to me 😄 any interest in writing some tests? Shouldn't be too bad thanks to time-mocking in the Cats Effect test runtime, that lets you make very precise assertions about temporal behavior while still running effectively instantly.

common/src/main/scala/explore/utils/package.scala
```diff
@@ -170,3 +176,66 @@ extension (bytes: NonNegLong)
...
+
+  /**
+   * Combine items within a given duration using Semigroup.
+   *
+   * When an item is received in the stream, a timer is started. Within that time, items are
+   * combined until the timer expires, at which point the combined item is emitted.
+   *
+   * If no items are received within the given duration, no item is emitted.
+   */
+  def reduceSemigroupWithin(
```
rpiaggio, line 202: Sorry I didn't get around to reviewing earlier.

Can the implementation be simplified by using `groupWithin`?

hugo-vrijswijk: Unfortunately no, the implementation of groupWithin always immediately outputs 1 element after the given duration has passed, instead of waiting for more, and it's not customizable

common/src/main/scala/explore/utils/package.scala
hugo-vrijswijk: Hmm, okay. It's mostly this way because I wanted the syntax of `combineAllOption`, which I would then have to write out manually (unless you have a suggestion for that)

hugo-vrijswijk:
@armanbilge  could you have a look at the tests? They work, but this is the first time I've used cats-effect-testkit

common/src/test/scala/explore/utils/StreamSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+import fs2.Stream
+import munit.Location
+
+import scala.concurrent.duration.*
+
+class StreamSpec extends munit.CatsEffectSuite:
+
+  // Stream that emits 1 every 100ms, sleeps for 1 second, and repeats
+  val stream =
+    (Stream.awakeEvery[IO](50.millis).as(1).take(4) ++ Stream.sleep_[IO](100.millis)).repeat
```
armanbilge, line 18: Right?
```suggestion
  // Stream that emits 1 every 50ms, sleeps for 100ms, and repeats
  val stream =
    (Stream.awakeEvery[IO](50.millis).as(1).take(4) ++ Stream.sleep_[IO](100.millis)).repeat
```

common/src/test/scala/explore/utils/StreamSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import cats.effect
+import cats.effect.IO
+import cats.effect.testkit.TestControl
+import fs2.Stream
+import munit.Location
+
+import scala.concurrent.duration.*
+
+class StreamSpec extends munit.CatsEffectSuite:
```
armanbilge, line 14: Sorry if I missed it, is there a test here that asserts the behavioral difference from `groupWithin`? Something to guard against future regression in case somebody reading the code thinks they can simplify it.

hugo-vrijswijk: I _think_ with groupWithin at least one of the tests would fail (it would emit more elements). But I'll confirm it tomorrow

hugo-vrijswijk: Changing the implementation to groupWithin does make at least 1 test fail, but I added another just to make it explicit

common/src/main/scala/explore/utils/package.scala
```diff
@@ -170,3 +176,81 @@ extension (bytes: NonNegLong)
...
+   *
+   * When an item is received in the stream, a timer is started. Within that time, items are
+   * combined until the timer expires, at which point the combined item is emitted.
+   *
+   * If no items are received within the given duration, no item is emitted.
+   *
+   * This might seem similar to `groupWithin`. The subtle difference is that in-between durations
+   * `groupWithin` enters a "timed out" state, where the next element will immediately be emitted,
+   * and only elements _after_ that will be grouped. This is not the case for `reduceWithin`, which
+   * will always wait for the timeout to expire before emitting the combined element.
```
armanbilge, line 206: Thanks for adding the comment, sorry for being dumb, but I'm still struggling to see which of your tests specifically asserts this behavior.

armanbilge (approved):
Thanks for updating the comments and tests 👍

common/src/main/scala/explore/utils/package.scala
hugo-vrijswijk: I've added a test that explicitly tests different behaviour. But another test already fails if I replace the implementation with groupWithin:

```
==> X explore.utils.StreamSpec.reduceWithin only combines within given time window 1.05s munit.ComparisonFailException: /Users/hvanrijswijk1/ws/explore/common/src/test/scala/explore/utils/StreamSpec.scala:32
31:
32:    TestControl.executeEmbed(program).assertEquals(Vector(4, 4, 4, 4, 4))
33:  }
values are not the same
=> Obtained
Vector(
  4,
  3,
  3,
  3,
  3
)
=> Diff (- obtained, + expected)
   4,
-  3,
-  3,
-  3,
-  3
+  4,
+  4,
+  4,
+  4
 )
```

PR #3442, opened 2023-11-27
Revert "Try to config fortawesome registry"

armanbilge:
This _did_ fix dependabot updates for FontAwesome. But also I think it caused us to go over our quota 🤔 hard to be certain. So let's revert for now.


toddburnside (approved):
It's worth a try. LGTM.

PR #3468, opened 2023-12-05
Fixes the `toHours` conversion from TimeSpan

toddburnside:
Super simple PR that fixes a bug that crept in during the switch from `NonNegDuration` to `TimeSpan`.


swalker2m (approved):
Oops.  :+1:

PR #3469, opened 2023-12-07
Move dynamic enum logic to lucuma-schemas

rpiaggio:


vite.config.mjs
```diff
@@ -205,6 +205,7 @@ export default defineConfig(async ({ mode }) => {
       host: '0.0.0.0',
       port: 8080,
       https: true,
+      cors: { origin: "*" },
```
rpiaggio, line 208: Unrelated. Useful for connecting to servers on other domains from dev.

toddburnside (approved):
LGTM

PR #3506, opened 2024-01-04
Improve the way tile resizing is enabled/disabled

cquiroz:
on layouts with multiple tiles it may take a while to maximize/minimize a tile and you can (if you click multiple times) end on a weird loop while minimizing and maximizing

This PR adds a state so the buttons are disabled until the size change gets completed


hugo-vrijswijk (approved):
LGTM

PR #3516, opened 2024-01-11
Migrate ExecutionEnvironment to core

rpiaggio:


.vscode/settings.json
```diff
@@ -1,13 +0,0 @@
-{
-  "files.watcherExclude": {
-    "**/target": true
-  },
-  "[scss]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode",
-    "editor.codeActionsOnSave": {
-      "source.fixAll.stylelint": true
```
hugo-vrijswijk, line 8: Personally I think these rules are useful to have committed to align editor settings across usage. It means you don't have to configure anything to get stylelint or prettier formatting for the scss files

toddburnside: Since 75% of the developers on this project use vscode, it does seem like it has some merit to have the editor setup to mirror the styling required by CI. I guess there could be issues if someone wants some other custom setting for this project.
 
It seems like we should at least keep the recommended extensions. I think that file is meant to be kept with a project.

rpiaggio: Ok, I'll revert.

Due to the fact that we cannot put comments here and my short-lived memory, we will probably have this conversation again in a couple of months.

hugo-vrijswijk: These json files actually do support comments

toddburnside (approved):
LGTM

PR #3524, opened 2024-01-15
Downgrade primereact to 10.2.1

toddburnside:
Downgrade primereact because the `TieredMenu` that we use in the popup menu in the upper right of explore is broken in 10.3.*. A fix has been merged and will be released in `10.4.*`. At that point, we will also need to switch our lucuma themes to the ones for the new `lucuma-primereact-sass-theme` repo due to changes in the the theming for (at least) the menus. These themes are brought in via `lucuma-ui`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "highcharts": "11.2.0",
     "loglevel": "1.8.1",
     "primeicons": "^6.0.1",
-    "primereact": "^10.2.1",
+    "primereact": "~10.2.1",
     "prop-types": "15.8.1",
     "react": "17.0.2",
     "react-beautiful-dnd": "13.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "highcharts": "11.2.0",
     "loglevel": "1.8.1",
     "primeicons": "^6.0.1",
-    "primereact": "^10.3.0",
+    "primereact": "^10.2.1",
```
hugo-vrijswijk, line 20: If we want to pin it to 10.2 this should be `~10.2.1` (`~` means in the range of the minor version, `^` is in the range of the major version, so a lockfile update could update it)

toddburnside: Ah, right. I always forget how those work.

hugo-vrijswijk (approved):

PR #3531, opened 2024-01-19
Fix tile controller when `storeLayout` is false

toddburnside:
I'm not sure exactly why the tilecontroller waits for a update from user preferences before saying it isn't resizing. However, if `storeLayout` is false, that means it always thinks it is resizing after any minimizing or mazimizing. This was making it very hard to test out layouts before going to the effort of adding the layouts to the user preferences.


cquiroz (approved):
LGTM

PR #3538, opened 2024-01-23
Button for deleting grid layouts from user preferences

toddburnside:
I was just going to make the button available in development for helping to test default layouts, but then I thought it could be handy for users if they really mess up all the tiles on, for example, the observation tab. 

I thought the subscription would reset the tiles, but it seems to need to have a reload to take effect. Maybe I should display a toast or something to that effect when the reset is complete?

Hint: Hiding whitespace changes will make this much easier to review. 😄

### DIFF
--- a/explore/src/main/scala/explore/users/UserPreferencesPopup.scala
+++ b/explore/src/main/scala/explore/users/UserPreferencesPopup.scala
@@ -11,6 +11,7 @@ import crystal.react.*
 import crystal.react.hooks.*
 import explore.DefaultErrorPolicy
 import explore.Icons
+import explore.common.UserPreferencesQueries.GridLayouts
 import explore.components.ui.ExploreStyles
 import explore.model.ApiKey
 import explore.model.AppContext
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/users/UserPreferencesPopup.scala
```diff
@@ -185,91 +187,122 @@ object UserPreferencesContent:
...
+              loading = isCleaningTheCache.value.value,
+              icon = Icons.Trash,
+              onClick = requestCacheClean.runAsyncAndForget
+            ).small.compact
 
-        val unsupportedRoles = Enumerated[RoleType].all.filterNot { rt =>
-          allRoles.exists(r => r.`type` === rt)
-        }
-        val currentKeyRoleId = allRoles.find(_.`type` === newRoleType.get).map(_.id)
+          val deleteGridLayoutsButton =
```
cquiroz, line 231: I wonder if it shoud be dev only. OTOH I can imagine more people wanting to reset their layouts

cquiroz (approved):
LGTM

PR #3541, opened 2024-01-29
Use program time estimates instead of total

hugo-vrijswijk:
Everywhere time estimates are shown the  time should be used, and not include partner or nonCharged


hugo-vrijswijk:
Closed in favour of Todd's PR

PR #3550, opened 2024-01-29
Bump to react 18

rpiaggio:


explore/src/main/scala/explore/Explore.scala
```diff
@@ -172,8 +170,11 @@ object ExploreMain {
         _                    <- setupReusabilityOverlay(appConfig.environment)
         r                    <- (ctx.sso.whoami, setupDOM[IO], showEnvironment[IO](appConfig.environment)).parTupled
         (vault, container, _) = r
-      } yield RootComponent(ctx, router, initialModel(vault, localPreferences))
-        .renderIntoDOM(container)
+      } yield ReactDOMClient
+        .createRoot(container)
+        .render:
+          RootComponent(ctx, router, initialModel(vault, localPreferences)).toUnmounted
```
toddburnside, line 176: I don't really understand what appears to be a type ascription to `RootComponent(ctx...`

toddburnside (approved):
LGTM

explore/src/main/scala/explore/Explore.scala
rpiaggio: You mean the `toUnmounted`? That has to do with new `Renderable` typeclass implemented in scalajs-react. I think the long-term strategy to avoid this is to create an instance of the typeclass for classes that extend `ReactProps`/`ReactFnProps`/etc.

toddburnside: I was thinking of the `.render: RootComponent(...`

hugo-vrijswijk: It's not a type annotation, it's the braceless syntax for functions

```scala
x.map: y =>
  y + 1
```

PR #3557, opened 2024-02-01
Fix problem with explore not loading with invalid program id

toddburnside:
Moving program details into the ProgramCache broke explore when trying to load with a missing or invalid program id in the URL. This is because the ProgramCache is loaded regardless and the list of program summaries within it is used to populated the program selection dialog. This makes the program details optional and deals with it downstream.


cquiroz (approved):
LGTM

PR #3561, opened 2024-02-02
API changes

toddburnside:
I changed all the places where we were building `input` or `WHERE' objects within the graphql documents, since this makes it less likely we'll catch API changes such as the ones here. It is a tradeoff for convenience, though.


rpiaggio (approved):
LGTM

PR #3568, opened 2024-02-06
Make the entire program readonly for submitted proposals

toddburnside:
Also adds a `Message` to the bottom of the MainBody of the layout to let the user know why everything is readonly.
image

### DIFF
--- a/explore/src/clue/scala/queries/common/TargetQueriesGQL.scala
+++ b/explore/src/clue/scala/queries/common/TargetQueriesGQL.scala
@@ -94,21 +94,6 @@ object TargetQueriesGQL {
   }
 
   @GraphQL
-  // This is not used for the moment, since we are getting all targets via
-  // AsterismQueriesGQL.AsterismGroupObsQuery.targetGroup.
-  // But we could use this instead and build the groups manually, which we now know
-  // how to do. Is it worth it?
-  trait AllProgramTargets extends GraphQLOperation[ObservationDB] {
-    val document = s"""
-      query($$where: WhereTarget!) {
-        targets(WHERE: $$where) {
-          matches $TargetWithIdSubquery
-        }
-      }
-    """
-  }
-
-  @GraphQL
   trait ProgramTargetsDelta extends GraphQLOperation[ObservationDB] {
     val document = s"""
       subscription($$input: TargetEditInput!) {
--- a/explore/src/main/scala/explore/EditableLabel.scala
+++ b/explore/src/main/scala/explore/EditableLabel.scala
@@ -36,7 +36,8 @@ case class EditableLabel(
   rightButtonIcon:      FontAwesomeIcon = Icons.Eraser,
   rightButtonTooltip:   Option[String] = None,
   okButtonTooltip:      Option[String] = None,
-  discardButtonTooltip: Option[String] = None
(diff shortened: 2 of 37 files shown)


hugo-vrijswijk (commented):
I'm wondering if it might be cleaner to use context to keep the readonly variable, or some other way that's not passing props throughout the entire application. Context changes require large re-renders but this isn't something that will change very often (I assume)

toddburnside:
> I'm wondering if it might be cleaner to use context to keep the readonly variable, or some other way that's not passing props throughout the entire application. Context changes require large re-renders but this isn't something that will change very often (I assume)

Currently, it will only happen with changes in the proposal status, although that could change in the future. I like the simplicity of the props, but we do skip a few levels. I'll try it out with Context and see how it feels.

rpiaggio:
> OK. I pushed a commit that switches to using a context for the readonly flag instead of passing it via props. I'm not really sure if it is an improvement for just the readonly flag. 🤷‍♂️ It seems to me that if we want to go this route, we should maybe make something like an `AppDataContext` that includes other values that change infrequently, such as the program id and user. What do ya'll think? @hugo-vrijswijk @rpiaggio @cquiroz

We could have a program-level context where we could put this property, the program id and even the program cache with all its data (user id wouldn't belong here). However, for coherence, I think we should fully go one way (all of these in props) or the other (all of these in context). I lean towards everything in props, which cleanly outlines everything the component has access to.

toddburnside:
> We could have a program-level context where we could put this property, the program id and even the program cache with all its data (user id wouldn't belong here).

I think putting data that changes frequently in a context that is used many places, might cause a lot of re-rendering. Which is why I though program id and user could go there because changing them essentially forces everything to re-render anyway. I'm not sure why you think user would not go there, since it is something we pass a lot of places. 

> However, for coherence, I think we should fully go one way (all of these in props) or the other (all of these in context). I lean towards everything in props, which cleanly outlines everything the component has access to.

I kind of lean that way myself. For the readonly flag, at least, I feel like all the additional `useContext` pretty much offset the few places where we didn't have to pass the flag in props. I like the Context idea for things like AppContext and HelpContext, but I'm not sure I like it as much for things like readonly, which affects the component behavior. Having it in the props is very clear.

hugo-vrijswijk:
Looking at the changes I agree, I thought there would be a bigger difference in lines-of-code but I suppose it doesn't matter that much. I thought there were more components that were just only passing through the props. 

> I think putting data that changes frequently in a context that is used many places, might cause a lot of re-rendering

It does. A change in context makes every component re-rendern. React I think now only uses Context for things that don't really ever change (like the AppContext). So, sorry for my bad suggestion but I guess props is the way to go 😅.

In the future it might be worth it to look at something like [jotai](https://jotai.org/) to share state between components efficiently. I feel like state management in explore could use some love

toddburnside:
I set this PR back to the commit that uses the props for passing the readonly flag.

rpiaggio:
> I think putting data that changes frequently in a context that is used many places, might cause a lot of re-rendering. Which is why I though program id and user could go there because changing them essentially forces everything to re-render anyway. I'm not sure why you think user would not go there, since it is something we pass a lot of places.

I was thinking of a "program context".

rpiaggio (approved):
LGTM

PR #3584, opened 2024-02-14
Delay loading of execution times in ProgramCache

toddburnside:
Don't include any of the `digest` or `execution` times in the original cache initialization or in the subscriptions. These are queried separately and provided via `Pot`s to the rest of the application.

### DIFF
--- a/common/src/main/scala/explore/cache/CacheComponent.scala
+++ b/common/src/main/scala/explore/cache/CacheComponent.scala
@@ -16,8 +16,19 @@ import japgolly.scalajs.react.vdom.html_<^.*
 trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
   private type F[T] = DefaultA[T]
 
+  // Initializes the cache.
+  // `P` is the Properties
+  // `(S => S) => F[Unit]` is a delayed update function. Things fed to this function
+  // will be merged with the main update stream. The idea is that this will be done
+  // asynchronously. BUT, it is the responsibility of the implementation to
+  // call `.start` on the IO returned by this function. This gives the implementation
+  // more control over when the IO will be started.
   protected val initial: (P, (S => S) => F[Unit]) => F[S]
 
+  // A stream of updates to the cache.
+  // See explanation of types above.
+  // Plus, in the output, the stream values must be `F[S => S]` in order to support
+  // the asnchronous updates.
   protected val updateStream: (P, (S => S) => F[Unit]) => Resource[F, fs2.Stream[F, F[S => S]]]
 
   val component =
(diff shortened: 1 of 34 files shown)


common/src/main/scala/explore/cache/CacheComponent.scala
```diff
@@ -15,9 +16,9 @@ import japgolly.scalajs.react.vdom.html_<^.*
 trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
   private type F[T] = DefaultA[T]
 
-  protected val initial: P => F[S]
+  protected val initial: (P, (S => S) => F[Unit]) => F[S]
```
rpiaggio, line 19: Can I request a comment explaining the type?

toddburnside: Sure. I also realized I forgot to rename the parameter in ProgramCache. At the time I was fresh out of good names and just wanted to get it working. I'll change that, too.

common/src/main/scala/explore/cache/CacheComponent.scala
```diff
@@ -26,24 +27,34 @@ trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
...
+            initialValue <- initial(props, effects.set)
             cache        <- SignallingRef[F].of(initialValue)
             _            <- latch.complete(cache) // Allow stream updates to proceed.
           yield cache
       )
       .useStreamBy((props, cache) => (props, cache.isReady))((props, cache) =>
-        _ => cache.toOption.map(_.discrete).orEmpty.evalTap(value => props.setState(value.some))
+        _ =>
+          cache.toOption.map(_.discrete).orEmpty.evalTap { value =>
+            cats.effect.Async[F].delay(println("Emitting a value")) >>
```
hugo-vrijswijk, line 55: Is this here for debugging purposes?

toddburnside: Ah, yeah. Good catch.

explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -274,6 +278,10 @@ object ProposalEditor:
       classView.set(newClass) >> minPct2.set(minPctTotalTime) >> totalHours.set(toHours(totalTime))
     }
 
+    extension [A](pot: Pot[A])
+      def orSpinner(f: A => VdomNode): VdomNode =
```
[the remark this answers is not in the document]
toddburnside: Seems like it could be generally useful. I can move it to lucuma.ui where renderPot is in another PR.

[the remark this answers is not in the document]
toddburnside: Wait. It can't go into lucuma.ui because the spinner icon is part of explore. I'll move it someplace in explore.

explore/src/main/scala/explore/cache/CacheModifierUpdaters.scala
```diff
@@ -41,7 +44,8 @@ trait CacheModifierUpdaters {
       )
 
   protected def modifyObservations(
-    observationEdit: ObservationEdit
+    observationEdit: ObservationEdit,
+    uuid:            UUID
```
rpiaggio, line 48: What's this `UUID` for?

toddburnside: All of the Pots in the cache are associated with UUIDs. They are used so that the async updates only actually update if they are the most recent async request. This is to prevent updating with values that are already out of date.

rpiaggio: Is this a mechanism to avoid UI glitches?

If I understand correctly, the UUID is generated when a request is made and somewhere the latest UUID is stored and the result only updates the model if its UUID still matches the latest?

Can use use an existing mechanisms for this? `useSingleEffect`?

PR #3588, opened 2024-02-15
Use programTime instead of total summed time for execution

hugo-vrijswijk:


hugo-vrijswijk:
Todd will include this in his PR

toddburnside:
I don't think the delayed loading is going to happen any time real soon, so I think you should go ahead and fix this now.

toddburnside (approved):
LGTM

PR #3594, opened 2024-02-20
Delay loading of execution times in ProgramCache (Alternate approach)

rpiaggio:
This is an alternate approach to #3584.

The rendering logic is exactly the same as in #3584, but the cache handling goes a different way.

The main purpose of the alternate approach is to avoid increasing the complexity of `CacheComponent` by keeping the signature of its abstract methods closer the original ones.

Furthermore, the concern of canceling long-running time queries is abstracted away into a `Pipe` instead of having `UUID`s in the model for that purpose.

### DIFF
--- a/common/src/main/scala/explore/cache/CacheComponent.scala
+++ b/common/src/main/scala/explore/cache/CacheComponent.scala
@@ -16,8 +16,19 @@ import japgolly.scalajs.react.vdom.html_<^.*
 trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
   private type F[T] = DefaultA[T]
 
+  // Initializes the cache.
+  // `P` is the Properties
+  // `(S => S) => F[Unit]` is a delayed update function. Things fed to this function
+  // will be merged with the main update stream. The idea is that this will be done
+  // asynchronously. BUT, it is the responsibility of the implementation to
+  // call `.start` on the IO returned by this function. This gives the implementation
+  // more control over when the IO will be started.
   protected val initial: (P, (S => S) => F[Unit]) => F[S]
 
+  // A stream of updates to the cache.
+  // See explanation of types above.
+  // Plus, in the output, the stream values must be `F[S => S]` in order to support
+  // the asnchronous updates.
   protected val updateStream: (P, (S => S) => F[Unit]) => Resource[F, fs2.Stream[F, F[S => S]]]
 
   val component =
@@ -27,15 +38,15 @@ trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
         props =>
           for
             latch        <- Deferred[F, SignallingRef[F, S]]
-            effects      <- SignallingRef[F].of(identity[S])
+            delayUpdate  <- SignallingRef[F].of(identity[S])
             // Start the update fiber. We want subscriptions to start before initial query.
             // This way we don't miss updates.
             // The update fiber Will only update the cache once it is initialized (via latch).
             // TODO: RESTART CACHE IN CASE OF INTERRUPTED SUBSCRIPTION.
             _            <-
-              updateStream(props, effects.set)
+              updateStream(props, delayUpdate.set)
                 .map(
-                  _.merge(effects.discrete.map(Async[F].pure))
+                  _.merge(delayUpdate.discrete.map(Async[F].pure))
                 )
                 .evalTap(
                   _.evalTap(fmod =>
@@ -44,7 +55,7 @@ trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
                 )
                 .useForever
                 .start
-            initialValue <- initial(props, effects.set)
+            initialValue <- initial(props, delayUpdate.set)
             cache        <- SignallingRef[F].of(initialValue)
             _            <- latch.complete(cache) // Allow stream updates to proceed.
           yield cache
@@ -52,8 +63,7 @@ trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
       .useStreamBy((props, cache) => (props, cache.isReady))((props, cache) =>
         _ =>
           cache.toOption.map(_.discrete).orEmpty.evalTap { value =>
-            cats.effect.Async[F].delay(println("Emitting a value")) >>
-              props.setState(value.some)
+            props.setState(value.some)
           }
       )
       // .useEffectWithDepsBy((_, _, value) => value.toOption)((props, _, _) =>
--- a/common/src/main/scala/explore/cache/CacheComponent.scala
+++ b/common/src/main/scala/explore/cache/CacheComponent.scala
@@ -45,7 +45,7 @@ trait CacheComponent[S, P <: CacheComponent.Props[S]: Reusability]:
             cache                        <- SignallingRef[F].of(initialValue)
             _                            <-
               delayedInits
-                .evalTap: mod =>
+                .evalMap: mod =>
                   cache.update(mod)
                 .compile
                 .drain
--- a/common/src/main/scala/explore/cache/CacheComponent.scala
+++ b/common/src/main/scala/explore/cache/CacheComponent.scala
@@ -3,7 +3,6 @@
 
 package explore.cache
 
-import cats.effect.Async
 import cats.effect.Resource
 import cats.effect.kernel.Deferred
 import cats.syntax.all.*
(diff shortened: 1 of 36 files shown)


common/src/test/scala/explore/utils/StreamSpec.scala
```diff
@@ -58,3 +58,23 @@ class StreamSpec extends munit.CatsEffectSuite:
 
     TestControl.executeEmbed(program).map((a, b) => assertNotEquals(a, b))
   }
+
+  test("keyedSwitchEvalMap cancels selectively") {
+    val stream                 = Stream(1, 2, 3).evalMap(IO.sleep(0.25.seconds).as(_))
```
hugo-vrijswijk, line 63:
```suggestion
    val stream                 = Stream(1, 2, 3).meteredStartImmediately(0.25.seconds)
```
I think this does the same

common/src/test/scala/explore/utils/StreamSpec.scala
```diff
@@ -58,3 +58,23 @@ class StreamSpec extends munit.CatsEffectSuite:
...
+    val effect: Int => IO[Int] = i => IO.sleep(1.second) >> IO.pure(i)
+    val pipe                   = keyedSwitchEvalMap[IO, Int, Int, Int](_ % 2, effect)
+
+    val program = stream.through(pipe).compile.toVector
+
+    TestControl.executeEmbed(program).assertEquals(Vector(2, 3))
```
hugo-vrijswijk, line 69: I don't really understand what is being asserted here. The effect sleeps for 1 second and emits the same element. So the effect is _always_ cancelled? How is it tested that it is being cancelled? If the effect raises an error or some other noticeable effect, that would be clearer, I think.

I think overall, the tests don't need to do anything with time. I'll admit the `List[Int]` from my tests also aren't super clear. But from a glance I don't really understand what is happening. Maybe they could be clearer if you, for example, have a `Ref[List[String]]` and the effect appends starting, closing and cancelling strings for each element. That way the assertions are clearer on _what_ behaviour is happening and in what order

explore/src/main/scala/explore/cache/CacheModifierUpdaters.scala
```diff
@@ -59,7 +60,11 @@ trait CacheModifierUpdaters {
     // TODO: this won't be needed anymore when groups are also updated through events of observation updates
     val groupsUpdate = updateGroupsMappingForObsEdit(observationEdit)
 
-    obsUpdate.andThen(groupsUpdate)
+    // the program times and the observation execution will share the same UUID, but that's fine.
```
hugo-vrijswijk, line 63: Share the same UUID?

explore/src/main/scala/explore/cache/ProgramCache.scala
```diff
@@ -120,68 +151,106 @@ object ProgramCache
...
+          ProgramSummaries.fromLists(pd, ts, observations, gs, oas, pas, ps, Pot.pending, obsPots)
+      }
+
+    def combineTimesUpdates(
+      observations: List[ObsSummary]
+    ): fs2.Stream[IO, ProgramSummaries => ProgramSummaries] =
+      // We want to update all the observations first, followed by the program, because
+      // the program requires all of the observation times be calculated. So, if we do
+      // it first, it could take a long time.
+      fs2.Stream(observations.map(_.id)*).evalMap(updateObservationExecution) ++
```
hugo-vrijswijk, line 167:
```suggestion
      fs2.Stream.emits(observations.map(_.id)).evalMap(updateObservationExecution) ++
```

explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -309,20 +314,30 @@ object ProposalEditor:
...
                 id = "maxTime"
               ),
-              timeSplits(splitsMap, maxExecutionTime),
-              minimumTime(minimumPct1View.get, maxExecutionTime),
+              maxExecutionPot.renderPot(
+                valueRender = maxExecutionTime =>
+                  React.Fragment(timeSplits(splitsMap, maxExecutionTime),
+                                 minimumTime(minimumPct1View.get, maxExecutionTime)
+                  ),
+                pendingRender = React.Fragment(<.span(), <.span())
```
hugo-vrijswijk, line 326: Why these two spans for the pending render? An empty span by default takes up no space, so if you wanted a similar space as the render value these should probably be divs

hugo-vrijswijk (commented):
Nice! I think overall the approach is much cleaner than the UUID. I agree keeping some complexity and implementation details outside the CacheComponent is worth it.

The Pipe is a bit complicated with the Pull and could maybe use some extra comments about the intended purpose? I think the tests could improve as well. 

I reviewed this PR more thoroughly than #3584, but this does seem a cleaner solution.

Also, congratulations on refactoring code that hasn't even been merged yet 😃

explore/src/main/scala/explore/cache/CacheModifierUpdaters.scala
rpiaggio: Old comment, removed it now.

explore/src/main/scala/explore/proposal/ProposalEditor.scala
rpiaggio: This was carried over from https://github.com/gemini-hlsw/explore/pull/3584. @toddburnside ?

rpiaggio:
> The Pipe is a bit complicated with the Pull and could maybe use some extra comments about the intended purpose? I think the tests could improve as well.

I agree, I will address that.

explore/src/main/scala/explore/proposal/ProposalEditor.scala
toddburnside: It is a grid layout, so the space will be determined by the grid. There just needs to be something to put in the cell to keep everything in the right cells.

common/src/test/scala/explore/utils/StreamSpec.scala
rpiaggio: I added some comments to the test. I might still be missing something.

common/src/main/scala/explore/cache/CacheComponent.scala
```diff
@@ -15,36 +15,46 @@ import japgolly.scalajs.react.vdom.html_<^.*
...
                 .useForever
                 .start
-            initialValue <- initial(props)
-            cache        <- SignallingRef[F].of(initialValue)
-            _            <- latch.complete(cache) // Allow stream updates to proceed.
+            (initialValue, delayedInits) <- initial(props)
+            cache                        <- SignallingRef[F].of(initialValue)
+            _                            <-
+              delayedInits
+                .evalTap: mod =>
```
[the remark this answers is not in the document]
rpiaggio: It's the same. The output is discarded by the `drain` anyway. But I agree it's a bit clearer to follow with `evalMap`, making it clear that we don't need the `mod` anymore.

common/src/main/scala/explore/utils/package.scala
```diff
@@ -254,3 +258,54 @@ extension [F[_], O](s: Stream[F, O])
...
+ * Produces a `Pipe` that evaluates an effect for each element in the input stream and emits its
+ * result, canceling previous effects that are still running with the same `key`.
+ */
+def keyedSwitchEvalMap[F[_]: Concurrent, I, O, K](
+  key: I => K,
+  f:   I => F[O]
+): Pipe[F, I, O] = {
+  def go(
+    stream: Stream[F, Either[Option[I], (K, O)]], // input? | key -> output
+    fibers: Map[K, Fiber[F, Throwable, Unit]],
```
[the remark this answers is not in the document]
rpiaggio: > There is always at max 1 running effect per key, and that is cancelled when pulling the new element.

Yes, that is the case.

> I think this could be simplified by using a [HotSwap](https://typelevel.org/cats-effect/docs/std/hotswap) instead of managing fibers directly?

`HotSwap` acts on `Resource`s. Do you mean running the fibers as `Resource`s with allocation as `.start` and releasing as `.cancel`? We would have to store a `HotSwap` for each key in a `Map` instead of `fibers`. That would probably work, but I'm not sure it's a simplification.

rpiaggio:
I addressed the PR comments and rebased.

I must note that the new behavior is a bit different than the original PR: in the original PR the initialization effects could be cancelled by an update (although I think there might have been a (fixable) bug causing initialization to never complete when this happened).

In the logic of this PR, explore waits (asynchronously) for initialization effects before it starts processing updates.

If we really want initialization effects to be cancellable in the case of an update, it can be done and I started implementing it in the `delayed-times-2` branch. However, it adds some complexity which IMHO is not worth for the gains.

toddburnside:
> If we really want initialization effects to be cancellable in the case of an update, it can be done and I started implementing it in the `delayed-times-2` branch. However, it adds some complexity which IMHO is not worth for the gains.

Probably isn't necessary. It would only be helpful in the extreme cases where a lot of observations need to be recalculated by the ODB, where it would allow the user to see times for whatever observation they were working on. However, my PR also had an issue with this case, because the program times would also be immediately called, even if there were many observations left to calculate. It is probably better this way.

toddburnside (approved):
LGTM - Updates to the program times from a program edit and from an observation edit happen independently, but I think that's fine. In fact, I'm not sure if, at the moment, any programEdit changes would affect the times or not.

toddburnside:
@hugo-vrijswijk Are we querying the group times now? If so, we need to add those into this. Could be another PR.

hugo-vrijswijk:
> @hugo-vrijswijk Are we querying the group times now? If so, we need to add those into this. Could be another PR.

Yes, but I can add it in a separate PR

common/src/main/scala/explore/utils/package.scala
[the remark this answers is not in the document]
hugo-vrijswijk: Hmm, agreed this is probably the simpler solution

PR #3611, opened 2024-02-26
Use Apple Silicon in CI runners

armanbilge:
Let's see if this makes anything faster ...


armanbilge:
Excellent. Looks like this is even slower than the ubuntu runners 🙃

PR #3619, opened 2024-03-01
Single program times updates pipeline

rpiaggio:
This PR pipelines all program times updates, whether triggered through an observation or a program edit.

This allows an ongoing update to be cancelled when a new one is triggered, regardless of what triggered it.


toddburnside (approved):
LGTM

PR #3622, opened 2024-03-01
Don't add reusability to aladin components

cquiroz:
We had reusability in Aladin because you'd get sometimes two renders for one change which was visible. This was supposed to be fixed in React 18 and it seems it is


toddburnside (approved):
LGTM

PR #3628, opened 2024-03-04
Upgrade dependencies - changes to Semester

toddburnside:
This gets our dependencies mostly updated. The biggest change is that obtaining a `Semester` returns an Option because the semester year is constrained.

There are still graphql errors on the observation tab due to the grackle changes in graphql interfaces.


cquiroz (approved):
Looks good to me :+1:

PR #3637, opened 2024-03-06
Allow adding new finder charts when the tile is closed

cquiroz:


common/src/main/scala/explore/syntax/ui/package.scala
```diff
@@ -93,3 +96,8 @@ extension [F[_]: MonadCancelThrow, A](f: F[A])
 extension [A](pot: Pot[A])
   def orSpinner(f: A => VdomNode): VdomNode =
     pot.renderPot(valueRender = f, pendingRender = Icons.Spinner.withSpin(true))
+
+extension [A](vdom: Option[A])(using c: Conversion[A, VdomNode])
+  inline def getOrEmpty: VdomNode = vdom.map(c).getOrElse(EmptyVdom)
+
+extension (vdom: Option[VdomNode]) inline def getOrEmpty: VdomNode = vdom.getOrElse(EmptyVdom)
```
rpiaggio, line 103: There's a `Monoid[VdomNode]` definition in `lucuma.ui.syntax`. Maybe we can use that and `orEmpty`.

rpiaggio (approved):
Can we use the existing `Monoid[VdomNode]`? Otherwise LGTM

common/src/main/scala/explore/syntax/ui/package.scala
cquiroz: I tried it but type inference failse because I try to applyy to a Component that can be converted to a `VdomNode`, like in this PR `FinderCharts`

PR #3642, opened 2024-03-07
Vite warmup of files that take long to process.

rpiaggio:
Warms up files that take a long time for Vite to transform.

Performance improvement is minimal, but at least explore always seems to load correctly on the first try now.

Reference:
https://vitejs.dev/guide/performance#warm-up-frequently-used-files
https://vitejs.dev/config/server-options#server-warmup


cquiroz (approved):
Thanks, let's give it a try

PR #3644, opened 2024-03-08
Remove program times update on `programEdit`

toddburnside:
After editing a program, the program times would be invalidated but the request for new times wouldn't fire, so we ended up with perpetual spinners.

However, I'm wondering if any change we would receive a `programEdit` subscription message for would actually change the program times... Maybe we could just remove the update on `programEdit`?

### DIFF
--- a/explore/src/main/scala/explore/cache/ProgramCache.scala
+++ b/explore/src/main/scala/explore/cache/ProgramCache.scala
@@ -227,7 +227,7 @@ object ProgramCache
               _.map: data => // Replace program and reset times.
                 ProgramSummaries.optProgramDetails.replace(data.programEdit.value.some) >>>
                   ProgramSummaries.programTimesPot.replace(Pot.pending)
-            .evalTap(_ => queryProgramTimes)
+              .evalTap(_ => queryProgramTimes)
 
         val updateTargets: Resource[IO, Stream[IO, ProgramSummaries => ProgramSummaries]] =
           TargetQueriesGQL.ProgramTargetsDelta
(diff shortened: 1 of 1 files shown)


toddburnside:
> Probably. Is there anything at the program level that takes part in computing the times?

Nothing that I know of. Looking at the `Program` returned from the API, it doesn't seem like it. I mean, the observations and groups are part of the program, but they have their own `Edit` subscriptions and will trigger a request for program times.

toddburnside:
> > Probably. Is there anything at the program level that takes part in computing the times?
> 
> Nothing that I know of. Looking at the `Program` returned from the API, it doesn't seem like it. I mean, the observations and groups are part of the program, but they have their own `Edit` subscriptions and will trigger a request for program times.

Maybe we should remove the time update on program edit and put it back if we discover anything in the future. It seems not good to have the times refresh and spinners flash for no reason. As it is, that will happen for a lot of observation updates because we have no way of knowing if they affect the times or not.

swalker2m:
> However, I'm wondering if any change we would receive a programEdit subscription message for would actually change the program times... Maybe we could just remove the update on programEdit?

What about group edits?  I'm not sure what events are fired but rearranging observations in groups could impact the computation.

toddburnside:
> What about group edits? I'm not sure what events are fired but rearranging observations in groups could impact the computation.

Right. Hugo is working on including the program times update on groupEdit. I `think` the groupEdit is fired for group rearranging now, but Hugo would know for sure.

PR #3645, opened 2024-03-08
Revert to vite 4.5.0

cquiroz:
Tested locally and react-grid-layout and clipboard works. Also workers are running in dev and feels a lot faster


rpiaggio (approved):
LGTM

PR #3646, opened 2024-03-09
Better looking loading of user preferences

cquiroz:
This is just a tiny nitpick where the loading indicator doesn't seem right when opening the user preferences dialog box


toddburnside (approved):
LGTM

PR #3658, opened 2024-03-15
Load group program times separately from main query

hugo-vrijswijk:


build.mjs
```diff
@@ -2,17 +2,22 @@ import path from 'path';
...
 import fs from 'fs/promises';
 import process from 'process';
-import { build } from 'vite';
+import { build, loadConfigFromFile } from 'vite';
 import { minify } from 'terser';
 import humanFormat from 'human-format';
 import brotliSize from 'brotli-size';
-import config from './vite.config.mjs';
+
+const config = await loadConfigFromFile({ mode: 'production' });
```
rpiaggio, line 10: Why this change?

explore/src/main/scala/explore/ExplorePWA.scala
```diff
@@ -138,8 +135,8 @@ object ExplorePWA {
     bc.messages
       .evalTap { me =>
         me.data.event match
-          case ExploreEvent.PWAReloadId      => IO.fromFuture(IO(updateSW(true).toFuture)) // .void
-          case ExploreEvent.ExploreUIReadyId => IO.fromFuture(IO(updateSW(()).toFuture))   // .void
+          case ExploreEvent.PWAReloadId      => IO.fromPromise(IO(updateSW(true))) // .void
```
rpiaggio, line 138: Nice

model/shared/src/main/scala/explore/model/GroupElement.scala
```diff
@@ -54,23 +58,50 @@ object GroupElement:
...
+  def findParentGroupIds(
+    groups:    GroupList,
+    elementId: Either[Observation.Id, Group.Id]
+  ): List[Group.Id] = {
+    @tailrec
+    def go(
+      acc:     List[Group.Id],
+      current: Either[Observation.Id, Group.Id]
+    ): List[Group.Id] =
+      val parentGroup = groups.find(
```
rpiaggio, line 73: To find the parent node, are we scanning the whole list of groups on each iteration?

rpiaggio (approved):
LGTM

I hadn't taken a look at the way Groups are implemented until now. I'd like to see if we can retrofit it into an `IndexedKeyTree` so that:
- We have separate structure from data.
- All nodes have a link to its parent.
- We can have positional undo, even in the face of intermediate remote edits/insertions/deletions. (I actually haven't checked if the current structure supports this).

build.mjs
hugo-vrijswijk: I moved the vite config to a typescript file, but that means you can't import it as plain js anymore

model/shared/src/main/scala/explore/model/GroupElement.scala
hugo-vrijswijk: Yes, if it's not in a tree structure, I think this is the fastest/only way. Since I imagine groups won't be _that_ deep, it won't be too many iterations. `IndexedKeyTree` is already used, but I can have a look at lifting it higher

explore/src/main/scala/explore/cache/ProgramCache.scala
```diff
@@ -213,6 +241,15 @@ object ProgramCache
               queryProgramTimes
         )
 
+        val groupTimeRangeUpdate: Pipe[
+          IO,
+          ProgramQueriesGQL.GroupEditSubscription.Data,
+          ProgramSummaries => ProgramSummaries
+        ] = keyedSwitchEvalMap(
+          _.groupEdit.value.map(_.id),
+          data => updateGroupTimeRange(props.programId)
```
toddburnside, line 250: Should this also flatTap to queryProgramTimes?

toddburnside (approved):
LGTM, with one question.

explore/src/main/scala/explore/cache/ProgramCache.scala
hugo-vrijswijk: Yes, I've added that now

PR #3681, opened 2024-03-21
Use KeyedIndexedTree for groups in ProgramSummaries

hugo-vrijswijk:
I'm impressed by just how much ugly tree-mapping code we can throw away because of this 🙏

### DIFF
--- a/model/shared/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
+++ b/model/shared/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
@@ -5,7 +5,9 @@ package explore.data.tree
 
 import cats.kernel.Eq
 import cats.syntax.all.*
+import eu.timepit.refined.types.numeric.NonNegShort
 
+import scala.annotation.tailrec
 import scala.collection.immutable.HashMap
 import scala.collection.immutable.HashSet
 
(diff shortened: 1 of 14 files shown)


model/shared/src/main/scala/explore/model/GroupTree.scala
```diff
@@ -0,0 +1,47 @@
...
+import cats.syntax.all.*
+import eu.timepit.refined.cats.*
+import explore.data.tree.KeyedIndexedTree
+import explore.data.tree.Node
+import explore.data.tree.Tree
+import explore.model.syntax.all.*
+import lucuma.core.model.Group
+import lucuma.core.model.Observation
+
+type GroupTree = KeyedIndexedTree[Either[Observation.Id, Group.Id], Either[GroupObs, Grouping]]
```
rpiaggio, line 16: `GroupObs` seems to be class wrapping an `Observation.Id` and a `groupIndex`. Is `groupIndex` the position in the group? If that's the case, `KeyedIndexedTree` already provides `childPos` in its `Index`. Can we use just `childPos` in the structure, and leave `GroupObs` just for the decoding the server's response?

model/shared/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
```diff
@@ -93,15 +95,55 @@ case class KeyedIndexedTree[K: Eq, A] private (
...
+    val newByKey = byKey.map((k, node) => (k, mapNode(node)))
+    KeyedIndexedTree(newByKey, newTree)(getKey)
+  }
+
 }
 
 object KeyedIndexedTree {
   case class Index[K](parentKey: Option[K], childPos: Int)
+  object Index:
+    inline def apply[K](parentKey: Option[K], childPos: NonNegShort): Index[K] =
```
rpiaggio, line 145: Maybe it makes sense to change `childPos` in the class to a non-negative type. Not sure about it being a `Short` though. Is that how it comes from the ODB for groups?

model/shared/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
```diff
@@ -93,15 +95,55 @@ case class KeyedIndexedTree[K: Eq, A] private (
     KeyedIndexedTree(buildKeyMap(newTree, getKey), newTree)(getKey)
   }
 
+  def updated(key: K, newNode: A, newIndex: Index[K]): KeyedIndexedTree[K, A] = {
```
rpiaggio, line 98: Nice additions.

model/shared/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
hugo-vrijswijk: Yeah I thought about this too. But because there are multiple operations done on the childPos when updating the tree you'd have to do a lot of `NonNegShort.unsafeFrom(i.toShort)` which is kind of ugly. I agree it makes sense for it to be NonNegX though.

`NonNegShort` comes from the odb types for indices, yes

model/shared/src/main/scala/explore/data/tree/KeyedIndexedTree.scala
hugo-vrijswijk: Now with tests!

rpiaggio: Nice!

model/shared/src/main/scala/explore/model/GroupTree.scala
rpiaggio: Ah I see all the tree information is still in `Grouping`. Is this redudant?

My suggestion for using `KeyedIndexedTree` involved using it for the tree structure, removing the need for `elements`, `parentId` and `parentIndex` in `Grouping` as well as `GroupObs` and `GroupingElement`.

hugo-vrijswijk: Yeah the information is redundant. I could make an 'intermediate' Grouping case class that's used for the API, but it is probably going to be quite a lot of changes

rpiaggio: Yes, it would be a largish PR. Maybe another PR?

About the intermediate class: could we decode directly to the model structure? Otherwise an intermediate class is fine. Maybe even the one generated by clue.

rpiaggio (approved):
LGTM, let's remove duplicate tree structure in another PR

hugo-vrijswijk:
I think it might even be possible to remove the `ObsNode` code completely. The tree could just use the KeyedIndexedTree in the ObsList component and only map to the primereact tree node when rendering. But I'll take a proper look tomorrow

PR #3682, opened 2024-03-22
Allow spaces on the description

cquiroz:
This is odd, it works fine on the react-demo and it has been working in the past
The fix here is the only way I found to do it

### DIFF
--- a/explore/src/main/scala/explore/EditableLabel.scala
+++ b/explore/src/main/scala/explore/EditableLabel.scala
@@ -20,6 +20,8 @@ import lucuma.ui.syntax.all.given
 import lucuma.ui.utils.given
 import org.scalajs.dom
 
+import scala.util.Random
+
 import scalajs.js.JSConverters.*
 
 case class EditableLabel(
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/EditableLabel.scala
```diff
@@ -139,13 +141,15 @@ object EditableLabel {
         if (editing.value.value)
           <.div(^.width := "100%", ^.display.flex)(
             InputText(
-              id = "editable-label-input", // won't necesessarily be unique...
+              id = s"editable-label-input-${Random.nextInt}",
```
rpiaggio, line 144: I think this should be memoized. Otherwise it will change on each render.

rpiaggio: Actually, can we pass de id as a required property?

Then the caller can pass the `obsId` for example.

explore/src/main/scala/explore/EditableLabel.scala
```diff
@@ -139,13 +141,15 @@ object EditableLabel {
...
               value = displayValue.value,
               onChange = (e: ReactEventFromInput) => displayValue.setState(e.target.value),
               clazz = props.inputClass
             ).mini.withMods(
               ^.onKeyUp ==> (e =>
-                if (e.key === "Enter") submitCB
+                // This is odd, it works fine in demo but not in explore
+                if (e.key === " ") displayValue.modState(_ + " ")
+                else if (e.key === "Enter") submitCB
```
rpiaggio, line 152: Really odd. Does it still happen if you comment out the whole `onKeyUp` parameter?

cquiroz: Yes I tried all kind of things and this is the only thing it worked. I thought it could be related to prime-react but it works fine on the demo

explore/src/main/scala/explore/EditableLabel.scala
```diff
@@ -139,13 +142,15 @@ object EditableLabel {
         if (editing.value.value)
           <.div(^.width := "100%", ^.display.flex)(
             InputText(
-              id = "editable-label-input", // won't necesessarily be unique...
+              id = s"editable-label-input-${props.id}",
```
hugo-vrijswijk, line 145: Has the new hooks in scalajs-react been released yet? If so you could also use the `useId` for a unique id per component, rather than having to pass id's as a prop. This situation is exactly what it was made for

rpiaggio: Yes! It should be in beta 3. I had forgotten about that.

rpiaggio (approved):
LGTM.

Maybe in the future we should `useId` inside the controls in `lucuma-ui`.

PR #3684, opened 2024-03-22
Move user details table to its own tile

cquiroz:


rpiaggio (approved):
LGTM. This looks like it will require migration in dev/staging/prod.

cquiroz:
> LGTM. This looks like it will require migration in dev/staging/prod.

Why? What migration?

rpiaggio:
> > LGTM. This looks like it will require migration in dev/staging/prod.
> 
> Why? What migration?

user prefs DB?

PR #3692, opened 2024-03-25
Remove index and parentId from GroupTree node data

hugo-vrijswijk:
The GroupTree node case classes `GroupObs`/`Grouping` had some duplicate data in `parentId`, `parentIndex` and `elements`. Since this data is already known by the tree, we can remove it from the case classes and only use the node data for this.

- Also remove intermediate ObsNode class that was used for ObsList (this greatly simplifies the Iso used!)
- Use `NonNegInt` for KeyedIndexed(Tree|List)


model/shared/src/main/scala/explore/data/KeyedIndexedList.scala
```diff
@@ -139,7 +145,7 @@ object KeyedIndexedList:
               .map(list =>
                 KeyedIndexedList.unsafeFromTreeSeqMap(
                   TreeSeqMap.from(
-                    list.sortBy { case (_, (_, i)) => i }
+                    list.sortBy { case (_, (_, i)) => i }(Order[NonNegInt].toOrdering)
```
hugo-vrijswijk, line 148: I'm not sure why, but I wasn't able to derive an `Ordering[NonNegInt]` here, so instead I summon the `Order[NonNegInt]` and convert that. If I'm missing an import then I'd be happy to replace it

rpiaggio: Usually, if you `import cats.Order.given`, it derives an `Ordering` from an implicit `Order` in scope. Not sure whether we can bring the `Order` given into scope here though.

explore/src/main/scala/explore/tabs/GroupEditTile.scala
```diff
@@ -242,6 +248,6 @@ object GroupEditTile:
     val andOrStr: VdomNode =
       if group.isAnd then if group.ordered then "Ordered" else "Any order"
       else
-        s"Choose ${group.minimumRequired.getOrElse(NonNegShort.unsafeFrom(1))} of ${group.elements.length}"
+        s"Choose ${group.minimumRequired.getOrElse(NonNegShort.unsafeFrom(1))} of ${elementsLength}"
```
rpiaggio, line 251: `1.refined` ?

model/shared/src/main/scala/explore/data/KeyedIndexedList.scala
hugo-vrijswijk: Yup, that fixed it

explore/src/main/scala/explore/tabs/GroupEditTile.scala
hugo-vrijswijk: Yep

PR #3695, opened 2024-03-26
Fix configuration filtering and slit width format

cquiroz:


model/shared/src/main/scala/explore/model/formats.scala
```diff
@@ -68,6 +68,17 @@ trait formats:
   val formatArcsec: Format[String, Angle] =
     Format(_.parseIntOption.map(Angle.arcseconds.reverseGet(_)), Angle.arcseconds.get(_).toString)
 
+  val formatDecimalArcsec: Format[String, Angle] =
```
rpiaggio, line 71: Should we test the `Format` laws?

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
```diff
@@ -401,7 +401,7 @@ case class SpectroscopyModesMatrix(matrix: List[SpectroscopyModeRow]) {
         wavelength.forall(w => w >= r.λmin.value && w <= r.λmax.value) &&
         resolution.forall(_ <= r.resolution) &&
         range.forall(_ <= r.λdelta.value) &&
-        slitWidth.forall(_.toMicroarcseconds <= r.slitLength.value.toMicroarcseconds) &&
+        slitWidth.forall(_.toMicroarcseconds <= r.slitWidth.value.toMicroarcseconds) &&
```
swalker2m, line 404: When the slit width is specified, it must be less than the configuration option slit width value?  In other words, in the requirements the slit width entered is the maximum value that the user will accept?

model-tests/jvm/src/test/scala/explore/modes/ModesSuite.scala
```diff
@@ -64,7 +64,7 @@ class ModesSuite extends CatsEffectSuite {
       )
       // .flatTap(_.traverse(IO.println))
       .map(_.length)
-      .assertEquals(85)
+      .assertEquals(48)
```
swalker2m, line 67: Maybe helpful to add a couple of tests to check that the results match.  For example, to check that all results have a slit width `<= 1`?

model/shared/src/main/scala/explore/model/formats.scala
cquiroz: I'll make a separate PR about this

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
andrewwstephens: I think slit *length* is the desired parameter here:
https://github.com/gemini-hlsw/explore-help-docs/edit/main/configuration/focal_plane.md

PR #3700, opened 2024-03-27
Filter out deleted observations in GroupTree

hugo-vrijswijk:


explore/src/clue/scala/queries/common/GroupQueriesGQL.scala
```diff
@@ -22,6 +22,7 @@ object GroupQueriesGQL:
         observation {
           id
           groupIndex
+          existence
```
hugo-vrijswijk, line 25: I'm not sure if there's any way to query to already filter deleted observations out, but that might be an even better solution

cquiroz: Right, we would get less data in the first place, @swalker2m ?

PR #3719, opened 2024-03-31
Don't save timingwindows if some are invalid

cquiroz:
image (1)

### DIFF
--- a/explore/src/main/scala/explore/common/TimingWindowsQueries.scala
+++ b/explore/src/main/scala/explore/common/TimingWindowsQueries.scala
@@ -14,6 +14,8 @@ import explore.syntax.ui.*
 import explore.utils.ToastCtx
 import japgolly.scalajs.react.util.Effect.Dispatch
 import lucuma.core.model.TimingWindow
+import lucuma.core.model.TimingWindowEnd
+import lucuma.react.primereact.Message
 import lucuma.schemas.ObservationDB
 import lucuma.schemas.ObservationDB.Types.*
 import lucuma.schemas.odb.input.*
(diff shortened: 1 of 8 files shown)


toddburnside:
Because this uses `DatePicker`, I guess we can't solve the problem with validators like we did for, for example, hour angles:
image

But, could we prevent it from happening by adding min/max dates to the DatePickers properties? In other words, prevent the error state from ever happening?

cquiroz:
Not sure if we can do that, maybe we can if there is a max or min date property?

cquiroz:
https://reactdatepicker.com/#example-min-date

cquiroz:
Great suggestion, we can use min/max dates and a bit of css

image

explore/src/main/scala/explore/common/TimingWindowsQueries.scala
```diff
@@ -14,29 +14,42 @@ import explore.syntax.ui.*
...
+                 WHERE = obsIds.toList.toWhereObservation.assign,
+                 SET = ObservationPropertiesInput(
+                   timingWindows = value.filter(_.isValid).map(_.toInput).assign
+                 )
+               )
+             )
+             .toastErrors
+             .void
+         else
+           ToastCtx[F].clear() *> ToastCtx[F]
```
toddburnside, line 50: Can we ever hit this now?

toddburnside (approved):
Nice. LGTM.

explore/src/main/scala/explore/common/TimingWindowsQueries.scala
cquiroz: I'd rather leave the error handling on just in case

PR #3723, opened 2024-04-02
Update to API with new proposal mutations.

toddburnside:
The relevant API changes here are the splitting out of the proposal mutations from the `updatePrograms` mutation.


cquiroz:
Slightly related, should the proposal data be in a cache similar to `ProgramCache`?

cquiroz (approved):
LGTM

toddburnside:
> Slightly related, should the proposal data be in a cache similar to `ProgramCache`?

It is in the ProgramCache.

PR #3726, opened 2024-04-03
Allow revoking invites

cquiroz:
Note that the invitations table does not get updated thus we keep a local copy to add/remove invites


toddburnside:
> Note that the invitations table does not get updated thus we keep a local copy to add/remove invites

I think this should be resolved when this PR gets merged? https://github.com/gemini-hlsw/lucuma-odb/pull/1106

explore/src/main/scala/explore/proposal/ProgramUserInvitations.scala
```diff
@@ -3,54 +3,115 @@
...
   private def column[V](
     id:       ColumnId,
-    accessor: UserInvitation => V
-  ): ColumnDef.Single[UserInvitation, V] =
+    accessor: CoIInvitation => V
+  ): ColumnDef.Single[CoIInvitation, V] =
     ColDef(id, accessor, columnNames(id))
 
-  private val columns: List[ColumnDef[UserInvitation, ?]] =
+  private def columns(active: View[IsActive], invitations: View[List[CoIInvitation]])(
```
toddburnside, line 55: `deleteConfirmation` modifies the `active` parameter, but I don't see it used elsewhere. Did you want to disable the Button while active was true?

explore/src/main/scala/explore/users/UserPreferencesPopup.scala
```diff
@@ -166,12 +139,27 @@ object UserPreferencesContent:
...
+                isActive
+              )
+
               <.div(
                 ExploreStyles.ApiKeyDelete,
                 Button(
                   icon = Icons.Trash,
                   severity = Button.Severity.Secondary,
-                  onClick = deleteKey(keyId, isAdding, newKey, props.vault),
+                  onClick = deleteKey,
```
toddburnside, line 162: Should this button be disabled when `isActive` is true, too?

toddburnside (approved):
LGTM, with a couple of questions.

explore/src/main/scala/explore/users/UserPreferencesPopup.scala
cquiroz: Not sure as is Active is used on  the popup for invites, though now that you mention it maybe it should be used while deleting too

explore/src/main/scala/explore/proposal/ProgramUserInvitations.scala
cquiroz: It is used on the Invite Popup

explore/src/main/scala/explore/users/UserPreferencesPopup.scala
cquiroz: Sorry I got confused, you're right we should use it in the button
This is also because thee confirmation dialog method takes a `View[IsActive]`

explore/src/main/scala/explore/proposal/ProgramUserInvitations.scala
toddburnside: Maybe I'm missing something, but I only see it being used by the delete confirmation dialog, which sets it to true and then false. But I don't see anything else using it.

cquiroz: Yeah deleteConfirmation needs it to call a common dialog box. That is also called in another part of the code  that does use it. I'll anyway disable the button based on `IsActive`

PR #3730, opened 2024-04-04
Hide the proposal tab for non-Science program types

toddburnside:
The button in the side tabs is removed for non-science program types. However, since it is possible to directly go to the `proposals` url, I also display a message on the proposal tab in that case. Since it is very unlikely for someone to try to go directly to the proposal tab for a non-science program, this seemed simpler than trying to prevent it in the router.

Hint: Hiding whitespace changes will make this simpler to compare.


cquiroz (approved):
LGTM

PR #3732, opened 2024-04-04
Show step id on dev mode, useful for debugging sequence execution

rpiaggio:
Also, remove pprinter which was moved to lucuma-ui.


cquiroz (approved):
LGTM

PR #3733, opened 2024-04-04
Do not use gaia g magnitude for the itc

cquiroz:
As the backend changed to remove gaia bands we have to do it on explore as well


toddburnside (approved):
LGTM

PR #3739, opened 2024-04-05
Add box to redeem invites

cquiroz:
This is a poor's man way to accept invitations
Take it as a first attempt to close the loop but I'm certain it can benefit from many improvements

https://github.com/gemini-hlsw/explore/assets/3615303/b959be28-ece4-4784-a35b-272958819d9b


toddburnside (approved):
Looks great! Does the user need to be logged in as a real user in order to redeem an invitation?

cquiroz:
Yes, has to be logged in simplifying the process, we may want to allow guests in the future though

PR #3750, opened 2024-04-08
Disable sky brightness and lunar elevation on the elevation plot

cquiroz:
I made a simple version so we wouldn't remove existing code in case users want to see them back


toddburnside (approved):
Good planning. LGTM

PR #3751, opened 2024-04-10
Hide unsupported modes

cquiroz:
We can now load the modes directly from the odb rather than from a file


explore/src/main/scala/explore/cache/ModesCache.scala
```diff
@@ -0,0 +1,48 @@
...
+import clue.StreamingClient
+import explore.DefaultErrorPolicy
+import explore.modes.SpectroscopyModesMatrix
+import fs2.Stream
+import japgolly.scalajs.react.*
+import lucuma.react.common.ReactFnProps
+import lucuma.schemas.ObservationDB
+import queries.common.ModesQueriesGQL
+
+case class ModesCache(
```
cquiroz, line 18: I wonder if this is overkill for this case

toddburnside: Do you mean caching the modes might be overkill?

cquiroz: We want to call it only once not  every time we open the modees tile, in that sense caching is important
However they also don't (often) change upstream, there is no subscription for example

toddburnside: We'll have to figure this out when there is filtering on the ODB for CfP, resource availability, etc. But for now this is 👍

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
```diff
@@ -332,65 +337,61 @@ object SpectroscopyModeRow {
...
-      }
 
-  given CellDecoder[Option[SpectroscopyCapabilities]] =
-    CellDecoder.stringDecoder
-      .map {
-        case "Nod&Shuffle" => SpectroscopyCapabilities.NodAndShuffle.some
-        case "coronagraph" => SpectroscopyCapabilities.Coronagraphy.some
-        case _             => none
-      }
+  // decodores for instruments are used locally as they are not lawful
```
toddburnside, line 341: Spelling of `decoders`.

toddburnside (approved):
LGTM. Thanks for doing this!

PR #3774, opened 2024-04-23
Remove times from ProgramDetailsSubquery

toddburnside:
The program times were removed from `ProgramDetails`, but remained in the query. This sort of defeated the purpose of removing them from ProgramDetails...


cquiroz (approved):
LGTM

PR #3777, opened 2024-04-24
Support removing users

cquiroz:


model/shared/src/main/scala/explore/model/ProgramDetails.scala
```diff
@@ -18,23 +19,27 @@ case class ProgramDetails(
...
 ) derives Eq:
-  val allUsers = pi.fold(users)(p => ProgramUserWithRole(p, None) :: users)
+  val allUsers: NonEmptyList[ProgramUserWithRole] = NonEmptyList(pi, users)
 
 object ProgramDetails:
-  val proposal: Lens[ProgramDetails, Option[Proposal]]     = Focus[ProgramDetails](_.proposal)
-  val proposalStatus: Lens[ProgramDetails, ProposalStatus] = Focus[ProgramDetails](_.proposalStatus)
+  val proposal: Lens[ProgramDetails, Option[Proposal]]               = Focus[ProgramDetails](_.proposal)
+  val proposalStatus: Lens[ProgramDetails, ProposalStatus]           = Focus[ProgramDetails](_.proposalStatus)
+  val users: Lens[ProgramDetails, NonEmptyList[ProgramUserWithRole]] =
```
toddburnside, line 31: It seems potentially a little confusing to have the `users` lens use the `allUsers` value.

toddburnside (approved):
LGTM

model/shared/src/main/scala/explore/model/ProgramDetails.scala
```diff
@@ -18,23 +19,27 @@ case class ProgramDetails(
...
+      b => b.copy(pi = a.head, users = a.tail)
+    )
 
   given Decoder[ProgramDetails] = Decoder.instance(c =>
     for {
       t  <- c.get[ProgramType]("type")
       p  <- c.get[Option[Proposal]]("proposal")
       ps <- c.get[ProposalStatus]("proposalStatus")
-      pi <- c.get[Option[ProgramUser]]("pi")
+      pi <- c.get[ProgramUser]("pi")
```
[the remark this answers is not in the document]
cquiroz: Yeah I was a bit surprised but I think it shouldn't be null

model/shared/src/main/scala/explore/model/ProgramDetails.scala
cquiroz: I'll rename it

PR #3780, opened 2024-04-25
Show the available call for proposals

cquiroz:


common/src/main/scala/explore/model/CallForProposal.scala
```diff
@@ -0,0 +1,13 @@
...
+package explore.model
+
+import cats.Eq
+import cats.derived.*
+import cats.syntax.all.*
+import io.circe.Decoder
+import lucuma.core.model.Semester
+import lucuma.schemas.decoders.given
+
+case class CallForProposal(semester: Semester) derives Eq, Decoder
```
swalker2m, line 13: This will grow to include the call type, etc. right?

cquiroz: Yes, I'm pretty sure  it will grow, we probably need partners and instruments to filter things on the UI

toddburnside (approved):
It seems like it would be nice to include the call type in the dropdown somehow. If there was a standard string format, the API might be able to send that and we wouldn't have to worry about the semester, type, etc. in explore - unless we want to display them somewhere. This looks fine for a first pass.

swalker2m:
> If there was a standard string format, the API might be able to send that 

That seems like a good idea, I'll add it.

PR #3789, opened 2024-04-30
Create observations/groups inside selected group

hugo-vrijswijk:


explore/src/main/scala/explore/common/GroupQueries.scala
```diff
@@ -54,10 +54,13 @@ object GroupQueries:
...
       .void
 
-  def createGroup[F[_]: Async](programId: Program.Id)(using
+  def createGroup[F[_]: Async](programId: Program.Id, parentId: Option[Group.Id])(using
     FetchClient[F, ObservationDB]
   ): F[Grouping] =
     CreateGroupMutation[F]
       .execute:
-        CreateGroupInput(programId = programId.assign)
+        CreateGroupInput(
```
cquiroz, line 62: Is this undoable?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/common/GroupQueries.scala
hugo-vrijswijk: No since removing groups is not supported yet

PR #3792, opened 2024-04-30
Set the semester from CfP and show submission errors on the screen

cquiroz:
These are some steps towards  supporting submission but we are still missing some bits on the backend

*  Submission errors are now displayed
* Hide proposal class selector and set its value internally to queue
* Use the CfP title on the UI
* Retrieve the program reference (not in use yet)


hugo-vrijswijk (approved):
LGTM

PR #3806, opened 2024-05-07
Fix problem of editing brightnesses table always cloning the target.

toddburnside:
This was also a problem in the EmissionsLineEditor.

Raul helped me track down the problem. The solution is not very satisfying, but it does solve the problem.


rpiaggio (approved):
LGTM

cquiroz (approved):
LGTM, great find

PR #3808, opened 2024-05-07
Tighten up brightness table to avoid horizontal scrolling

toddburnside:
I narrowed the minimum width of the columns so it will go to this point before requiring scroll bars:
image
Any more and we will not be able to see all the information or be able to distinguish the units or have the sortable header (Band).

If we want to be able to go narrow, we could do some additional work to allow stackable tables in ReactTable, but I'm not sure that's a great option because of the vertical space it would require.


cquiroz (approved):
LGTM

PR #3809, opened 2024-05-10
Remember configuration choice after revert

toddburnside:
This seems to fix the issue. It is a complex section, so hopefully it doesn't cause additional problems...

The second commit fixes a problem where a local, unlawful Order instance was being used for an === comparison. The result was that a view was not getting updated and the `Accept Configuration` button was not getting enabled sometimes when reverting a configuration.

### DIFF
--- a/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
+++ b/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
@@ -5,7 +5,6 @@ package explore.config
 
 import boopickle.DefaultBasic.*
 import cats.Eq
-import cats.Order
 import cats.data.*
 import cats.effect.*
 import cats.implicits.catsKernelOrderingForOrder
@@ -68,6 +67,7 @@ import lucuma.ui.utils.*
 import java.text.DecimalFormat
 import scala.collection.decorators.*
 import scala.concurrent.duration.*
+import scala.math.Ordering.OptionOrdering
 
 import scalajs.js
 import scalajs.js.JSConverters.*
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -294,7 +294,10 @@ private object SpectroscopyModesTable:
       column(AvailablityColumnId, row => row.rowToConf(cw))
         .setCell(_.value.fold("No")(_ => "Yes"))
         .setColumnSize(FixedSize(66.toPx))
-        .sortable
+        .sortableBy(
+          // None are first, just like in Ordering[Option[?]]
+          _.fold("")(_.configuration.configurationSummary)
```
rpiaggio, line 299: Can we just use `sortableBy(_.map(_.configuration.configurationSummary))` and avoid the `fold`?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -465,7 +468,15 @@ private object SpectroscopyModesTable:
         table.getSortedRowModel().rows.map(_.original).toList
       }
       // selectedRow
-      .useState(none[SpectroscopyModeRow])
+      .useStateBy((props, _, _, rows, _, _, _, _, _, _, _) =>
+        props.selectedConfig.get
```
rpiaggio, line 472: `props.selectedConfig` is still set after we revert then?

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
toddburnside: Yeah. Seems like that should work. I'll try.

rpiaggio (approved):
LGTM

PR #3810, opened 2024-05-11
Extra spaces between observations

cquiroz:
Before:
image

After:
image

### DIFF
--- a/explore/src/main/scala/explore/observationtree/ObsList.scala
+++ b/explore/src/main/scala/explore/observationtree/ObsList.scala
@@ -352,7 +352,14 @@ object ObsList:
               <.div(
                 ^.overflow := "auto",
                 Tree(
-                  treeNodes.get,
+                  treeNodes.get.filter(_.data match {
+                    case Left(GroupTree.Obs(id)) =>
+                      props.observations.get
+                        .getValue(id)
+                        .map(o => !o.deleted)
+                        .getOrElse(false)
+                    case _                       => true
+                  }),
                   renderItem,
                   expandedKeys = expandedGroups.get,
                   onToggle = expandedGroups.set,
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -1854,6 +1854,10 @@ input[type='file'].explore-fileupload {
     border: none;
     overflow-y: auto;
   }
+
+  .p-tree .p-treenode-droppoint {
+    height: 0;
+  }
 }
 
 // -------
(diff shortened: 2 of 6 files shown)


rpiaggio (approved):
LGTM, but out of curiosity: why is the ODB returning deleted observations if the default is `includeDeleted: Boolean! = false` ?

cquiroz:
I also wondered that as well. We have two queries one for obs which works as expected but the one getting groups brings them all, I couldn't find a way to filter them

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -352,7 +352,14 @@ object ObsList:
               <.div(
                 ^.overflow := "auto",
                 Tree(
-                  treeNodes.get,
+                  treeNodes.get.filter(_.data match {
+                    case Left(GroupTree.Obs(id)) =>
+                      props.observations.get
+                        .getValue(id)
+                        .map(o => !o.deleted)
+                        .getOrElse(false)
```
hugo-vrijswijk, line 360:
```suggestion
                        .exists(o => !o.deleted)
```

hugo-vrijswijk (commented):
Interesting, I thought nodes without existence were already filtered out in GroupTree.scala. Maybe it makes more sense to fix the bug there and not create the nodes in the first place?

Update:
We do check for existence, but I forgot to add the check for root groups. 

image

cquiroz:
> Interesting, I thought nodes without existence were already filtered out in GroupTree.scala. Maybe it makes more sense to fix the bug there and not create the nodes in the first place?
> 
> Update: We do check for existence, but I forgot to add the check for root groups.
> image

Do you think it maybe better to fix it this way?
We can also consider whether the backend should not return deleted nodes in the first place. It would imply less traffic as well

hugo-vrijswijk:
> > Interesting, I thought nodes without existence were already filtered out in GroupTree.scala. Maybe it makes more sense to fix the bug there and not create the nodes in the first place?
> > Update: We do check for existence, but I forgot to add the check for root groups.
> > image
> 
> Do you think it maybe better to fix it this way? We can also consider whether the backend should not return deleted nodes in the first place. It would imply less traffic as well

It seems like a cleaner solution to not _create_ the nodes we don't want to show, rather than trying to hide them and adding extra CSS. Existence is already queried in the group elements query, so that saves adding it to the query in this PR. 

I think the cleanest would definitely be not returning non-existent elements in the group elements query.

cquiroz:
I agree, let's make a sample query to let the odb team fix it

cquiroz:
https://github.com/gemini-hlsw/lucuma-odb/issues/1182

cquiroz:
This was fixing upstream, closing...

PR #3813, opened 2024-05-15
Update pwa plugin

cquiroz:
Supersedes #3800

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,10 @@ http://local.lucuma.xyz:5000/
 
 Besides doing a full deployment you need to comment out the condition about local.lucuma.xyz on main.jsx
 
-Normally you want to test changes thus you need to do even a tiny change and build again to trigger an update prompt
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -60,6 +60,12 @@ npx vite build && npx vite preview
 As before you can now see the app locally but in a different port
 http://local.lucuma.xyz:5000/
 
+## Testing PWA locally
+
+Besides doing a full deployment you need to comment out the condition about local.lucuma.xyz on main.jsx
+
+Normally you want to test changes thus you need to do even a tiny change and build again to trigger an update prompt
```
toddburnside, line 67: Should the service worker be unregistered manually after testing is done or something? I don't remember what I did.

toddburnside (approved):
LGTM

README.md
cquiroz: you could, since it gets registered in a different port it should be ok, but I could mention it

PR #3814, opened 2024-05-16
Fix issue of broken Semester Plot

toddburnside:
This fixes an issue where switching between GS and GN on a semester plot would break the semester plot and no elevation line would appear even when switching back to the original site. In the case of breakage, the console would show this error:
```
Uncaught TypeError: Cannot read properties of undefined (reading '0')
```
Even reloading would not fix the problem, and in addition to the above error you would also get:
```
scala.scalajs.js.JavaScriptException: TypeError: Cannot read properties of undefined (reading 'data')
```
The fact that reloading wouldn't fix the issue was surprising. The code involves one of the service workers, so maybe that has something to do with it?

However, minimizing the elevation plot tile or switching to `Night` and then back to `Semester` would "fix" it.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
+++ b/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
@@ -159,7 +159,7 @@ object ElevationPlotSection:
                   options.get,
                   props.coords,
                   windowsNetExcludeIntervals
-                )
+                ).withKey(s"${options.get}-${props.coords}")
           },
           <.div(
             ExploreStyles.ElevationPlotControls,
--- a/explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
+++ b/explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
@@ -24,6 +24,7 @@ import lucuma.react.highcharts.ResizingChart
 import lucuma.react.resizeDetector.hooks.*
 import lucuma.typed.highcharts.highchartsStrings.line
 import lucuma.typed.highcharts.mod.*
+import lucuma.ui.reusability.given
 import lucuma.ui.syntax.all.given
 
 import java.time.Duration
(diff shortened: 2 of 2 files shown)


explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
```diff
@@ -159,7 +159,7 @@ object ElevationPlotSection:
                   options.get,
                   props.coords,
                   windowsNetExcludeIntervals
-                )
+                ).withKey(s"${options.get}-${props.coords}")
```
toddburnside, line 162: This is the change that actually fixed the issue.

explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
```diff
@@ -57,111 +58,108 @@ object ElevationPlotSemester:
       .useContext(AppContext.ctx)
       .useResizeDetector()
       .useSerialState(none[Chart_])
-      .useEffectWithDepsBy((props, _, _, chartOpt) => chartOpt.value.isDefined)(
-        (props, ctx, _, chartOpt) =>
-          _ =>
-            chartOpt.value.value
-              .map { chart =>
-                import ctx.given
+      .useEffectWithDepsBy((props, _, _, chartOpt) =>
```
toddburnside, line 61: The changes in this file did not seem to affect this problem, but it seems more correct to me to have the effect depend on the values that it uses. Using `Hide whitespace` helps show what actually changed.

explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
```diff
@@ -57,111 +58,108 @@ object ElevationPlotSemester:
...
-                  // 2pm of the selected day, same as for semester start and end
-                  val localDateTime: LocalDateTime =
-                    LocalDateTime.of(date, LocalTime.MIDNIGHT).plusHours(14)
-                  val zonedDateTime: ZonedDateTime =
-                    ZonedDateTime.of(localDateTime, props.options.site.timezone)
+      .useEffectWithDepsBy((props, _, _, chartOpt) =>
+        (props.options.date, props.options.site, chartOpt)
+      )((_, _, _, _) =>
+        (date, site, chartOpt) =>
+          CallbackTo(chartOpt.value.value) >>=
```
cquiroz, line 138: Why is this in a callback?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
toddburnside: I didn't change this, but since the whole body needs to be an effect, this was one way to do that.

cquiroz: can it be just inlined in the next block?

toddburnside: Yeah, it should be able to work like the `useEffect` above. I'll do that when I can get back on this branch.

explore/src/main/scala/explore/targeteditor/ElevationPlotSemester.scala
```diff
@@ -57,111 +58,108 @@ object ElevationPlotSemester:
...
+      )((_, ctx, _, chartOpt) =>
+        (semester, site, coords, _) =>
+          chartOpt.value.value
+            .map { chart =>
+              import ctx.given
 
-                val series = chart.series(0)
-                val xAxis  = chart.xAxis(0)
+              val series = chart.series(0)
+              val xAxis  = chart.xAxis(0)
```
toddburnside, line 71: All right, this is fun! I'm still getting the errors if I do a browser reload directly on a page with the semester plot, such as `https://local.lucuma.xyz:8080/p-103/targets/obs/o-213/target/t-505` or `https://local.lucuma.xyz:8080/p-103/observation/o-213/target/t-505`. However, if I then switch sites or observations or anything, it works from then on. I put some `println`s here, and it turns out that under circumstances where we get the errors, we have times where `chartOpt` is defined, but `chart.series` and `chart.xAxis` are NOT. 

It really seems like this function and the render function are called way too many times when loading the page OR just changing the site or something. Since the `chart` in `chartOpt` is mutable, maybe we're having problems with the repeated calls?

I switched back to `react 17` and the problem on initial load went away. However, without the new `withKey` in `ElevationPlotSection`, the problem with switching sites, etc. still persisted. In this case, however, there were never cases where `series` and `xAxis` were undefined here, so it is a slightly different problem. But, it did work when changing observations. Even with react 17 this method and render are called dozens of times on load and many times when switching sites, etc.

In summary, `react 17` plus the change in ElevationPlotSection completely fixes the problem. So, I'm guessing that another `withKey` might be needed somewhere for react 18? But, if we have to keep chasing `withKey`s for react 18, we could be in for a long ride.

toddburnside: Well, I just went back to react 17 to test something again, and this time it did NOT fix the error on load. 🤯 

However, if I switch to react 17, or just go back to `renderIntoDom` in Explore.scala, I have been unable to reproduce the "double click on target table" issue that Carlos is investigating.

toddburnside:
Superseded by #3820

PR #3816, opened 2024-05-19
Improve reset layouts toast

cquiroz:


explore/src/main/scala/explore/users/UserPreferencesPopup.scala
```diff
@@ -194,6 +194,24 @@ object UserPreferencesContent:
       ) =>
         import ctx.given
 
+        val layoutsPopup = ctx.toastRef.get.flatMap(
```
cquiroz, line 197: @rpiaggio I tried `useToastRef` and it all compiles but no toast is shown 🤷

rpiaggio: I'll take a look

rpiaggio: After taking a look, I remembered that `useToastRef` is only used to create a ref to use wherever the `Toast` is defined.

After that, a wrapper `ToastCtx` is stored in the `AppContext`, which can be obtained with the `ToastCtx[F]` summoner when an implicit `AppContext` is in scope.

I changed the code to make use of this, as well as adding a method to the `ToastCtx` API that receives a `MessageItem`.

I inadvertently pushed to this branch instead of creating a new one, sorry for that.

rpiaggio (approved):
LGTM, but someone should review the changes I made.

explore/src/main/scala/explore/users/UserPreferencesPopup.scala
cquiroz: Cool thanks

cquiroz:
> LGTM, but someone should review the changes I made.

They seem fine to me

PR #3818, opened 2024-05-19
Keep expanded groups when deleting an observation

cquiroz:


explore/src/main/scala/explore/components/FocusedStatus.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+object FocusedStatus:
+  private type Props = FocusedStatus
+
+  private val component =
+    ScalaFnComponent
+      .withHooks[Props]
+      .useContext(AppContext.ctx)
+      .render: (props, ctx) =>
+        if (LinkingInfo.developmentMode)
```
rpiaggio, line 32: I would put the `if (LinkingInfo.developmentMode)` at use sites. That will tree-shake the whole component off in production.

rpiaggio (approved):
LGTM

PR #3819, opened 2024-05-20
Use angle from PosAngleConstraint for finder chart `Align to PA`

toddburnside:
@cquiroz You put in this shortcut issue. The `Align to PA` button seemed to work, with a couple of issues:

1. It didn't take into account the angle specified in the PosAngleConstraint. So, I added a check for that before trying to calculate the PA for the Viz Time.
2. If there was no PA available (can't be calculated), the `Align to PA` did nothing. I removed the button if there isn't a PA so that it isn't confusing. At least it seems less confusing to me.

Another thing we could do would be to use the current time if the Viz Time isn't set.

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -44,10 +44,8 @@ import japgolly.scalajs.react.extra.router.SetRouteVia
 import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.ags.AgsAnalysis
 import lucuma.core.math.Angle
-import lucuma.core.math.Coordinates
 import lucuma.core.math.Offset
 import lucuma.core.math.skycalc.averageParallacticAngle
-import lucuma.core.math.skycalc.parallacticAngle
 import lucuma.core.model.ConstraintSet
 import lucuma.core.model.CoordinatesAtVizTime
 import lucuma.core.model.ObsAttachment as ObsAtt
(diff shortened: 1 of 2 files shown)


cquiroz:
Yes I think we should default to current time if VizTime isn't set

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -340,9 +346,13 @@ object ObsTabTiles:
             }
 
           val pa: Option[Angle] =
-            (basicConfiguration.map(_.siteFor), asterismAsNel, vizTime)
-              .mapN((site, asterism, vizTime) =>
-                parallacticAngle(site, asterism.baseTracking, vizTime)
+            PosAngleConstraint.angle
```
toddburnside, line 349: We could also use the default viz time for the `targetCoords` above, which is used for a bunch of the other tiles. Not sure if that makes sense or not. Currently base coords are used if there isn't a viz time.

cquiroz: I think it is better coordinates corrected at the current time than the base coords

toddburnside: So, I should change all of them to use the current time if there isn't a viz time?

cquiroz: in general I'd say yes, note that there is a tagged type for `CoordinatesAtVizTime` so you know if the coordinates have been corrected.
Now in certain cases the effect of pm correction is quite small and may not matter as much. I think a case are the elevation plots where coordinate changes have little effect on the elevation

cquiroz (approved):
LGTM

PR #3820, opened 2024-05-20
Rework charts to new facade

rpiaggio:
This PR is based on #3814, rebased on master.

The new facade requires `Reusable[Options]`, so now we keep each charts options in `useMemo` blocks.

Keys for remounting are not necessary now. Neither is resize tracking.


rpiaggio:
I recomment "Hide Whitespace" for reviewing.

toddburnside (approved):
LGTM. This appears to have fixed both the semester plot issue and the double click on the target summary table issue.

And, the semester plots seem to load faster, too!

Thanks for taking care of this!

cquiroz (approved):
LGTM

PR #3821, opened 2024-05-21
Remove existence checks from allGroupElements query

hugo-vrijswijk:
~~Alternative way to fix #3810~~

Removes the checks that are no longer needed since the backend filters out deleted groups/observations


hugo-vrijswijk:
Actually I think since ODB no longer returns deleted observations this should no longer be necessary

rpiaggio:
> Actually I think since ODB no longer returns deleted observations this should no longer be necessary

Agreed. Has the ODB behavior changed already?

hugo-vrijswijk:
> > Actually I think since ODB no longer returns deleted observations this should no longer be necessary
> 
> Agreed. Has the ODB behavior changed already?

I think so yes. I'll update this PR tomorrow to remove the existence from the query and the existing check

rpiaggio (approved):
LGTM

PR #3827, opened 2024-05-21
Revert "Bump primereact from 10.5.1 to 10.6.5"

hugo-vrijswijk:
Reverts gemini-hlsw/explore#3825

image


cquiroz:
Could this be fixed in css instead?

PR #3829, opened 2024-05-22
Remove logic to keep resizing state

cquiroz:
The logic was somewhat convoluted and had accumulated for a while.
This PR removes having an isResizing state simplifying the logic and apparently fixing a bug
This PR needs more user testing


model/shared/src/main/scala/explore/model/enums/TileSizeState.scala
```diff
@@ -8,12 +8,12 @@ import cats.Eq
 enum TileSizeState:
   case Maximized extends TileSizeState
   case Minimized extends TileSizeState
-  case Normal    extends TileSizeState
+  // case Normal    extends TileSizeState
```
toddburnside, line 11: Commented code?

toddburnside (approved):
LGTM

PR #3830, opened 2024-05-22
Adds strict mode around the main panel.

toddburnside:
The strict mode is around the part of the app rendered via the ROUTER. 

NOTE: This is for testing only!!! It breaks Explore.


rpiaggio (approved):
Now that we have solved the issues, I think we should have this enabled for development.

LGTM

toddburnside:
> Now that we have solved the issues, I think we should have this enabled for development.

I wonder if we should put StrictMode around the entire app, not just the main panel. I just put it around the main panel because it was easy and produced the errors we had been seeing.

rpiaggio:
> > Now that we have solved the issues, I think we should have this enabled for development.
> 
> I wonder if we should put StrictMode around the entire app, not just the main panel. I just put it around the main panel because it was easy and produced the errors we had been seeing.

Yes, good idea.

PR #3831, opened 2024-05-23
Temporary fix until the CfP changes in the API can be integrated.

toddburnside:
This makes explore usable with the new API changes for Calls for Proposals. The exception is Proposals. Trying to do anything with proposals in the UI will be a source of pain and suffering.


cquiroz (approved):
LGTM

PR #3833, opened 2024-05-24
Support deleting groups

hugo-vrijswijk:
- New delete button for groups (disabled if the group is non-empty, with a tooltip to explain)
- Improves UX for moving elements by increasing the dropzone height and using different colours when hovering

https://github.com/gemini-hlsw/explore/assets/10114577/7fa7dff1-d2ba-48e5-8777-d79cacdbf00f


toddburnside (approved):
LGTM

PR #3841, opened 2024-05-27
Fix crash when double clicking on the target table

toddburnside:
There were actually (at least) 2 issues that caused the target table to crash when clicking.

Raúl fixed one by rewriting our highcharts facade. This is the one that was actually demonstrated in the video in the referenced Shortcut ticket.

However, a second issue was discovered. This PR seems to fix that one.

### DIFF
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -455,7 +455,7 @@ object TargetTabContents extends TwoPanels:
       .useContext(AppContext.ctx)
       // Two panel state
       .useStateView[SelectedPanel](SelectedPanel.Uninitialized)
-      .useEffectWithDepsBy((props, _, state) => props.focused) { (_, _, selected) => focused =>
+      .useEffectWithDepsBy((props, _, _) => props.focused) { (_, _, selected) => focused =>
         (focused, selected.get) match
           case (Focused(Some(_), _, _), _)                    => selected.set(SelectedPanel.Editor)
           case (Focused(None, Some(_), _), _)                 => selected.set(SelectedPanel.Editor)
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -457,14 +455,13 @@ object TargetTabContents extends TwoPanels:
...
       .useStateView[SelectedPanel](SelectedPanel.Uninitialized)
-      .useEffectWithDepsBy((props, _, state) => (props.focused, state.reuseByValue)) {
-        (_, _, selected) => (focused, _) =>
-          (focused, selected.get) match
-            case (Focused(Some(_), _, _), _)                    => selected.set(SelectedPanel.Editor)
-            case (Focused(None, Some(_), _), _)                 => selected.set(SelectedPanel.Editor)
-            case (Focused(None, None, _), SelectedPanel.Editor) =>
-              selected.set(SelectedPanel.Summary)
-            case _                                              => Callback.empty
+      .useEffectWithDepsBy((props, _, state) => props.focused) { (_, _, selected) => focused =>
```
cquiroz, line 458: Weird, we don't want to run this if `state` changes? Not sure how this ends up producing the infinite loop

toddburnside: Well, `state` is the same as `selected`, which we set in the method. So, we're both depending on it and setting it. I guess it usually `settles down` to a value, but the double clicking upsets that somehow. I should change `state` to `_`.

toddburnside: I haven't seen a case where I think removing it has affected the behavior. But, I might be missing something. 

We actually do the same thing in `ObsTabContents`. I'll see if that still works if I remove it.

cquiroz (approved):
LGTM, thanks a lot for tracking this down

PR #3843, opened 2024-05-28
Keep minimized target tiles from persisting.

toddburnside:
As mentioned in the code comment, the `minWidth larger than item width/maxWidth` still happens on occasion on the targets tab when doing things like double clicking on the target summary table. The tiles minimize but it is transient and they restore to normal, probably from the local saved state. And, it is never persisted to User Preferences.

It would be nice to get rid of the error and transient minimization, but it seems to be happening within the ResponsiveReactGridLayout.

### DIFF
--- a/common/src/main/scala/explore/model/layout.scala
+++ b/common/src/main/scala/explore/model/layout.scala
@@ -89,6 +89,15 @@ object layout {
       }(l)
     }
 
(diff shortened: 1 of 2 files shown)


cquiroz (approved):
Thanks for tracking this. LGTM

PR #3858, opened 2024-05-30
Basic support for the new proposal model

cquiroz:
There is quite a bit of commented code used for proposal times and partner splits


explore/src/clue/scala/queries/common/ProgramDetailsSubquery.scala
```diff
@@ -14,8 +14,9 @@ object ProgramDetailsSubquery
   override val subquery: String = s"""
     {
       type
-      proposalStatus
       pi $ProgramUserSubquery
+      proposal $ProposalSubquery
+      proposalStatus
```
toddburnside, line 19: Get rid of the comment below?

explore/src/main/scala/explore/common/ProposalQueries.scala
```diff
@@ -0,0 +1,122 @@
...
+import lucuma.schemas.ObservationDB.Types.FastTurnaroundInput
+import lucuma.schemas.ObservationDB.Types.LargeProgramInput
+import lucuma.schemas.ObservationDB.Types.PoorWeatherInput
+import lucuma.schemas.ObservationDB.Types.ProposalPropertiesInput
+import lucuma.schemas.ObservationDB.Types.ProposalTypeInput
+import lucuma.schemas.ObservationDB.Types.QueueInput
+import lucuma.schemas.ObservationDB.Types.SystemVerificationInput
+
+trait ProposalQueries:
+  private def toOAUpdater(f: Input[ToOActivation] => Input[ToOActivation]) =
```
rpiaggio, line 24: Nit: This can be a `val` with type `Endo[Input[ToOActivation]] => Endo[ProposalTypeInput]`.

That way, you can omit the `f` altogether, just `X.modify >>> Y.modify ...`.

PR #3867, opened 2024-05-31
Final (hopefully) fix for minimized tile bug on target tab

toddburnside:
This should address the `minWidth larger than item width/maxWidth` error on the targets tab. I had more extensive changes, but it turns out that this is all that is needed. 

This reverts the changes from #3843 as they are no longer necessary.

### DIFF
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -155,7 +155,7 @@ object TargetTabContents extends TwoPanels:
     /**
      * Render the summary table.
      */
-    def renderSummary(single: Boolean): Tile =
+    def renderSummary: Tile =
       Tile(
         ObsTabTilesIds.TargetSummaryId.id,
         "Target Summary",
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -366,48 +366,60 @@ object TargetTabContents extends TwoPanels:
...
         GridLayoutSection.TargetLayout,
         backButton.some,
-        Option.when(tileList.isEmpty)(ExploreStyles.SingleTileMaximized),
-        storeLayout = tileList.nonEmpty
-      ).withKey(if (obsSelected) "target-obs-controller" else "target-summary-controller"): VdomNode
-      // We need different tile controller keys when in observations than when in target summary,
-      // so that it clears its internal state.
+        Option.when(tileListKeyOpt.isEmpty)(ExploreStyles.SingleTileMaximized),
+        storeLayout = tileListKeyOpt.nonEmpty
+      ).withKey(key): VdomNode
```
cquiroz, line 420: I guess this is the key finding. Could it happen somewhere else?

cquiroz (approved):
Thanks a lot. LGTM

explore/src/main/scala/explore/tabs/TargetTabContents.scala
toddburnside: Well, yes. We've been adding a lot of 'withKey`s lately.

toddburnside: Ah, you mean in other instances of the tile controller. Currently, this is the only place where we switch between layouts in the same tile controller. If we do it elsewhere, it could be a problem there, as well.

PR #3873, opened 2024-06-03
Add partner to invite form

hugo-vrijswijk:


explore/src/main/scala/explore/proposal/InviteUserPopup.scala
```diff
@@ -93,11 +99,17 @@ object InviteUserPopup:
...
                   id = "email-invite".refined,
                   value = emailView,
-                  label = "email",
+                  label = "Email",
                   disabled = inviteState.get === CreateInviteProcess.Running,
                   validFormat = MailValidator.optional,
                   onValidChange = v => validEmail.setState(v)
-                )(^.autoComplete := "off")
+                )(^.autoComplete := "off"),
+                FormEnumDropdownOptionalView(
```
cquiroz, line 107: I wonder if we can show the flags of each partner, It can be done later though

cquiroz (approved):
LGTM

explore/src/main/scala/explore/proposal/InviteUserPopup.scala
hugo-vrijswijk: Yeah that would be great. But I'm not sure how since a `Display` works with strings, not VDomElements

rpiaggio: @hugo-vrijswijk There's a `Render` typeclass defined in `lucuma-ui`. It's the equivalent of `Display` but returns a `VdomNode`.

toddburnside: But `FormEnumDropdownOptionalView` takes a `Display`

rpiaggio: Shall we change that to a `Render` ?

hugo-vrijswijk: It feels like that would be a breaking change for a lot of places. Or is any Display a valid Render?

rpiaggio: It's not, but we can add an implicit conversion.

PR #3874, opened 2024-06-03
Replace lint-action

hugo-vrijswijk:
Annotations in github still work as that's built-in, but I'm not sure if that ever worked for prettier:

afbeelding


cquiroz (approved):
LGTM

PR #3876, opened 2024-06-05
More support new proposal model

cquiroz:
This PR adds support to change CfP, partners and too activation
Thanks to @rpiaggio to get some of the difficult parts of storing changes in db working


explore/src/main/scala/explore/common/ProposalQueries.scala
```diff
@@ -73,22 +61,25 @@ trait ProposalQueries:
             LargeProgramInput(
               toOActivation = toOActivation.assign,
               minPercentTime = minPercentTime.assign,
-              minPercentTotalTime = minPercentTotalTime.assign
-              // totalTime = totalTime.toInput.assign
+              minPercentTotalTime = minPercentTotalTime.orUnassign
+              // totalTime = totalTime.map(_.toInput).orUnassign
```
rpiaggio, line 65: Is this commented code to be addressed in a future PR?

explore/src/main/scala/explore/proposal/package.scala
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.proposal
+
+import lucuma.core.util.NewType
+
+object PartnersDialogState extends NewType[Boolean]:
```
rpiaggio, line 8: Can we use the `Visible` newtype instead of defining a new one? Or at least define the newtype on top of `Visible`.

rpiaggio (approved):
LGTM, with a comment regarding the newtype.

explore/src/main/scala/explore/common/ProposalQueries.scala
cquiroz: Yes, times is the last piece

explore/src/main/scala/explore/proposal/package.scala
cquiroz: Good idea

PR #3881, opened 2024-06-05
Create TimeSpanView

hugo-vrijswijk:


common/src/main/scala/explore/components/TimeSpanView.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package explore.components
+
+import explore.model.syntax.all.toHoursMinutes
+import japgolly.scalajs.react.*
+import japgolly.scalajs.react.vdom.html_<^.*
+import lucuma.core.util.TimeSpan
+import lucuma.react.common.ReactFnProps
+
+case class TimeSpanView(timespan: TimeSpan, modifiers: Seq[TagMod] = Seq.empty)
```
cquiroz, line 12: I guess this will eventually go to `lucuma-ui`

cquiroz (approved):
LGTM

common/src/main/scala/explore/components/TimeSpanView.scala
hugo-vrijswijk: Yeah good idea I should probably put this there

PR #3882, opened 2024-06-06
Time controls

cquiroz:
This PR restores the time controls though one related LP is pending clarification
I may want to do some UI work but it will done later so we can test up to this point


toddburnside (approved):
LGTM!

PR #3892, opened 2024-06-09
A few UI touches

cquiroz:
make finder charts and target search read only when needed

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -2541,7 +2541,13 @@ a.explore-upgrade-link {
   display: flex;
 
   .p-dropdown {
-    min-width: 20ch;
+    min-width: 5ch;
+    max-width: 12ch;
+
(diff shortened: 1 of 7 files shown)


common/src/main/webapp/sass/explore.scss
```diff
@@ -2541,7 +2541,13 @@ a.explore-upgrade-link {
   display: flex;
 
   .p-dropdown {
-    min-width: 20ch;
+    min-width: 5ch;
+    max-width: 12ch;
+
+    @media (min-width: lucumaUICommon.$mobile-responsive-cutoff) {
```
rpiaggio, line 2547: I'm not sure it would help here, but CSS now supports container queries.

rpiaggio (approved):
LGTM

common/src/main/webapp/sass/explore.scss
cquiroz: Not for this but I'm considering container queries for the tiles

PR #3911, opened 2024-06-14
Initial observation validations

toddburnside:
The first step in display observation validations in explore. Currently, if there are validation errors, it shows an error icon with a tooltip in the observation badges. 
image

The next step will be showing the full list in the `Warnings and Errors` tile.


cquiroz:
I seem to remember in the mockups there would be a tile with errors and warnings. Which can be besides your implementation not necessarily instead

toddburnside:
> I seem to remember in the mockups there would be a tile with errors and warnings. Which can be besides your implementation not necessarily instead

Right. It is on the Overview tab and is what I am working on next.

PR #3913, opened 2024-06-15
Support opening programs without a pi

cquiroz:
Staff programs don't have a pi

### DIFF
--- a/model/shared/src/main/scala/explore/model/ProgramDetails.scala
+++ b/model/shared/src/main/scala/explore/model/ProgramDetails.scala
@@ -4,7 +4,6 @@
 package explore.model
 
 import cats.Eq
-import cats.data.NonEmptyList
 import cats.derived.*
 import io.circe.Decoder
 import io.circe.generic.semiauto.*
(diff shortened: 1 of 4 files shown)


model/shared/src/main/scala/explore/model/ProgramDetails.scala
```diff
@@ -18,30 +17,30 @@ case class ProgramDetails(
...
   proposal:       Option[Proposal],
   proposalStatus: ProposalStatus,
-  pi:             ProgramUserWithRole,
+  pi:             Option[ProgramUserWithRole],
   users:          List[ProgramUserWithRole],
   invitations:    List[CoIInvitation],
   reference:      Option[ProgramReference]
 ) derives Eq:
-  val allUsers: NonEmptyList[ProgramUserWithRole] = NonEmptyList(pi, users)
+  val allUsers: List[ProgramUserWithRole] = pi.fold(users)(_ :: users)
```
hugo-vrijswijk, line 25: Isn't this concatenating users with users?

cquiroz: pi with co-pi

toddburnside (approved):
LGTM

PR #3918, opened 2024-06-17
Update resizer, enable in sequence.

rpiaggio:
The maximum update depth exceeded error seems to be fixed now.


toddburnside (approved):
LGTM - Thanks for figuring this out and fixing it! So, usage is the same as the previous hook?

rpiaggio:
> LGTM - Thanks for figuring this out and fixing it! So, usage is the same as the previous hook?

Yes, it's pretty much the same. It has some limitations (like measuring the bounding box instead of the content box), but we are not using any of the missing functionality. If need be, we can add it.

PR #3920, opened 2024-06-18
Add Observation Validations Table

toddburnside:
Adds the observation validations table to the Overview tab. Each observation can potentially have errors in multiple categories and each category can have multiple messages. I wanted to do this in a table in some way that made sense and didn't take up more rows than necessary. I came up with the table below. Note that this is a bit of a contrived example - in most cases an observation probably won't have this many validations.
image

All the (potential) subrows for a validation expand and collapse as a unit. When collapsed the categories and messages are concatenated into the one line. There is a tooltop for the `Category` with a longer description. The tile title has buttons for expanding all and collapsing all. 
If an observation has a single validation with a single message, it occupies one row and there is nothing to expand. I think this will be a fairly common case.

### DIFF
--- a/common/src/main/scala/explore/Icons.scala
+++ b/common/src/main/scala/explore/Icons.scala
@@ -326,6 +326,14 @@ object Icons {
   @JSImport("@fortawesome/pro-regular-svg-icons", "faAngle")
   val faAngle: FAIcon = js.native
 
+  @js.native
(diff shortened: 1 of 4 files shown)


cquiroz:
Looks good

cquiroz (approved):
LGTM

toddburnside:
Oh, I'm also not currently saving the table state. I thought I'd get some more feedback on it before doing that.

PR #3921, opened 2024-06-19
Avoid table remounts in ObsSummaryTable and ObsAttachmentsTable

rpiaggio:
In ObsSumaryTable all data was moved to rows, in ObsAttachmentsTable we make use of `meta`.

### DIFF
--- a/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
+++ b/explore/src/main/scala/explore/attachments/ObsAttachmentsTable.scala
@@ -65,14 +65,21 @@ case class ObsAttachmentsTable(
   authToken:                NonEmptyString,
   obsAttachments:           View[ObsAttachmentList],
   obsAttachmentAssignments: ObsAttachmentAssignmentMap,
-  readonly:                 Boolean,
+  readOnly:                 Boolean,
   renderInTitle:            Tile.RenderInTitle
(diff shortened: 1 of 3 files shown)


hugo-vrijswijk (approved):
LGTM. I guess the main thing is not having to remount because only the meta changes?

rpiaggio:
> LGTM. I guess the main thing is not having to remount because only the meta changes?

The components are remounted if the columns definitions change. So, the idea here is to avoid redefining the columns based on dependencies. If either the meta or the rows change, it will induce a rerender but not a remount.

toddburnside (approved):
LGTM. Thanks for taking care of this. It's nice to have an example for using table metadata now.

The summary seems to be: Recalculating rows or the TableOptions in `useReactTableBy` doesn't force a remount, but changing the columns does.

toddburnside:
> LGTM. Thanks for taking care of this. It's nice to have an example for using table metadata now.
> 
> The summary seems to be: Recalculating rows or the TableOptions in `useReactTableBy` doesn't force a remount, but changing the columns does.

I see you already answered this above.

rpiaggio:
Right. Ideally, the cols should be defined in a static `val`. Come thing of it, in that case maybe we don't even need to `useMemo` on it. I'll try that in the next PR.

PR #3922, opened 2024-06-19
Fix aladin container with new resizer

rpiaggio:
I don't know how this was working before. The same div that was being measured got remounted when size changed because the size was in the key.

I removed the key now, avoiding remounts, and instead calling `fixLayoutDimensions` when size changes (or `showCatalog` changes, which requires a redraw).

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
+++ b/explore/src/main/scala/explore/targeteditor/AladinContainer.scala
@@ -396,25 +396,23 @@ object AladinContainer extends AladinCommon {
                     _
                   )
                 ),
-              AladinComp
-                .withRef(aladinRef) {
-                  // .withKey(key) {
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -258,173 +254,168 @@ object AladinContainer extends AladinCommon {
...
+                    _,
+                    _,
+                    _,
+                    screenOffset,
+                    _
+                  )
+                ),
+              AladinComp
+                .withRef(aladinRef) {
+                  // .withKey(key) {
```
toddburnside, line 401: remove comment?

toddburnside (approved):
LGTM
> I don't know how this was working before.

Gotta love magic!

PR #3927, opened 2024-06-19
Avoid remounts in FinderChartLinker

rpiaggio:
I also removed the `useResizeDetector`. In other tables we use it for `useDynTable`, but here it seemed we were not using the size at all.


toddburnside (approved):
LGTM

PR #3930, opened 2024-06-20
Use container queries instead of ResponsiveComponent

cquiroz:
We had a custom component attached to tiles to change style depending on tile size
This can now be done with container queries and we also avoid one use of a resizer hook


common/src/main/webapp/sass/explore.scss
```diff
@@ -43,6 +41,10 @@ $tree-section-width: 300px;
 }
 /* stylelint-enable selector-class-pattern */
 
+:root {
+  color-scheme: dark;
+}
+
```
rpiaggio, line 47: Do we want this commited or did you use it locally?

rpiaggio (approved):
LGTM, nice!

common/src/main/webapp/sass/explore.scss
cquiroz: It was moved from styles.less, I wanted to put the widths in css variables but that doesn't work

Issue #1483, opened 2022-01-25
Should we switch to sass


cquiroz:
It seems tooling is falling behind for `less`. We can't really abandon it due to the dependency with semantic ui but perhaps we can switch some/most of our less code to sass?

PR #3933, opened 2024-06-21
Remove old styles.less

cquiroz:
We still have some stylesheets defined with less.
This PR remove most of them


rpiaggio:
Can we close #1483  with this PR?

cquiroz:
We have one less file still present that uses more advanced features which needs to be ported

PR #3934, opened 2024-06-21
Avoid table remounts in SpectroscopyModesTable

rpiaggio:
This is the last one. It was a bit trickier than the previous ones.


cquiroz (approved):
LGTM, thanks a lot

PR #3935, opened 2024-06-21
Fix chart resize

rpiaggio:
Sizes as `100%` don't seem to play well together with flexboxes. The recommended way to do "`100%`" in flex seems to be `flex: 1 1 0;`.

This seems to fix at least the resizing issue with charts, where controls were sometimes lost when shrinking. Also, I removed a nesting level.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
+++ b/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
@@ -142,9 +142,9 @@ object ElevationPlotSection:
                 .map(BoundedInterval.fromInterval)
                 .flattenOption
 
-        <.div(ExploreStyles.ElevationPlotSection)(
+        React.Fragment(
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
```diff
@@ -142,9 +142,9 @@ object ElevationPlotSection:
                 .map(BoundedInterval.fromInterval)
                 .flattenOption
 
-        <.div(ExploreStyles.ElevationPlotSection)(
+        React.Fragment(
           HelpIcon("target/main/elevation-plot.md".refined, ExploreStyles.HelpIconFloating),
-          <.div(ExploreStyles.ElevationPlot) {
+          <.span(ExploreStyles.ElevationPlot)(
```
cquiroz, line 147: Why a `span`?

rpiaggio: I was just trying things, this can be reverted, I'll push a commit

PR #3938, opened 2024-06-22
Better UI for tiny chart tiles

cquiroz:
Progressively hide items as it gets smaller

https://github.com/gemini-hlsw/explore/assets/3615303/0d439b12-de77-42ba-b5d4-5c3114baadbb


rpiaggio (approved):
LGTM

PR #3943, opened 2024-06-23
Remove last less file

cquiroz:


package.json
```diff
@@ -48,13 +46,10 @@
     "unplugin-fonts": "^1.1.1",
     "vite": "4.5.0",
     "vite-plugin-mkcert": "^1.17.5",
-    "vite-plugin-pwa": "^0.20.0"
+    "vite-plugin-pwa": "^0.20.0",
+    "vite-plugin-env-compatible": "^2.0.1"
   },
   "scripts": {
-    "lint-dark": "sed '/^[\\.\\}\\@]/d;/^[[:blank:]]*\\..*/d;/\\/\\/.*/d' common/src/main/webapp/less/variables-dark.less > vars.css;trap \"rm vars.css\" EXIT;npx stylelint common/src/main/webapp/sass",
```
hugo-vrijswijk, line 54: Is the `sed` trick here no longer necessary? Because the script still seems useful to have

hugo-vrijswijk (commented):
LGTM other than 1 question 🎉

package.json
cquiroz: the variables file no longer exists. I could keep one for reference

hugo-vrijswijk: The `vars.css` is no longer needed? The CI check should probably also be updated then. But I meant the `stylelint` call

cquiroz: The current check is
`npx stylelint --formatter github common/src/main/webapp/sass`
It checks whatever is in that dir

hugo-vrijswijk: Yes but the script can still be useful to run locally, or to call it with `--fix` without having to remember what the exact command is

cquiroz: ok I'll restore them

PR #3947, opened 2024-06-25
Fix ITC chart display on resize

rpiaggio:
It seems that when we have a chart in a `flex` or `grid`, the `div` containing the chart should be a direct child of the `flex`/`grid` `div`. Otherwise, strange things seem to happen when HighCharts attempts to resize itself.

This PR rearranges things in the ITC tile to achieve this. The help icon is moved so that it's rendered together with the controls instead of the chart.

Also, the grid definition was simplified. Hiding of the details panel is now done by setting its size to `0` or `max-content` in `grid-template-columns`.


cquiroz (approved):
Thanks a lot for getting this done
LGTM

PR #3965, opened 2024-07-02
Don't allow targets for executed observations to be edited

toddburnside:
This implements clone on demand for targets when some associated observations have been executed, or readonly if all have been. It uses the same space that was in the Asterism for the `current only` or `all instances of target` choice. But, now the messages are a bit different depending on what would need cloning.

To be addressed in following PRs:

1. Neither the old nor this new cloning implementation play well with undo/redo. I will try to make the cloning undoable/redoable. This will require treating target undo/redo with obs assignment undo/redo as a single step. 
2. Prevent deleting of targets from the target summary table if any of the observations have been executed.
3. Prevent editing of asterisms for observations that have been cloned. This may require additional cloning behavior.

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -25,6 +25,8 @@ import explore.model.*
 import explore.model.AppContext
 import explore.model.LoadingState
 import explore.model.ObsSummary.observingMode
+import explore.model.ProgramSummaries
+import explore.model.TargetEditObsInfo
 import explore.model.display.given
 import explore.model.enums.AgsState
 import explore.model.enums.AppTab
@@ -38,6 +40,7 @@ import explore.modes.SpectroscopyModesMatrix
 import explore.observationtree.obsEditAttachments
 import explore.syntax.ui.*
 import explore.timingwindows.TimingWindowsPanel
+import explore.undo.UndoContext
 import explore.undo.UndoSetter
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.extra.router.SetRouteVia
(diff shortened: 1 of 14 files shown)


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -76,27 +79,31 @@ import java.time.Instant
...
+  searching:         View[Set[Target.Id]],
+  defaultLayouts:    LayoutsMap,
+  layouts:           LayoutsMap,
+  resize:            UseResizeDetectorReturn,
+  globalPreferences: View[GlobalPreferences],
+  readonly:          Boolean
 ) extends ReactFnProps(ObsTabTiles.component):
-  val obsId: Observation.Id = observation.get.id
+  val obsId: Observation.Id                                         = observation.get.id
+  val allConstraintSets: Set[ConstraintSet]                         = programSummaries.get.constraintGroups.map(_._2).toSet
```
cquiroz, line 98: Nice, I guess theses extra params accumulated over time

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -265,11 +260,45 @@ object TargetTabContents extends TwoPanels:
...
+        val obsIds4Url = ObsIdSet.fromSortedSet(idsToEdit.idSet.intersect(obsIdsToClone.idSet))
+        // all of the current groups that contain any of the obsIdsToClone
+        val allGroups  = props.programSummaries.model.get.asterismGroups
+          .filterNot(_._1.idSet.intersect(obsIdsToClone.idSet).isEmpty)
+          .map(_._1)
+          .toList
+          .distinct
+        // update the programSummaries
+        props.programSummaries.model.mod { ps =>
+          val newPs = ps.cloneTargetForObservations(oldTid, newTarget, obsIdsToClone)
```
cquiroz, line 286: I guess this could be inlined

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -265,11 +260,45 @@ object TargetTabContents extends TwoPanels:
...
+          newPs
+        } *>
+          setCurrentTarget(obsIds4Url)(newTarget.id.some, SetRouteVia.HistoryReplace) *>
+          // Deal with the expanded groups - we'll open all affected groups
+          allGroups.traverse { ids =>
+            val intersect = ids.idSet.intersect(obsIdsToClone.idSet)
+            println(s"Group Ids: $ids intersection: $intersect")
```
cquiroz, line 293: println

explore/src/main/scala/explore/targeteditor/TargetCloneSelector.scala
```diff
@@ -0,0 +1,69 @@
...
+        if (info.noMessages) <.div()
+        else
+          <.div(
+            ExploreStyles.SharedEditWarning,
+            info.message.map(_.value),
+            info.choice
+              .map((currentText, allText) =>
+                BooleanRadioButtons(
+                  view = editScope.as(EditScope.value),
+                  idBase = "editscope".refined,
```
cquiroz, line 62: Great name

cquiroz (approved):
LGTM

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
toddburnside: Part if it, I think, was trying to pass just what was needed. But, since we need to pass the ProgramSummaries now, might as well do it here.

explore/src/main/scala/explore/tabs/TargetTabContents.scala
toddburnside: Yeah, I think I had more printlns here for debugging purposes.

PR #3966, opened 2024-07-02
Restrict theme menu to the dev environment

cquiroz:


explore/src/main/scala/explore/TopBar.scala
```diff
@@ -115,43 +115,52 @@ object TopBar:
...
+                MenuItem.Item(
+                  label = "Debug",
+                  command = setLogLevel(LogLevelDesc.DEBUG),
+                  disabled = level === LogLevelDesc.DEBUG,
+                  icon = Icons.Bug
+                )
+              )
+              .some,
+            if (ctx.environment === ExecutionEnvironment.Development)
+              ThemeSubMenu(props.theme).some
```
toddburnside, line 152: Seems like this could be done with `visible` like for the other menu items. But, I guess that's more of a pain since ThemeSubMenu is in lucuma-ui.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/TopBar.scala
cquiroz: Right, we could do that, I just made it into an Option

PR #3968, opened 2024-07-03
Show the proposal id on the read-only message

cquiroz:
image

### DIFF
--- a/explore/src/main/scala/explore/ExploreLayout.scala
+++ b/explore/src/main/scala/explore/ExploreLayout.scala
@@ -172,7 +172,7 @@ object ExploreLayout:
 
           given Display[AppTab] = _.title
 
-          val (showProgsPopup, msg, isSubmitted, ref) =
+          val (showProgsPopup, msg, isSubmitted, proposalReference) =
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -172,15 +172,16 @@ object ExploreLayout:
 
           given Display[AppTab] = _.title
 
-          val (showProgsPopup, msg, isSubmitted) =
-            props.view.get.programSummaries.fold((false, none, false)) { pss =>
-              routingInfo.optProgramId.fold((true, none, false)) { id =>
+          val (showProgsPopup, msg, isSubmitted, ref) =
```
toddburnside, line 175: A different name than `ref` might be nice. Maybe `proposalRef`? When I read it I thought it referred to a react ref.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/ExploreLayout.scala
cquiroz: 👍

PR #3969, opened 2024-07-03
Fix deletion of magnitudes

toddburnside:


cquiroz (approved):
Nice, LGTM

explore/src/main/scala/explore/targeteditor/BrightnessesEditor.scala
```diff
@@ -66,7 +66,11 @@ sealed abstract class BrightnessesEditorBuilder[T, Props <: BrightnessesEditor[T
 
   private type RowValue = (Band, View[BrightnessMeasure[T]])
 
-  protected[targeteditor] case class TableMeta(disabled: Boolean)
+  protected[targeteditor] case class TableMeta(
+    // needs to be in the table meta because it is needed by the `delete` column.
+    brightnesses: View[SortedMap[Band, BrightnessMeasure[T]]],
```
rpiaggio, line 71: Thank you for fixing this. I overlooked this case when passing data to the meta.

One small comment: it seems that only the mod function is needed in the meta, not the actual value.

PR #3971, opened 2024-07-03
Improve sorting of Target and Duration columns

cquiroz:


explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -199,10 +202,28 @@ object ObsSummaryTable:
...
+                case s: String => <.span(s)
+                case (a, b)    => targetUrl(a, b)
+              }
+            }
+            .setSortingFn { (a, b, c) =>
+              val aʹ: String | (Observation.Id, TargetWithId) = a.getValue(c)
+              val aʹʹ                                         = aʹ match {
+                case s: String => s
+                case (_, b)    => b.target.name.value
+              }
```
rpiaggio, line 219: Maybe this can be in an `extension (name: String | (Observation.Id, TargetWithId))`?

Then we can simplify this to `a.getValue(c).compareTo(b.getValue(c))` (and avoid duplicate logic).

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -254,9 +275,14 @@ object ObsSummaryTable:
...
             DurationColumnId,
             _.execution.map(_.programTimeEstimate)
-          ).setCell: cell =>
+          ).setCell { cell =>
             cell.value.map:
               _.orSpinner(_.map(_.toHoursMinutes).orEmpty)
+          }.setSortingFn { (a, b, c) =>
+            val aʹ: Option[Pot[Option[TimeSpan]]] = a.getValue(c)
+            val bʹ: Option[Pot[Option[TimeSpan]]] = b.getValue(c)
+            tsOrder.compare(aʹ.flatMap(_.toOption).flatten, bʹ.flatMap(_.toOption).flatten)
```
rpiaggio, line 284: Does it work if we just `aʹ.flatMap(_.toOption).flatten.compareTo(bʹ.flatMap(_.toOption).flatten)` and avoid defining `tsOrder`?

Also, consider an extension for the `.flatMap(_.toOption).flatten` ?

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -199,10 +212,18 @@ object ObsSummaryTable:
...
+            r => (r.obsId, r.targetWithId)
           )
-            .setCell(_.value),
+            .setCell { c =>
+              c.value match {
+                case s: String => <.span(s)
+                case (a, b)    => targetUrl(a, b)
+              }
+            }
+            .setSortingFn { (a, b, c) =>
```
rpiaggio, line 224: I think this can be simplified to `.sortableBy(_.sortableValue)`.

rpiaggio (approved):
LGTM, consider using `sortableBy` (if it works).

PR #3976, opened 2024-07-07
Show the program id label if present

cquiroz:
image

### DIFF
--- a/explore/src/main/scala/explore/ExploreLayout.scala
+++ b/explore/src/main/scala/explore/ExploreLayout.scala
@@ -18,6 +18,7 @@ import explore.events.ExploreEvent.LogoutEventId
 import explore.model.*
 import explore.model.AppContext
 import explore.model.enums.AppTab
+import explore.model.reusability.given
 import explore.programs.ProgramsPopup
 import explore.shortcuts.*
 import explore.shortcuts.given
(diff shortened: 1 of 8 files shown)


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -64,6 +66,12 @@ object ExploreLayout:
       .withHooks[Props]
       .useContext(HelpContext.ctx)
       .useContext(AppContext.ctx)
+      .useEffectWithDepsBy((p, _, _) => p.view.zoom(RootModel.programReference).get)((_, _, _) =>
+        // Set the title of the page to the program reference
+        // scalajs-react suggest to use setTitle for this but we'd need to put the
+        // reference in the url for that
+        _.map(r => Callback(document.title = s"Explore - ${r.label}")).orEmpty
```
hugo-vrijswijk, line 73: Is it possible to get an update that removes the program id? If so this would keep the old one in the title

toddburnside: This would also be a problem when switching from a program with a program reference to a program without a reference, wouldn't it?

swalker2m:
I'm guessing that during the time after a proposal has been submitted but before it has been accepted, Explore should show the proposal id.  In other words, show the program id when present or else the proposal id if present.

PR #3983, opened 2024-07-10
Avoid captures and provide stable undo modders

rpiaggio:
It's desirable that state/view modification functions are stable, but this was not the case when undoer was involved, since it relied on capturing the old values.

This PR reworks the undo so that no capture is performed, relying on the new `View` functionality which now reports the previous value when it is modified.

We were also capturing values in a couple of other places, which are now fixed.

This PR also reverts #3969 and #3970, which implemented workarounds to propagate updated modder functions.

Review hints:
- Start with `UndoContext` and then `UndoSetter`.
- The changes in `TargetTable`, `TargetColumns` and `BrightnessEditor` are just reversals of the workarounds.
- The changes in `TargetTabContents` avoid capture of the target in the target editor shown below the target summary table.
- Other changes are adjustments for the new `View` API (basically reporting the previous value on mod).

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -5,3 +5,4 @@ GraphQLGen.scalaJSReactReuse=false
 GraphQLGen.catsShow=false
 
 OrganizeImports.removeUnused = false
+OrganizeImports.targetDialect = Scala3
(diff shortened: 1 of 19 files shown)


toddburnside (approved):
👍

PR #3991, opened 2024-07-13
Cleaning up tile controls

cquiroz:
Simplifies a bit the tile code and make the controls in different style consistest across each other and when minimized/normal

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
+++ b/explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
@@ -108,19 +108,19 @@ object AsterismEditor extends AsterismModifier:
 
         <.div(
           ExploreStyles.AladinFullScreen.when(fullScreen.get.value),
-          if (props.readonly) EmptyVdom
-          else
-            props.renderInTitle(
-              targetSelectionPopup(
-                "Add",
-                props.programId,
-                props.obsIds,
-                props.asterismIds,
-                props.allTargets.model,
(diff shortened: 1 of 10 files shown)


explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
```diff
@@ -108,19 +108,19 @@ object AsterismEditor extends AsterismModifier:
 
         <.div(
           ExploreStyles.AladinFullScreen.when(fullScreen.get.value),
-          if (props.readonly) EmptyVdom
-          else
```
toddburnside, line 112: Why do we want to display the target selection popup if they can't do anything with it?

cquiroz: It makes the layout simpler if we always know there is an element.
But yes we could hide it as well and tweak the css a bit more

toddburnside: Not a big deal, I guess. Keeping CSS simpler is a good goal.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/AsterismEditor.scala
toddburnside: Would an empty `span` work as well as the button for keeping the CSS simple?

PR #3995, opened 2024-07-15
Undo/Redo for target cloning

toddburnside:
The following things need to happen in all cases when a clone is created

1. Create the clone via the API, assign it to the specified observations, and make whatever changes to the target triggered the cloning. This is all done via a single call to `cloneTarget` mutation with the appropriate `Input`.
2. Add the target target to the local list of targets and swap the original target id for the cloned target id in the local list of ObsSummaries.

Depending on where the cloning was initiated from (target tab with one or more observations selected, target tab with a row in the summary table selected, or an observation selected on the observation tab) various other things need to be done. This always includes setting the URL, but can also include expanding groups in the AsterismGroupList. This is done via the `onCloneTarget` function passed to the SIderealTargetEditor. On creation this is called directly so that it happens in the same cycle as the cloning. If I waited to call this in the `onSet` of the action, when editing a asterism from the target tab, the URL would usually be incorrect. This is because the page would be set to the summary table in the meantime, and it would force the URL back to it.

For `undo`, everything is handled in the `TargetCloneAction`:

1. Remotely, the target needs to be deleted (I avoid that if someone in another session has assigned the new target to a different observation) and the asterism of all the affected observations need to have the clone removed and the original target added. 
2. Locally the clone is deleted and the asterisms are updated according. This is via the `onCloneTarget` method mentioned above.

`redo` is similar to undo, but in the reverse.

The core of the changes are in the cloning in `SiderealTargetEditor` and the `TargetCloneAction` that is used there. The `Fix problem of updating undostacks multiple times` commit also changes how the `Aligner` is created so that the `UndoSetter[Target.Sidereal]` that is passed in the props does not get updated when a clone will be performed. Before this change, the undo stacks got modified both by the change to the target AND by the cloning.

Next are the `onCloneTarget*` functions in `TargetTabContents` and `ObsTabTiles` that are passed to `SidereralTargetEditor` and on to `TargetCloneAction`.

For the initial clone and redoing, the URL seems to work correctly, however when undoing something about the cycle of events and the async nature of the undo often results in the TargetSummaryTable being displayed instead of the original target editor. I have decided that this is enough of a `best effort`.

### DIFF
--- a/explore/src/main/scala/explore/tabs/AsterismEditorTile.scala
+++ b/explore/src/main/scala/explore/tabs/AsterismEditorTile.scala
@@ -16,10 +16,13 @@ import explore.model.GlobalPreferences
 import explore.model.ObsConfiguration
 import explore.model.ObsIdSet
 import explore.model.ObsTabTilesIds
+import explore.model.OnCloneParameters
+import explore.model.ProgramSummaries
 import explore.model.TargetEditObsInfo
 import explore.model.TargetList
 import explore.model.enums.TileSizeState
 import explore.targeteditor.AsterismEditor
+import explore.undo.UndoContext
 import explore.undo.UndoSetter
 import japgolly.scalajs.react.Callback
 import japgolly.scalajs.react.extra.router.SetRouteVia
@@ -29,7 +32,6 @@ import lucuma.core.model.Target
 import lucuma.core.model.User
 import lucuma.schemas.ObservationDB
 import lucuma.schemas.model.BasicConfiguration
-import lucuma.schemas.model.TargetWithId
 import lucuma.ui.syntax.all.given
 import org.typelevel.log4cats.Logger
 import queries.schemas.odb.ObsQueries
@@ -44,12 +46,13 @@ object AsterismEditorTile:
     obsIds:            ObsIdSet,
     asterismIds:       View[AsterismIds],
     allTargets:        UndoSetter[TargetList],
+    programSummaries:  UndoContext[ProgramSummaries],
     configuration:     Option[BasicConfiguration],
     vizTime:           View[Option[Instant]],
     obsConf:           ObsConfiguration,
     currentTarget:     Option[Target.Id],
     setTarget:         (Option[Target.Id], SetRouteVia) => Callback,
-    onCloneTarget:     (Target.Id, TargetWithId, ObsIdSet) => Callback,
+    onCloneTarget:     OnCloneParameters => Callback,
     obsInfo:           Target.Id => TargetEditObsInfo,
     searching:         View[Set[Target.Id]],
     title:             String,
@@ -80,6 +83,7 @@ object AsterismEditorTile:
           obsIds,
           asterismIds,
           allTargets,
+          programSummaries,
           vizTime,
           obsConf,
           currentTarget,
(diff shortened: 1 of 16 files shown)


explore/src/main/scala/explore/tabs/AsterismEditorTile.scala
```diff
@@ -44,12 +46,13 @@ object AsterismEditorTile:
     obsIds:            ObsIdSet,
     asterismIds:       View[AsterismIds],
     allTargets:        UndoSetter[TargetList],
+    programSummaries:  UndoContext[ProgramSummaries],
```
rpiaggio, line 49: I'd rather we limit write access to only what the lower components are supposed to change. Unless we need the lower components to be able modify any part of `ProgramSummaries`.

Also, passing `UndoContext[ProgramSummaries]` makes the `allTargets` property redundant.

Finally, `UndoContext` is an `UndoSetter` that adds the ability to `undo` and `redo`. If we don't plan to allow those in the lower components, it's enough to pass an `UndoSetter`.

toddburnside: Lower down there is an `Action` that modifies both the observations and the targets in the ProgramSummaries. I guess I could make a `Lens` into a tuple of `(ObservationList, TargetList)` and use that to zoom into the ProgramSummaries. 

It does have to be an undo context, because the `Action` requires `UndoContext`

rpiaggio: Yes, we can do a `disjointZip` to zoom into observations and targets if we need them together. Then a `zoom` into the first element of the tuple would give us `allTargets`.

> the Action requires UndoContext

Not sure I follow what you meant. The `set` and `mod` methods in `Action` take an `UndoSetter`.

toddburnside: > Not sure I follow what you meant. The `set` and `mod` methods in `Action` take an `UndoSetter`.

My bad. I just glanced at the class and the parameters are named `undoCtx`, but they are `UndoSetter`s.

explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
```diff
@@ -122,6 +138,30 @@ object SiderealTargetEditor:
     attemptCombine(ra, dec)
       .map(ProperMotion.apply.tupled)
 
+  // An UndoSetter that doesn't really update any undo stacks
+  private def noopUndoSetter[M](view: View[M]): UndoSetter[M] =
```
toddburnside, line 142: This seems to work and be useful here. I don't know if it is generally useful enough to put somewhere like the UndoSetter companion object?

rpiaggio (commented):
Thank you very much for the detailed and clear explanation. It helped me understand the challenge.

I'm not sure I understand why this happens:
> If I waited to call this in the onSet of the action, when editing a asterism from the target tab, the URL would usually be incorrect. This is because the page would be set to the summary table in the meantime, and it would force the URL back to it.

The Under works with getter/setter functions such that the whole action is captured as a single value `A`, and the setter acts as an interpreter of sorts to apply an `A` to the model. `A` needs not be part of the model, it could be an auxiliary structure that extracts parts of model or fictitious values to signal inclusion/exclusion in collections, for example. That's what `KIListMod` and `KITreeMod` do, for example.

If we were to create a similar structure to represent cloning (or maybe it's enough to add specific methods in the `KI*Mod`s), I feel that it would be more undo-friendly and logic could be streamlined and compartmentalized a bit more.

We can explore this further if you feel like it. Anyway, inasmuch as I followed the logic, this PR LGTM.

PR #4001, opened 2024-07-17
Time award table

rpiaggio:
image


model/shared/src/main/scala/explore/model/allocations.scala
```diff
@@ -0,0 +1,47 @@
...
+import lucuma.core.enums.ScienceBand
+import lucuma.core.util.NewType
+import lucuma.core.util.TimeSpan
+import lucuma.odb.json.time.decoder.given
+import monocle.Focus
+import monocle.Lens
+
+import scala.collection.immutable.SortedMap
+
+case class Allocation(partner: Partner, scienceBand: ScienceBand, duration: TimeSpan) derives Eq
```
cquiroz, line 21: Should this be in core?

cquiroz (approved):
LGTM

model/shared/src/main/scala/explore/model/allocations.scala
rpiaggio: I'm not sure. Will it be used from somewhere else? We also have some model ADTs in schemas.

PR #4013, opened 2024-07-19
Reinstate floatinguUI styles

rpiaggio:
In a previous PR I had removed these styles but it was a mistake since they are still widely used.


cquiroz (approved):
LGTM

PR #4017, opened 2024-07-22
Show the submission deadline for a proposal

cquiroz:


common/src/main/webapp/sass/explore.scss
```diff
@@ -3011,6 +3011,19 @@ div.spectroscopy-table-emptymessage {
...
   }
 
+  .explore-proposal-submission-line {
+    display: flex;
+    align-items: center;
+  }
+
+  .explore-proposal-deadline {
+    .p-inline-message-text {
+      /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
```
rpiaggio, line 3021: Why is this directive necessary?

explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -611,6 +613,25 @@ object ProposalEditor:
             else
               ps.set(cfpPartners.map(p => PartnerSplit(p, 0.refined)))
           )
+      // Update the deadline when the call or partner splits change
+      .useEffectWithDepsBy((props, _, cfps, _, _, splits) =>
+        (props.proposal.get.callId, cfps.toOption.orEmpty, splits.get)
+      ): (props, _, _, _, _, ps) =>
+        (callId, cfps, splits) =>
+          callId.fold(Callback.empty)(cid =>
```
rpiaggio, line 621: Doesn't `foldMap` work with `Callback`? Otherwise `.map(..).getOrEmpty` should work.

rpiaggio (approved):
LGTM

common/src/main/webapp/sass/explore.scss
cquiroz: because of the variable, at least I got an error with stylelint locally

explore/src/main/scala/explore/proposal/ProposalEditor.scala
cquiroz: Good question, I'll check

PR #4018, opened 2024-07-22
Display the default deadline if there are no partners and dynamic dea…

cquiroz:
…dline countdown

### DIFF
--- a/explore/src/main/scala/explore/proposal/ProposalTabContents.scala
+++ b/explore/src/main/scala/explore/proposal/ProposalTabContents.scala
@@ -167,7 +167,7 @@ object ProposalTabContents:
                   ExploreStyles.ProposalSubmissionBar,
                   Button(label = "Submit Proposal",
                          onClick = updateStatus(ProposalStatus.Submitted),
-                         disabled = isUpdatingStatus.get.value
+                         disabled = isUpdatingStatus.get.value || proposalView.get.callId.isEmpty
                   ).compact.tiny,
                   (deadline.get, now).mapN((t, n) =>
                     val deadline = t.toLocalDateTime
--- a/explore/src/main/scala/explore/proposal/ProposalTabContents.scala
+++ b/explore/src/main/scala/explore/proposal/ProposalTabContents.scala
@@ -59,6 +59,42 @@ import java.time.LocalDateTime
 import java.time.ZoneOffset
 import scala.concurrent.duration.*
 
+case class CallDeadline(
+  deadline: Timestamp
+) extends ReactFnProps(CallDeadline.component)
+
+object CallDeadline:
+  private type Props = CallDeadline
+
+  private val component =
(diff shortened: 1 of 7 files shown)


explore/src/main/scala/explore/proposal/ProposalTabContents.scala
```diff
@@ -229,7 +246,10 @@ object ProposalTabContents:
     )
     .useState(none[String])        // Submission error message
     .useStateView(none[Timestamp]) // CFP/Proposal Deadline
-    .render { (props, ctx, isUpdatingStatus, readonly, errorMsg, deadline) =>
+    .useStreamOnMount(
```
rpiaggio, line 249: This causes the whole tab contents to be recomputed and rerendered every second. Maybe it's better to move the timer to its own component?

explore/src/main/scala/explore/proposal/ProposalTabContents.scala
```diff
@@ -59,6 +59,42 @@ import java.time.LocalDateTime
 import java.time.ZoneOffset
 import scala.concurrent.duration.*
 
+case class CallDeadline(
+  deadline: Timestamp
+) extends ReactFnProps(CallDeadline.component)
```
rpiaggio, line 64:
```suggestion
) extends ReactFnProps(CallDeadline.component):
  val deadlineLDT: LocalDateTime = p.deadline.toLocalDateTime
  val deadlineStr: String = s"${Constants.GppDateFormatter.format(deadline)} ${Constants.GppTimeTZFormatterWithZone
                  .format(deadline)}
```

explore/src/main/scala/explore/proposal/ProposalTabContents.scala
```diff
@@ -59,6 +59,42 @@ import java.time.LocalDateTime
...
+          val now      = LocalDateTime.ofInstant(n, ZoneOffset.UTC)
+          val diff     = java.time.Duration.between(now, deadline)
+          val left     = Constants.DurationLongWithSecondsFormatter(diff)
+          val dateFmt  =
+            if (diff.isNegative)
+              s"${Constants.GppDateFormatter.format(deadline)} ${Constants.GppTimeTZFormatterWithZone
+                  .format(deadline)}"
+            else
+              s"${Constants.GppDateFormatter.format(deadline)} ${Constants.GppTimeTZFormatterWithZone
+                  .format(deadline)} [$left]"
```
rpiaggio, line 87:
```suggestion
          val dateFmt  =
            if (diff.isNegative)
              props.deadlineStr
            else
              val left = Constants.DurationLongWithSecondsFormatter(diff)
              s"${props.deadlineStr} [$left]"
```

rpiaggio (approved):
LGTM. I added a couple of suggestions to avoid recomputing the format every second.

PR #4023, opened 2024-07-24
Avoid remount on observation time change

rpiaggio:
Also, fix observation time picker display over other tiles when target tile is collapsed.

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -1010,6 +1010,13 @@ thead tr th.sticky-header {
   }
 }
 
+.target-tile-controller {
+  // Container size queries provoke new stacking contexts at the tile controller level.
+  // In the observation tile, the target tile controller's z-index must be greater than 
+  // the z-index of the other tiles' controllers for the obs time picker to be visible over them.
+  z-index: 20;
+}
+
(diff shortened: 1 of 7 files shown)


common/src/main/webapp/sass/explore.scss
```diff
@@ -1010,6 +1010,13 @@ thead tr th.sticky-header {
   }
 }
 
+.target-tile-controller {
+  // Container size queries provoke new stacking contexts at the tile controller level.
+  // In the observation tile, the target tile controller's z-index must be greater than 
+  // the z-index of the other tiles' controllers for the obs time picker to be visible over them.
+  z-index: 20;
```
cquiroz, line 1017: So this gives a index for the container? should all containers have a z-insdex?

cquiroz (approved):
LGTM Thanks for fixing this

common/src/main/webapp/sass/explore.scss
rpiaggio: Right, below this point every tile has their own context, so we cannot set relative z-indices across elements from different tiles. But we can set the z-index of each tile, which will make the whole tile render in front of/behind another tile.

I don't think it's necessary to set the z-indices of the other tiles for the moment, they just have the default (0).

PR #4033, opened 2024-07-30
Avoid blinks when selecting in target summary

rpiaggio:


cquiroz (approved):
LGTM

toddburnside (approved):
LGTM. Thanks

explore/src/main/scala/explore/components/TileController.scala
```diff
@@ -93,34 +93,31 @@ object TileController:
...
-      )
-      .render { (p, ctx, breakpoint, currentLayout) =>
+      .render: (p, ctx, breakpoint, currentLayout) =>
         import ctx.given
 
         def sizeState(id: Tile.TileId) = (st: TileSizeState) =>
           currentLayout
             .zoom(allTiles)
-            .mod {
+            .mod:
```
hugo-vrijswijk, line 110: Slowly but surely all of explore will use braceless syntax :)

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -423,18 +430,28 @@ object TargetTabContents extends TwoPanels:
...
+                      )
               case Right(obsIds)  =>
                 findAsterismGroup(obsIds, props.programSummaries.get.asterismGroups)
                   .map: asterismGroup =>
-                    (renderAsterismEditor(resize, obsIds, asterismGroup), "target-obs-controller")
+                    (renderAsterismEditor(resize, obsIds, asterismGroup),
+                     TargetTabControllerIds.AsterismEditor.id
+                    )
+
+      val justSummaryTiles: List[Tile] = List(
```
hugo-vrijswijk, line 445: So the fix is to always have the same number of tiles, and then react won't remount existing tiles when tiles are added?

hugo-vrijswijk (approved):

explore/src/main/scala/explore/tabs/TargetTabContents.scala
rpiaggio: The remount happened because we changed the key of the `TileController`. Now we only use 2 keys: one for when the summary is shown and when for when it isn't.

We did this to prevent the tile controller tripping up when the displayed tiles changed (the saved information would go bonkers). This PR prevents that by still having the unused tiles in the controller, but hidden.

PR #4034, opened 2024-07-31
Add hasura-cli to flake.nix

toddburnside:
I also tried to add `firebase-tools`, but although it works for me in home manager, I couldn't get it to work here.


cquiroz (approved):
LGTM

PR #4036, opened 2024-08-02
Enable trace logging

rpiaggio:
Adds the option in the menu to switch to trace logging. Useful in particular for debugging issues with clue.


hugo-vrijswijk (approved):
LGTM

PR #4038, opened 2024-08-02
Display calibrations as read only

rpiaggio:
- Calibration obervations are now readonly.
- So are system groups.
- `ObsSummary` was renamed to `Observation`.
- System groups and calibration observations cannot be dragged or dropped upon.
- Add button in asterism editor is now hidden when in readonly mode.
- Active status is disabled in badges, as well as cloning and deletion buttons.
- New group and new observation are disabled when a system group is selected.
- Remove calibration observations and targets from trees in targets, constraints and scheduling tabs.
- Remove calibration observations and targets from target summary and observation summary tables.


common/src/test/scala/explore/common/arb/ArbTargetObsQueries.scala
```diff
@@ -11,7 +11,7 @@
 // import lucuma.core.enums.ObsActiveStatus
 // import lucuma.core.enums.ObsStatus
 // import lucuma.core.model.Asterism
-// import lucuma.core.model.Observation
+// import explore.model.Observation
```
cquiroz, line 14: 🤔

rpiaggio: `lucuma.core.model.Observation` only has the `Id` type, which is now forwarded from `explore.model.Observation`.

We use this same technique in other cases (eg: `Visit`).

cquiroz (approved):
LGTM

PR #4043, opened 2024-08-05
Change allocations from Partners to TimeAccountingCategories

toddburnside:
Time allocations are not assigned to a `TimeAccountingCategory` instead of a `Partner`.  Not all time accounting categories are partners. The time allocation table now displays a flag if and only if the time accounting category refers to a partner, and it looks like:
image


swalker2m (approved):
Looks right to me. :+1:

cquiroz (commented):
LGTM

PR #4044, opened 2024-08-06
Allow changing S/N at in the advanced conf panel

cquiroz:
image


explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -96,7 +96,7 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
   ReadMode: Enumerated: Display,
   Gain: Enumerated: Display,
   Roi: Enumerated: Display
-] {
+] extends ConfigurationFormats {
```
toddburnside, line 99: Does this need to extend `ConfigurationFormats`?

toddburnside (approved):
LGTM

PR #4046, opened 2024-08-08
Break race condition when selecting newly created target and previous focus was on observation

rpiaggio:
Fixes https://app.shortcut.com/lucuma/story/3057/select-new-target


toddburnside (approved):
Nice

PR #4047, opened 2024-08-09
Expand program selection table to available height, scroll to bottom when new program added

rpiaggio:
Kapture 2024-08-09 at 11 46 45


cquiroz (approved):
LGTM

PR #4048, opened 2024-08-09
Reset error message when changing cfp

cquiroz:


explore/src/main/scala/explore/proposal/ProposalTabContents.scala
```diff
@@ -260,6 +261,9 @@ object ProposalTabContents:
       p => p === ProposalStatus.Submitted || p === ProposalStatus.Accepted
     )
     .useState(none[String])        // Submission error message
+    .useEffectWithDepsBy((props, _, _, _, _) =>
```
rpiaggio, line 264: In this particular case, it doesn't change much, but when updating UI state in response to external changes, it's usually more appropriate to `useLayoutEffect`. At the very least, it avoids an extra render; but it can avoid race conditions if there are other effects triggered. (which, as said, doesn't seem to be the case here).

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/proposal/ProposalTabContents.scala
cquiroz: Interesting, what would be an external change?

rpiaggio: Anything that causes a rerender. It can be a prop or some other state changing. That will cause a rerender, which will cause the effect to run, changing the state, and triggering another rerender immediately. The intermediate render may cause a flicker, or it may be an invalid state. `useLayoutEffect` blocks the next render until the effect completes (as long as it's a sync effect).

PR #4049, opened 2024-08-09
Make explore work with the latest schema

cquiroz:


explore/src/clue/scala/queries/common/ProgramUserWithRoleSubquery.scala
```diff
@@ -14,7 +14,12 @@ object ProgramUserWithRoleSubquery
   override val subquery: String = s"""
     {
       user $ProgramUserSubquery
-      partner
+      partner {
```
swalker2m, line 17: What type in the schema is this a query for?  `ProgramUser`?  Shouldn't it be `programLink:` here?

cquiroz: Yes, how did this work 🤔

cquiroz: Fixed

PR #4056, opened 2024-08-13
Prevent asterism updates for executed observations, add undo/redo

toddburnside:


explore/src/clue/scala/queries/common/AsterismQueriesGQL.scala
```diff
@@ -20,4 +20,25 @@ object AsterismQueriesGQL {
       }
     """
   }
+
+  // Group these 2 mutations together so that the targetEdit and observationEdit subscriptions come
```
cquiroz, line 24: Is this guaranteed to work?

explore/src/main/scala/explore/tabs/AsterismEditorTile.scala
```diff
@@ -59,6 +59,7 @@ object AsterismEditorTile:
     backButton:        Option[VdomNode] = none
   )(using FetchClient[IO, ObservationDB], Logger[IO]): Tile = {
     // Save the time here. this works for the obs and target tabs
+    // It's OK to save the viz time for executed observations, I think.
     val vizTimeView =
       vizTime.withOnMod(t => ObsQueries.updateVisualizationTime[IO](obsIds.toList, t).runAsync)
```
cquiroz, line 64: This needs an update I think

cquiroz (approved):
This is quite an undertaking! thanks
LGTM

explore/src/clue/scala/queries/common/AsterismQueriesGQL.scala
toddburnside: Well, we reduce the transit time involved, but they still aren't real close together. That timing depends on when the database updates happen for the 2 mutations, I guess. Still seemed better than doing them individually. 🤷‍♂️

explore/src/main/scala/explore/tabs/AsterismEditorTile.scala
toddburnside: Actually, it looks like it is still called `ObsQueries.updateVisualizationTime` in your PR. 😄 So, I don't think this needs to change. Unless there is another reason.

PR #4057, opened 2024-08-13
Update pi to support the latest schema

cquiroz:


toddburnside (approved):
LGTM. It seems like at some point we could add the role and partnerLink to ProgramUser and get rid of ProgramUserWithRole.

PR #4059, opened 2024-08-14
Show the add target button when the tile is closed

cquiroz:


explore/src/main/scala/explore/tabs/AsterismEditorTile.scala
```diff
@@ -56,15 +59,28 @@ object AsterismEditorTile:
     globalPreferences: View[GlobalPreferences],
     readonly:          Boolean,
     sequenceChanged:   Callback = Callback.empty,
+    adding:            View[AreAdding],
     backButton:        Option[VdomNode] = none
-  )(using FetchClient[IO, ObservationDB], Logger[IO]): Tile = {
+  )(using FetchClient[IO, ObservationDB], Logger[IO], ToastCtx[IO]): Tile = {
+
+    val targetAdd   =
+      targetSelectionPopup(
```
toddburnside, line 67: I think this will bypass the protection for observations that have been executed.

PR #4060, opened 2024-08-17
Tile Simplification

cquiroz:
Tiles can render controls on the title that may interact with the body.
At the moment this can be done in two ways which is confusing and it is a bit hard to share state
This PR changes this, now a Tile can take two components for body and tile (providing a single way to do this) and a mechanism for them to share state.

This is a relatively large PR and though it seems to add lines it is clearer to understand now as components for tile and body are clearly separated instead of being intermeshed. Also we were able to move some the shared state down.

I tried to keep a naming convention such that we can more or less easily identify the role of each component (Body/Title)


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -15,57 +17,103 @@ import lucuma.react.common.ReactFnProps
...
   controllerClass:   Css = Css.Empty, // applied to wrapping div when in a TileController.
   bodyClass:         Css = Css.Empty, // applied to tile body
   tileClass:         Css = Css.Empty, // applied to the tile
   tileTitleClass:    Css = Css.Empty  // applied to the title
-)(val render: Tile.RenderInTitle => VdomNode)
-    extends ReactFnProps[Tile](Tile.component) {
-  def showMaximize: Boolean =
-    state === TileSizeState.Minimized || (canMaximize && state === TileSizeState.Minimized)
+)(
+  val tileBody:      View[A] => VdomNode,
```
cquiroz, line 76: This is the key change, components for body and title are passed as a param

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -75,47 +123,44 @@ object Tile:
...
                     minimizeButton.when(p.showMinimize),
                     maximizeButton.when(p.showMaximize)
-              )
+              ).unless(p.fullSize)
             ),
             // Tile body
-            infoRef.value
-              .map(node =>
-                <.div(ExploreStyles.TileBody |+| p.bodyClass,
-                      p.render(info => ReactPortal(info, node))
```
cquiroz, line 116: No need for a Portal anymore

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -263,20 +261,11 @@ object ObsTabTiles:
             .whenA(itcProps.isExecutable)
             .runAsyncAndForget
       }
-      // ITC selected target. Here to be shared by the ITC tile body and title
```
cquiroz, line 266: This is an example of a state that was pushed down to the tile where it makes more sense

explore/src/main/scala/explore/attachments/ObsAttachmentsTableBody.scala
```diff
@@ -323,54 +327,17 @@ object ObsAttachmentsTable extends ObsAttachmentUtils:
...
           getRowId = (row, _, _) => RowId(row.get.id.toString),
           meta = TableMeta(client, props.obsAttachmentAssignments, urlMap.get, props.readOnly)
         )
-      .useStateView(Enumerated[ObsAttachmentType].all.head)
-      .render: (props, ctx, client, action, _, _, _, table, newAttType) =>
-        import ctx.given
-
+      .render: (props, _, _, _, _, _, table) =>
         React.Fragment(
-          props.renderInTitle(
```
cquiroz, line 340: This is an example where a new component for the title was extracted out from the body and it is now an independent component for the title

explore/src/main/scala/explore/components/ColumnSelectorInTitle.scala
```diff
@@ -0,0 +1,39 @@
...
+import lucuma.react.common.ReactFnProps
+import lucuma.react.table.ColumnId
+import lucuma.react.table.Table
+import lucuma.ui.table.ColumnSelector
+
+case class ColumnSelectorState[A, B](
+  table: Option[Table[A, B]] = None
+)
+
+case class ColumnSelectorInTitle[A, B](
```
cquiroz, line 20: This was extracted to be common across several tiles

common/src/main/scala/explore/components/Tile.scala
```diff
@@ -15,57 +17,103 @@ import lucuma.react.common.ReactFnProps
 import lucuma.react.common.style.*
 import lucuma.react.primereact.Button
 import lucuma.ui.syntax.all.given
-import org.scalajs.dom
 
-case class Tile(
+/**
```
rpiaggio, line 21: Thanks for the doc.

explore/src/main/scala/explore/components/ColumnSelectorInTitle.scala
rpiaggio: Nice

explore/src/main/scala/explore/attachments/ObsAttachmentsTableBody.scala
```diff
@@ -60,17 +60,22 @@ import org.typelevel.log4cats.Logger
 
 import scala.collection.immutable.SortedSet
 
-case class ObsAttachmentsTable(
+object ObsAttachmentsTableTileState extends NewType[Action]:
+  def apply(): ObsAttachmentsTableTileState = ObsAttachmentsTableTileState(Action.None)
+type ObsAttachmentsTableTileState = ObsAttachmentsTableTileState.Type
+
+case class ObsAttachmentsTableBody(
```
rpiaggio, line 67: Maybe we want to keep the name of the file `ObsAttachmentsTable` (without `Body`).

Also, have you considered the following structure?
``` scala
object ObsAttachmentsTable:
  object State extends NewType[Action]
  
  case class Body(...)
  object Body {
    ...
  }
  
  case class Title(...)
  object Title {
    ...
  }
```

rpiaggio (approved):
I haven't gone in detail into all the individual tiles, but the general concept LGTM.

Thank you for tackling this!

explore/src/main/scala/explore/attachments/ObsAttachmentsTableBody.scala
cquiroz: Seems good, maybe calling it `ObsAttachmentsTile`?

cquiroz:
Thanks to @rpiaggio for the great ideas on how to implement this

toddburnside (approved):
I must admit my eyes glazed over in some of the tile refactoring... But this looks great. It also look like a lot of work. Thanks for doing this.

PR #4077, opened 2024-08-21
Support selecting a users' partner

cquiroz:
https://github.com/user-attachments/assets/6daa2b71-ec86-4a27-b837-f3f45e2b7d2d


explore/src/main/scala/explore/proposal/ProgramUsersTable.scala
```diff
@@ -70,29 +76,78 @@ object ProgramUsersTable:
...
       column(NameColumnId, _.name),
       ColDef(
         PartnerColumnId,
-        _.partnerLink.flatMap {
-          case PartnerLink.HasPartner(partner)   => Some(partner)
-          case PartnerLink.HasNonPartner         => None
-          case PartnerLink.HasUnspecifiedPartner => None
-        },
-        "",
+        identity,
```
rpiaggio, line 132: This breaks sorting. We probably want what you have further down as `pl` as the column value.

Then, in the cell render function, you can still access the row as `cell.row.original` to get the `usersView`.

explore/src/main/scala/explore/proposal/ProgramUsersTable.scala
```diff
@@ -138,8 +193,8 @@ object ProgramUsersTable:
       .withHooks[Props]
       .useContext(AppContext.ctx)
       .useStateView(IsActive(false))
-      .useMemoBy((_, _, _) => ()): (_, ctx, _) => // cols
-        _ => columns(ctx)
+      .useMemoBy((_, _, _) => ()): (p, ctx, _) => // cols
```
rpiaggio, line 196: Can `users` change externally? Eg: editing in another session.

If so, it should go in a `TableMeta`.

toddburnside (approved):
LGTM

PR #4081, opened 2024-08-22
Correct for PM on the target editor's elevation plot

cquiroz:


explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -167,6 +167,13 @@ object TargetTabContents extends TwoPanels:
         )
     )
 
+    def getObsTime(idsToEdit: ObsIdSet): ProgramSummaries => Option[Instant] = a =>
```
rpiaggio, line 170: Why move this here? It doesn't seem to be called from anywhere else.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/tabs/TargetTabContents.scala
cquiroz: I was going to share it with another function but I didn't need it after all

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -391,7 +394,6 @@ object TargetTabContents extends TwoPanels:
                 props.userId,
                 targetId.some,
                 none,
-                // TODO PM correct the coordinates
                 CoordinatesAtVizTime(Target.Sidereal.baseCoordinates.get(target.get)).some,
```
toddburnside, line 397: Are these PM corrected?

cquiroz: There is no time in this case, we can only fallback to base

PR #4082, opened 2024-08-23
Remove matrix from 'full' GH action

hugo-vrijswijk:
The 'required' workflow includes the scala version in the name, which makes upgrading scala versions more annoying because you have to disable the required action, merge, and enable the new workflow with the new scala version in the name


cquiroz (approved):
Thanks, LGTM

PR #4092, opened 2024-08-29
Show the deadline for the selected partner

cquiroz:


model/shared/src/main/scala/explore/CallForProposal.scala
```diff
@@ -36,13 +37,24 @@ case class CallForProposal(
...
 
-  def deadline(proposalPartners: List[Partner]): Option[Timestamp] =
-    val callPartners = partners
-    proposalPartners
-      .map(p => callPartners.find(_.partner === p).flatMap(_.submissionDeadline))
-      .minimumOption
-      .flatten
-      .orElse(submissionDeadlineDefault)
+  def deadline(piPartner: Option[PartnerLink]): Option[Timestamp] =
+    piPartner.flatMap {
```
cquiroz, line 41: Maybe this logic should go to core

model/shared/src/main/scala/explore/CallForProposal.scala
```diff
@@ -36,13 +37,24 @@ case class CallForProposal(
...
+      _.fold(
+        None,
+        partners
+          .find(_.partner === Partner.US)
```
cquiroz, line 45: This is unfortunate

model/shared/src/main/scala/explore/CallForProposal.scala
```diff
@@ -36,13 +37,24 @@ case class CallForProposal(
...
+          .flatMap(_.submissionDeadline)
+          .orElse(submissionDeadlineDefault),
```
swalker2m, line 47: The call for proposals has a `nonPartnerDeadline` field.  Also the partner `submissionDeadline` already does the explicit deadline or else default in the view from which the data comes.

model/shared/src/main/scala/explore/CallForProposal.scala
swalker2m: So I think you can ignore / delete the `submissionDeadlineDefault` unless you're making an editor for the CfP.  Just ask each partner for its deadline and it will handle it. You can also add the `nonPartnerDeadline` and get it from the query instead of computing it.

swalker2m (approved):
Looks good to me at least. :+1:

PR #4101, opened 2024-09-01
Allow resizing tiles

cquiroz:
It was requested to remove the restrictions on resizing tiles. A min width is enforced or else it is too difficult to layout components


rpiaggio (approved):
LGTM

PR #4104, opened 2024-09-02
Migrate to new ITC API with asterism

rpiaggio:


model/shared/src/main/scala/explore/model/GlobalPreferences.scala
```diff
@@ -32,6 +32,9 @@ object ElevationPlotScheduling extends NewType[Boolean]:
 
 type ElevationPlotScheduling = ElevationPlotScheduling.Type
 
+given Decoder[GraphType] = Decoder.decodeString.map:
+  case "S2N_CHART" => GraphType.S2NGraph
```
cquiroz, line 36: What about the ohter cases?

workers/src/main/scala/workers/itc/ITCRequests.scala
```diff
@@ -9,105 +9,94 @@ import cats.data.*
...
-                mode = mode
-              ),
-              false
-            )
-            .map(r => Map(params -> itcResults(r)))
-            .handleError { e =>
-              Map(
-                params -> processExtension(e).leftNec[ItcResult]
+          s"ITC: Request for mode: ${params.mode}, centralWavelength: ${params.wavelength} and target count: ${params.asterism.length}"
+        ) *> // selectedBand(params.target.sourceProfile, params.wavelength.value),
```
cquiroz, line 78: Commented line

model/shared/src/main/scala/explore/model/GlobalPreferences.scala
rpiaggio: Good catch. This was meant to be temporary. The idea is to use a migration to rename in the DB and use the standard codecs. I'll add a commit with that.

PR #4105, opened 2024-09-02
Support editing educational status and thesis

cquiroz:
image

### DIFF
--- a/explore/src/main/scala/explore/proposal/ProgramUsersTable.scala
+++ b/explore/src/main/scala/explore/proposal/ProgramUsersTable.scala
@@ -8,7 +8,7 @@ import cats.syntax.all.*
 import crystal.react.*
 import crystal.react.hooks.*
 import explore.Icons
-import explore.common.ProgramQueries.updateProgramUsers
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/proposal/ProgramUsersTable.scala
```diff
@@ -150,6 +170,52 @@ object ProgramUsersTable:
...
+          val userId = cell.user.id
+
+          c.table.options.meta.map: meta =>
+            val view = meta.users
+              .zoom(thesisLens(userId))
+              .withOnMod(th =>
+                updateUserThesis[IO](meta.programId,
+                                     userId,
+                                     th.headOption.flatten
+                ).runAsyncAndForget
```
rpiaggio, line 212: In other tables, we have turned a `View[List[Foo]]` to a `List[View[Foo]]` by using `toListOfViews`. Then each row is actually a `View`. Maybe that would simplify things here?

rpiaggio (approved):
LGTM, but let's consider switching the users table to a list of views.

PR #4109, opened 2024-09-04
Fix undo with effectful get

rpiaggio:
I really don't know how this was working before. It probably wasn't and nobody noticed.


cquiroz (approved):
LGTM

PR #4112, opened 2024-09-04
Add editing of observation duration

toddburnside:
This adds a duration editor alongside the observation time editor in the asterism editor tile. I tweaked the CSS a bit to give all the control in that title bar some space (Add Button, time and duration editor, column selector). 
When observation duration has a value:
image
When it is empty:
image

We may need to play with the UI a bit. Because of the amount of space it takes up, I had to hide the duration editor when the tile width got down to medium...


rpiaggio (approved):
LGTM

PR #4115, opened 2024-09-06
Deglitch

rpiaggio:
`ProgramSummaries` has been removed from `RootModel` into another `View` which is wrapped in a `ThrottlingView`. Both are now in a new class `RootModelViews` which is the new root state of the app (instead of `View[RootModel]`).

Then the `throttledView` of the `ProgramSummaries` is passed to the `CacheController`, while the `throttlerView` is passed to the rest of the app.


toddburnside (approved):
Very cool! LGTM

PR #4124, opened 2024-09-09
Apply migrations on promotion

rpiaggio:


cquiroz (approved):
Excellent, thanks

rpiaggio:
I'm merging, but I should note that this is untested. We will find out in the next promotion.

promoteDevToStaging.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 
+# WARNING: This will apply CURRENT LOCAL migrations to the staging server,
+# which may not necessarily be in sync with the dev server.
+cd hasura/user-prefs
+hasura migrate apply --endpoint https://user-prefs-staging.herokuapp.com
```
toddburnside, line 6: Do we also need to run `hasura metadata apply ...` and maybe `hasura metadata reload`?

PR #4125, opened 2024-09-09
Clean observation attachments when copying observation

rpiaggio:


cquiroz (approved):
LGTM.. I wonder if this should be enforced on the backend

rpiaggio:
> LGTM.. I wonder if this should be enforced on the backend

Not sure. In the backend we invoke `cloneObservation`. Is there a scenario in which somebody wants to invoke `cloneObservation` in the API but keep the observation attachments? @andrewwstephens

andrewwstephens:
It seems like `cloneObservation` should give an exact copy - including the attachment association.
The issue that we want to avoid is when an observation is copied to a *different* target we don't want to retain an association with the previous finder chart attachment.

PR #4129, opened 2024-09-10
Fix promotion scripts

cquiroz:


cquiroz:
Maybe, I can add that too. I tested these today deploying to staging and production

rpiaggio:
Thank you, still LGTM 👍

PR #4132, opened 2024-09-11
Action buttons for target tab

rpiaggio:
Addresses https://app.shortcut.com/lucuma/story/1708/action-buttons only for the Target Tab. I'm sending separate PRs to keep them more manageable.

- Copy copies from either the tree or the target summary table.
- Paste pastes either in the tree or the summary table.
- Delete only deletes from the tree, and only if a single observation is selected.
  - Supporting multiple obs delete seems complicated, mainly due to undo.
  - Unifying with deletion in the summary table is an option, probably desired, to keep the UI consistent. I took a cursory look and it's not completely straightforward since the delete action interacts with the table (which is not available when the delete button in the tree is rendered).

Kapture 2024-09-11 at 17 24 26

### DIFF
--- a/common/src/main/scala/explore/Icons.scala
+++ b/common/src/main/scala/explore/Icons.scala
@@ -275,6 +275,10 @@ object Icons {
   val faClone: FAIcon = js.native
 
   @js.native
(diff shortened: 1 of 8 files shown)


cquiroz (commented):
Thanks! LGTM

PR #4138, opened 2024-09-12
Action buttons in observations tab

rpiaggio:


common/src/main/webapp/sass/explore-common.scss
```diff
@@ -1,5 +1,5 @@
 $tile-title-height: 36px;
-$tree-section-width: 300px;
+$tree-section-width: 320px;
```
cquiroz, line 2: Could yoeu verify this number is not used inother places in the code?

cquiroz (approved):
LGTM

common/src/main/webapp/sass/explore-common.scss
rpiaggio: Seems to be safe.

PR #4141, opened 2024-09-13
Action buttons in constraints tab

rpiaggio:
Hotkeys have also been implemented.

As in targets, deleting only works for single observations.


cquiroz (approved):
LGTM

PR #4147, opened 2024-09-15
Fix broken band label

cquiroz:
The right hand side label of the elevation plot looks incorrectly starting on 11.4.8.
After some debugging it was noted that the width of the label was set to 1 as the band is empty. this leads to the label being broken into words.

The fix is to make the band match the width or the whole range though perhaps a more appropriate solution would be to use a plot line, however the styling doesn't match


rpiaggio (approved):
LGTM

PR #4156, opened 2024-09-18
Edit support users in program tab

rpiaggio:
~Invitations can now have roles. Only PIs and CoIs are now shown in the Proposal tab. This PR also paves the way to editing support users in the Program tab.~

Edit support users in the program tab. Only Admin and Staff users can do this. Support users were removed from the Proposal tab.

image


cquiroz (approved):
LGTM

PR #4161, opened 2024-09-21
Simplifications on the hooks to do AGS

cquiroz:
It also fixes a bug where vizTime was not updated when it changed locally


explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -230,29 +226,28 @@ object AladinCell extends ModelOptics with AladinCommon:
       // target options, will be read from the user preferences
       .useStateView(Pot.pending[AsterismVisualOptions])
       // to get faster reusability use a serial state, rather than check every candidate
```
toddburnside, line 228: Should the `serial state` comment be removed, too?

cquiroz: Yeah, it is removed in my next PR

toddburnside (approved):
LGTM

PR #4166, opened 2024-09-23
Copy observations into groups

rpiaggio:
Previous functionality only allowed cloning observations within same group.


toddburnside (approved):
LGTM

PR #4167, opened 2024-09-23
Invoke ITC chart only once when Image Quality changes

rpiaggio:
Changing the Image Quality can change the CCD binning. We have logic to compute the new binning but it was being ignored, in favor of getting the value from the server.

This caused the ITC chart to be invoked twice when the IQ changed: once with the old binning, and again with the fixed new binning once the server roundtrip completed.

This PR does 2 things:
- Invoke ITC with the new, computed in the client, binning.
- Change reusability so that the server binning doesn't take part in it (otherwise the chart was still being invoked twice, albeit both times with the right values).

### DIFF
--- a/common/src/main/scala/explore/model/reusability.scala
+++ b/common/src/main/scala/explore/model/reusability.scala
@@ -5,6 +5,7 @@ package explore.model
 
 import cats.Eq
 import cats.data.NonEmptyChain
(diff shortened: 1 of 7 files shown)


cquiroz (approved):
LGTM

toddburnside (approved):
LGTM

PR #4168, opened 2024-09-24
Delete obsExecutionPot when observation is deleted.

toddburnside:
Just something I noticed while looking for something else.


cquiroz (approved):
LGTM

PR #4169, opened 2024-09-24
Unify delete buttons in target tab

rpiaggio:
Now targets can be deleted via the action button above the tree. This results in a more consistent experience with respect to the copy/paste buttons behavior.


cquiroz (approved):
LGTM

PR #4180, opened 2024-09-30
Add help icon for the calibrations group

cquiroz:


explore/src/main/scala/explore/observationtree/GroupBadge.scala
```diff
@@ -58,5 +61,9 @@ object GroupBadge:
     )(
       if group.isAnd then "AND" else "OR",
       group.name.map(n => <.em(n.value)),
-      deleteButton
+      deleteButton,
+      NonEmptyString
+        .from(s"groups/system-$name.md")
+        .toOption
+        .map(HelpIcon(_, ExploreStyles.GroupHelp))
```
toddburnside, line 68: I must be missing something. Won't this result in there always being a help link, even for non-system groups?

cquiroz: your're right, I'm missing a `when`

PR #4183, opened 2024-10-02
Rerender ITC graph when basic config is overriden

rpiaggio:
The actual fix is adding the basicConfiguration to the Reusability of ItcProps.

The other changes are things I found while debugging, notably that the parameters to `toItcClientMode` were not being used since the last ITC refactor.

### DIFF
--- a/explore/src/main/scala/explore/itc/ItcProps.scala
+++ b/explore/src/main/scala/explore/itc/ItcProps.scala
@@ -96,13 +96,13 @@ case class ItcProps(
       .map:
         case BasicConfiguration.GmosNorthLongSlit(grating, filter, fpu, _) =>
           val gmosOverride: Option[GmosSpectroscopyOverrides] = modeOverrides match
-            case Some(o: GmosSpectroscopyOverrides) => o.some
-            case _                                  => none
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/itc/ItcProps.scala
```diff
@@ -164,6 +164,7 @@ object ItcProps:
        p.observation.constraints,
        p.observation.scienceRequirements,
        p.observation.wavelength,
+       p.observation.basicConfiguration,
```
cquiroz, line 167: I may have deleted this by mistake

model/shared/src/main/scala/queries/schemas/itc/syntax.scala
```diff
@@ -26,7 +24,7 @@ import lucuma.itc.client.TargetInput
 trait syntax:
 
   extension (row: InstrumentRow)
-    def toItcClientMode(ps: NonEmptyList[SourceProfile], iq: ImageQuality): Option[InstrumentMode] =
+    def toItcClientMode: Option[InstrumentMode] =
```
cquiroz, line 27: Strange this wasn't flagged before

cquiroz (approved):
LGTM

PR #4185, opened 2024-10-02
Setup honeycomb telemetry

cquiroz:
I learned we can use honeycomb from web clients. There is a provided sdk to  add it without disturbing the code too much

I think we can benefit of getting some insights though we need to balance this with the cost of the service

Note that we are including a key in the open. This is an ingest key that can only be used to submit traces though we could use an extra proxy in between


toddburnside:
What all is this sending to honeycomb. Do we have to start traces, etc.? Or is it automagically sending stuff? I see something about `getWebAutoInstrumentations` and `UserInteractionInstrumentation`

cquiroz:
it is quite a bit of magic. `getWebAutoInstrumentations` will send stats about page loads, requests, etc. `UserInteractionInstrumentation` will send things like `clicks`. I 'm not sure how useful this is but it seems better to have it than not?

We can add custom traces if we want but we get quite a bit out of the basics

toddburnside (approved):
LGTM - let's give it a try. Should we enable it for `development` during a testing period, too?

cquiroz:
I'm not sure, we could use it on development but there will be more noise on the network tab
OTOH it wouldd ensure that nothing breaks by adding Observability

PR #4187, opened 2024-10-03
Remove undroppable workaround now that it's fixed in PrimeReact

rpiaggio:
There was a bug with undroppable nodes in PrimeReact, which is now fixed. We can therefore remove the workaround.


cquiroz (approved):
LGTM

PR #4189, opened 2024-10-04
Show science bands and allow setting/changing them.

toddburnside:
The science band is displayed as an column in the Observation Summary table and on the Observation Badge when there is more than one band allocated to a program. On the badge it looks like 
image
On all the tabs except for the Observation Tab the science band is read only, although there is a tool tip. 
image
On the Observation Tab, the tool tip includes an indication it is clickable
image
Which brings up a popup menu
image

Setting the visibility in the summary table was more work than I thought it would be due to the column visibility selector AND the saving of the table state in user preferences.

Note that the menu items do not automatically update if the allocations are changed in the ODB because changing allocations does not result in a subscription event. There is a story for that [here](https://app.shortcut.com/lucuma/story/3649/setting-program-allocations-does-not-raise-a-subscription-event)

There is currently no way to make bulk updates to the observation science bands via Explore. This may be possible if multi-select is enabled in the summary table.


rpiaggio (approved):
LGTM

cquiroz (approved):
LGTM

PR #4191, opened 2024-10-04
Show ITC errors on tile, as well as sequence errors

rpiaggio:
[sc-3107]

### DIFF
--- a/explore/src/main/scala/explore/Telemetry.scala
+++ b/explore/src/main/scala/explore/Telemetry.scala
@@ -95,7 +95,7 @@ object Observability:
       .withHooks[Props]
       .useEffectOnMountBy: props =>
         val sdk = new HoneycombWebSDK(props.options)
-        Callback { println(sdk); sdk.start() } *> CallbackTo(
+        Callback.log(sdk) *> Callback(sdk.start()) *> CallbackTo(
           Callback.log("shutdown") *> Callback(sdk.shutdown())
         )
       .render(_ => EmptyVdom)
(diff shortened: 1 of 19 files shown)


explore/src/main/scala/explore/Telemetry.scala
```diff
@@ -95,7 +95,7 @@ object Observability:
       .withHooks[Props]
       .useEffectOnMountBy: props =>
         val sdk = new HoneycombWebSDK(props.options)
-        Callback { println(sdk); sdk.start() } *> CallbackTo(
+        Callback.log(sdk) *> Callback(sdk.start()) *> CallbackTo(
```
cquiroz, line 98: I guess we should remove the log

workers/src/main/scala/workers/itc/ITCGraphRequests.scala
```diff
@@ -39,7 +39,7 @@ object ITCGraphRequests:
     targets:         NonEmptyList[ItcTarget],
     mode:            InstrumentRow,
     cache:           Cache[F],
-    callback:        GraphResponse => F[Unit]
+    callback:        ItcAsterismGraphResults => F[Unit]
```
cquiroz, line 42: Do we need to invalidate the cache?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/Telemetry.scala
rpiaggio: I didn't know if you wanted to leave it, the way it was before it just printed `object Object`. I'll remove.

workers/src/main/scala/workers/itc/ITCGraphRequests.scala
rpiaggio: This is used to send the results back to the client. Here it was just a type rename, the type was moved to `shared` into the `explore.model.itc` package.

PR #4192, opened 2024-10-04
Change SASS deprecated color function

toddburnside:
We were getting a deprecation error for this in CI.

There is also another SASS related deprecation error (repeated many times)

> DEPRECATION WARNING: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.
More info: https://sass-lang.com/d/legacy-js-api

Vite defaults to the legacy API, but provides a config setting to use the `modern` API. However, this setting is only available in vite 5, which is still very slow to load explore during development.


cquiroz (approved):
LGTM

PR #4196, opened 2024-10-08
Migrate to pragmatic-drag-and-drop

rpiaggio:
For the moment, we keep the `react-beatiful-dnd` facade through their migration layer.


toddburnside (approved):
LGTM

PR #4199, opened 2024-10-09
Disable submission/retraction when program has a CfP that is due

rpiaggio:
If the user set a CfP for a program before the deadline, they could still submit it after the deadline.

The CfP still apprared set, but the deadline was None, which was interpreted as the CfP having no deadline.

We are now disallowing submission/retraction if there's no deadline.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -253,6 +253,12 @@ def anyConds(conds: String*) = conds.mkString("(", " || ", ")")
 
 val faNpmAuthToken = "FONTAWESOME_NPM_AUTH_TOKEN" -> "${{ secrets.FONTAWESOME_NPM_AUTH_TOKEN }}"
 
+lazy val setupSbt = WorkflowStep.Use(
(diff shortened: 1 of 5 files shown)


cquiroz (approved):
LGTM

build.sbt
```diff
@@ -253,6 +253,12 @@ def anyConds(conds: String*) = conds.mkString("(", " || ", ")")
 
 val faNpmAuthToken = "FONTAWESOME_NPM_AUTH_TOKEN" -> "${{ secrets.FONTAWESOME_NPM_AUTH_TOKEN }}"
 
+lazy val setupSbt = WorkflowStep.Use(
+  UseRef.Public("sbt", "setup-sbt", "v1"),
+  name = Some("Setup SBT"),
+  params = Map("sbt-runner-version" -> "1.10.2")
+)
```
toddburnside, line 260: I think I prefer manually updating ci.yml and removing the workflow check. That way, when it is repaired upstream the workflows will regenerate and we don't have to undo anything.

PR #4200, opened 2024-10-10
Plots use their set date/semester instead of obs time

rpiaggio:
Coordinates were always corrected for the observation time (or not corrected if they were shown outside of an observation).

This PR always corrects them for the date selected in the plot tile, or the start of the semester.

When plotting an observation, it now uses the center of the asterism instead of the first target.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
+++ b/explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
@@ -25,7 +25,10 @@ import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.enums.Site
 import lucuma.core.enums.TimingWindowInclusion
 import lucuma.core.math.BoundedInterval
+import lucuma.core.math.Coordinates
 import lucuma.core.model.CoordinatesAtVizTime
+import lucuma.core.model.ObjectTracking
+import lucuma.core.model.Semester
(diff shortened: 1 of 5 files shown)


explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
```diff
@@ -147,22 +150,31 @@ object ElevationPlotSection:
...
+                  coords,
                   props.visualizationTime,
                   windowsNetExcludeIntervals,
                   props.pendingTime,
                   options
                 )
               case PlotRange.Semester =>
+                val coords: CoordinatesAtVizTime =
+                  props.tracking
+                    .at(semesterView.get.start.atSite(siteView.get).toInstant)
```
cquiroz, line 168: We maybe want to use middle of semester here

cquiroz (approved):
LGTM

toddburnside (approved):
LGTM

explore/src/main/scala/explore/targeteditor/ElevationPlotSection.scala
rpiaggio: I doubt it makes much of a difference for sidereal targets. For non-sidereal targets, I expect we will have to implement a different way of computing this, since coordinates can change each day (or even faster).

PR #4201, opened 2024-10-10
Hide the message for empty trees in the obs list

toddburnside:
When either the main observation tree or the calibration observation tree are empty, a message that is particularly unhelpful to users is displayed. This hides those messages.
image


cquiroz (approved):
LGTM

PR #4210, opened 2024-10-14
Minimal changes to handle BandedTimes for timeCharge

toddburnside:


model/shared/src/main/scala/explore/model/ProgramTimes.scala
```diff
@@ -6,9 +6,12 @@ package explore.model
 import cats.Eq
 import cats.derived.*
 import io.circe.Decoder
+import lucuma.core.util.TimeSpan
 
 final case class ProgramTimes(
   timeEstimateRange: Option[ProgramTimeRange],
-  timeCharge:        ProgramTime
+  timeCharge:        List[BandedProgramTime]
```
rpiaggio, line 13: Would a `SortedMap` better serve our purposes? Anyway, it can be revised later.

rpiaggio (approved):
LGTM

model/shared/src/main/scala/explore/model/ProgramTimes.scala
toddburnside: I'm not sure. It will have to be processed for the table anyway. (If we ever get to the point where we can make the table - it requires time estimates by band, which is less than simple for the ODB if the face of groups.)

PR #4220, opened 2024-10-17
Initialize selected target even if the ITC tile is closed

cquiroz:


explore/src/main/scala/explore/itc/ItcPanelTitle.scala
```diff
@@ -27,64 +28,67 @@ case class ItcPanelTitle(
...
   private type Props = ItcPanelTitle
 
   private val component =
-    ScalaFnComponent[Props] { props =>
-      def newSelected(p: Int): Option[ItcTarget] =
-        props.itcPanelProps.targets.lift(p)
+    ScalaFnComponent
+      .withHooks[Props]
+      // Reset the selected target if it changes
+      .useEffectWhenDepsReadyBy(props => props.itcGraphResults.map(_.brightestTarget)): props =>
```
rpiaggio, line 41: We probably want `useEffectWhenDepsReadyOrChange` if we want it to execute when the brightest target changes.

PR #4233, opened 2024-10-22
Workaround for twilight sequence

cquiroz:
This is good for a demo but needs a proper solution

### DIFF
--- a/explore/src/main/scala/explore/config/sequence/GeneratedSequenceBody.scala
+++ b/explore/src/main/scala/explore/config/sequence/GeneratedSequenceBody.scala
@@ -61,10 +61,11 @@ object GeneratedSequenceBody:
         obs.execution.config.map: config =>
           SequenceData(
             config,
-            Map(
-              ObserveClass.Science     -> obs.itc.science.selected.signalToNoise,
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/config/sequence/GeneratedSequenceBody.scala
```diff
@@ -61,10 +61,11 @@ object GeneratedSequenceBody:
         obs.execution.config.map: config =>
           SequenceData(
             config,
-            Map(
-              ObserveClass.Science     -> obs.itc.science.selected.signalToNoise,
-              ObserveClass.Acquisition -> obs.itc.acquisition.selected.signalToNoise
-            )
+            Map.empty
```
toddburnside, line 64: So, we're not displaying science and acquisition signal to noise values somewhere now? If Andy is OK with that for the demo, I'm fine, too. :)

toddburnside (approved):
LGTM

explore/src/main/scala/explore/config/sequence/GeneratedSequenceBody.scala
cquiroz: we can ask @andrewwstephens

PR #4235, opened 2024-10-22
Explicit selection of target editor in summary table

rpiaggio:
With this PR, the target editor is only shown below the target summary table if the user clicks on the target id. Also in that case, the plot is hidden.

The plot is restored if any row is selected by clicking anywhere else except the target id.

Also, the plot is always show even if no rows are selected (unless the target editor is shown).


toddburnside (approved):
LGTM

toddburnside (approved):
LGTM

PR #4238, opened 2024-10-23
Reinstante actions over focused target

rpiaggio:
With the last changes, copying and deleting through action buttons wouldn't work when a target was selected via clicking its id in the summary table. This PR fixes it.


toddburnside (approved):
LGTM

PR #4252, opened 2024-10-25
Observation summary table improvements

rpiaggio:
Kapture 2024-10-25 at 15 23 55

This PR implements most of the changes requested in https://app.shortcut.com/lucuma/story/3334/observation-summary-table-improvements, namely multiple selection in the observation summary table.

It was a bit tricky to allow selection of observations when they can span multiple table rows, since the rows are expandable, showing the asterism targets.

I was also tricky to make the action buttons work, since now we can copy, paste and delete multiple observations. I took the opportunity to group all observation actions into an object, with the sole purpose of organizing code.

Also, I reworked a bit the mechanism to share tables between tile headers and bodies (used for select all/none and for column visibility).

Finally, the row click handler for selecting multiple rows was generalized and moved to lucuma-ui.

Some requested behaviors are not implemented yet and will be addressed in future PRs. These are:
- Showing the observation group in the table.
- Elevation plot for observations below the summary table.


toddburnside (approved):
LGTM

PR #4253, opened 2024-10-25
Adds configuration requests and new observation workflow

toddburnside:
Adds configuration requests to Explore. For now, the only place you can request approval is on the summary tab in the validations table. This will change will a future PR. When an approval is requested, we optimistically update the affected observations, but they don't get updated via the API. This will change when the `ConfigurationRequest` in the API has the affected observations added to it and we can request them as part of the `configurationRequestEdit` event.

It also incorporates the new observation workflow, which removes `obsStatus` and `obsActiveStatus`. A new `ObservationWorkflow` is present on the `Observation` in the API, which contains a new calculated `ObservationWorkflowState`. It also contains a list of the valid states to which the observation can be transitioned, and the observation validations have been moved within the workflow. The `ObsBadge` has been updated to remove the Active State and change the status dropdown to update the new state where invalid transitions have been disabled. Currently there is no mutation for updating the state via the API, so any changes made locally will be overwritten on reload of the observation.


explore/src/clue/scala/queries/common/ConfigurationSubquery.scala
```diff
@@ -0,0 +1,40 @@
...
+      }
+      referenceCoordinates {
+        ra $RASubquery
+        dec $DecSubquery
+      }
+      observingMode {
+        instrument
+        mode
+        gmosNorthLongSlit {
+          grating
```
cquiroz, line 33: Why do we care only about grating?

explore/src/main/scala/explore/cache/ProgramCacheController.scala
```diff
@@ -132,12 +134,28 @@ object ProgramCacheController
...
+    val configurationRequests: IO[List[ConfigurationRequest]] =
+      drain[ConfigurationRequest,
+            ConfigurationRequest.Id,
+            ProgramSummaryQueriesGQL.AllProgramConfigurationRequests.Data
+      ](
+        offset =>
+          ProgramSummaryQueriesGQL
+            .AllProgramConfigurationRequests[IO]
+            .query(props.programId, offset.orUnassign),
+        _.program.fold(List.empty)(_.configurationRequests.matches),
```
cquiroz, line 154: `foldMap`?

explore/src/clue/scala/queries/common/ConfigurationSubquery.scala
toddburnside: 🤷 Apparently that is the only piece of a gmos long slit observation that matters when they are handing out approvals.

explore/src/main/scala/explore/cache/ProgramCacheController.scala
toddburnside: Ah. Good catch.

explore/src/main/scala/explore/cache/ProgramCacheController.scala
```diff
@@ -132,12 +134,28 @@ object ProgramCacheController
         offset =>
           ProgramSummaryQueriesGQL
             .AllProgramObservations[IO]
-            .query(props.programId.toWhereObservation, offset.orUnassign),
+            .query(props.programId.toWhereObservation, offset.orUnassign)(using
+              ErrorPolicy.IgnoreOnData
```
rpiaggio, line 138: Why is IgnoreOnData necessary?

model/shared/src/main/scala/explore/model/ProgramSummaries.scala
```diff
@@ -200,8 +204,10 @@ object ProgramSummaries:
     programs:            List[ProgramInfo],
     programTimesPot:     Pot[ProgramTimes],
     obsExecutionPots:    Map[Observation.Id, Pot[Execution]],
-    groupTimeRangePots:  Map[Group.Id, Pot[Option[ProgramTimeRange]]]
+    groupTimeRangePots:  Map[Group.Id, Pot[Option[ProgramTimeRange]]],
+    configRequests:      List[ConfigurationRequest]
   ): ProgramSummaries =
+    println(s"stati: ${obsList.map(_.workflow.state)}")
```
cquiroz, line 210: println

cquiroz (approved):
Lots of queries! LGTM

explore/src/main/scala/explore/observationtree/ObsList.scala
```diff
@@ -301,13 +300,9 @@ object ObsList:
                       props.obsExecutionTimes.getPot(obsId).map(_.programTimeEstimate),
                       ObsBadge.Layout.ObservationsTab,
                       selected = selected,
-                      setStatusCB = obsEditStatus(obsId)
+                      setStateCB = obsEditStatus(obsId)
```
rpiaggio, line 303: Do we want to rename to `obsEditState` too?

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -85,7 +85,7 @@ object ObsSummaryTable:
   private val GroupsColumnId          = ColumnId("groups")
   private val ObservationIdColumnId   = ColumnId("observation_id")
   private val ValidationCheckColumnId = ColumnId("validation_check")
-  private val StatusColumnId          = ColumnId("status")
+  private val StateColumnId           = ColumnId("status")
```
rpiaggio, line 88: Do we want to rename to "state" too?

explore/src/main/scala/explore/cache/ProgramCacheController.scala
toddburnside: I'm trying to remember. I'll get back to you shortly...

explore/src/main/scala/explore/observationtree/ObsList.scala
toddburnside: I'll do that after I rebase on your PR. 👅

model/shared/src/main/scala/explore/data/KeyedIndexedList.scala
```diff
@@ -99,6 +99,10 @@ case class KeyedIndexedList[K, A] private (private val list: TreeSeqMap[K, (A, N
   def updatedValueWith(key: K, f: A => A): KeyedIndexedList[K, A] =
     updatedWith(key, (v, i) => (f(v), i))
 
+  // WARNING - Do not use to update the part of the value used as a key
+  def unsafeMapValues(f: A => A): KeyedIndexedList[K, A] =
```
rpiaggio, line 103: You could make this safe by passing a lens for the key so that it gets reinstated in case it's modified. In fact, this would be automatic if you use `KIListMod(keyLens).mod`

model/shared/src/main/scala/explore/model/Observation.scala
```diff
@@ -179,33 +183,76 @@ case class Observation(
   lazy val constraintsSummary: String =
     s"${constraints.imageQuality.label} ${constraints.cloudExtinction.label} ${constraints.skyBackground.label} ${constraints.waterVapor.label}"
 
+  private val executedStates =
```
rpiaggio, line 186: `ExecutedStates`? Would it make sense to put it in `Constants`, or will it only be used here?

model/shared/src/main/scala/explore/model/ProgramSummaries.scala
rpiaggio, line 210: println

explore/src/main/scala/explore/cache/ProgramCacheController.scala
toddburnside: It's because we now get some errors with the observations for some things.
image
I guess we could review the ODB and see if that is really necessary. But, that is the way it is now,

rpiaggio (approved):
Looks awesome, thank you.

LGTM, with some comments.

model/shared/src/main/scala/explore/model/Observation.scala
toddburnside: As far as I know, it's only needed here. I guess that could change in the future?

model/shared/src/main/scala/explore/model/ProgramSummaries.scala
toddburnside: Too late. Carlos already reported that.

explore/src/main/scala/explore/cache/ProgramCacheController.scala
rpiaggio: Yeah, that one at least sounds more like a validation message than a query error.

But I agree that we want the IgnoreOnData in the presence of this. Maybe add a comment?

model/shared/src/main/scala/explore/model/Observation.scala
rpiaggio: Ok, anyway we use CamelCase for constants, it's in the new code guidelines ;)

toddburnside: We have guidelines? 😮 Anyway, their just guidelines - kind of like the Pirates Code. 😆

model/shared/src/main/scala/explore/data/KeyedIndexedList.scala
toddburnside: While digging through the KeyedIndexexList stuff to figure out how to make this work, I realized where I'm using this isn't participating in `undo`. 😢 I'll take care of that, and then figure this out.

toddburnside: Actually, I'm not sure this should participate in `undo`. We can't get rid of a ConfigurationRequest via the API. I guess once we get a mutation we could Withdraw the request on `undo` and set it back to `Requested` on `redo`? It's not quite the same as fully undoing it.

rpiaggio: Sounds like we want to make those actions explicit. Like submit/retract a proposal.

toddburnside: Yeah. I think you're right. Undo doesn't seem appropriate here. Thanks for the input.

toddburnside: I'll deal with the unsafeMapValues in a new PR. I want to get this merged so Rob can continue cleaning up the odb/core.

PR #4262, opened 2024-10-28
Elevation plots in observation summary

rpiaggio:
image

A new layout id is now used for the observation summary+plot. I renamed some tile ids so as to have a unified naming scheme, hence the large number of changed files.


USERPREFS.md
```diff
@@ -27,6 +27,14 @@ We have four environments with a respective db and heroku app:
 - staging
 - production
 
+## Console
+
+You can spin up a server that servers a console app for an environment by using the CLI from the `hasura/user-prefs` directory. For example:
+
+```
+hasura console --endpoint https://user-prefs-development.herokuapp.com
```
cquiroz, line 35: maybee we should recommend using  master

common/src/main/webapp/css/react-grid-layout.css
```diff
@@ -64,7 +64,8 @@
 }
 
 .rgl-tile-overlay {
-  display: none; /* Enable for grid debugging */
+  display: none;
```
cquiroz, line 67: remove debugging

cquiroz:
Could you squash the migrations?

common/src/main/scala/explore/model/ExploreGridLayouts.scala
```diff
@@ -269,6 +270,40 @@ object ExploreGridLayouts:
       ).withMinWidth
   end observations
 
+  object observationList:
```
toddburnside, line 273: Because the observation tree on the left is still name `ObsList`, I wonder if this should be named something like `observationSummary` for clarity?

toddburnside (approved):
LGTM

USERPREFS.md
rpiaggio: good point, I'll change it

common/src/main/webapp/css/react-grid-layout.css
rpiaggio: It's removed, but prettier doesn't let me put the comment in the same line anymore.

common/src/main/scala/explore/model/ExploreGridLayouts.scala
rpiaggio: `observationSummary` is just the tile, not the view with the tile+plot. I'd rather change `ObsList` to `ObsTree`

rpiaggio:
I just realized that the Hasura migration is missing from this PR. Hasura Console seems to create it (which I checked) but then deletes it after a few seconds, for some mysterious reason unknown to mere mortals.

I will add it to the next PR.

PR #4264, opened 2024-10-28
Final obs summary table improvements

rpiaggio:
- Include missing migration from previous PR.
- Show column with observation group. The name of the group is shown. If missing, then the id is shown. It's clickable and redirects to the group edit.
- Show the number of observations in the list header.

image

### DIFF
--- a/explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
+++ b/explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
@@ -20,6 +20,7 @@ import explore.model.AppContext
 import explore.model.Asterism
 import explore.model.Execution
 import explore.model.Focused
+import explore.model.Group
 import explore.model.GroupTree
 import explore.model.Observation
 import explore.model.ObservationExecutionMap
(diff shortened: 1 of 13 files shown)


explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
```diff
@@ -184,24 +185,59 @@ object ObsSummaryTable:
...
+                e.preventDefaultCB >> e.stopPropagationCB >>
+                  ctx.pushPage(
+                    AppTab.Observations,
+                    props.programId,
+                    Focused.singleObs(obsId)
+                  )
+              }
+            )(obsId.toString)
+
+          // TODO Groups are a dependency now, check that this updates is obs is moved around
```
toddburnside, line 222: TODO still needed?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/observationtree/ObsSummaryTable.scala
rpiaggio: nope, good catch, I'll remove

PR #4265, opened 2024-10-28
Remove KeyedIndexedList.unsafeMapValues in favor of a safe one.

toddburnside:


cquiroz (approved):
LGTM

model/shared/src/main/scala/explore/data/KeyedIndexedList.scala
```diff
@@ -99,9 +99,15 @@ case class KeyedIndexedList[K, A] private (private val list: TreeSeqMap[K, (A, N
...
 
-  // WARNING - Do not use to update the part of the value used as a key
-  def unsafeMapValues(f: A => A): KeyedIndexedList[K, A] =
-    KeyedIndexedList(TreeSeqMap.from(list.map { case (k, (a, idx)) => (k, (f(a), idx)) }))
+  def mapValues(keyLens: Lens[A, K])(f: A => A): KeyedIndexedList[K, A] =
+    KeyedIndexedList(
+      TreeSeqMap.from(
+        list.map { case (_, (a, idx)) =>
+          val newA = f(a)
+          (keyLens.get(newA), (newA, idx))
```
rpiaggio, line 107: I thought the idea was to get the `key` from the old value and force it back again into the new value.

It's a bit nicer this way, where we do allow the `key` to change, but in this case we don't need a `Lens` since we are not using the `setter` at all, it suffices to just pass `getKey: A => K`.

model/shared/src/main/scala/explore/data/KeyedIndexedList.scala
```diff
@@ -99,9 +99,15 @@ case class KeyedIndexedList[K, A] private (private val list: TreeSeqMap[K, (A, N
   def updatedValueWith(key: K, f: A => A): KeyedIndexedList[K, A] =
     updatedWith(key, (v, i) => (f(v), i))
 
-  // WARNING - Do not use to update the part of the value used as a key
-  def unsafeMapValues(f: A => A): KeyedIndexedList[K, A] =
-    KeyedIndexedList(TreeSeqMap.from(list.map { case (k, (a, idx)) => (k, (f(a), idx)) }))
+  def mapValues(keyLens: Lens[A, K])(f: A => A): KeyedIndexedList[K, A] =
```
hugo-vrijswijk, line 102: Minor suggestion, but should this be called `updateValues`? Since there's no mapping from type A to B?

rpiaggio: Good catch. Or it could actually map into `B`.

model/shared/src/main/scala/explore/data/KeyedIndexedList.scala
toddburnside: I'll change that. With our current usages, I don't really see us ever updating the key, but you never know...

rpiaggio (approved):
LGTM

PR #4267, opened 2024-10-28
Elevation plot improvements

rpiaggio:
- Make sure areas are painted below lines
- Draw faint line with minimum elevation at 30 degrees (only if plotting elevations).
- Make twilight texts a bit lighter.


cquiroz (approved):
LGTM

PR #4270, opened 2024-10-30
Simplify and unify ITC model

rpiaggio:
A bug was reported where invocations to ITC did not include CCD mode information when invoked before selecting a configuration.

This PR attempts to fix that by computing the necessary overrides for the current observations. This is now computed and stored in `InstrumentConfig` (previously known as `InstrumentRow`) when computing each possible row of the modes table. Also, the selected configuration is now on `InstrumentConfig` instead of a `BasicConfig`. This simplifies keeping track of the overrides since now they are in the selected object. In particular, `ItcProps` is simplified.

ITC results before and after configuration selection are now consistent.


explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -726,13 +723,7 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
...
               icon = Icons.ListIcon,
               severity = Button.Severity.Secondary,
-              onClick = props.selectedConfig.mod(c =>
-                BasicConfigAndItc(
-                  props.observingMode.get.toBasicConfiguration,
-                  c.flatMap(_.itcResult.flatMap(_.toOption.map(_.asRight)))
-                ).some
-              )
-                >> props.deleteConfig
+              onClick = props.deleteConfig
```
cquiroz, line 726: Should this be `revertConfig`?

explore/src/main/scala/explore/itc/ItcProps.scala
```diff
@@ -132,13 +120,16 @@ case class ItcProps(
...
         t    <- itcTargets
-        mode <- instrumentRow
+        mode <- instrumentConfig
       yield ItcClient[IO]
         .requestSingle:
           ItcMessage.GraphQuery(w, sn, snAt, constraints, t, mode)
         .map:
           _.toRight(new Throwable("No response from ITC server."))
         .rethrow
+        .flatTap: result =>
```
cquiroz, line 130: debugging?

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -175,12 +170,12 @@ object ObsTabTiles:
...
-      .useStateView(none[BasicConfigAndItc])
-      .useStateWithReuseBy: (props, _, _, _, selectedConfig) =>
-        itcQueryProps(props.observation.get, selectedConfig.get, props.allTargets)
+      .useStateView(none[InstrumentConfigAndItcResult])
+      .useStateWithReuseBy: (props, _, _, _, selectedConfig) => // itcQueryProps
+        ItcProps(props.observation.get, selectedConfig.get, props.obsTargets)
       .useState(Pot.pending[ItcAsterismGraphResults]) // itcGraphResults
       .useAsyncEffectWithDepsBy((props, _, _, _, selectedConfig, _, _) =>
-        itcQueryProps(props.observation.get, selectedConfig.get, props.allTargets)
+        ItcProps(props.observation.get, selectedConfig.get, props.obsTargets)
```
cquiroz, line 178: Caan this be combined in one hook? Like one of those useResourceWhen...

rpiaggio: Let me check. At the very least, we can use a `localVal` so that we don't build `ItcProps` twice.

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
```diff
@@ -216,22 +100,60 @@ case class SpectroscopyModeRow(
...
+      val instrumentConfig: Option[InstrumentConfig] =
+        instrument.instrument match
+          case Instrument.GmosNorth | Instrument.GmosSouth =>
+            instrument match
+              case i @ InstrumentConfig.GmosNorthSpectroscopy(grating, fpu, _, None) =>
+                i.copy(modeOverrides =
+                  InstrumentOverrides
+                    .GmosSpectroscopy(
+                      cw,
+                      GmosCcdMode.defaultGmosNorth(profiles, fpu, grating, imageQuality),
```
cquiroz, line 125: Nice, I didn't notice this method. Is it from `lucuma-core`?

model/shared/src/main/scala/explore/modes/syntax.scala
```diff
@@ -0,0 +1,63 @@
...
+import lucuma.core.enums.GmosSouthGrating
+import lucuma.core.enums.GmosXBinning
+import lucuma.core.enums.GmosYBinning
+import lucuma.core.enums.ImageQuality
+import lucuma.core.model.SourceProfile
+import lucuma.core.model.sequence.gmos.GmosCcdMode
+import lucuma.core.model.sequence.gmos.longslit.*
+
+object syntax:
+  // TODO Actually move to lucuma-core, to GmosCcdMode
```
cquiroz, line 20: 👍 so odb can use it too

workers/src/main/scala/workers/itc/ITCGraphRequests.scala
```diff
@@ -61,12 +59,13 @@ object ITCGraphRequests:
...
           m
         ).some
-      case _                                        =>
+      case _                                                      =>
         none
 
     def doRequest(request: ItcGraphRequestParams): F[ItcAsterismGraphResults] =
       request.mode.toItcClientMode
         .map: mode =>
+          println(s"request for mode $mode")
```
cquiroz, line 68: println

workers/src/main/scala/workers/itc/ITCRequests.scala
```diff
@@ -73,7 +71,7 @@ object ITCRequests:
       params: ItcRequestParams
     ): F[Option[Map[ItcRequestParams, EitherNec[ItcTargetProblem, ItcResult]]]] =
       Logger[F]
-        .debug(
+        .info(
```
cquiroz, line 74: Should it go back to `debug`?

model/shared/src/main/scala/explore/modes/SpectroscopyModesMatrix.scala
rpiaggio: It's an extension I created. But I think it should be moved to `lucuma-core` in the future.

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
rpiaggio: I was able to simplify quite a bit by using `localVal` and `useEffectResult`. Thanks for the suggestion!

PR #4271, opened 2024-10-30
Migrate to latest ODB changes regarding sequence offsets

rpiaggio:


explore/src/clue/scala/queries/common/ObsQueriesGQL.scala
```diff
@@ -28,21 +28,10 @@ object ObsQueriesGQL:
   @GraphQL
   trait SequenceOffsets extends GraphQLOperation[ObservationDB]:
     val document = s"""
-      fragment stepDataGN on GmosNorthStep {
+      fragment stepData on GmosNorthStep {
```
toddburnside, line 31: This works for GmosSouth, too, even though it specifies it is for GmosNorthStep?

toddburnside (approved):
LGTM

explore/src/clue/scala/queries/common/ObsQueriesGQL.scala
rpiaggio: Good catch. It does seem to work. But I'll revert, just in case.

toddburnside: Maybe grackle is not enforcing that.

toddburnside (approved):
LGTM

PR #4281, opened 2024-11-04
Add Unrequested Configurations table

toddburnside:
Creates the new `Unrequested Coordinates + Configurations + Constraints` table which allows the user to create new configuration requests. It also abstracts out the common columns with the `Requested Coordinates + Configurations + Constraints` table.

The next PR will save the states of these 2 tables and the observation validations table, as well as remove the ability to create a new configuration request from the observations validations table (it turns out to be not such a great place to make the requests from).

image
image

### DIFF
--- a/explore/src/main/scala/explore/model/AppContext.scala
+++ b/explore/src/main/scala/explore/model/AppContext.scala
@@ -19,7 +19,9 @@ import fs2.dom.BroadcastChannel
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.extra.router.SetRouteVia
 import japgolly.scalajs.react.feature.Context
+import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.enums.ExecutionEnvironment
+import lucuma.core.model.Observation
 import lucuma.core.model.Program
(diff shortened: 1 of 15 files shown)


explore/src/main/scala/explore/model/AppContext.scala
```diff
@@ -53,6 +55,27 @@ case class AppContext[F[_]](
   def replacePage(appTab: AppTab, programId: Program.Id, focused: Focused): Callback =
     setPageVia(appTab, programId, focused, SetRouteVia.HistoryReplace)
 
+  def routingLink(
```
rpiaggio, line 58: Good idea! There are a number of places where this could be used.

rpiaggio: I see you have already refactored to use it.

toddburnside: I refactored some of the areas, but I probably missed some.

toddburnside:
> Screenshot?

Added to the description.

I should also add that the text entered in the request editor is currently ignored, but the message should be added to the API at some point soon.

PR #4282, opened 2024-11-04
Hide Inactive obs and config request errors in validations table

toddburnside:
Also removes the Actions column from the table, since the only thing it allowed at this point was to request approval and that is no longer relevant since those observations aren't shown anyway.


cquiroz (approved):
LGTM

PR #4283, opened 2024-11-04
Hide unavailable states in ObsBadge dropdown

toddburnside:
Andy wanted the unavailable states removed from the dropdown instead of being grayed out.

Also changes the format of the messages for `ConfigurationRequest`* errors.


cquiroz (approved):
LGTM

PR #4291, opened 2024-11-06
Add margin to last item in the obs lists to provide room for url preview

toddburnside:
As suggested in the shortcut story, this adds a margin to the last item in the trees/lists in the left panel of explore so that the user can scroll them up above the url preview in the browser.


cquiroz (approved):
LGTM

PR #4293, opened 2024-11-06
Exclude inactive and calibration obs from config request tables

toddburnside:
Calibration observations are ignored for both the `Requested` and `Unrequested` tables since they don't need approval.

Inactive observations are filtered out of the `Unrequested` table so requests can't be generated for a configuration if it only has inactive observations. They are still kept in the list of observation ids in the `Requested` table, but the text in the link is strike throughed (struck through?).


cquiroz (approved):
LGTM

PR #4295, opened 2024-11-08
Tile reorganization, column selector fix

rpiaggio:
This PR started out as an attempt to fix https://app.shortcut.com/lucuma/story/3639/column-selector-drop-downs-for-summary-tables-don-t-update, which it does, but then I went into the rabbit hole of aligning the controls in the tile titles and from there to the even deeper rabbit hole of organizing tile code.

The result is that each tile now has all of its code in only one file, with `Body` and `Title` as private components, a `TileState` for the shared state among them, and an `apply` method that builds the tile and passes to each subcomponent only what they need from the state.

This eliminates the files that only contained glue code to build the tiles (that code is now the new `apply` methods).

Styling of titles has been simplified a bit too.

Oh, and now the column selector checkboxes show whether the column is visible :D


common/src/main/scala/explore/components/Tile.scala
```diff
@@ -135,31 +135,22 @@ object Tile:
...
           )(
             // Tile title
-            <.div(ExploreStyles.TileTitle)(
-              ^.key := s"tileTitle-${p.id.value}",
+            <.div(ExploreStyles.TileTitle, ^.key := s"tileTitle-${p.id.value}")(
               // Title and optional back button
-              <.div(
-                ExploreStyles.TileTitleMenu |+| p.tileTitleClass,
+              <.div(ExploreStyles.TileTitleMenu |+| p.tileTitleClass)(
                 p.back.map(b => <.div(ExploreStyles.TileButton, b)),
```
cquiroz, line 141: is it correct to move the back  button to the title text?

explore/src/main/scala/explore/config/ConfigurationTile.scala
```diff
@@ -0,0 +1,327 @@
...
+        selectedConfig,
+        revertedInstrumentConfig,
+        modes,
+        sequenceChanged,
+        readonly,
+        units
+      )
+    )
+
+  private case class Body(
```
cquiroz, line 99: Seems a good refactoring in general

explore/src/main/scala/explore/observationtree/ObsSummaryTile.scala
```diff
@@ -74,87 +78,134 @@ object ObsSummaryTable:
...
-      PriorityColumnId      -> "Priority",
       RAColumnId            -> "RA",
       DecColumnId           -> "Dec",
-      TimingWindowsColumnId -> "Scheduling Windows",
       SEDColumnId           -> "SED",
-      ChargedTimeColumnId   -> "ChargedTime"
+      ConstraintsColumnId   -> "Constraints",
+      ConfigurationColumnId -> "Configuration",
+      DurationColumnId      -> "Duration"
+      // ValidationCheckColumnId -> " ",
```
cquiroz, line 146: Commented code

cquiroz (approved):
In general I'm ok with the refactor. It is hard to tell about the change on the dropdown among the other changes

LGTM

explore/src/main/scala/explore/observationtree/ObsSummaryTile.scala
rpiaggio: We have commented the columns that are pending to add. I just consolidated them.

common/src/main/scala/explore/components/Tile.scala
rpiaggio: This didn't change. I just separated style and content into two different parameter lists (which I find it resembles HTML's structure of attributes together with the tag and then content in a different block).

PR #4296, opened 2024-11-08
Rename ItcProps to ItcGraphQuerier, fix the reusability

toddburnside:


explore/src/main/scala/explore/itc/ItcGraphQueryer.scala
```diff
@@ -98,10 +98,16 @@ case class ItcProps(
 
   val targets: List[ItcTarget] = itcTargets.foldMap(_.toList)
 
-  private val queryProps: List[Option[?]] =
-    List(itcTargets, finalConfig, wavelength, instrumentConfig, signalToNoise)
+  private val queryProps =
+    (wavelength,
+     signalToNoise,
+     signalToNoiseAt,
+     constraints.some,
```
rpiaggio, line 105: I don't think `constraints` is needed. All that we need from it is already factored into `instrumentConfig`.

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/itc/ItcGraphQuerier.scala
```diff
@@ -98,10 +98,16 @@ case class ItcProps(
...
   val targets: List[ItcTarget] = itcTargets.foldMap(_.toList)
 
-  private val queryProps: List[Option[?]] =
-    List(itcTargets, finalConfig, wavelength, instrumentConfig, signalToNoise)
+  private val queryProps =
+    (wavelength,
+     signalToNoise,
+     signalToNoiseAt,
+     constraints.some,
+     itcTargets,
```
cquiroz, line 106: Is it nott enough to query only for target ids?

toddburnside: Does the itc know the target ids?

toddburnside: This would be a bigger change. I'm going to merge, but we can continue discussing.

cquiroz: If I recall this is used only to tell  if the properties have changed, and using the ids would bee more efficient that comparing the whole target

toddburnside: The ItcTargets, with the gaia bands removed, are passed to `ItcClient.spectroscopyIntegrationTimeAndGraphs`. So, I assume the itc needs the target information?

cquiroz: I see, for reusability would the `id` be enough?

toddburnside: I don't think so. If a target is edited, it keeps the same id but the itc results could be affected.

PR #4297, opened 2024-11-08
Get target ITC band from ITC results instead of computing it in the client

rpiaggio:


cquiroz:
Is this because the ITC server calculates the band out of the requested wavelength?
Don't they match?

rpiaggio:
> Is this because the ITC server calculates the band out of the requested wavelength? Don't they match?

Yes, we have duplicated logic, which can cause maintainability issues in case it changes. This unifies the logic on the server side.

cquiroz (approved):
LGTM

PR #4299, opened 2024-11-09
Correctly plot observation time and refresh visualization if it changes

rpiaggio:


explore/src/main/scala/explore/targeteditor/SiderealTargetEditor.scala
```diff
@@ -82,8 +81,6 @@ case class SiderealTargetEditor(
 object SiderealTargetEditor:
   private type Props = SiderealTargetEditor
 
-  private given Reusability[Instant] = siderealTargetReusability
```
rpiaggio, line 85: I'm not sure why we were using this reusability logic here. It only refreshed if obs time changed a month or more, which prevented the visualization to update for smaller changes.

cquiroz: We didn't want to recalculate AGS unless `obsTime` changed more than a month. I can see why it  could be a problem for visualization. Can we be smarter and make it only use this criteria if not using average PA?

rpiaggio: Good idea, I'll attempt that. I'll also try to make the logic a bit more explicit, it took me a while to realize we were using a non-standard implicit `Reusability`.

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -114,8 +113,8 @@ object AladinContainer extends AladinCommon {
       // View coordinates base coordinates with pm correction + user panning
       .useStateBy: (p, baseCoordinates) =>
         baseCoordinates.value._1.value.offsetBy(Angle.Angle0, p.options.viewOffset)
-      // Update coordinates if asterism or vizTime changes
-      .useEffectWithDepsBy((p, _, _) => (p.asterism, p.vizTime)):
+      // Update coordinates if asterism or obsTime changes
+      .useEffectWithDepsBy((props, _, _) => (props.asterism, props.obsTime)):
```
cquiroz, line 117: So this allows a re render anytime obsTime changes

explore/src/main/scala/explore/targeteditor/SiderealDiscretizedObsTime.scala
```diff
@@ -0,0 +1,30 @@
...
+// If it changes over a month we'll request the data again and recalculate.
+// This way we avoid recalculating PM for example if only pos angle or conditions change.
+// However, if `AverageParallactic` is used, we need to recompute GS candidates for each obsTime.
+case class SiderealDiscretizedObsTime(
+  obsTime:            Instant,
+  posAngleConstraint: Option[PosAngleConstraint]
+)
+
+object SiderealDiscretizedObsTime:
+  given Reusability[SiderealDiscretizedObsTime] = Reusability[SiderealDiscretizedObsTime]: (a, b) =>
```
cquiroz, line 26: nice, thanks for the comment

explore/src/main/scala/explore/targeteditor/AladinCell.scala
```diff
@@ -225,31 +225,35 @@ object AladinCell extends ModelOptics with AladinCommon:
...
-                candidates <- CatalogClient[IO]
-                                .requestSingle:
-                                  CatalogMessage.GSRequest(baseTracking, obsTime)
-              yield candidates)
-                .guarantee:
-                  props.obsConf.flatMap(_.agsState).foldMap(_.async.set(AgsState.Idle))
-            else none.pure
+          if (props.needsAGS)
+            (for
+              _          <- props.obsConf
```
cquiroz, line 236: So we run ags only if time changes more thaan a month, except for average pa
that seems right

explore/src/main/scala/explore/targeteditor/AladinContainer.scala
rpiaggio: Yes, I wasn't sure whether to use the discretized time here. But it's a cheap computation anyway.

cquiroz: yes, cheaper than ags

PR #4304, opened 2024-11-11
Use stateInViewHandler where possible, enable restore table state

rpiaggio:
This fixes the issue where column selection was lost when navigating away from the table.


cquiroz (approved):
LGTM

PR #4307, opened 2024-11-13
Allow setting X and Y Binning independently

toddburnside:
image

The `Read Mode` now also handles when either `AmpReadMode` or `AmpGain` is explicitly set via the API and the other is not. It does not allow for setting them independently - if the Read Mode is set via Explore, both are explicitly set.

### DIFF
--- a/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
@@ -133,11 +133,17 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
     Logger[IO]
   ): View[Fpu]
 
-  @inline protected def explicitBinning(aligner: AA)(using
+  @inline protected def explicitXBinning(aligner: AA)(using
     MonadError[IO, Throwable],
     Effect.Dispatch[IO],
     Logger[IO]
-  ): View[Option[(XBinning, YBinning)]]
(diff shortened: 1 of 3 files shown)


rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -681,19 +689,30 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
...
+            <.div(
+              ExploreStyles.AdvancedConfigurationBinning,
+              customizableEnumSelectOptional(
+                id = "explicitXBin".refined,
+                view = explicitXBinning(props.observingMode).withDefault(defaultXBinning),
+                original = defaultXBinning.some,
+                disabled = disableAdvancedEdit
+              ),
+              FormLabel(htmlFor = "explicitYBin".refined)("x"),
+              customizableEnumSelectOptional(
```
hugo-vrijswijk, line 701: Shouldn't this have a formLabel too?

toddburnside: Both the dropdowns do have labels. However, your comment made me realize that the current labels are not very helpful for screen readers, so I disassociated the labels from the dropdowns and used `aria-label` directly on the dropdowns.

PR #4308, opened 2024-11-13
Simplify group model

rpiaggio:
A while ago I separated the group tree structure from the information of observations.

However, when attempting to fix https://app.shortcut.com/lucuma/story/3871/difficult-to-delete-groups it came to my attention that having observation information (the observation itself and its position in the group tree) is just not friendly at all with undo.

This PR rolls back the previous change and goes a step further: we no longer keep a tree in the model. Instead, we just keep a list of groups and observations with their parent info (parent group and index in the group), mimicking the data provided by the server. The group tree can be reconstructed from this info just for displaying it in the UI. This means we don't have to deal with a tree in the model, nor with disperse bits of information.

This also means that `KeyedIndexedList` and `KeyedIndexedTree` are not used anymore. They were useful for treating the position if an element in a list or tree as a property, so that it played nice with undo, but it turns out that this information is already properties in the model, so we don't actually need them. I'll consider removing the code in a future PR.


common/src/main/scala/explore/utils/package.scala
```diff
@@ -309,3 +310,16 @@ def keyedSwitchEvalMap[F[_]: Concurrent, I, O, K](
 
 def timeDisplay(name: String, time: TimeSpan, sep: String = ": ") =
   <.span(<.span(ExploreStyles.SequenceTileTitleItem)(name, sep), TimeSpanView(time))
+
+extension (s: NonNegShort)
+  def +|(i: Int): NonNegShort =
```
toddburnside, line 315: Is this safe? You could add a negative number and get a number less than 0. Similarly, you could subtract a negative number and overflow the short.

explore/src/main/scala/explore/observationtree/ConstraintGroupObsList.scala
```diff
@@ -131,10 +130,9 @@ object ConstraintGroupObsList:
 
     val traversal = Iso
       .id[ObservationList]
-      .filterIndex((id: Observation.Id) =>
+      .filterIndex: (id: Observation.Id) =>
         oData.exists((_, _, draggedIds, _) => draggedIds.contains(id))
-      )
-      .andThen(KeyedIndexedList.value)
+      // .andThen(KeyedIndexedList.value)
```
toddburnside, line 135: Commented code

toddburnside (approved):
LGTM

explore/src/main/scala/explore/tabs/ObsGroupTiles.scala
```diff
@@ -34,85 +33,53 @@ case class ObsGroupTiles(
...
-                    .map(modF)
-                    .map((newValue, newIndex) =>
-                      groups.upserted(groupTreeKey, newValue.value, newIndex)
-                    )
-                    .getOrElse(groups)
+      val editTile: Tile[?] =
+        Tile(
+          GroupEditTileIds.GroupEditId.id,
+          s"${
+              if props.group.get.system then props.group.get.name.foldMap(_.value)
```
hugo-vrijswijk, line 45: Does this no longer append the type of group after the name?

hugo-vrijswijk (approved):
Nice! I really like having the tree structure only as some view data, and keeping the data similar to the server in the rest of the application

explore/src/main/scala/explore/tabs/ObsGroupTiles.scala
rpiaggio: Not for system groups. This logic was already in place.

rpiaggio:
Forgot to mention that we now have undo/redo for drag and drop in the obs tree. With the new model it was almost trivial to support it.

PR #4314, opened 2024-11-15
Fix scheduling group update, expand group with focused observation

rpiaggio:
The scheduling group update was victim of a memoized callback in the table column definition: since the callback depends on the selected observations, delete operations were always applied to the original selection rather than the current one. Solved it by use of table meta.

Also, it was requested that the focused observation is still shown when edited, even if changes groups. Therefore, logic was added so that when the groups change, the group containing the focused obs is expanded (and a group from which observations are removed is not collapsed).

### DIFF
--- a/explore/src/main/scala/explore/observationtree/SchedulingGroupObsList.scala
+++ b/explore/src/main/scala/explore/observationtree/SchedulingGroupObsList.scala
@@ -40,6 +40,7 @@ import lucuma.react.fa.FontAwesomeIcon
 import lucuma.react.floatingui.syntax.*
 import lucuma.schemas.ObservationDB
 import lucuma.ui.primereact.*
+import lucuma.ui.reusability.given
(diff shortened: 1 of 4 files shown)


toddburnside (approved):
LGTM

toddburnside:
> Also, it was requested that the focused observation is still shown when edited, even if changes groups. Therefore, logic was added so that when the groups change, the group containing the focused obs is expanded (and a group from which observations are removed is not collapsed).

I hate to ask, but does this work OK with undo/redo? 🙉

rpiaggio:
> > Also, it was requested that the focused observation is still shown when edited, even if changes groups. Therefore, logic was added so that when the groups change, the group containing the focused obs is expanded (and a group from which observations are removed is not collapsed).
> 
> I hate to ask, but does this work OK with undo/redo? 🙉

Yes, it does! Since it's recomputed from scratch on every change.

PR #4321, opened 2024-11-18
Autofocus name field in TargetSelectionPopup

toddburnside:
I tried using `Ref`s to focus the input, and nothing would work. Then I set `closable` to false and the Refs worked, but so did the `^.autoFocus`...

The `X` icon in the upper right is now not displayed, and the user can't close it with Escape. But closing by clicking on the close button in the footer or outside of the dialog works.


cquiroz (approved):
LGTM

rpiaggio (approved):
LGTM, is this worth reporting to PrimeReact?

PR #4322, opened 2024-11-18
Configuration drop down menu

rpiaggio:
Kapture 2024-11-18 at 20 06 51


model/shared/src/main/scala/explore/model/ObservingModeSummary.scala
```diff
@@ -0,0 +1,143 @@
...
+    observingMode match
+      case ObservingMode.GmosNorthLongSlit(
+            _,
+            grating,
+            _,
+            filter,
+            _,
+            fpu,
+            _,
+            centralWavelength,
```
cquiroz, line 80: It seems there maybe plenty of unused params

model/shared/src/main/scala/explore/model/ObservingModeSummary.scala
```diff
@@ -0,0 +1,143 @@
...
+    grating:           GmosSouthGrating,
+    filter:            Option[GmosSouthFilter],
+    fpu:               GmosSouthFpu,
+    centralWavelength: CentralWavelength,
+    ampReadMode:       GmosAmpReadMode,
+    roi:               GmosRoi
+  ) extends ObservingModeSummary
+
+  def toInput: ObservingModeInput = this match
+    case GmosNorthLongSlit(grating, filter, fpu, centralWavelength, ampReadMode, roi) =>
```
cquiroz, line 46: Is  this matching the grouping on the odb?
How do we keep them in sync?

model/shared/src/main/scala/explore/model/ObservingModeSummary.scala
rpiaggio: Yes, the story requests:

```
The components that should be included are:
Grating, Filter, FPU, Central Wavelength, Read Mode, ROI
```

model/shared/src/main/scala/explore/model/ObservingModeSummary.scala
rpiaggio: No, this is independent of the groupings in the ODB. The ODB groupings take into account all of the `ObservationMode` parameters.

model/shared/src/main/scala/explore/model/ObservingModeSummary.scala
toddburnside: I think he was referring to params being named in the match but not used?

toddburnside (approved):
LGTM!

model/shared/src/main/scala/explore/model/ObservingModeSummary.scala
rpiaggio: Ah right, I did a first version which used everything. I'm removing them.

model/shared/src/main/scala/explore/model/ObservingModeSummary.scala
swalker2m: What happens when two configurations differ only by parameters that are not present in the summary?

rpiaggio: > What happens when two configurations differ only by parameters that are not present in the summary?

They are lumped together. Then when one of them is selected, only those parameters are applied. The rest of the parameters are not sent in the mutation. Does the ODB keep the old values in this case? Should we set them to null?

swalker2m: Yes, with a mutation in general only the fields that you set are modified and the rest are left alone.  If there are fields that default based on the values in the part being modified, they may change.  That would be the case for the binning.  On the other hand, if the binning is explicitly set it will not change even if it doesn't make sense for the new instrument configuration.

rpiaggio: It sounds to me then that we want to reset the overrides. Does this sound right @andrewwstephens ?

PR #4329, opened 2024-11-21
Default slit observations to use parallactic angle plus more undo/redo

toddburnside:
In addition to the setting the Position Angle Constraint (PAC) to Average Parallactic when a long slit observation is selecting, and preparing for other modes with other defaults, this PR does a number of other things. Including, but maybe not limited to:

- When a guide star is selected in Allow Flip mode, it does not update the actual PAC if the guide star is for the flipped angle. Instead, this information is shown alongside the PAC in the Configuration Tile.
- PAC now takes part in undo/redo
- Selecting or reverting a configuration via the configurations dropdown now takes part in undo/redo
- Reverting a configuration via the configurations dropdown better supports setting the appropriate line in the spectroscopy table when the revert is done in a different session.
- We were sending 2 requests to the ODB when creating a configuration. It should now only send one request.

### DIFF
--- a/explore/src/main/scala/explore/config/ConfigurationTile.scala
+++ b/explore/src/main/scala/explore/config/ConfigurationTile.scala
@@ -29,11 +29,15 @@ import explore.model.ObsTabTileIds
 import explore.model.Observation
 import explore.model.ObservingModeGroupList
 import explore.model.ObservingModeSummary
+import explore.model.PosAngleConstraintAndObsMode
 import explore.model.ScienceRequirements
 import explore.model.ScienceRequirements.Spectroscopy
 import explore.model.TargetList
+import explore.model.enums.AgsState
+import explore.model.enums.PosAngleOptions
 import explore.model.enums.WavelengthUnits
 import explore.model.itc.ItcTarget
+import explore.model.syntax.all.*
 import explore.modes.InstrumentConfig
 import explore.modes.SpectroscopyModesMatrix
 import explore.syntax.ui.*
@@ -41,6 +45,7 @@ import explore.undo.*
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.Callback
 import japgolly.scalajs.react.vdom.html_<^.*
+import lucuma.core.math.Angle
 import lucuma.core.model.CoordinatesAtVizTime
 import lucuma.core.model.PosAngleConstraint
 import lucuma.core.model.Program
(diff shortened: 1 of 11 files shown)


explore/src/main/scala/explore/config/ConfigurationTile.scala
```diff
@@ -192,6 +292,29 @@ object ConfigurationTile:
...
+              props.posAngle,
+              UpdateObservationsInput(
+                WHERE = props.obsId.toWhereObservation.assign,
+                SET = ObservationPropertiesInput()
+              ),
+              ObsQueriesGQL
+                .UpdateObservationMutation[IO]
+                .execute(_)
+                .void
+                .switching(props.agsState.async, AgsState.Saving, AgsState.Idle)
```
toddburnside, line 307: @cquiroz I'm not sure exactly what the `agsState` is for. Should it be `switched` when updating the configuration, too?

cquiroz:
Does this need a backend counterpart? So that when the config is set via api, PAC is also set accordingly?

explore/src/main/scala/explore/config/ConfigurationTile.scala
cquiroz: `agsState` is used to show where we are in the calculation process, loading stars, calculating, etc
I think it should recalculate if conf changes

toddburnside:
> Does this need a backend counterpart? So that when the config is set via api, PAC is also set accordingly?

I was thinking it was a UI only thing, but perhaps @andrewwstephens could chime in. Maybe API users don't need the encouragement to exhibit good behavior (to paraphrase the story)? The backend would need to be changed to even know what the current PAC is when setting the configuration. Or maybe to even know that we are setting a new configuration rather than updating and existing one.

explore/src/main/scala/explore/config/ConfigurationTile.scala
toddburnside: So, should it be set/unset for any configuration changes?  But, doesn't the `switching` here set it back to `Idle` when we get the response back from the API mutation? That seems it wouldn't have anything to do with the actual AGS calculation cycle.

toddburnside: It seems that the "calculating" gets displayed for a PAC change, even without this switching here. So, I'll remove it.

explore/src/main/scala/explore/config/PAConfigurationPanel.scala
```diff
@@ -93,7 +93,10 @@ object PAConfigurationPanel:
               )
             )
             .orElse(<.label("Not Visible").some)
-        case _                                     => None
+        case PosAngleConstraint.AllowFlip(af) if props.selectedPA.exists(_ =!= af) =>
+          props.selectedPA
+            .map(a => <.label(f"Flipped to ${a.toDoubleDegrees}%.0f °"))
```
cquiroz, line 98: so we only show it if flipped right?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/config/PAConfigurationPanel.scala
toddburnside: Correct

PR #4330, opened 2024-11-22
Migrate to new ITC API

rpiaggio:


rpiaggio:
> Do we need to bust the cache?

Good question. Existing cache entries will be ignored since input changed. Will the cache get auto busted if we increase the version number?

cquiroz:
Ok, no need then, the cache will delete old entries over time

cquiroz (approved):
LGTM

PR #4336, opened 2024-11-26
Hide some of the tiles until the proposal has been accepted.

toddburnside:


cquiroz (approved):
LGTM I don't think this messes with the save tiled state if you accept the proposal, does it?

toddburnside:
> LGTM I don't think this messes with the save tiled state if you accept the proposal, does it?

It shouldn't because of the dummy tiles. This is the technique Raul used for the target tiles and it seems to be working. 🤞

PR #4352, opened 2024-12-02
Fix change editor

cquiroz:
This is an odd bug, it manifest itself in safari, where the signal to noise fields gets right away the focus and you can't jump to another control, After checking a bit I think the problem as on the text input with a change  auditor
Now the change auditor has a different set of decimals than the validator and fixing that solves the proble


toddburnside (approved):
Wow. Good catch. Seems like it must have been hard to find... LGTM

PR #4361, opened 2024-12-03
Update to use the new schema for users

cquiroz:


toddburnside (approved):
LGTM. Thanks for taking care of this!

model/shared/src/main/scala/explore/model/ProgramUser.scala
```diff
@@ -7,34 +7,27 @@ import cats.Eq
...
+import lucuma.core.model.UserProfile
+import lucuma.sso.client.codec.userProfile.given
 
 // Case class for the 'pi' and `users` in the Program. These must be
 // actual users, not service users.
 case class ProgramUser(
   id:      User.Id,
-  profile: Option[OrcidProfile]
+  orcidId: Option[OrcidId],
+  profile: Option[UserProfile]
```
swalker2m, line 20: There should now be a `fallbackProfile` on `ProgramUser` in the ODB schema.  The idea is to use the fallback profile if the information extracted from ORCiD is not public or not set.

PR #4366, opened 2024-12-06
Remove notes tile for groups

toddburnside:
Removed as per the discussion with @andrewwstephens in the shortcut story.

I kept the time manager, etc. for the groups even though there is only one tile left, but we seem to have a habit of adding more things to the UI...


rpiaggio (approved):
LGTM

PR #4369, opened 2024-12-07
Handle initialization errors, fix leak when switching programs

rpiaggio:
If there was an error when initializing the cache, explore would just hang. If it happened in an update, explore would silently break synchronization.

This PR fixes both issues by using `Pot` for the cache instead of `Option` as well as `attempt` for the initialization and update stream processing.

Also, if the user switched programs, the cache-updater fibers for the old program would still run. This PR also fixes that by starting the fibers as `Resource`s using `background`.


cquiroz (approved):
LGTM

PR #4373, opened 2024-12-08
Use central wavelength for AGS thus able to calculate guide star for …

cquiroz:
…calibrations

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -55,6 +55,7 @@ import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.enums.Site
 import lucuma.core.math.Angle
 import lucuma.core.math.Offset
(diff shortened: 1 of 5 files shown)


cquiroz:
@toddburnside we may want to verify we are using the same on the odb ags calculation

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -146,6 +148,19 @@ object ObsTabTiles:
     acquisition: Option[NonEmptyList[Offset]]
   )
 
+  // TODO Move to core
+  extension (om: ObservingMode)
+    def centralWavelength: Option[Wavelength] =
```
rpiaggio, line 153: We have a `CentralWavelength` type alias, do we want to return that instead?

cquiroz: Yeah, I think that would be better, I'll change it

rpiaggio (approved):
LGTM

PR #4378, opened 2024-12-09
Fix class for BandedProgramTime to consider the band as optional

cquiroz:
The field is not in use but it will be requested in the near future to display charges per band


toddburnside (approved):
LGTM

rpiaggio (approved):
LGTM

PR #4383, opened 2024-12-10
Final leak fix + restore updates

rpiaggio:
A previous PR that promised to fix a leak when changing programs not only didn't live up to its promise but also broke subscription updates.

This PR includes the missing finishes touches of said PR and fixed both issues.


cquiroz (approved):
LGTM

PR #4384, opened 2024-12-11
Add support for readonly CoIs

toddburnside:
[sc-4271] Adds ability to create invitations for read only CoIs and makes the program readonly for them.
image


cquiroz (approved):
LGTM

PR #4385, opened 2024-12-12
Allow withdrawing a configuration request

toddburnside:
Also prevents readonly COIs from making/changing requests


cquiroz (approved):
LGTM

PR #4386, opened 2024-12-12
Support proposal and program id in the URL

rpiaggio:
Resolves proposal, program and observation ids in the URL. Redirects to the usual-style URL.


explore/src/main/scala/explore/ObsReferenceResolver.scala
```diff
@@ -0,0 +1,43 @@
...
+        import ctx.given
+
+        ResolveObsReference[IO]
+          .query(props.obsRef.assign)
+          .flatMap: data =>
+            data.observation
+              .map: o =>
+                ctx.pushPage(AppTab.Observations, o.program.id, Focused.singleObs(o.id)).to[IO]
+              .getOrElse:
+                IO.raiseError:
```
cquiroz, line 40: Should we instead be redirected to the program selector?

rpiaggio: Good idea, I'll try to show the error and the program selector.

explore/src/main/scala/explore/Routing.scala
```diff
@@ -52,135 +60,151 @@ object Routing:
...
+                  .get
+                  .toOption
+                  .getOrElse(SpectroscopyModesMatrix.empty),
+                routingInfo.focused,
+                model.rootModel.zoom(RootModel.searchingTarget),
+                model.rootModel.zoom(RootModel.expandedIds.andThen(ExpandedIds.obsListGroupIds)),
+                programSummaries.get.proposalIsSubmitted || model.userIsReadonlyCoi
+              )
+            .toOption
+      .orEmpty
```
cquiroz, line 124: I guess this is the case we can't parse the url, would it present an empty page then?

cquiroz (approved):
Looks great. LGTM with some comments inline

explore/src/main/scala/explore/Routing.scala
rpiaggio: It's the case where a `RoutingInfo` cannot be extracted from the `page`.

Up until now, all `Page`s produced a `RoutingInfo`, but now there are `Page`s that don't.

However, at the moment, we are only calling these methods with `Page`s from which `RoutingInfo` can always be extracted, and the `orEmpty` should never be hit. This is not future-proof tough.

We have some duplicate logic in the routing, we should revise things at some point and have better logic that doesn't require those `orEmpty`s, but it's not trivial.

PR #4390, opened 2024-12-16
Forbid submitting a proposal with targets outside the allowed windows

cquiroz:


rpiaggio (approved):
LGTM

toddburnside:
I might be missing it, but it seems like we should exclude inactive observations from this check. 

Also, shouldn't this case result in an observation error in the workflow? Couldn't we just forbid submitting a proposal if there are any errors? That would catch instruments and other problems as well without us needing to check for each one.
image

PR #4393, opened 2024-12-17
Correctly estimate row heights in sequence table

rpiaggio:
This avoids the flash when scrolling.


toddburnside (approved):
LGTM

rpiaggio:
> Isn't this common with observe? Should it be fixed in lucuma-ui?

Observe uses a lot of common code but renders the sequence a bit different: it has a `CurrentRow` for special rendering of the currently loaded row, and it renders visit headers.

However, I just realized that the heights are determined in lucuma-ui, so maybe we should have constants there that observe and explore can use. I'll merge and then refactor.

cquiroz (approved):
LGTM

PR #4396, opened 2024-12-17
Don't show executing step in visit, show all nextAtom steps

rpiaggio:
The currently executing step is not shown anymore as part of the visit, but instead as it comes in nextAtom. This avoids showing dataset info until it completes.

Also, a bug is fixed where some steps were hidden from nextAtom in cases where it contained more than 2 steps and it was currently executing the 2nd or later step.


rpiaggio:
Fixes https://app.shortcut.com/lucuma/story/4319/explore-shows-file-names-before-execution-is-complete

cquiroz (approved):
LGTM

PR #4399, opened 2024-12-18
Use a ref for TargetName input instead of direct dom lookup

toddburnside:
While working on an intractable story, I saw this comment and thought "hey, here's something I can atually fix!"

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/SearchForm.scala
+++ b/explore/src/main/scala/explore/targeteditor/SearchForm.scala
@@ -30,7 +30,6 @@ import lucuma.ui.reusability.given
 import lucuma.ui.syntax.all.given
 import org.scalajs.dom
 
-import scalajs.js.timers
 import scalajs.js.JSConverters.*
 
 case class SearchForm(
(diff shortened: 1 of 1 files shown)


explore/src/main/scala/explore/targeteditor/SearchForm.scala
```diff
@@ -55,22 +54,19 @@ object SearchForm:
...
-                Callback(timers.setTimeout(0)(node.asInstanceOf[dom.HTMLInputElement].select()))
+      .useRefToVdom[dom.HTMLButtonElement]
+      .useRefToVdom[dom.HTMLInputElement]
+      .useEffectWithDepsBy((props, _, _, _, _, _, inputRef) => (props.id, inputRef))(
+        (props, _, _, _, _, _, _) =>
+          // Auto select name field on new targets
+          (_, inputRef) =>
+            inputRef.get
+              .flatMap(
+                _.foldMap(i => Callback(timers.setTimeout(50)(i.select())))
```
toddburnside, line 65: With a timeout of zero, this solution had about the same reliability as the previous one. That is to say, the input always got selected but sometimes the cursor was at the end of the field rather than selecting all of the text. With the 50ms delay it has worked every time I tried it.

cquiroz: Can this be done as `Callback.delayMs(50)`?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/targeteditor/SearchForm.scala
toddburnside: I'm not sure. `delayMs` returns an `AsyncCallback` which seemed harder to deal with. I'd have to turn it back into a `Callback`.

PR #4400, opened 2024-12-19
Add a word counter for abstract and indicate if the abstract is too long

cquiroz:
image

### DIFF
--- a/common/src/main/webapp/sass/explore.scss
+++ b/common/src/main/webapp/sass/explore.scss
@@ -2014,6 +2014,11 @@ svg.fa-triangle-exclamation.explore-error-icon {
 // ------
 // Proposals and Partner Splits
 // ------
+
+.explore-abstract-too-long {
+  color: var(--error-background-color);
+}
+
 .partner-splits-grid {
   --gap: 0.5em;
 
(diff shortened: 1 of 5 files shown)


common/src/main/webapp/sass/explore.scss
```diff
@@ -2014,6 +2014,11 @@ svg.fa-triangle-exclamation.explore-error-icon {
 // ------
 // Proposals and Partner Splits
 // ------
+
+.explore-abstract-too-long {
+  color: var(--error-background-color);
```
rpiaggio, line 2019: There's a recently added `--negative-row-color` that you may want to use here instead of using a background color for the text. (There's also a `--negative-row-background`)

rpiaggio (approved):
LGTM, minor comment

common/src/main/webapp/sass/explore.scss
cquiroz: I tried but couldn't find the new variable even after a rebase

rpiaggio: > I tried but couldn't find the new variable even after a rebase

It's defined in lucuma-ui since it's shared with observe.

PR #4404, opened 2024-12-19
Avoid calling average PA with duration 0

rpiaggio:
Fixes https://app.shortcut.com/lucuma/story/4380/explore-crash-when-viewing-executed-sequence

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -289,7 +289,10 @@ object ObsTabTiles:
               PAProperties(props.obsId, guideStarSelection, agsState, props.posAngleConstraint)
 
             val averagePA: Option[AveragePABasis] =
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -289,7 +289,10 @@ object ObsTabTiles:
               PAProperties(props.obsId, guideStarSelection, agsState, props.posAngleConstraint)
 
             val averagePA: Option[AveragePABasis] =
-              (basicConfiguration.map(_.siteFor), asterismAsNel, obsDuration)
+              (basicConfiguration.map(_.siteFor),
+               asterismAsNel,
+               obsDuration.filter(_ > TimeSpan.Zero)
```
cquiroz, line 294: so the calculation crashes if time is zero? Don't we have a `NonNegDuration` somewhere?

cquiroz: I see

cquiroz (approved):
LGTM

PR #4407, opened 2024-12-20
Handle target cloning on target replacement better.

toddburnside:
[sc-4277]
The cloning of the target when editing a shared target was causing the issues. These became apparent now that odb-dev is slower. This fixes that by delaying the cloning of the target until necessary. Also, in the current code, the target name was changed, and THEN the target was replaced - so it was 2 steps. This eliminates the first step.

Readonly state of the target editor and other obs tiles was improved when an observation is being added/duplicated or a target is being cloned.

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/SearchForm.scala
+++ b/explore/src/main/scala/explore/targeteditor/SearchForm.scala
@@ -30,14 +30,17 @@ import lucuma.ui.reusability.given
 import lucuma.ui.syntax.all.given
 import org.scalajs.dom
 
+import scala.concurrent.duration.*
+
 import scalajs.js.JSConverters.*
 
 case class SearchForm(
-  id:         Target.Id,
-  targetName: View[NonEmptyString],
-  targetSet:  Target.Sidereal => Callback,
-  searching:  View[Set[Target.Id]],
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/targeteditor/SearchForm.scala
```diff
@@ -87,29 +109,37 @@ object SearchForm:
...
         FormInputTextView(
           id = "search".refined,
-          value = term.withOnMod(props.targetName.set),
+          value = term.withOnMod(nes =>
+            // We submit to the singleEffect with a delay. If the user hit Enter or they
+            // click on the button before leaving the input field, this will get cancelled
+            // so that the name doesn't get updated and, more importantly, no target cloning
+            // occurs, which messed everything up.
+            singleEffect
+              .submit(IO.sleep(200.milliseconds) >> props.targetName.set(nes).toAsync)
```
toddburnside, line 140: The Enter key scenario worked with much smaller sleeps, but it needed to be increased to handle the button clicking. The UI still feels fine.

cquiroz (approved):
LGTM

rpiaggio (approved):
LGTM, glad this worked

PR #4419, opened 2024-12-22
Allow accessing menu when no program selected

rpiaggio:
This PR makes the program selection non modal when no program is defined.

This allows users to access the menu, and therefore change roles, logout, redeem invitations, etc. without the need to select a program. Consequently, the logout button is removed from the program selection popup.

The popup is still modal when accessed from "Manage programs" from inside a program. BTW, "Manage programs" is hidden when no program is defined.


cquiroz:
Great, I've been thinking about this for a whlie as there are scenarios where you don't want/need to select a program

cquiroz (approved):
LGTM

PR #4420, opened 2024-12-23
Better identification of new programs

rpiaggio:
When creating a new program, its name will automatically become editable. Also, upon submission of the name, it will be automatically selected.

Kapture 2024-12-22 at 21 03 02


cquiroz:
Very nice, does it work if you just press enter? I mean going to the program?

rpiaggio:
> Very nice, does it work if you just press enter? I mean going to the program?

Yes

PR #4423, opened 2024-12-23
Internalize japgolly's IndexedDB library

rpiaggio:
This way we don't depend on the `webapp-utils` library.

### DIFF
--- a/workers/src/main/scala/japgolly/webapputil/binary/BinaryData.scala
+++ b/workers/src/main/scala/japgolly/webapputil/binary/BinaryData.scala
@@ -1,0 +1,242 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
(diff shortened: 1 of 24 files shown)


workers/src/main/scala/japgolly/webapputil/binary/BinaryData.scala
```diff
@@ -0,0 +1,242 @@
...
+package japgolly.webapputil.binary
+
+import japgolly.webapputil.general.ErrorMsg
+import java.io.OutputStream
+import java.lang.{StringBuilder => JStringBuilder}
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+import java.util.{Arrays, Base64}
+import scala.collection.immutable.ArraySeq
+import cats.Eq
```
cquiroz, line 13: Cool, you ported it to cats too

rpiaggio: Just the `Eq` and the `NonEmpty*`.

cquiroz (approved):
LGTM

PR #4449, opened 2024-12-30
Improve user experience in the ConfigurationRequestEditorPopup

toddburnside:
This enables the `submit` an`clear` buttons without needing to unfocus the input. Also migrates to the new hook syntax.


cquiroz (approved):
LGTM

PR #4454, opened 2025-01-02
Allow ProgramUser without a real user

toddburnside:
Because invitations are now made for a program user, the invitations table is gone.
image


explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -79,156 +76,132 @@ object ProposalEditor:
...
-            undoCtx,
-            UpdateProposalInput(
-              programId = props.programId.assign,
-              SET = ProposalPropertiesInput()
+  private val Component = ScalaFnComponent[ProposalEditor](props =>
+    for {
+      ctx             <- useContext(AppContext.ctx)
+      abstractCounter <-
+        useState(props.proposal.get.abstrakt.map(_.value).foldMap(_.wordCount))
+      _               <- useEffectWithDeps(props.proposal.get.abstrakt.map(_.value)) { // _ =>
```
cquiroz, line 84: Commented param

model/shared/src/main/scala/explore/model/User.scala
```diff
@@ -0,0 +1,35 @@
...
+import cats.derived.*
+import cats.syntax.all.*
+import io.circe.Decoder
+import lucuma.core.model.OrcidId
+import lucuma.core.model.UserProfile
+import lucuma.sso.client.codec.userProfile.given
+
+// Case class for the 'pi' and `users` in the Program. These must be
+// actual users, not service users.
+case class User(
```
cquiroz, line 16: Do we have something similar in core or sso already?

cquiroz (approved):
Great work, thanks for fixing this

model/shared/src/main/scala/explore/model/User.scala
toddburnside: No. Core has a trait for the various users, but not what we need here.

PR #4457, opened 2025-01-03
Remove AddProgramUserButton for support users

toddburnside:
Andy says the support users will be managed directly via the API or via the mythical `Admin UI` without the use of invitations, so he wanted the buttons for adding users removed for support users. I also made the table readonly and made a few minor changes to the users table.


cquiroz (approved):
LGTM

PR #4468, opened 2025-01-07
Implement AsyncAction, use Toast.showDuring for pasting observations

rpiaggio:
This PR introduces `AsyncAction`. It is very similar to `Action` but allows to run an effect to obtain the value to set. This attempts to be useful for operations where we don't know the whole data to add to the model until we have a response from the server. For example, we don't know the new ids of observations or targets when adding or cloning them.
 
`AsyncAction` allows us to define an effect to run initially and its result will determine the definition of the `getter`/ `setter`/`onSet`/`onRestore` regularly used in `Action`. This way, we can encapsulate the whole logic for insert and clone operations in an `AsyncAction` instead of it being scattered between components and `Action`s. Note that I only migrated one such action so far, the rest are pending.

Also, we now have a `effect.showToastDuring(...)` which will show an uncloseable toast before running the `effect`, replacing it with a regular toast when the operation completes or fails. Again, this is only called once for the moment, there are other instances where this can be used. We probably want to use it everywhere we end up using an `AsyncAction`.

Finally, this PR also fixes https://app.shortcut.com/lucuma/story/4421/copying-observations-onto-targets-yields-wrong-name-fixed-by-reload.

Kapture 2025-01-07 at 19 51 15

### DIFF
--- a/explore/src/main/scala/explore/tabs/TargetTabContents.scala
+++ b/explore/src/main/scala/explore/tabs/TargetTabContents.scala
@@ -6,7 +6,6 @@ package explore.tabs
 import cats.Order.given
 import cats.effect.IO
 import cats.syntax.all.*
-import clue.FetchClient
 import crystal.*
 import crystal.react.*
 import crystal.react.hooks.*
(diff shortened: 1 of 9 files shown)


rpiaggio:
[sc-4421]

explore/src/main/scala/explore/tabs/TargetTabContents.scala
```diff
@@ -223,11 +194,12 @@ object TargetTabContents extends TwoPanels:
...
 
                 IO.whenA(obsAndTargets.nonEmpty): // Apply the obs to selected targets on the tree
-                  applyObs(
-                    obsAndTargets,
-                    props.programSummaries,
-                    props.expandedIds
-                  ).withToast(s"Pasting obs ${copiedObsIdSet.idSet.toList.mkString(", ")}")
+                  ObservationPasteIntoAsterismAction(obsAndTargets, props.expandedIds.async.mod)(
+                    props.programSummaries
+                  ).withToastDuring(
```
cquiroz, line 199: Very nice

cquiroz (approved):
LGTM

PR #4472, opened 2025-01-08
Requery workflows on CfP change

rpiaggio:
Addresses https://app.shortcut.com/lucuma/story/4412/update-warnings-without-page-reload

### DIFF
--- a/explore/src/main/scala/explore/cache/ProgramCacheController.scala
+++ b/explore/src/main/scala/explore/cache/ProgramCacheController.scala
@@ -250,16 +250,24 @@ object ProgramCacheController
           def updateObservationsWorkflows(
             whereObservation: WhereObservation
           )(using StreamingClient[IO, ObservationDB]): IO[ProgramSummaries => ProgramSummaries] =
-            ProgramSummaryQueriesGQL
-              .ObservationsWorkflowQuery[IO]
-              .query(whereObservation)
-              .map:
-                _.observations.matches
-                  .map: m =>
-                    ProgramSummaries.observations
-                      .modify:
-                        _.updatedWith(m.id)(_.map(Observation.workflow.replace(m.workflow)))
-                  .combineAll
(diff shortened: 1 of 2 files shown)


cquiroz (approved):
LGTM

explore/src/main/scala/explore/cache/ProgramCacheController.scala
```diff
@@ -245,36 +247,46 @@ object ProgramCacheController
...
+
+          // Changing the proposal's CfP can change the validations for observations,
+          // eg: coordinates bounds, so we requery them all.
+          // https://app.shortcut.com/lucuma/story/4412/update-warnings-without-page-reload
+          val allObservationsValidationsUpdate: Pipe[
+            IO,
+            ProgramQueriesGQL.ProgramEditDetailsSubscription.Data,
+            ProgramSummaries => ProgramSummaries
+          ] =
+            _.map(_.programEdit.value.proposal.flatMap(_.callId)).changes.void
```
toddburnside, line 272: Nice!

toddburnside (approved):
LGTM

toddburnside:
I just realized there could be a problem for a large number of observations if it exceeds the limit of 1000 observations. I haven't really looked, but could `drain` be reused here?

We `shouldn't` have a timeout issue. Since this is only happening in response to a change to the CfP, the workflow shouldn't have to do any of the time consuming steps like build the digest.

rpiaggio:
> I just realized there could be a problem for a large number of observations if it exceeds the limit of 1000 observations. I haven't really looked, but could `drain` be reused here?

Good catch, we should `drain`, yes. I'll hold merging until I implement that.

rpiaggio:
> Good catch, we should `drain`, yes. I'll hold merging until I implement that.

Done

toddburnside (approved):
Looks great!

PR #4473, opened 2025-01-09
Gracefully handle errors in subscriptions.

toddburnside:
If this seems like a good approach we could apply it to other subscriptions as well. Maybe pop a toast if there is no data so the user knows they might be out of date and could reload?

For subscriptions where we regularly expect data + errors, like observations, we could only log in the case of no data. That would clean up the console. Maybe we want to only log for no data for all subscriptions? If there is data, we'll assume we're OK?

I don't expect this situation to happen often (no data). If the ODB is sending a subscription event, it seems like there should be data. But, it was happening for awhile until a bug got fixed in the ODB. Just suddenly going to a big screen of spinning planets is not very user friendly behavior. 😄

### DIFF
--- a/explore/src/main/scala/explore/cache/ProgramCacheController.scala
+++ b/explore/src/main/scala/explore/cache/ProgramCacheController.scala
@@ -9,6 +9,8 @@ import cats.syntax.all.*
 import clue.ErrorPolicy
 import clue.StreamingClient
 import clue.data.syntax.*
+import clue.model.GraphQLErrors
+import clue.model.GraphQLResponse
 import crystal.Pot
 import crystal.syntax.*
 import explore.DefaultErrorPolicy
@@ -299,6 +301,15 @@ object ProgramCacheController
               updateObservationExecution(data.observationEdit.observationId) <* queryProgramTimes
           )
 
+          def logGraphQLErrors(errors: GraphQLErrors): IO[Unit] =
+            val msg = errors.map(_.message).toList.mkString(": ")
+            Logger[IO].error(s"Error in subscription: $msg")
+
+          def logAndIgnoreErrors[D]: Pipe[IO, GraphQLResponse[D], D] =
+            _.evalTap(_.errors.foldMap(logGraphQLErrors))
+              .map(_.data)
+              .flattenOption
+
           extension (data: ProgramQueriesGQL.ConfigurationRequestSubscription.Data)
             private def obsIdList: List[Observation.Id] =
               data.configurationRequestEdit.configurationRequest.toList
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/cache/ProgramCacheController.scala
```diff
@@ -354,12 +365,16 @@ object ProgramCacheController
           val updateConfigurationRequests
             : Resource[IO, Stream[IO, ProgramSummaries => ProgramSummaries]] =
             ProgramQueriesGQL.ConfigurationRequestSubscription
-              .subscribe[IO](ConfigurationRequestEditInput(props.programId.assign))
+              .subscribe[IO](ConfigurationRequestEditInput(props.programId.assign))(using
+                summon,
```
rpiaggio, line 369: Can the  `summon` be omitted?

explore/src/main/scala/explore/cache/ProgramCacheController.scala
```diff
@@ -299,6 +301,15 @@ object ProgramCacheController
               updateObservationExecution(data.observationEdit.observationId) <* queryProgramTimes
           )
 
+          def logGraphQLErrors(errors: GraphQLErrors): IO[Unit] =
+            val msg = errors.map(_.message).toList.mkString(": ")
+            Logger[IO].error(s"Error in subscription: $msg")
+
+          def logAndIgnoreErrors[D]: Pipe[IO, GraphQLResponse[D], D] =
```
rpiaggio, line 308: Nice

rpiaggio (approved):
LGTM

I agree to make this pattern more widespread.

Maybe we can use a method or extension to encapsulate the behavior of passing `ReturnAlways` (or just using that as given) and passing the stream through the logging pipe?

explore/src/main/scala/explore/cache/ProgramCacheController.scala
toddburnside: I'll try again, but I don't think so. I think you need something in every position of the using parameter list. The existing subscriptions specifying a error policy are doing this. But, I guess I could maybe just do a `given` prior to the subscription,

toddburnside:
> Maybe we can use a method or extension to encapsulate the behavior of passing `ReturnAlways` (or just using that as given) and passing the stream through the logging pipe?

I'll look and see if I can come up with something more ergonomic.

explore/src/main/scala/explore/cache/ProgramCacheController.scala
rpiaggio: I was under the impression that Scala 3 was now smart enough to type-match givens in case some were not declared. If that's not the case, just leave the `summon`.

toddburnside: I thought so true, so I tried that earlier. It may `supposed to be` that smart, but apparently it is not.

toddburnside:
> Maybe we can use a method or extension to encapsulate the behavior of passing `ReturnAlways` (or just using that as given) and passing the stream through the logging pipe?

I wasn't able to find a nice solution to this. The main issue was that the `subscribe`  methods are concrete methods on the GraphQLOperation objects that take as parameters the parameters of the specific `Variable` case classes. As an alternative, I made a PR in clue that provides this functionality. https://github.com/gemini-hlsw/clue/pull/654

I will push another commit that shows how that could be used in explore. It, of course, won't compile at this time.

toddburnside:
Closed in favor of #4477

PR #4476, opened 2025-01-13
Keep sequence tile at bottom of obs tab

toddburnside:
I had to give the sequence tile it's own tile controller in order to accomplish this.

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -561,7 +561,6 @@ object ObsTabTiles:
                 constraintsTile.some,
                 schedulingWindowsTile.some,
                 configurationTile.some,
-                sequenceTile.some,
                 itcTile.some
               ).flattenOption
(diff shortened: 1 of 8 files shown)


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -570,12 +569,24 @@ object ObsTabTiles:
...
+              TileController(
+                props.vault.userId,
+                props.resize.width.getOrElse(0),
+                defaultLayout,
+                layout,
+                tiles,
+                section,
+                props.backButton.some
+              ),
+              TileController(
```
cquiroz, line 582: So you can't drag it to the other controller right?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
toddburnside: That is correct.

PR #4477, opened 2025-01-14
Clue error rework

rpiaggio:


toddburnside (approved):
LGTM. Thanks for doing this. I like the explicit acknowledgement of error handling.

toddburnside:
I think this line in .scalafix.conf needs to be updated:

> GraphQLGen.scalaJSReactReuse=false

rpiaggio:
> I think this line in .scalafix.conf needs to be updated:
> 
> > GraphQLGen.scalaJSReactReuse=false

Good catch!

rpiaggio:
It seems we need to recompile sbt-lucuma

PR #4478, opened 2025-01-15
Focus newly created target and allow navigation to a target via URL

toddburnside:
Simple fix but it took me a long time to find it....

### DIFF
--- a/explore/src/main/scala/explore/targets/TargetSummaryTile.scala
+++ b/explore/src/main/scala/explore/targets/TargetSummaryTile.scala
@@ -183,7 +183,6 @@ object TargetSummaryTile:
                     ),
                     ^.onClick ==> (e =>
                       e.preventDefaultCB >> e.stopPropagationCB >>
-                        Callback.log("onClick") >>
                         props.focusTargetId(cell.value.some)
                     )
                   )(
--- a/explore/src/main/scala/explore/targets/TargetSummaryTile.scala
+++ b/explore/src/main/scala/explore/targets/TargetSummaryTile.scala
@@ -91,6 +91,7 @@ object TargetSummaryTile:
           readonly,
           tileState.zoom(TileState.filesToImport),
           tileState.zoom(TileState.columnVisibility),
+          focusTargetId,
           tileState.get.toggleAllRowsSelected
         )
     )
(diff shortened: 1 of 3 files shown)


explore/src/main/scala/explore/targets/TargetSummaryTile.scala
```diff
@@ -261,8 +263,16 @@ object TargetSummaryTile:
...
-              onRowSelectionChange =
-                stateInViewHandler(rowSelection.mod(_) >> props.focusTargetId(none))
+              onRowSelectionChange = stateInViewHandler(
+                rowSelection
+                  .withOnMod: rs =>
+                    // We'll only unfocus if something is selected. Otherwise this is
+                    // called on initial load and prevents direct navigation to a url for
+                    // a target, and also doesn't allow focusing of a newly created target
+                    // while an observation is selected. See https://app.shortcut.com/lucuma/story/4425/select-newly-created-target
+                    props.focusTargetId(none).unless_(rs.value.isEmpty)
```
toddburnside, line 273: This is the actual fix.

explore/src/main/scala/explore/targets/TargetSummaryTile.scala
```diff
@@ -365,7 +376,7 @@ object TargetSummaryTile:
                       size = Button.Size.Small,
                       icon = Icons.SquareXMark,
                       label = "None",
-                      onClick = toggleAllRowsSelected(false)
+                      onClick = props.focusTargetId(none) >> toggleAllRowsSelected(false)
```
toddburnside, line 379: This was needed to allow unselecting all rows after the above fix.

explore/src/main/scala/explore/targets/TargetSummaryTile.scala
```diff
@@ -182,6 +183,7 @@ object TargetSummaryTile:
                     ),
                     ^.onClick ==> (e =>
                       e.preventDefaultCB >> e.stopPropagationCB >>
+                        Callback.log("onClick") >>
```
cquiroz, line 186: debugging?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/targets/TargetSummaryTile.scala
toddburnside: Good catch.

PR #4505, opened 2025-01-22
Unify program name/description and proposal title/abstract.

toddburnside:
This will need Shane's ODB PR and a schemas update in order to compile.

In a separate PR I will add the ability to edit the program description on the overview tab for non-science programs.  I will probably convert more components to the new hooks style at that time, but this seemed enough to review in one bite. This will make everything work and unblock Shane's PR.

This also makes significant updates to the programs popup. Some changes were needed in order to make programs undeleteable or unrenameable under various conditions, so I talked to Andy and also added some more columns to make it more useful until the mythical `Browse` is available.

### DIFF
--- a/explore/src/main/scala/explore/common/ProposalQueries.scala
+++ b/explore/src/main/scala/explore/common/ProposalQueries.scala
@@ -114,9 +114,7 @@ trait ProposalQueries:
     def toInput: ProposalPropertiesInput =
       ProposalPropertiesInput(
         callId = proposal.callId.orUnassign,
-        title = proposal.title.orUnassign,
         category = proposal.category.orUnassign,
-        `abstract` = proposal.abstrakt.orUnassign,
         `type` = proposal.proposalType.map(_.toInput).orUnassign
       )
 
--- a/explore/src/main/scala/explore/programs/ProgramsPopup.scala
+++ b/explore/src/main/scala/explore/programs/ProgramsPopup.scala
@@ -27,6 +27,7 @@ import explore.undo.UndoStacks
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.model.Program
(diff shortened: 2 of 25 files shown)


explore/src/main/scala/explore/programs/ProgramsPopup.scala
```diff
@@ -124,7 +127,7 @@ object ProgramsPopup:
       val programInfoViewListOpt: Option[List[View[ProgramInfo]]] =
         programInfosViewOpt.map:
           _.toListOfViews
-            .map(_._2.withOnMod(pi => newProgramId.set(none) >> doSelectProgram(pi.id)))
```
toddburnside, line 127: Unrelated bug fix. Whenever any program was edited on the dialog, new or old, including the name or existence status, the program was immediately selected and the popup went away. Among other oddities, this would select deleted programs. I moved the `select program` bit to where the name is edited, and only select the program when the name is saved if the program was just created.

explore/src/main/scala/explore/common/ProposalQueries.scala
```diff
@@ -114,9 +114,7 @@ trait ProposalQueries:
     def toInput: ProposalPropertiesInput =
       ProposalPropertiesInput(
         callId = proposal.callId.orUnassign,
-        title = proposal.title.orUnassign,
         category = proposal.category.orUnassign,
-        `abstract` = proposal.abstrakt.orUnassign,
```
cquiroz, line 119: no abstract?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/common/ProposalQueries.scala
toddburnside: The program description is now used as the abstract. That way non-science programs, which don't have proposals, can have a description field, too.

PR #4524, opened 2025-01-27
Add overview tile for edit program description for non-science programs.

toddburnside:
This adds a `Description` tile to the Overview tab for non-Science programs. It also displays the observation attachments tile for non-Science programs.


cquiroz (approved):
LGTM

PR #4525, opened 2025-01-28
Allow readonly CoIs to edit their own information

toddburnside:


explore/src/main/scala/explore/proposal/ProposalTabContents.scala
```diff
@@ -91,14 +90,14 @@ object ProposalTabContents:
 
   private val component = ScalaFnComponent[Props]: props =>
     for {
-      ctx      <- useContext(AppContext.ctx)
-      readonly <-
-        useMemo((props.programDetails.get.proposalStatus, props.userIsReadonlyCoi)):
-          (status, roCoi) =>
-            status === ProposalStatus.Submitted || status === ProposalStatus.Accepted || roCoi
+      ctx <- useContext(AppContext.ctx)
```
cquiroz, line 93: I guess you could do just `useContext(AppContext.ctx).map: ctx =>`

cquiroz (approved):
LGTM

explore/src/main/scala/explore/proposal/ProposalTabContents.scala
toddburnside: True. I'm not sure that is clearer, though.

PR #4526, opened 2025-01-28
Allow for changing between readonly and full access for CoIs.

toddburnside:
Also, Andy told me that the default for a CoI should be Readonly. So, now all CoIs are initially added as Readonly, but can be changed to Full Access via the dropdown in the program users table.


explore/src/main/scala/explore/users/ProgramUsersTable.scala
```diff
@@ -395,7 +398,27 @@ object ProgramUsersTable:
       ).sortableBy(_.get.toString),
       column(Column.OrcidId, _.get.user.flatMap(_.orcidId).foldMap(_.value)).sortable,
       // TODO: Make editable between COI and Readonly COI, if user is not this one
-      column(Column.Role, _.get.role.shortName).sortable,
+      // column(Column.Role, _.get.role.shortName).sortable,
```
cquiroz, line 401: Commented line

cquiroz (approved):
LGTM

PR #4529, opened 2025-01-30
Changes to proposal submission

toddburnside:
This PR:

- Uses observation validations with a code of `CallForProposalsError` to determine if the user can submit the proposal rather than calculating the coordinates out of range error locally. 
- Adds an error message to the proposals tab for the above condition.
- Fixes an issue where submission errors were not displayed. Although the error message was set, if the proposal status was updated, the components seem to reload so the error message was lost. The status is only set locally (in the view) if there was data in the response.


cquiroz (approved):
LGTM

PR #4543, opened 2025-02-03
Decode ExposureTimeMode correctly

toddburnside:
The fact that everything is Option hid the fact that the decoder was not working.

This makes a temporary fix until such time as Explore supports editing ExposureTimeMode. This only works for `signalToNoise`, not `timeAndCount`.


hugo-vrijswijk (approved):
LGTM. Does everything being Option mean it quietly fails? Id expect that it would only fallback to None on a missing value, not an incorrect one

PR #4545, opened 2025-02-06
Prevent proposal submission with undefined observations

toddburnside:
Updates the UI to match the new limit set in the API.

### DIFF
--- a/explore/src/main/scala/explore/proposal/ProposalSubmissionBar.scala
+++ b/explore/src/main/scala/explore/proposal/ProposalSubmissionBar.scala
@@ -38,12 +38,12 @@ import queries.common.ProposalQueriesGQL.SetProposalStatus
 import scala.concurrent.duration.*
 
 case class ProposalSubmissionBar(
-  programId:      Program.Id,
-  proposalStatus: View[ProposalStatus],
-  deadline:       Option[Timestamp],
-  callId:         Option[CallForProposals.Id],
-  canSubmit:      Boolean,
-  hasObsErrors:   Boolean
+  programId:                Program.Id,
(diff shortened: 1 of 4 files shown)


explore/src/main/scala/explore/proposal/ProposalSubmissionBar.scala
```diff
@@ -79,13 +79,15 @@ object ProposalSubmissionBar:
...
-                                  "One or more observations has an error. See Overview tab for details.".some
-                                )
-                              else errorMessage.set(none) // Reset error message on CfP change
+      _                <-
+        useLayoutEffectWithDeps(
+          (props.proposalStatus.get, props.callId, props.hasUndefinedObservations)
+        ): (ps, _, he) =>
+          if (he && ps === ProposalStatus.NotSubmitted)
+            errorMessage.set(
+              "Proposal cannot be submitted with undefined observations. Define them or mark them as inactive.".some
```
swalker2m, line 88: Is there any use in pointing them to the Overview tab?  That is, should we put that bit back?  Not suggesting it -- I just don't know the answer.

swalker2m (approved):
Looks great.  :+1:

explore/src/main/scala/explore/proposal/ProposalSubmissionBar.scala
toddburnside: I'm going to link to this message in the shortcut story and ask Andy that very question. :) I guess I could do it before merging.

toddburnside:
I am merging this because I have additional work to do in the same error. I will change the message in a new PR if it needs changed.

PR #4546, opened 2025-02-06
Add help icon to investigator tile

toddburnside:


explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -151,13 +152,14 @@ object ProposalEditor:
...
-                  ExploreStyles.AddProgramUserButton,
+            <.div(
+              ExploreStyles.AddProgramUserButton,
+              Option
+                .unless[VdomNode](props.proposalOrUserIsReadonly):
                   AddReadonlyCoiButton(props.programId, props.users)
-                )
-              .orEmpty
+                .orEmpty,
+              HelpIcon("proposal/main/investigators.md".refined)
```
swalker2m, line 161: So, the idea is that even if the `proposalOrUserIsReadonly` you swapped this around to show the div anyway because you're putting in the help icon?

swalker2m (approved):
Brilliant. :+1:

explore/src/main/scala/explore/proposal/ProposalEditor.scala
toddburnside: Correct. I figure the help information might be useful, anyway, even if they can't add users.

PR #4551, opened 2025-02-10
Fix missing Program tab

toddburnside:
The Program tab was not displaying because it required the semester from the call for proposals. We looked up the CfP using the id in the proposal of the program details, but since we were only downloading the `open` CfPs, if the proposal specified a CfP whose deadline had passed we wouldn't find it and wouldn't display the tab.

I initially just downloaded all the CfP (didn't filter for open in the query) and filtered for `open` locally. Then I realized it was silly to download an ever expanding list of CfPs that we didn't really need, and just downloaded the whole CfP for the proposal instead of just the CfP Id.

The first commit just switches to monadic style hooks and cleans up some commented out code. The second is where the real changes are.

### DIFF
--- a/explore/src/main/scala/explore/programs/ProgramDetailsTile.scala
+++ b/explore/src/main/scala/explore/programs/ProgramDetailsTile.scala
@@ -20,102 +20,101 @@ import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.model.Program
 import lucuma.core.syntax.display.*
 import lucuma.core.util.DateInterval
+import lucuma.react.common.ReactFnComponent
 import lucuma.react.common.ReactFnProps
 import lucuma.refined.*
 import lucuma.ui.primereact.CheckboxView
 import lucuma.ui.primereact.FormInfo
 import lucuma.ui.primereact.given
 
-import java.time.LocalDate
-
 case class ProgramDetailsTile(
   programId:         Program.Id,
   programDetails:    View[ProgramDetails],
   programTimes:      Pot[ProgramTimes],
   userIsReadonlyCoi: Boolean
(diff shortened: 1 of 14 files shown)


explore/src/main/scala/explore/programs/ProgramDetailsTile.scala
```diff
@@ -41,19 +42,25 @@ object ProgramDetailsTile:
...
+      val thesis: Boolean                       = details.allUsers.exists(_.thesis.exists(_ === true))
+      val users: View[List[ProgramUser]]        = props.programDetails.zoom(ProgramDetails.allUsers)
+      val newDataNotificationView               =
         props.programDetails
           .zoom(ProgramDetails.shouldNotify)
           .withOnMod(b => ProgramQueries.updateGoaShouldNotify[IO](props.programId, b).runAsync)
+      val cfpActivePeriod: Option[DateInterval] = details.proposal.flatMap(_.call).map(_.active)
+
+      // We `should` always have a call for proposal if we get here, but...
+      def dateOrMissing(o: Option[LocalDate], label: String) =
```
rpiaggio, line 55: Maybe we can just omit the element if the date is missing?

Also, could we add explicit types?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/programs/ProgramDetailsTile.scala
toddburnside: We could do that, but I thought it might be easier to debug in the future if we ever run into this again. It `should` never happen since the program tab is only for ACCEPTED proposals, which should have a CfP. But, omitting them would be cleaner, I guess.

PR #4554, opened 2025-02-11
Show CfP parameters on proposal details editor

toddburnside:
image

### DIFF
--- a/explore/src/main/scala/explore/proposal/ProposalDetailsTile.scala
+++ b/explore/src/main/scala/explore/proposal/ProposalDetailsTile.scala
@@ -25,6 +25,7 @@ import explore.components.ui.*
 import explore.components.undo.UndoButtons
 import explore.model.AppContext
 import explore.model.CallForProposal
+import explore.model.Constants
 import explore.model.ExploreModelValidators
 import explore.model.Hours
 import explore.model.PartnerSplit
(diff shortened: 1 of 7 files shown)


explore/src/main/scala/explore/proposal/ProposalDetailsTile.scala
```diff
@@ -317,6 +319,14 @@ object ProposalDetailsBody:
       )
     }
 
+    extension (limits: SiteCoordinatesLimits)
+      def format: String =
+        val raStart  = limits.raStart.toHourAngle.toDoubleHours
+        val raEnd    = limits.raEnd.toHourAngle.toDoubleHours
+        val decStart = limits.decStart.toAngle.toSignedDoubleDegrees
+        val decEnd   = limits.decEnd.toAngle.toSignedDoubleDegrees
+        f"$raStart%.1f ≤ RA ≤ $raEnd%.1fh    $decStart° ≤ Dec ≤ $decEnd°"
```
rpiaggio, line 328: Will this produce decimals with any number of digits in `Dec`?

rpiaggio (approved):
LGTM

explore/src/main/scala/explore/proposal/ProposalDetailsTile.scala
rpiaggio: Can this value be specified in sexagesimal anywhere (eg the API)?  If so, we might want to truncate somewhere. If "40°20'" is specified, it will be shown as 40.333333.......

toddburnside: Hmmm. Yeah. The could specify in `hms` format. I'll force a maximum of one decimal, I guess.

toddburnside: I chatted with Andy. He said we'll never need decimals in the Dec limits, so I'll just not print any, ever.

PR #4557, opened 2025-02-13
Filter calibrations out of constraints summary table

toddburnside:


explore/src/main/scala/explore/constraints/ConstraintsSummaryTile.scala
```diff
@@ -103,177 +110,195 @@ object ConstraintsSummaryTile:
...
+                                     obsId.toString
+                                   )
+                                 )
+                                 .mkReactFragment(", ")
+                             )
+                           )
+                           .setEnableSorting(false)
+                       )
+          // Memo rows
+          rows  <- useMemo((props.constraintList, props.calibrationObservations)): (cl, calibs) =>
```
toddburnside, line 266: The only real change is the calculation of the rows.

explore/src/main/scala/explore/constraints/ConstraintsSummaryTile.scala
```diff
@@ -103,177 +110,195 @@ object ConstraintsSummaryTile:
...
-        .useMemoBy((_, _) => ()): // Cols never changes, but needs access to props
-          (props, ctx) =>
-            _ =>
-              def column[V](id: ColumnId, accessor: ConstraintGroup => V)
-                : ColumnDef.Single.NoMeta[ConstraintGroup, V] =
-                ColDef(id, accessor, ColumnNames(id))
+      ScalaFnComponent[Body]: props =>
+        for {
+          ctx   <- useContext(AppContext.ctx)
+          cols  <- useMemo(()): // Cols never changes, but needs access to props
```
rpiaggio, line 132: Unrelated to the PR: in other places where we have stable columns definitions, we have moved them to a val with `.reuseAlways`, which declutters the hook section.

explore/src/main/scala/explore/constraints/ConstraintsSummaryTile.scala
```diff
@@ -103,177 +110,195 @@ object ConstraintsSummaryTile:
...
+                           )
+                           .setEnableSorting(false)
+                       )
+          // Memo rows
+          rows  <- useMemo((props.constraintList, props.calibrationObservations)): (cl, calibs) =>
+                     cl
+                       .map: (obsIdSet, cs) =>
+                         (obsIdSet -- calibs)
+                           .map: filtered =>
+                             ConstraintGroup(cs, filtered)
```
rpiaggio, line 271: Why not filter in the caller rather than passing the filtered observations list all the way up to here?

toddburnside: > filtered observations list

Do you mean the calibration observations list?

rpiaggio: > I wonder if there should just be a lazy val in programSummaries that does this?

Sounds good to me

explore/src/main/scala/explore/constraints/ConstraintsSummaryTile.scala
rpiaggio: Oh right, it needs the programId.

explore/src/main/scala/explore/constraints/ConstraintsSummaryTile.scala
toddburnside: This was a good call because it turns out that we don't want the calibration observations in any of the groupings that we calculate in the ProgramSummaries. We were filtering them out way down the chain and not always completely. Which messed up, for example, selections in some of the trees because the calibration obs ids were not being filtered out of the tree representation.

PR #4559, opened 2025-02-13
Add 5-second throttling to all subscriptions

rpiaggio:


toddburnside:
I'm not sure if this is quite what we want. It doesn't stop up from receiving the subscriptions and we already have the View mechanism in place for trying to avoid issues with updates from the server as well as the groupWithin. I'm afraid those could behave poorly in combination with this. I think it would be better to just throttle the queries that are triggered by the subscriptions - such as observation, program and group times updates.

rpiaggio:
I completely agree with your comment, I'll have another go.

rpiaggio:
Closing in favor of https://github.com/gemini-hlsw/explore/pull/4560

PR #4560, opened 2025-02-13
Throttle general updates

rpiaggio:
Applies throttling to queries involving general updates (program times, etc).


toddburnside (approved):
LGTM. Thanks for do this!

PR #4565, opened 2025-02-17
Fix some glitches when sequence is executing

rpiaggio:
- Requery visits/sequence also when a step starts. This way, when the sequence moves to science, we can adjust accordingly (mainly hiding acquisition).
- Show the current step from the visits and not from the sequence. Note that this mechanism is failing because we are currently unable to match the `generatedId` of the `StepRecord` with the `id` of the `Step`, since observe seems to be recording outdated `generatedId`s.
- Consider the last step of an executing atom as the current step, even if the step is not executing. The sequence could be paused and therefore the last step could be completed. But if the atom is not complete, we want to show the step from the visits, not from the future sequence.

Last commit is just monadic hook migration.


toddburnside (approved):
LGTM

PR #4567, opened 2025-02-18
Add Data Users table to program tab

toddburnside:
Includes a refactoring of the ProgramUsersTable to handle the new requirements.


explore/src/main/scala/explore/proposal/ProposalEditor.scala
toddburnside, line 1: The changes in this file are mostly refactoring to use monadic hooks.

explore/src/main/scala/explore/users/ProgramUsersTable.scala
```diff
@@ -61,43 +61,100 @@ import queries.common.InvitationQueriesGQL.RevokeInvitationMutation
...
 
 case class ProgramUsersTable(
-  userVault:          Option[UserVault],
-  users:              View[List[ProgramUser]],
-  filterRoles:        NonEmptySet[ProgramUserRole],
-  proposalIsReadonly: Boolean,
-  userIsReadonlyCoi:  Boolean,
-  hiddenColumns:      Set[ProgramUsersTable.Column] = Set.empty
+  users: View[List[ProgramUser]],
+  mode:  ProgramUsersTable.Mode
```
toddburnside, line 65: There are a lot of different rules with respect to editing in the different modes.Passing in the Modes enum which contains the information necessary for each mode simplified the call sites so they did not have to know as much about the table or pass in data that isn't necessary for that mode.

rpiaggio (approved):
LGTM

PR #4568, opened 2025-02-18
Avoid duplicate row when executing, without resorting to generatedId

rpiaggio:
If the last step of the last atom of the last visit is Ongoing, we remove the first step from the nextAtom.

Also, refactor sequence component into smaller files, by instrument.


toddburnside (approved):
LGTM

PR #4569, opened 2025-02-19
ObsBadge updates

toddburnside:
- [sc-4777] Displays the `ObservationReference` index in the ObsBadge when it is available, instead of the Observation.Id
- [sc-4774] Displays the filter and central wavelength in the configuration summary of the ObsBadge. Also changes the string in the configuration selection dropdown in the configuration tile title to match.
- [sc-3897] For the Data Users table, adds the count of users who have data access in the tile title. Also adds the role column which I omitted by accident.


toddburnside:
After discussions with Andy, I changed the name of the `Data Users` table to `Data Sharing` and sorted the data for it by `Role` in Enumerated order so that support users are the bottom by default.

rpiaggio (approved):
LGTM

PR #4570, opened 2025-02-19
Show obs reference index on obs summary table when available

toddburnside:
An extension of using the observation reference index on the Obs Badge, this shows it instead of the observation id in the observation summary table when it is available. 

The first commit is just switching to monadic hooks. The "real" changes are in the second commit.


rpiaggio (approved):
LGTM

Do you think it would be easy and worthwhile to move the columns definitions to a separate file `ObsSummaryColumns` ? I'm thinking about maintainability.

toddburnside:
> Do you think it would be easy and worthwhile to move the columns definitions to a separate file `ObsSummaryColumns` ? I'm thinking about maintainability.

It should be easy and might make maintenance a bit easier. I'll do that.

explore/src/main/scala/explore/observationtree/ObsSummaryTile.scala
```diff
@@ -191,305 +192,307 @@ object ObsSummaryTile:
...
+          _.value.map:
+            _.orSpinner(_.map(HoursMinutesAbbreviation.format).orEmpty)
+        .sortableBy(_.sortableTimeSpan)
+        // TODO: PriorityColumnId
+        // TODO: ChargedTimeColumnId
+      )
+    end columns
+
+    private val component = ScalaFnComponent[Body]: props =>
+      for {
```
hugo-vrijswijk, line 117: I still can't get over how much cleaner this looks 😅

toddburnside: Yeah. The old style is hard to look at now. 😄

PR #4571, opened 2025-02-19
Remove brackets for atoms in sequence, show thicker border between atoms instead

rpiaggio:
Brackets spanning multiple rows had some issues:
- If the bracket spanned too many rows, it could get removed by the virtualizer when scrolling since it was mounted on the first row of the atom.
- They wouldn't span executed steps.

Drawing a thicker border between atoms, just like observe does, solves both problems.


toddburnside (approved):
LGTM

cquiroz:
I wonder if we could differentiate atoms using a slightly different border or background. I'll see how it looks now first

PR #4572, opened 2025-02-20
Fix data sharing counter

toddburnside:
The counter was not including support users. When I added support users to the table, I neglected to update the counter.


rpiaggio (approved):
LGTM

PR #4573, opened 2025-02-20
Add tooltip to ObsBadge when observation reference index is shown

toddburnside:


rpiaggio (approved):
LGTM

explore/src/main/scala/explore/observationtree/ObsBadge.scala
```diff
@@ -86,7 +86,10 @@ object ObsBadge:
       val obs    = props.obs
       val layout = props.layout
 
-      val identifierStr = obs.reference.fold(obs.id.show)(_.observationIndex.toString)
+      val identifier: VdomNode = obs.reference.fold(s"[${obs.id.show}]": VdomNode): ref =>
+        Tooltip.Fragment(content = s"${ref.label} (${obs.id})")(
+          <.span(s"[${ref.observationIndex.toString}]")
```
rpiaggio, line 91: IIRC, there's a `withTooltip` extension.

toddburnside: There is. But, Tooltip.Fragment was used elsewhere in this file, and I didn't have to find the extension. Also, doesn't `withTooltip` use the floatingui tooltip? It looks slightly different than the prime ones. Or, is there a `withTooltip` for prime tooltips also?

toddburnside: Ah. I searched in Explore and found a `withTooltip` usage that uses prime. I can switch to that.

PR #4577, opened 2025-02-24
Select new observation on creation

toddburnside:
This PR changes both observation creation and cloning to use `AsyncAction`s instead of the `Action` that required calls to the API outside of the action. There is also a little cleanup on toasts


cquiroz (approved):
LGTM

rpiaggio (approved):
LGTM

PR #4580, opened 2025-02-27
Show warnings for group violations

toddburnside:
Recursively calculates warnings for observation groups according to SC-3328. It displays a warning icon in the Group "Badge" like this 
image
And messages at the bottom of the group editor like this
image

### DIFF
--- a/model/shared/src/main/scala/explore/model/ProgramSummaries.scala
+++ b/model/shared/src/main/scala/explore/model/ProgramSummaries.scala
@@ -10,6 +10,7 @@ import cats.derived.*
 import cats.implicits.*
 import crystal.Pot
 import eu.timepit.refined.cats.given
+import explore.model.enums.GroupWarning
 import explore.model.syntax.all.*
 import lucuma.core.enums.ObservationValidationCode
 import lucuma.core.enums.ObservationWorkflowState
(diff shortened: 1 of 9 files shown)


model/shared/src/main/scala/explore/model/ProgramSummaries.scala
```diff
@@ -212,6 +213,57 @@ case class ProgramSummaries(
       .toMap
   end groupsChildren
 
+  private def allObservationsForGroup(groupId: Group.Id): List[Observation] =
```
toddburnside, line 216: I don't know if we need to worry about stack safety here or not. I'm not sure if the ODB enforces a limit to the depth of the group tree, but:
1. It seems unlikely someone will build a tree hundreds of groups deep, and
2. If they do, it will probably bring the scheduler to it's knees and we'll need to implement a limit.

cquiroz: also would kill time calculation

model/shared/src/main/scala/explore/model/ProgramSummaries.scala
```diff
@@ -267,3 +319,10 @@ object ProgramSummaries:
       GroupTimeRangeMap(groupTimeRangePots),
       configRequests.toSortedMap(_.id)
     )
+    println(
```
cquiroz, line 322: Leftover debugging?

cquiroz (approved):
LGTM

model/shared/src/main/scala/explore/model/ProgramSummaries.scala
toddburnside: Doh... 🤦

model/shared/src/main/scala/explore/model/ProgramSummaries.scala
toddburnside: I decided to limit the recursion level just in case. It will only affect warning generation.

PR #4590, opened 2025-03-02
Update sequence on dataset change

rpiaggio:


explore/src/main/scala/explore/config/sequence/SequenceTileHelper.scala
```diff
@@ -108,6 +108,14 @@ trait SequenceTileHelper:
                 List(StepStage.StartStep, StepStage.EndStep)
                   .contains_(data.executionEventAdded.value.stepStage)
               .evalMap(_ => (refreshSequence >> refreshVisits).to[IO])
+      _                                       <-
+        useEffectStreamResourceOnMount: // Subscribe to dataset edits
```
cquiroz, line 112: Any benefit if you merge both on mount effects?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/config/sequence/SequenceTileHelper.scala
rpiaggio: Not really. On the contrary, if we keep them separate one will go on in case one of them errors.

PR #4608, opened 2025-03-05
Custom Sed selector - Part 1

rpiaggio:
This was a bit challenging since the SED selector used to directly show and edit the model we have in memory (and send mutations to the server).

However, we cannot do this anymore because when selecting a custom SED we may have an inconsistent state in the UI while the user selects and/or uploads a new SED. Therefore, state was created in the component: we now edit that state and forward it to the model and ODB when it's well defined.

This PR just presents the user with a dropdown showing all the custom SED attachments currently in the program. There's no functionality yet to upload new ones (that can be done via the REST API for testing purposes). I'll implement upload in a next PR, but this one's complex enough as it is, so I decided to split them.

image

### DIFF
--- a/common/src/main/scala/explore/model/enums/SEDType.scala
+++ b/common/src/main/scala/explore/model/enums/SEDType.scala
@@ -22,7 +22,6 @@ import lucuma.core.math.Wavelength
 import lucuma.core.math.dimensional.*
 import lucuma.core.math.dimensional.Units.*
 import lucuma.core.math.units.*
-import lucuma.core.model.Attachment
 import lucuma.core.model.SpectralDefinition
 import lucuma.core.model.UnnormalizedSED
 import lucuma.core.util.*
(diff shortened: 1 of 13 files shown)


common/src/main/scala/explore/model/enums/SEDType.scala
```diff
@@ -49,59 +56,68 @@ sealed abstract class SedTypeEnum[T](
       case BandNormalized(_, bs) => BandNormalized(sed.some, bs)
       case EmissionLines(_, _)   => BandNormalized(sed.some, SortedMap.empty)
 
-  protected sealed abstract class BandNormalizedSED(name: String, sed: UnnormalizedSED)
-      extends SedType[T](name, toBandNormalized(sed))
+  // protected sealed abstract class BandNormalizedSED(name: String) extends SedType[T](name)
```
toddburnside, line 59: Commented code

explore/src/main/scala/explore/targeteditor/spectralDefinition/SpectralDefinitionEditorBuilder.scala
```diff
@@ -76,12 +79,49 @@ private abstract class SpectralDefinitionEditorBuilder[
...
+          case (Some(sed @ SedType.Immediate(_, convert)), _) if !currentSedType.contains_(sed) =>
+            props.spectralDefinition
+              .view(props.toInput)
+              .mod(convert)
+          case (None, _) if currentSedType.isDefined                                            =>
+            props.spectralDefinition
+              .view(props.toInput)
+              .mod(SpectralDefinition.unnormalizedSED.replace(none))
+          case _                                                                                =>
+            Callback.empty
```
toddburnside, line 118: 😮‍💨

toddburnside (approved):
LGTM. Not complicated at all!

PR #4614, opened 2025-03-08
Custom SED - Part 2 - File upload

rpiaggio:
Allow uploading of Custom SED from target editor.

Generalized the file upload button so that code is shared with finder chart upload.

### DIFF
--- a/workers/src/main/scala/workers/ItcServer.scala
+++ b/workers/src/main/scala/workers/ItcServer.scala
@@ -36,7 +36,7 @@ object ItcServer extends WorkerServer[IO, ItcMessage.Request] with ItcPicklers {
 
   private def createClient[F[_]: Async]: Client[F] =
     FetchClientBuilder[F]
-      .withRequestTimeout(5.seconds)
+      .withRequestTimeout(10.seconds)
       .create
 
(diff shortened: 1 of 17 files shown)


workers/src/main/scala/workers/ItcServer.scala
```diff
@@ -36,7 +36,7 @@ object ItcServer extends WorkerServer[IO, ItcMessage.Request] with ItcPicklers {
 
   private def createClient[F[_]: Async]: Client[F] =
     FetchClientBuilder[F]
-      .withRequestTimeout(5.seconds)
+      .withRequestTimeout(10.seconds)
```
cquiroz, line 39: is this because of the time to upload a sed?

cquiroz (approved):
LGTM

workers/src/main/scala/workers/ItcServer.scala
rpiaggio: I haven't profiled how much it is because of the time ITC needs to pull the custom SED out of S3 and how much is due to ITC maybe needing more time to process a custom SED, if that's the case; but I was getting timeouts with 5 seconds.

PR #4621, opened 2025-03-11
Fix duration used for parallactic angle calculation

toddburnside:
When an observation duration was not explicitly set, the program time was used instead of the full remaining time.


cquiroz (approved):
LGTM

PR #4624, opened 2025-03-11
The UI is not showing the flipped angle if AGS decides to do a flip

cquiroz:


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -340,26 +346,28 @@ object ObsTabTiles:
               .mapN: (site, asterism, duration) =>
                 props.posAngleConstraint match
                   case PosAngleConstraint.AverageParallactic =>
-                    // See also `anglesToTestAt` in AladinCell.scala.
-                    averageParallacticAngle(
+                    val avpa = averageParallacticAngle(
```
toddburnside, line 349: Ah. So for `AverageParallacticAngle`, the flip of the angle is allowed. Good detective work.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
cquiroz: yeah it was all well documented of course

PR #4651, opened 2025-03-17
Promotion scripts

cquiroz:


toddburnside (approved):
Very cool. This will be very helpful!
LGTM

PR #4657, opened 2025-03-19
Add Program Notes tile to Program tab

toddburnside:
The notes support markdown and the markdown editor uses a tabbed view, similar to on github or shortcut. 
image
The notes are sorted by title so that the user has a way to hack a specific order for notes if they desire.

The observer notes tile has also been updated to use the tabbed markdown editor.
[sc-2419]

### DIFF
--- a/explore/src/main/scala/explore/notes/NotesTile.scala
+++ b/explore/src/main/scala/explore/notes/NotesTile.scala
@@ -24,31 +24,9 @@ object NotesTile:
       ObsTabTileIds.NotesId.id,
       s"Note for Observer",
       bodyClass = ExploreStyles.NotesTile
-    )(_ => NotesTileBody(notes), (_, tileSize) => NotesTileTitle(notes, tileSize))
-
-case class NotesTileBody(
-  notes: View[Option[NonEmptyString]]
-) extends ReactFnProps(NotesTileBody.component)
-
-object NotesTileBody:
-  private type Props = NotesTileBody
-
-  val themeAttr = VdomAttr("data-theme")
-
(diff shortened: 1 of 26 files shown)


explore/src/main/scala/explore/notes/NotesTile.scala
```diff
@@ -6,96 +6,41 @@ package explore.tabs
...
-    )(NotesTileBody(obsId, notes, _), NotesTileTitle(obsId, notes, _, _))
-
-case class NotesTileState(notes: String, editing: Editing)
-
-object NotesTileState:
-  val notes: Lens[NotesTileState, String]    = Focus[NotesTileState](_.notes)
-  val editing: Lens[NotesTileState, Editing] = Focus[NotesTileState](_.editing)
+    )(_ => NotesTileBody(notes), (_, tileSize) => NotesTileTitle(notes, tileSize))
 
 case class NotesTileBody(
```
cquiroz, line 29: Is it worth a component for this? could it be just inlined?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/notes/NotesTile.scala
toddburnside: I guess it could be inlined. But ObsTabTiles is a pretty busy file so I think this is cleaner.

toddburnside: Oh, I guess you mean inlining `within` this file. Yeah, that would be simpler now that I look at the file. I'll do that.

PR #4659, opened 2025-03-20
Always show finder charts tile for non-science programs

toddburnside:


explore/src/main/scala/explore/notes/NotesTile.scala
```diff
@@ -18,15 +18,17 @@ import lucuma.react.common.*
...
-      bodyClass = ExploreStyles.NotesTile
-    )(_ => MarkdownEditor(notes),
-      (_, tileSize) =>
-        Option.unless(tileSize === TileSizeState.Minimized)(
-          HelpIcon("observer-notes.md".refined)
-        )
-    )
+  def apply(notes: View[Option[NonEmptyString]], hidden: Boolean): Tile[Unit] =
+    if (hidden)
+      Tile(ObsTabTileIds.NotesId.id, "", hidden = true)(_ => EmptyVdom)
```
toddburnside, line 23: We have been creating these dummy tiles, but looking at the code it doesn't seem any different that specifying `hidden = true` for the regular tile. https://github.com/gemini-hlsw/explore/blob/d463d48e08903f29650a27f1a00bfc3f8bbf4bab/common/src/main/scala/explore/components/Tile.scala#L132

cquiroz: so we could just pass the flag? did you try it?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/notes/NotesTile.scala
toddburnside: I'm trying that now

toddburnside: Seems to work just as well. There are still cases where the tile layouts get a bit messed up when the tiles return, but it's the same as before.

PR #4660, opened 2025-03-20
Add note count to the program notes tile title.

toddburnside:


cquiroz (approved):
LGTM, would you special case "no notes"?

toddburnside:
> LGTM, would you special case "no notes"?

We're doing this same thing for the data sharing tile and we don't special case it there. But, maybe we should in both cases.

PR #4661, opened 2025-03-21
Add controls to select time x count mode

cquiroz:
They are disabled until we get the ITC working

### DIFF
--- a/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
@@ -56,6 +56,7 @@ import lucuma.react.primereact.Button
 import lucuma.react.primereact.PrimeStyles
 import lucuma.refined.*
 import lucuma.schemas.ObservationDB.Types.*
+import lucuma.schemas.ObservationDB.Types.WavelengthInput
 import lucuma.schemas.model.CentralWavelength
 import lucuma.schemas.model.ObservingMode
 import lucuma.schemas.odb.input.*
@@ -69,6 +70,7 @@ import lucuma.ui.utils.given
 import monocle.Lens
 import mouse.boolean.*
 import org.typelevel.log4cats.Logger
+import explore.model.enums.ExposureTimeModeType
 
 import scalajs.js
 import scalajs.js.JSConverters.*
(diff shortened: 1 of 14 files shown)


explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -365,79 +367,49 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
...
-        { case (reqsWavelength, rows, _, _, _, _) =>
-          findMatrixData(props.observingMode.get, reqsWavelength, rows)
-        }
-      }
-      .useStateView(ConfigEditState.View)
-      .render { (props, ctx, _, modeData, editState) =>
+    ScalaFnComponent[Props]: props =>
+      for
+        ctx       <- useContext(AppContext.ctx)
+        emv       <- useStateView(
```
toddburnside, line 373: I'm a little confused. Does this mean the exposure mode drop down changes this view but doesn't really change anything?

explore/src/main/scala/explore/config/SpectroscopyConfigurationPanel.scala
```diff
@@ -92,7 +105,16 @@ object SpectroscopyConfigurationPanel extends ConfigurationFormats:
...
+          FormEnumDropdownView(
+            id = "exposureMode".refined,
+            value = emv,
+            label = ReactFragment(
+              "Exposure Mode",
+              HelpIcon("configuration/exposure-mode.md".refined)
+            ),
+            disabled = true
+          ),
+          snMode.asView.map(SignalToNoiseAt(_, p.readonly, p.units, p.calibrationRole)),
```
toddburnside, line 117: Here we actually switch between the 2 modes but only show editing controls for S/N?

toddburnside (approved):
LGTM
Do you think it's possible to extract a component that can be used in both places?

explore/src/main/scala/explore/config/SpectroscopyConfigurationPanel.scala
cquiroz: Yes because this is the initial PR, the selector is disable, I have to add the time and count editor and make it common to both editors

toddburnside:
Ah, never mind. I see the drop downs are disabled. I should have read the description better.

PR #4667, opened 2025-03-21
Add controls to set time and count on the

cquiroz:
Next step, this PR adds a time x count editor:
image

It is still disabled though


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -302,382 +302,329 @@ private object SpectroscopyModesTable:
     )
 
   private val component =
-    ScalaFnComponent
```
cquiroz, line 305: The only change here is to switch to monadic style

toddburnside (approved):
LGTM

PR #4668, opened 2025-03-21
Display custom SEDs in attachment table

toddburnside:
The custom SED attachments are now shown in the Attachments tile on the Overview tab. Because of this, the tile is always shown. However, observation attachments are only displayed/uploaded for science programs when the proposal has been accepted.

There are 2 commits.
The first commit is some initial renaming, switching to monadic hooks and some other refactoring. I intended to make this 2 separate commits to make it easier to follow, but apparently I created a branch but didn't commit anything for the first part of it. 😦 

The second commit has the actual changes related to the shortcut story.


cquiroz (approved):
LGTM

PR #4704, opened 2025-03-31
Changes to Average Parallactic Angle calculation and obs duration

toddburnside:
For sc-5061 (https://app.shortcut.com/lucuma/story/5061/average-parallactic-angle-calculation-should-only-include-science-portion) this changes the average parallactic angle calculation to ignore the setup time and calculate with science time (obs time + setup time) and science duration (obs duration - setup time).

For [sc-5193] (https://app.shortcut.com/lucuma/story/5193/do-not-automatically-recalculate-average-parallactic-angles) a number of changes were made so that Explore will not recalculate the average parallactic angle and the available guide stars while an observation is being executed.

On the API side, obs duration is now required and must be greater than the setup time and no greater than the estimated remaining time.

In explore:
1.  the `Now` button for setting the observation time now also sets the observation duration to the remaining estimated time.
2. Once an duration has been set, the button for unsetting the duration has been changed to one that will set it to the remaining time. If it is already set to the remaining time, the button is hidden.
3. The duration editor coerces the value to be in the interval between setup time + 1 second and the remaining time.

There were also some CSS changes to give the time/duration editor more space in the title bar. Prior to this, the add button, the obs time editor, and the column selector split the title bar evenly and the obs time editor could wrap on occassion.

### DIFF
--- a/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
+++ b/explore/src/main/scala/explore/tabs/ObsTabTiles.scala
@@ -319,7 +319,8 @@ object ObsTabTiles:
               obsEditAttachments(props.obsId, ids).runAsync
             }
 
-          val pendingTime = props.obsExecution.toOption.flatMap(_.fullTimeEstimate)
+          val pendingTime = props.obsExecution.toOption.flatMap(_.remainingObsTime)
+          val setupTime   = props.obsExecution.toOption.flatMap(_.fullSetupTime)
           val obsDuration =
             props.observation.get.observationDuration
               .orElse(pendingTime)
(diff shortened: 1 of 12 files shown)


explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -333,22 +334,28 @@ object ObsTabTiles:
...
-                      asterism.baseTracking,
-                      obsTimeOrNow,
-                      duration
-                    )
-                    flipIfNeeded(avpa)
-                      .map(AveragePABasis(obsTimeOrNow, duration, _))
-                  case _                                     => none
+            (basicConfiguration.map(_.siteFor), props.asterismAsNel, obsDuration, setupTime)
+              .flatMapN: (site, asterism, fullDuration, setupDuration) =>
+                // science duration is the obsDuration - setup time
```
cquiroz, line 339: and acq time?

model/shared/src/main/scala/explore/model/ExploreModelValidators.scala
```diff
@@ -54,15 +51,8 @@ object ExploreModelValidators:
       n => Try(n.shortName).toOption.orEmpty
     )
 
-  // TODO Move to lucuma core
```
cquiroz, line 57: Did you move these to core?

cquiroz (approved):
LGTM

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
toddburnside: Yes, the setup time includes the acquisition time.

explore/src/main/scala/explore/tabs/ObsTabTiles.scala
```diff
@@ -333,22 +334,28 @@ object ObsTabTiles:
...
+                fullDuration
+                  .subtract(setupDuration)
+                  .filter(_ > TimeSpan.Zero)
+                  .map: scienceDuration =>
+                    // scienceStartTime is the obsTimeOrNow + setup time
+                    val scienceStartTime =
+                      obsTimeOrNow.plusNanos(setupDuration.toMicroseconds * 1000)
```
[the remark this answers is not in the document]
toddburnside: No, TimeSpan has microsecond resolution, and you can add milliseconds or nanonseconds to a java duration, but not microseconds. 🤷‍♂️

model/shared/src/main/scala/explore/model/ExploreModelValidators.scala
toddburnside: Yes.

PR #4705, opened 2025-03-31
Don't allow decimal for exposure time on gmos

cquiroz:
I can foresee different instruments will have different rules, thus the format is instrument-dependent.


toddburnside (approved):
LGTM

PR #4722, opened 2025-04-04
Use NewBoolean new convenience methods

rpiaggio:
Simplifies NewBoolean definitions


cquiroz (approved):
LGTM

model/shared/src/main/scala/explore/model/enums/Visible.scala
toddburnside, line 1: I guess this isn't really an enum anymore. Maybe it should be moved to `model`?

toddburnside (approved):
LGTM

PR #4723, opened 2025-04-04
Invalid the cache and reload itc/sequence when custom seds are updated

toddburnside:


common/src/main/public/environments.conf.json
```diff
@@ -2,10 +2,10 @@
   {
     "hostName": "*",
     "environment": "DEVELOPMENT",
-    "odbURI": "wss://lucuma-postgres-odb-dev.herokuapp.com/ws",
-    "odbRestURI": "https://lucuma-postgres-odb-dev.herokuapp.com",
+    "odbURI": "ws://localhost:8082/ws",
```
cquiroz, line 5: We probably should revert these changes

cquiroz (approved):
LGTM

common/src/main/public/environments.conf.json
toddburnside: Ooops. Inadvertant file inclusion. :(

PR #4724, opened 2025-04-05
Fix formats for exp time and s/n

cquiroz:


rpiaggio (approved):
LGTM

explore/src/main/scala/explore/config/TimeAndCountEditor.scala
```diff
@@ -74,9 +74,9 @@ object TimeAndCountEditor extends ConfigurationFormats:
           changeAuditor = timeAuditor,
           disabled = props.readonly
         ).clearable(^.autoComplete.off),
-        FormLabel("signal-to-noise".refined)("Number of Exp."),
+        FormLabel("exposure-count".refined)("Number of Exp."),
         FormInputTextView(
-          id = "count".refined,
+          id = "exposure-count".refined,
```
toddburnside, line 79: Just a note that using the `label` parameter of `FormInputTextView` should be the same as explicitly putting a `FormLabel` before it and it will set the `htmlFor` for you.

PR #4730, opened 2025-04-07
Rework of exposure time mode editors 

cquiroz:
it improves how the editors keep their local state and how they react to remote changes.

### DIFF
--- a/explore/src/main/scala/explore/common/ScienceQueries.scala
+++ b/explore/src/main/scala/explore/common/ScienceQueries.scala
@@ -95,7 +95,7 @@ object ScienceQueries:
           _ <-
             SpectroscopyScienceRequirementsInput.exposureTimeMode := op.exposureTimeMode
               .map(_.toInput)
-              .orIgnore
+              .orUnassign
           _ <- SpectroscopyScienceRequirementsInput.wavelengthCoverage := op.wavelengthCoverage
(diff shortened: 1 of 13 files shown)


explore/src/main/scala/explore/common/ScienceQueries.scala
```diff
@@ -117,7 +93,9 @@ object ScienceQueries:
                  .orUnassign
           _ <- SpectroscopyScienceRequirementsInput.resolution         := op.resolution.orUnassign
           _ <-
-            SpectroscopyScienceRequirementsInput.exposureTimeMode := op.exposureTimeMode.toInput.orUnassign
+            SpectroscopyScienceRequirementsInput.exposureTimeMode := op.exposureTimeMode
+              .map(_.toInput)
+              .orIgnore
```
toddburnside, line 98: You switched from `orUnassign` to `orIgnore`. I just want to make sure that is intentional and not an autocomplete error. 😆

toddburnside (approved):
LGTM

explore/src/main/scala/explore/common/ScienceQueries.scala
cquiroz: Good question. I think `orUnassign` though I can't see how the UI would set it to null
I'll change it nonetheless

PR #4740, opened 2025-04-09
Migrate core model

rpiaggio:


rpiaggio:
Failing because of bundlemon, but everything else seems to pass

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
```diff
@@ -55,36 +54,35 @@ case class ConstraintsPanel(
 object ConstraintsPanel:
   private type Props = ConstraintsPanel
 
-  private enum ElevationRangeType(val label: String):
-    case AirMass   extends ElevationRangeType("Air Mass")
-    case HourAngle extends ElevationRangeType("Hour Angle")
+  private enum ElevationRangeType(val label: String, val tag: String) derives Enumerated:
```
toddburnside, line 57: Do we actually need the separate tag here, or is it just a convenience to be able to use `derives Enumerated`?

toddburnside (approved):
LGTM

explore/src/main/scala/explore/constraints/ConstraintsPanel.scala
rpiaggio: It's just to derive `Enumerated`. We could assign it to be the same as `label`, it doesn't matter much since this is all local. I just separated it to avoid spaces in the tag.

I wondered if we needed the `Enumerated`, but we have it in order to use `EnumDropdownView`.

PR #4744, opened 2025-04-11
Perishable type - mark obs and group times as stale instead of pending

toddburnside:
This PR introduces a `Perishable` type that represents a value that can be marked as stale. There is also a type alias `PerishablePot` for `Pot[Perishable[A]]` that provides helpful syntax and is used in the observation and group times maps. 

The initial impetus for this was the comment in [sc-5193] regarding the average parallactic angle going invalid while waiting for new execution times to arrive from the ODB. Now, instead of setting the execution values to pending while waiting, they are marked stale. The values are still displayed in the UI, but in most cases they are grayed out and a tooltip provided stating that it is waiting for new data. The APA remains based on the stale values until new data arrives. If there is an error with the execution times query, the PerishablePot is set to Error so that it can be displayed in the UI.

Some work that will be done in future PRs:
1. Improve the display of the errors. They are displayed now and it is better than previously, but is not always pretty.
2. Apply the same PerishablePot treatment to the program execution times.
3. Possibly improve the ergonomics of using PerishablePot.

### DIFF
--- a/model/shared/src/main/scala/explore/model/Perishable.scala
+++ b/model/shared/src/main/scala/explore/model/Perishable.scala
@@ -1,0 +1,21 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats.Eq
+
+// A value, A, that can be marked as stale or errored.
+opaque type Perishable[A] = (A, Boolean)
+
+object Perishable:
+  // it is fresh when created
+  def apply[A](a: A): Perishable[A] = (a, false)
+  given [A: Eq]: Eq[Perishable[A]]  = Eq.by(e => (e._1, e._2))
+
+  extension [A](entry: Perishable[A])
+    def value: A                              = entry._1
+    def isStale: Boolean                      = entry._2
+    def map[B](f:      A => B): Perishable[B] = (f(value), entry._2)
+    def mapValue[B](f: A => B): B             = f(value)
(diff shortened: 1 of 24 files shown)


model/shared/src/main/scala/explore/model/Perishable.scala
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.model
+
+import cats.Eq
+
+// A value, A, that can be marked as stale or errored.
+opaque type Perishable[A] = (A, Boolean)
```
cquiroz, line 9: so the bool is true/false for stale/errored? or false means either stale or errored?

cquiroz (approved):
Great idea. LGTM

model/shared/src/main/scala/explore/model/Perishable.scala
toddburnside: Yes, true is stale. But, that isn't visible outside the type so it doesn't really matter. Users of the class use `isStale` and `setStale`.

rpiaggio: Maybe we can use a `NewBoolean` ?

rpiaggio (approved):
Looks great, thank you for doing this :+1:

model/shared/src/main/scala/explore/model/Perishable.scala
rpiaggio: It's OK then since it's not visible from outside.

PR #4749, opened 2025-04-13
Port aladin container to mondaic style

cquiroz:


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 1.10.11
```
rpiaggio, line 1: why the dial back?

rpiaggio (approved):
LGTM

project/build.properties
cquiroz: Sorry I'll revert
For some reason this doesn't work for me, but only in explore

PR #4756, opened 2025-04-14
Fix transition from emission lines -> custom sed

rpiaggio:
Also, when choosing the "User Defined" option now, the brightness editor will disappear until an the attachment is chosen. This is because internally we can't assign a new SED value until we have an attachment and therefore the old value is still recorded.


cquiroz (approved):
LGTM

PR #4759, opened 2025-04-17
Allow failure for bundlemon action

hugo-vrijswijk:


cquiroz:
I added this to sbt lucuma https://github.com/gemini-hlsw/sbt-lucuma/pull/420

hugo-vrijswijk:
> I added this to sbt lucuma https://github.com/gemini-hlsw/sbt-lucuma/pull/420

The `params` add the `with` block, which is not the correct place for the continue-on-error

PR #4762, opened 2025-04-20
Remove warnings triggered on scala 3.7

cquiroz:
I started testing scala 3.7 but it produces lots of warnings

This PR fixes those warnings in anticipation

### DIFF
--- a/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
+++ b/explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
@@ -48,7 +48,6 @@ import lucuma.core.syntax.all.*
 import lucuma.core.util.Display
 import lucuma.core.util.Enumerated
 import lucuma.core.validation.*
-import lucuma.react.common.Css
 import lucuma.react.common.ReactFnProps
 import lucuma.react.fa.IconSize
 import lucuma.react.floatingui.syntax.*
@@ -225,7 +224,7 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
     reqsWavelength: Option[Wavelength],
     row:            SpectroscopyModeRow
   ): Option[ModeData] =
-    reqsWavelength.flatMap(cw =>
+    reqsWavelength.flatMap(_ =>
       (mode, row.instrument) match
         case (m: ObservingMode.GmosNorthLongSlit,
               InstrumentConfig.GmosNorthSpectroscopy(rGrating, rFpu, rFilter, _)
@@ -324,8 +323,8 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
     changeAuditor: ChangeAuditor,
     label:         TagMod,
     originalValue: A,
-    units:         Option[String] = None,
-    disabled:      Boolean = false
+    units:         Option[String],
+    disabled:      Boolean
   ) =
     val isCustom    = value.get =!= originalValue
     val customAddon =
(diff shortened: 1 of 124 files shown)


explore/src/main/scala/explore/config/AdvancedConfigurationPanel.scala
```diff
@@ -324,8 +323,8 @@ sealed abstract class AdvancedConfigurationPanelBuilder[
     changeAuditor: ChangeAuditor,
     label:         TagMod,
     originalValue: A,
-    units:         Option[String] = None,
-    disabled:      Boolean = false
+    units:         Option[String],
```
cquiroz, line 326: Not sure why, but giving a default to the parameters, produced a warning.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -285,15 +282,15 @@ private object SpectroscopyModesTable:
       column(TimeColumnId, _.totalItcTime)
         .withHeader(progressingCellHeader("Time"))
         .withCell: cell =>
-          cell.table.options.meta.map: meta =>
+          cell.table.options.meta.map: _ =>
```
cquiroz, line 285: We are not using meta here, Maybe it should be removed altogether?

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
```diff
@@ -313,7 +313,7 @@ object AsterismGroupObsList:
           val list        = obsIds.toList
           val div: TagMod = <.div(
             <.div(
-              list.toTagMod(id => <.div(id.show))
+              list.toTagMod(using id => <.div(id.show))
```
cquiroz, line 316: It is strange that these methods take a function as a implicit. I'll send a PR to scalajs-react it ha fix

explore/src/main/scala/explore/observationtree/ViewCommon.scala
```diff
@@ -101,13 +101,12 @@ trait ViewCommon {
     ExploreStyles.DraggingOver.when(isDragging)
 
   def undoableDeleteObs(
-    obsId:        Observation.Id,
-    observations: UndoSetter[ObservationList],
-    afterUndo:    Observation.Id => Callback,
-    afterDelete:  Callback
+    obsId:             Observation.Id,
+    observations:      UndoSetter[ObservationList],
+    @unused afterUndo: Observation.Id => Callback,
```
cquiroz, line 106: This is unused but it is defined in many places it is called, I think we are missing something here

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
rpiaggio: By "here", you mean this column, or the table?

explore/src/main/scala/explore/observationtree/AsterismGroupObsList.scala
rpiaggio: The idea of the implicit is that if you pass a string or a component for example, it will convert automatically to vdom; but you can override like we do here.

explore/src/main/scala/explore/observationtree/ViewCommon.scala
rpiaggio: We might have used this at some point and removed its usage later.

rpiaggio (approved):
LGTM, thank you. We should probably remove those unused parameters.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: that particular cell

PR #4764, opened 2025-04-21
Add a tooltip with the slit width in pixels

cquiroz:
This PR is bigger than expected because of a class name change. The core change is smaller

### DIFF
--- a/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
+++ b/explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
@@ -84,18 +84,18 @@ import scalajs.js
 
 // TODO: These constants should likely go to lucuma-core
 type ArcSecondPerPixel = ArcSecond / Pixels
-type PlateScale        = Quantity[BigDecimal, ArcSecondPerPixel]
+type PixelScale        = Quantity[BigDecimal, ArcSecondPerPixel]
 
-val GmosPlateScale: PlateScale = BigDecimal(0.0807).withUnit[ArcSecondPerPixel]
-val F2PlateScale: PlateScale   = BigDecimal(0.18).withUnit[ArcSecondPerPixel]
+val GmosPixelScale: PixelScale = BigDecimal(0.0807).withUnit[ArcSecondPerPixel]
+val F2PixelScale: PixelScale   = BigDecimal(0.18).withUnit[ArcSecondPerPixel]
 
 def gmosSlitWidthPixels(slitWidth: Angle, xBin: GmosXBinning): Quantity[BigDecimal, Pixels] =
   val widthArcSeconds = Angle.decimalArcseconds.get(slitWidth).withUnit[ArcSecond]
-  widthArcSeconds / (BigDecimal(xBin.count) * GmosPlateScale)
+  widthArcSeconds / (BigDecimal(xBin.count) * GmosPixelScale)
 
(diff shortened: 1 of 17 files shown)


explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -67,13 +73,30 @@ import lucuma.ui.table.*
...
 
 import java.text.DecimalFormat
 import scala.collection.decorators.*
 import scala.concurrent.duration.*
+import scala.language.implicitConversions
 
 import scalajs.js
 
+// TODO: These constants should likely go to lucuma-core
+type ArcSecondPerPixel = ArcSecond / Pixels
```
cquiroz, line 86: These are methods to calculate slit width in pixels

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -267,6 +290,33 @@ private object SpectroscopyModesTable:
         )
     )
 
+  private def slitWidthCell(config: ItcInstrumentConfig, slitWidth: ModeSlitSize): VdomNode =
+    val width = config match {
```
cquiroz, line 294: Show the width on a tooltip

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
```diff
@@ -67,13 +73,30 @@ import lucuma.ui.table.*
...
+import scala.language.implicitConversions
 
 import scalajs.js
 
+// TODO: These constants should likely go to lucuma-core
+type ArcSecondPerPixel = ArcSecond / Pixels
+type PixelScale        = Quantity[BigDecimal, ArcSecondPerPixel]
+
+val GmosPixelScale: PixelScale = BigDecimal(0.0807).withUnit[ArcSecondPerPixel]
+val F2PixelScale: PixelScale   = BigDecimal(0.18).withUnit[ArcSecondPerPixel]
```
rpiaggio, line 90: This looks like it should be part of the `Instrument` enum. Or maybe we can put the next 2 methods in extensions for each instrument? It seems the constants are not needed elsewhere.

workers/src/main/scala/workers/itc/ITCRequests.scala
```diff
@@ -42,7 +42,7 @@ object ITCRequests:
       ga.parTraverse(a => s.permit.use(_ => f(a)))
 
   // Wrapper method to match the call in ItcServer.scala
-  def queryItc[F[_]: Concurrent: Parallel: Logger](
+  def queryItc[F[_]: {Concurrent, Parallel, Logger as L}](
```
rpiaggio, line 45: Ouch. I admit I like the summoner pattern better rather than naming implicits.

rpiaggio (approved):
LGTM, with a couple of suggestions.

explore/src/main/scala/explore/config/SpectroscopyModesTable.scala
cquiroz: Yes, I'll move them to core

workers/src/main/scala/workers/itc/ITCRequests.scala
cquiroz: scalafmt doesn't like it either, I'll revert for now

PR #4766, opened 2025-04-22
Simplification

cquiroz:
We have a class `ObsModeSummary` which is almost the same as `ObservingMode` in schemas.
This PR will remove that class and use the one from schemas

I tried to improve a bit the code using name tuples but I couldn't make it work due to
https://github.com/scala/scala3/issues/22890


rpiaggio:
I'm not fully convinced for 3 reasons:

1) Usage of type checking in pattern matching. This is sugar for `isInstanceOf` and I think we should avoid it in favor of `unapply`. We discussed this and agreed to this temporal solution until we can use named tuples in Scala 3.7, but after a bit of thought, I think I'd rather use `unapply` as temporal solution. It's much more likely that it will bother us in the future enough to make the switch.
2) The logic to resolve the `ampReadMode` and `roi` is now duplicated (in the `Order` and `Display` instances) whereas before it was only once in the `fromObservingMode` converter. I'd rather not duplicate the logic, this makes maintenance more difficult and brittle. We could add `resolved*` methods to the `lucuma-schemas` class.
3) I don't think the `toInput` conversion is working the same way: the `ampReadMode` and `roi` are being lost now if they are not overridden. Eg: Configuration 1 has only default values for `ampReadMode` and `roi`; if it is applied to an observation that has other default values, they should be overridden with the ones from Configuration 1. Before they were, now they are not. Again, we could add a `toResolvedInput` or similar to the `lucuma-schemas`class.

In summary, we could add methods to the `lucuma-schemas` class to deal with the resolved values, but then it may become too crowded, more so when we add more instruments. This was the reasoning for creating a new class. Maybe `ObservingModeSummary` just needs a better comment explaining its reason of being. Sorry that I didn't have all this fresh in my memory to stop you in time when you told me you would attempt this, now that I see the code it came back to me.

cquiroz:
> I'm not fully convinced for 3 reasons:
> 
>     1. Usage of type checking in pattern matching. This is sugar for `isInstanceOf` and I think we should avoid it in favor of `unapply`. We discussed this and agreed to this temporal solution until we can use named tuples in Scala 3.7, but after a bit of thought, I think I'd rather use `unapply` as temporal solution. It's much more likely that it will bother us in the future enough to make the switch.
Our codebase is full of those, lucuma-odb uses this pattern a lot

> 
>     2. The logic to resolve the `ampReadMode` and `roi` is now duplicated (in the `Order` and `Display` instances) whereas before it was only once in the `fromObservingMode` converter. I'd rather not duplicate the logic, this makes maintenance more difficult and brittle. We could add `resolved*` methods to the `lucuma-schemas` class.
I'm doing than on my changes on lucuma-schemas. I agree thy should be in a single place
> 
>     3. I don't think the `toInput` conversion is working the same way: the `ampReadMode` and `roi` are being lost now if they are not overridden. Eg: Configuration 1 has only default values for `ampReadMode` and `roi`; if it is applied to an observation that has other default values, they should be overridden with the ones from Configuration 1. Before they were, now they are not. Again, we could add a `toResolvedInput` or similar to the `lucuma-schemas`class.
I'll review that

> 
> 
> In summary, we could add methods to the `lucuma-schemas` class to deal with the resolved values, but then it may become too crowded, more so when we add more instruments. This was the reasoning for creating a new class. Maybe `ObservingModeSummary` just needs a better comment explaining its reason of being. Sorry that I didn't have all this fresh in my memory to stop you in time when you told me you would attempt this, now that I see the code it came back to me.

PR #4774, opened 2025-04-23
Update explore itc to support f2 long slit

cquiroz:
image


explore/src/main/scala/explore/config/TimeAndCountEditor.scala
```diff
@@ -43,19 +43,14 @@ object TimeAndCountEditor extends ConfigurationFormats:
...
 
-      val timeFormat = props.instrument
+      val (timeFormat, timeAuditor) = props.instrument
         .map {
-          case Instrument.GmosSouth | Instrument.GmosNorth => durationS.optional
-          case _                                           => durationMs.optional
+          case Instrument.GmosSouth | Instrument.GmosNorth | Instrument.Flamingos2 =>
+            (durationS.optional, ChangeAuditor.int.optional)
+          case _                                                                   =>
+            (durationMs.optional, ChangeAuditor.posBigDecimal(2.refined).optional)
```
rpiaggio, line 51: I see that the `_` case is the same as the default. I think  we `collect` instead of `map` and avoid the `_` clause.

model/shared/src/main/scala/explore/model/Observation.scala
```diff
@@ -202,21 +203,22 @@ case class Observation(
...
           )
+      case _: ObservingMode.F2LongSlit => None
 
   def toInstrumentConfig(targets: TargetList): Option[ItcInstrumentConfig] =
-    (toModeOverride(targets), observingModeSummary)
+    (toModeOverride(targets), observingMode)
       .mapN:
         case (overrides @ InstrumentOverrides.GmosSpectroscopy(_, _, _),
-              ObservingModeSummary.GmosNorthLongSlit(grating, filter, fpu, _, _, _)
+              n: ObservingMode.GmosNorthLongSlit
```
rpiaggio, line 212: Can't wait for Scala 3.7 :D

rpiaggio (approved):
LGTM, with a suggestion

model/shared/src/main/scala/explore/model/Observation.scala
cquiroz: shorter is better

explore/src/main/scala/explore/config/TimeAndCountEditor.scala
cquiroz: good, I'll add it

cquiroz: hmm collect returns `Option[B]`. it makes it more complicated I think

PR #4780, opened 2025-04-27
Improve error messages

cquiroz:
The core error was resolved on the ITC server side
But we can also improve a bit thee error messages.


toddburnside (approved):
LGTM

PR #4781, opened 2025-04-28
Fix conversion of wavelength <-> wavelengthdelta

cquiroz:


model/shared/src/main/scala/explore/model/ExploreModelValidators.scala
```diff
@@ -141,14 +141,8 @@ object ExploreModelValidators:
...
-      Iso[Wavelength, WavelengthDelta](w =>
-        WavelengthDelta
-          .fromIntNanometers(w.nm.value.value.toInt)
-          .getOrElse(sys.error("Invalid Wavelength"))
-      )(wc =>
-        Wavelength
-          .fromIntNanometers(wc.nm.value.value.toInt)
-          .getOrElse(sys.error("Invalid Wavelength"))
+      Iso[Wavelength, WavelengthDelta](w => WavelengthDelta(w.pm.value))(wd =>
+        Wavelength(wd.pm.value)
```
toddburnside, line 145: Ah. I guess the `wavelengthNanoValidWedge` takes care of the `Nano` part so the conversion can be exact here?

toddburnside (approved):
LGTM

model/shared/src/main/scala/explore/model/ExploreModelValidators.scala
cquiroz: yes, the problem was that the Iso above wasn't really lawful, but it can be done so converting all to the minimal unit

PR #4782, opened 2025-04-28
Improve loading of the service worker

cquiroz:
If there is an initial error, e.g. the ODB schema changed but Explorer hasn't, we may not install the service worker, which could be very bad. 

This PR changes the order and it also updates the vite pwa plugin

### DIFF
--- a/explore/src/main/webapp/main.jsx
+++ b/explore/src/main/webapp/main.jsx
@@ -47,12 +47,14 @@ fetch('/environments.conf.json').then((response) => {
       // Set it globally so it can be read in Scala code.
       window.enumMetadataString = enumMetadataModule.enumMetadata;
 
-      Explore.runIOApp();
-
       // Setup the Service Worker, after Explore is started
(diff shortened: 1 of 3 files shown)


toddburnside (approved):
LGTM. How did you even think to change this?

cquiroz:
I was testing it locally, to check the lib update and just happened.
I did several tests, breaking a decoder which replicates the exact issue

explore/src/main/webapp/main.jsx
```diff
@@ -47,13 +47,15 @@ fetch('/environments.conf.json').then((response) => {
...
 
-      Explore.runIOApp();
-
       // Setup the Service Worker, after Explore is started
       if ('serviceWorker' in navigator && !/local.lucuma.xyz/.test(window.location)) {
         ExplorePWA.runServiceWorker();
       }
 
+      // IMOPORTANT: Starte explore after the PWA service worker
+      // Otherwiste, errors on load may swallow the service worker
```
rpiaggio, line 56: A couple of typos in the comment

PR #4800, opened 2025-05-02
F2 Longslit Configuration Panel

toddburnside:
As per requirements, staff users can edit the `Decker`
image
This requires clicking `Customize` and then `Advanced Customization`.

For non-staff, Decker shows up as a static field.
image

All the values in the leftmost column are editable by pressing the `Customize` button.

One slight oddity is that if the staff modifies the `Decker` value, non-staff will still have the option to `Revert Modifications`.  Since reverting modifications is a pre-requisite to be able to `Revert Configuration`, I'm not sure what, if anything, to do about it. We can start here and modify as we figure things out.

Since `AdvancedConfigurationPanel` is really specific to GMOS longslit, I will submit a separate PR renaming it and it's contents.

### DIFF
--- a/explore/src/main/scala/explore/components/CustomizableEnumSelect.scala
+++ b/explore/src/main/scala/explore/components/CustomizableEnumSelect.scala
@@ -6,7 +6,8 @@ package explore.components
 import cats.syntax.all.*
 import crystal.react.View
 import eu.timepit.refined.types.string.NonEmptyString
-import japgolly.scalajs.react.ScalaFnComponent
+import explore.model.Help
+import japgolly.scalajs.react.*
 import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.syntax.all.*
 import lucuma.core.util.Display
@@ -14,6 +15,7 @@ import lucuma.core.util.Enumerated
 import lucuma.react.common.ReactFnProps
 import lucuma.react.primereact.PrimeStyles
 import lucuma.ui.primereact.FormEnumDropdownView
+import lucuma.ui.primereact.FormLabel
 import lucuma.ui.primereact.LucumaPrimeStyles
 import lucuma.ui.primereact.given
 import lucuma.ui.syntax.all.given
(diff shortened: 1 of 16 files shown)


explore/src/main/scala/explore/components/CustomizableEnumSelect.scala
```diff
@@ -0,0 +1,64 @@
...
+import lucuma.ui.primereact.given
+import lucuma.ui.syntax.all.given
+
+/**
+ * An enum select component that allows the user to select an enum value from a list of options. But
+ * it has the concept of a default value that can be reverted to. If the current value is not the
+ * same as the default value, then an addon icon with a tooltip is displayed. Clicking on the icon
+ * reverts the value to the default value.
+ */
+final case class CustomizableEnumSelect[A: Enumerated: Display](
```
cquiroz, line 29: Will this go to lucuma-ui?

explore/src/main/scala/explore/config/package.scala
```diff
@@ -50,3 +62,95 @@ trait ConfigurationFormats:
         case WavelengthUnits.Nanometers  => wvDeltaNanoInput
 
 object ConfigurationFormats extends ConfigurationFormats
+
+case class ModeData private (
```
toddburnside, line 66: This was moved from AdvancedConfigurationPanel and modified a bit. There, it had `centralWavelength` and `interval` parameters. But, it seems the algorithm for finding the ModeData was changed to use the spectroscopy requirements wavelength instead of central wavelength, so those parameters were no longer valid nor used.

cquiroz:
What about `reads` and `readoutMode`?

explore/src/main/scala/explore/components/CustomizableEnumSelect.scala
toddburnside: It could go there, but to date it was just in AdvancedConfigurationPanel. I'm not sure how likely it is to be used elsewhere.

toddburnside:
> What about `reads` and `readoutMode`?

Per the shortcut story, Andy does not want to display those.

cquiroz (approved):
LGTM

PR #4805, opened 2025-05-03
Support showing visualization of the mode before assignig it

cquiroz:
From the mode selection we can make a basic configuration that can be used for viz before actually assigning a mode


model/shared/src/main/scala/explore/model/ConfigurationForVisualization.scala
```diff
@@ -0,0 +1,31 @@
...
+import explore.model.syntax.all.*
+import lucuma.core.math.Angle
+import lucuma.core.math.Offset
+import lucuma.core.model.PosAngleConstraint
+import lucuma.schemas.model.BasicConfiguration
+
+// Yet another config class. This one is has the minmimal set of params to visualize the configuration
+// It is a subset of ObsConfiguration such that can be built out of either the db config
+// Or the minimal config from the modes table
+case class ConfigurationForVisualization(
```
rpiaggio, line 19: Have you considered moving all the constructors to the companion object here?

Eg: `fromItcInstrumentConfig` and `fromObsConfiguration`.

That way, those classes don't have to know about this one. Also you can make the default constructor `private`.

model/shared/src/main/scala/explore/modes/ItcInstrumentConfig.scala
```diff
@@ -38,7 +41,17 @@ sealed trait ItcInstrumentConfig derives Eq:
...
 
+  // Return a configuration enough to be used for visualization
+  def toVizConfig(selectedPosAngle: Option[Angle]): Option[ConfigurationForVisualization] =
+    None
+
 object ItcInstrumentConfig:
+  // GMOS suporta a total wavelength range of 360-1030 nm
+  // https://www.gemini.edu/instrumentation/gmos
+  // the center is 360 + (1030 - 360) / 2 = 695
+  private val gmosFallbackCW: CentralWavelength =
```
rpiaggio, line 52: Capital letter for constant?

model/shared/src/main/scala/explore/model/ConfigurationForVisualization.scala
cquiroz: Seems a good idea, I'll implement it

rpiaggio (approved):
LGTM

PR #4808, opened 2025-05-05
Rename AdvancedConfigurationPanel to GmosLongslitConfigPanel

toddburnside:
I spoke with Andy and verified that it is highly unlikely we'll have another instrument with the same configuration parameters as Gmos Longslit, so I renamed `AdvancedConfigurationPanel` to `GmosLongslitConfigPanel` to reflect this fact, and removed a no longer necessary level of abstraction. 

(Probably best viewed with `ignore whitespace`.


rpiaggio (approved):
LGTM

PR #4809, opened 2025-05-05
Extract configuration OffsetsControl

toddburnside:
This will need to shared with `F2 Longslit` when offsets for F2 Longslit are added to the API.


cquiroz (approved):
LGTM

PR #4812, opened 2025-05-07
New ODB API layer

rpiaggio:
The idea here is that the application doesn't invoke the ODB directly anymore, but does so through an `OdbApi` layer that concentrates all queries and therefore makes it easier to implement error handling than doing it in each invocation scattered through the app.

The `OdbApi` is built in `AppContext` and is passed implicitly to keep invocation simpler.

For the moment, only targets and asterisms queries and mutations have been moved to the API layer. The idea is to eventually move the rest of the `explore.common.*Queries` here. I'm splitting it to make review more manageable.

No special error handling has been implemented yet.

Some components were converted to monadic hook style in the process.


explore/src/main/scala/explore/RootComponent.scala
```diff
@@ -26,34 +26,20 @@ case class RootComponent(
...
-      .render: (props, rootModel, programSummariesPot) =>
-        AppContext.ctx.provide(props.ctx):
-          React.Fragment(
-            props.ctx.tracing.map: c =>
-              val attr = rootModel.get.vault.map(ResourceAttributes.fromUserVault)
-              Observability(
-                HoneycombOptions(
-                  c.key,
-                  c.serviceName,
-                  instrumentations = instrumentations,
```
cquiroz, line 51: Why was instrumentations removed?

cquiroz (approved):
Seems like a good change. We call queries from all over the place. Having an API layer is much better.

explore/src/main/scala/explore/RootComponent.scala
rpiaggio: It wasn't used. It had an `@unused` annotation in the called site. Do we plan to use it in the future?

cquiroz: Right, I'm trying to remember why we had it

PR #4814, opened 2025-05-08
Move program and observation queries to Api layer

rpiaggio:
Subscriptions not included yet.

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -1,6 +1,6 @@
 {
   "projects": {
-    "default": "explore-gemini"
+    "default": "explore-gemini-dev"
   },
   "targets": {
     "explore-gemini": {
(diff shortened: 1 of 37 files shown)


.firebaserc
```diff
@@ -1,6 +1,6 @@
 {
   "projects": {
-    "default": "explore-gemini"
+    "default": "explore-gemini-dev"
```
rpiaggio, line 3: Unrelated change, attempt to erase the unused deployment in firebase.

toddburnside (approved):
Wow. We have a lot of api calls... LGTM

.firebaserc
rpiaggio: Yeah, that won't work. It's the name of the overarching project, not the site. We can't really remove the default site. I'll revert.

PR #4817, opened 2025-05-11
Move group, sequence and visit queries/subscriptions to API layer

rpiaggio:
Again, some components were moved to monadic style hooks in the process.


toddburnside (approved):
LGTM

PR #4821, opened 2025-05-12
Move almost all remaining ODB calls to API layer

rpiaggio:


explore/src/main/scala/explore/services/OdbObservationApiImpl.scala
```diff
@@ -362,3 +364,4 @@ trait OdbObservationApiImpl[F[_]: Sync](using StreamingClient[F, ObservationDB])
       .subscribe[F](obsId.toObservationEditInput)
       .raiseFirstNoDataError
       .ignoreGraphQLErrors
+      .map(_.throttle(5.seconds)) // TODO Do we want this in all subscriptions? Parametrize timeout?
```
toddburnside, line 367: What do we want the throttle for?

toddburnside: Ah, I see. This subscription only returns the obs id and is used as a trigger. The throttle works here, but I don't thing it would be good for subscriptions where we actually use the data from the subscription.

toddburnside (approved):
LGTM

explore/src/main/scala/explore/services/OdbObservationApiImpl.scala
rpiaggio: I agree, in the comment I mean "signal subscriptions".

PR #4822, opened 2025-05-13
OptionalEnumerated

toddburnside:
I started down the path of making a custom control for this situation, but thought this might be better. What do you think @rpiaggio and @cquiroz?

Also makes use of this for the F2ReadMode, which has a "default" of None.


toddburnside:
> I think the default is "Auto"

Well, the `display` for the default is "Auto", but the actual value is `None`.

rpiaggio (approved):
LGTM

PR #4823, opened 2025-05-13
Disable editing of observations in an "Executed" state.

toddburnside:
It turns out we were preventing editing of targets and asterisms for executed observations in explore, but not the observations themselves...

Unfortunately, we need to prevent editing the timing windows and constraints for executed observations on their respective tabs. I'll submit a new story for that.


rpiaggio (approved):
LGTM

PR #4829, opened 2025-05-14
[WIP] Establish mechanism to halt UI and reset cache in case of global errors

rpiaggio:


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -63,99 +66,113 @@ object ExploreLayout:
...
-                    <.span(
-                      "A new version of ",
-                      <.a(
-                        ExploreStyles.UpgradeLink,
-                        "Explore",
-                        ^.href   := s"https://github.com/gemini-hlsw/explore/compare/${utils.gitHash.orEmpty}...HEAD",
-                        ^.target := "_blank"
-                      ),
-                      " is available!"
+  private def useMountProgressiveWebApp(
```
cquiroz, line 111: Is this like a local hook?

rpiaggio: Yes, to organize the code a bit.

explore/src/main/scala/explore/services/OdbApiImpl.scala
```diff
@@ -10,7 +10,7 @@ import explore.utils.ToastCtx
 import lucuma.schemas.ObservationDB
 import org.typelevel.log4cats.Logger
 
-case class OdbApiImpl[F[_]: Async]()(using
+case class OdbApiImpl[F[_]: Async](resetCache: String => F[Unit])(using
```
cquiroz, line 13: is that function beig used?

rpiaggio: Not yet, but it will. I sent this cutoff PR to keep the size manageable.

cquiroz:
This is looking good, much better than the forever spinning wheel. 
In some cases, errors come because the backend changed but the local explorer build hasn't. I wonder if we should have a button there to force a reload of the PWA.
We can do that later.

cquiroz (approved):
LGTM

PR #4830, opened 2025-05-15
Process errors in mutations and queries

rpiaggio:
Bootstrap queries with cache resetting mechanism.

In most cases this is overkill and we could reload parts of the program cache instead of the whole thing. We can address that in the future.

Still missing:
- Handling of errors in edit events subscriptions.


explore/src/main/scala/explore/services/OdbApiHelper.scala
```diff
@@ -0,0 +1,61 @@
...
+      fa.raiseGraphQLErrorsOnNoData.processOdbErrors.resetCacheOnError
+
+  protected def drain[A, Id, R](
+    fetch:      Option[Id] => F[R],
+    getList:    R => List[A],
+    getHasMore: R => Boolean,
+    getId:      A => Id
+  ): F[List[A]] = {
+    def go(id: Option[Id], accum: List[A]): F[List[A]] =
+      fetch(id).flatMap(result =>
```
cquiroz, line 53: tailrec?

cquiroz (approved):
LGTM

cquiroz:
can't wait? you know we may need to fix those erros

explore/src/main/scala/explore/services/OdbApiHelper.scala
rpiaggio: I think that's done automatically in `F` land.

PR #4831, opened 2025-05-16
Process subscription errors, remove reset on some errors

rpiaggio:
This is the final PR in the series.

Some queries should not trigger a cache reload, this addresses it; as well as triggering a reset on errors in cache subscriptions.


toddburnside (approved):
LGTM

PR #4832, opened 2025-05-16
Add Imaging Modes Table - Step 1

toddburnside:
This is the first step in the Imaging Modes table. It simply adds the table and shows all of the rows. 

In the following steps I will do things like add the calls to the ITC, allow selection and multi-selection of rows, scroll to selected rows, etc. I will try to extract out more logic common to the spectroscopy modes table.

This PR also updates the `hasura` endpoints for our migration and promotion scripts.

image

### DIFF
--- a/explore/src/main/scala/explore/config/ImagingModesTable.scala
+++ b/explore/src/main/scala/explore/config/ImagingModesTable.scala
@@ -1,0 +1,227 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.config
+
+import cats.Order
+import cats.data.EitherNec
+import cats.implicits.catsKernelOrderingForOrder
+import cats.syntax.all.*
+import crystal.Pot
+import explore.common.UserPreferencesQueries.TableStore
+import explore.components.HelpIcon
+import explore.components.ui.ExploreStyles
+import explore.model.AppContext
+import explore.model.Progress
+import explore.model.display.*
+import explore.model.enums.TableId
+import explore.model.enums.WavelengthUnits
+import explore.model.itc.*
+import explore.modes.*
+import explore.syntax.ui.*
+import japgolly.scalajs.react.*
+import japgolly.scalajs.react.vdom.html_<^.*
+import lucuma.core.enums.*
+import lucuma.core.math.Angle
+import lucuma.core.math.BoundedInterval
+import lucuma.core.math.BoundedInterval.*
+import lucuma.core.math.SignalToNoise
+import lucuma.core.math.Wavelength
+import lucuma.core.math.WavelengthDelta
+import lucuma.core.model.User
+import lucuma.core.syntax.all.*
+import lucuma.core.util.Display
+import lucuma.core.util.TimeSpan
+import lucuma.react.common.ReactFnProps
+import lucuma.react.syntax.*
+import lucuma.react.table.*
+import lucuma.refined.*
+import lucuma.ui.reusability.given
+import lucuma.ui.table.*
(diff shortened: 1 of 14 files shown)


explore/src/main/scala/explore/config/ImagingModesTable.scala
```diff
@@ -0,0 +1,227 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.config
+
+import cats.Order
```
cquiroz, line 6: Nice set of imports, was it hard?

explore/src/main/scala/explore/config/ImagingModesTable.scala
```diff
@@ -0,0 +1,227 @@
...
+  matrix: ImagingModesMatrix,
+  units:  WavelengthUnits
+) extends ReactFnProps(ImagingModesTable.component)
+
+object ImagingModesTable extends ModesTableCommon:
+
+  private given Reusability[ImagingModesMatrix] = Reusability.by(_.matrix.length)
+
+  extension (interval: BoundedInterval[Wavelength])
+    // The 'z' filter, at least, has an upper bound of Int.MaxValue picometers. However,
```
cquiroz, line 55: Should we fix this in core?

explore/src/main/scala/explore/config/ImagingModesTable.scala
```diff
@@ -0,0 +1,227 @@
...
+        row => ImagingModeRow.filter.get(row.entry).wavelengthRangeAndDelta
+      )
+        .withHeader(s"Δλ ${units.symbol}")
+        .withCell: cell =>
+          val (range, delta) = cell.value
+          <.span(delta.fold("-")(_.shortName))
+            .withOptionalTooltip(range.map(r => s"${r.shortName} ${units.symbol}"))
+        .withColumnSize(FixedSize(100.toPx))
+        .sortableBy(_._2),
+      column(FovColumnId, _.entry.fov)
```
cquiroz, line 169: will we display if the filter is narrow/broad?

model/shared/src/main/scala/explore/modes/ItcInstrumentConfig.scala
```diff
@@ -24,6 +24,7 @@ sealed trait ItcInstrumentConfig derives Eq:
   val grating: Grating
   def gratingDisplay: Display[Grating]
   def gratingStr: String = gratingDisplay.shortName(grating)
```
cquiroz, line 26: I was tempted to make a deeper structure with ImagingIItcINstrumentConfig and SpectroscopyItcINstrurmentConfig and remove `gratingStr` but it seemed overkill

cquiroz (approved):
LGTM

explore/src/main/scala/explore/config/ImagingModesTable.scala
toddburnside: They were all very obvious. 👅

explore/src/main/scala/explore/config/ImagingModesTable.scala
toddburnside: We can. It wasn't in the mockup. I suppose it probably should be. It can be `Narrow` `Broad` or `Combination` for the imaging filters. (so far)

explore/src/main/scala/explore/config/ImagingModesTable.scala
swalker2m: The values come from the [GMOS instrument components page](https://www.gemini.edu/instrumentation/gmos/components#Filters).  If it makes sense to limit them to the effective upper limit that sounds good to me.  I don't believe we actually _use_ the width for anything in the ODB.

PR #4835, opened 2025-05-19
Search both simbad sites in case one is down

toddburnside:
At some point it might be nice to make a Catalog client, but I don't think it is a priority now.


cquiroz (approved):
LGTM

PR #4849, opened 2025-06-02
Validate proposal for submission and show errors in a tile.

toddburnside:
Displays the errors in a very simple format in a new tile on the Proposals Tab.
image

 Also:
- Fixes the "Download Templates" link in the attachments tile title to look like a button and prevents it from being the full width of the title bar. Before, the tooltip would appear with the mouse anywhere in the middle section of the title bar and was confusing.
- Don't reload the program cache for errors while updating the proposal status. We don't update anything locally unless the call is successful.
- Display errors from updating the proposal status in a toast instead of the error message in the submission bar. We should have eliminated most error responses from the API with our new validations (or will have when the API is updated to properly validate proposal submission). This results in better behavior of the error message as reported in https://app.shortcut.com/lucuma/story/5699/proposal-error-only-cleared-by-page-reload.

### DIFF
--- a/common/src/main/scala/explore/components/Tile.scala
+++ b/common/src/main/scala/explore/components/Tile.scala
@@ -101,6 +101,9 @@ case class Tile[A](
 object Tile:
   type TileId = NonEmptyString
 
+  def dummyTile(id: TileId): Tile[Unit] =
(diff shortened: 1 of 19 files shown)


rpiaggio (approved):
LGTM

PR #4856, opened 2025-06-09
Add additional proposal validation

toddburnside:
It also keeps the `Submit` button enabled even if there are errors so that Andy can more easily compare against the ODB validations. After he completes testing, this part will be reverted.


cquiroz (approved):
LGTM

PR #4861, opened 2025-06-11
Use CalculatedValue[ExecutionDigest] from ODB

toddburnside:
I still need to do the CalculatedValue for the workflows, but figured this was enough for one PR.


model-testkit/shared/src/main/scala/explore/model/arb/ArbObservation.scala
```diff
@@ -126,7 +131,7 @@ trait ArbObservation:
        Option[Configuration],
        ObservationWorkflow,
        Option[Group.Id],
-       Short
+       (Short, Execution)
```
toddburnside, line 134: Apparently there is a tuple size limit of 22 here. It took me awhile to figure that out. 😦

explore/src/main/scala/explore/observationtree/ObsSummaryColumns.scala
```diff
@@ -124,9 +121,8 @@ trait ObsSummaryColumns:
           case s: String => s
           case (_, b)    => b.target.name.value
 
-    extension (a: Option[PerishablePot[Option[TimeSpan]]])
-      def sortableTimeSpan =
-        a.flatMap(_.asValuePot.toOption).flatten
+    extension (a: Option[CalculatedValue[Option[TimeSpan]]])
```
cquiroz, line 124: nice type

model-testkit/shared/src/main/scala/explore/model/arb/ArbObservation.scala
cquiroz: I got thit by that recently as well

PR #4862, opened 2025-06-11
Fix exposure time mode editing in the advanced config panels

toddburnside:
Only the local model was getting updated when editing on the advanced configuration panels. The API was not updated.


cquiroz:
is this for imaging?

toddburnside:
> is this for imaging?

This is for the GMOS and F2 spectroscopy config panels, but will be for all config panels. 
The exposure time mode is independent of spectroscopy or imaging.

cquiroz:
Weird we didn't notice this before

toddburnside:
> Weird we didn't notice this before

I only noticed it because I was editing S/N in order to trigger observation duration changes to test my last PR. And nothing happened.

PR #4877, opened 2025-06-14
Switch to GaiaClient

rpiaggio:


workers/src/main/scala/workers/CatalogCache.scala
```diff
@@ -31,18 +25,14 @@ import java.time.temporal.ChronoField
 import java.time.temporal.ChronoUnit
 
 trait CatalogQuerySettings {
-  val proxy = uri"https://cors-proxy.lucuma.xyz"
```
cquiroz, line 34: can we remove the proxy then?

cquiroz (approved):
LGTM

PR #4881, opened 2025-06-17
Use calculatedWorkflow

toddburnside:
Most of the time when the workflow is stale, it won't really be changing so I don't make the observation readonly or anything like that. If the user edits it when it WILL be changing (for example to OnGoing) it will still be caught by validations in the ODB. This is really no different than before this PR because we wouldn't be informed of the new workflow until after it was recalculated.

I do disable the dropdown for selecting a new workflow state and put a tooltip there while it is stale. I didn't think this would be too intrusive and would give some indication of what is going on.


rpiaggio (approved):
LGTM

PR #4884, opened 2025-06-17
Implement missing part of global error handling

rpiaggio:
When I implemented global error handling, I changed the way the UI informs the user when it's populating the cache. But I forgot to hook the new mechanism to the cache controller.

I finally remembered how I intended to do this and implemented it here.


toddburnside (approved):
LGTM

PR #4894, opened 2025-06-23
Add Fast turnaround reviewer and mentor.

toddburnside:
It might even follow all the rules...

The API says it will default the reviewer to the PI. Explore will always explicitly set it to the PI by default, but it can be unset via the API. I wonder if, instead of requiring the ODB to figure out the default, we should require it to be set in order to submit a proposal, and have the only defaulting be in the UI.

### DIFF
--- a/explore/src/main/scala/explore/proposal/ProposalEditor.scala
+++ b/explore/src/main/scala/explore/proposal/ProposalEditor.scala
@@ -24,6 +24,8 @@ import explore.model.ProgramDetails
 import explore.model.ProgramUser
 import explore.model.Proposal
 import explore.model.ProposalTabTileIds
+import explore.model.ProposalType
+import explore.model.ProposalType.FastTurnaround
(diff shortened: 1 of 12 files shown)


explore/src/main/scala/explore/proposal/ProposalEditor.scala
```diff
@@ -125,13 +128,115 @@ object ProposalEditor
...
+              val newMentorId =
+                if (reviewer.orElse(props.pi).exists(_.hasPhd))
+                  none
+                else
+                  (ft.mentorId, mentor) match
+                    case (Some(_), None)                  =>
+                      // mentor was deleted
+                      none
+                    case (Some(id), Some(m)) if !m.hasPhd =>
+                      // The mentor has probably been downgraded from a PhD, so remove them
```
cquiroz, line 208: that would be odd 😄

PR #4905, opened 2025-06-30
Remove double planets

cquiroz:


explore/src/main/scala/explore/ExploreLayout.scala
```diff
@@ -205,8 +204,7 @@ object ExploreLayout:
                   position = Sidebar.Position.Bottom,
                   content = error,
                   clazz = ExploreStyles.GlobalErrorDialog
-                ),
```
rpiaggio, line 208: This will remove the planets in the case of error while Explore refreshes its cache. We want them there to tell the user that the application is not stuck.

toddburnside (approved):
LGTM. Thanks for fixing this.

PR #4906, opened 2025-06-30
Improve ITC error message for target problems.

toddburnside:
This changes a couple of things with respect to errors:

1. Instead of filtering out targets that aren't valid for a call to ITC and continuing on with the remaining ones (if any), an invalid target will prevent a call to the ITC.
2. Instead of just having a `-` in the `Time` or `S/N` columns in the modes tables when the ITC is not called, the warning triangle is placed on each row of the table to make it clear that the modes are not valid. A message is also placed next to the target on the upper right of the table.

image
 
I started refactoring the graph querying code (`ItcGraphQuerier` and related), but it was going to complicate this PR a lot. I may do this is a separate PR. I think as the code has changed, the graph querying has become more complicated than it needs to be.


cquiroz (approved):
LGTM

PR #4907, opened 2025-06-30
New itc

cquiroz:


workers/src/main/scala/workers/itc/ITCRequests.scala
```diff
@@ -53,9 +53,10 @@ object ITCRequests:
     cache:               Cache[F],
     callback:            Map[ItcRequestParams, EitherNec[ItcTargetProblem, ItcResult]] => F[Unit]
   )(using Monoid[F[Unit]], ItcClient[F]): F[Unit] = {
-    def itcResults(r: ClientCalculationResult): EitherNec[ItcTargetProblem, ItcResult] =
+    def itcResults(r: ClientAllResults): EitherNec[ItcTargetProblem, ItcResult] =
```
toddburnside, line 56: This will return a `NonEmptyList[EitherNec[ItcTargetProblem, ItcResult]]`?

toddburnside (approved):
LGTM

workers/src/main/scala/workers/itc/ITCRequests.scala
toddburnside: Yes, but that will be the behavior for multiple modes, right? Just loop through the `all`?

cquiroz: Yes I expect that you'd loop over all the `modes`

PR #4912, opened 2025-07-02
Improve itc graph error handling and flow

toddburnside:
The main thing I wanted to achieve was having all of the necessary information in the return from `ItcGraphQuerier.requestGraphs`. Before the ItcGraphQuerier itself had to be passed around as well and error messages checked for separately, etc. 

 Also:
- Some errors were handled as an IO error. They are now represented in an EitherNec.
- The target dropdown in the tile title was sometimes empty. It now always has a value.
- The target specific error messages displayed in the high charts chart did not actually updated when the selected target changed.They are now displayed as a Message, like the other errors are.


cquiroz (approved):
LGTM
Thanks a lot, yeah error handling has been spread all over while trying to work out the kinks of ITC
we should be ready to add warnings 😄

PR #4915, opened 2025-07-03
Ensure container login for itc promotion

rpiaggio:


cquiroz (approved):
LGTM let's also delete the scripts that only promote explore

PR #4928, opened 2025-07-08
Show correct value for workflow state in obs summary table

toddburnside:
This is related to the change to a CalculatedValue for the workflow. The table was showing the state of the calculation, not the state of the workflow. Both of them have `Ready` states. So, in this case it is good they were all going to `Ready`.


cquiroz (approved):
LGTM

PR #4931, opened 2025-07-08
Update lucuma-core 0.136.3 to add units to ImageQuality

toddburnside:
I also added tooltips to the `Constraints` and `Configuration` columns in the observation summary table since these cells tend to be quite long.


model/shared/src/main/scala/explore/model/display.scala
```diff
@@ -63,16 +62,6 @@ trait DisplayImplicits:
     case ScienceMode.Imaging      => "Imaging"
     case ScienceMode.Spectroscopy => "Spectroscopy"
 
-  @targetName("given_Display_ImageQuality_Preset")
```
toddburnside, line 66: These were already defined in the companion objects in lucuma-core.

PR #5013, opened 2025-08-08
Updated promote script to send slack messages

cquiroz:


toddburnside (approved):
As far as I can tell, it seems fine. Thanks for doing this.

PR #5015, opened 2025-08-08
Use the ProgramUser fallback profile as the preferred profile

toddburnside:


model/shared/src/main/scala/explore/model/ProgramUser.scala
```diff
@@ -30,18 +30,18 @@ case class ProgramUser(
   thesis:            Option[Boolean],
   gender:            Option[Gender],
   affiliation:       Option[NonEmptyString],
-  fallbackProfile:   UserProfile,
+  preferredProfile:  UserProfile,
   invitations:       List[UserInvitation],
   hasDataAccess:     Boolean
 ) derives Eq:
   val name: String          =
-    user.fold(fallbackProfile.displayName.orEmpty)(_.name)
```
toddburnside, line 38: The change in name and email are some of the few real changes.

explore/src/main/scala/explore/users/ProgramUsersTable.scala
```diff
@@ -326,24 +327,24 @@ object ProgramUsersTable:
...
             val canEdit                       = meta.canEditUserFields(pu)
 
-            // We'll allow editing the name if there is no REAL user
-            // AND the fallback display name is the same as the credit name.
             // In explore, we'll always set the credit name, but a user could have
-            // updated the fallback profile via the API, and we won't mess with that.
-            if canEdit && pu.user.isEmpty && pu.fallbackProfile.creditName === pu.fallbackProfile.displayName
+            // updated the preferred profile via the API and set given and family names.
+            // These would be the fallback before the user.profile.
+            if canEdit
```
toddburnside, line 333: The rules for when the name and email can be edited has changed such that the user can always override the orcid profile. Before, if the orcid profile existed it would use those and prevent additional editing.

PR #5026, opened 2025-08-12
Setup sops to encode the secrets

cquiroz:


toddburnside (approved):
LGTM, after you fix the linting.
You may also want to update the README to say if you are using nix you need to have the keys file. Otherwise you'll have the bogus environment variables.

PR #5091, opened 2025-09-09
Avoid ITC chart redraws

cquiroz:
and filter out empty warnings

### DIFF
--- a/explore/src/main/scala/explore/config/ModesTableCommon.scala
+++ b/explore/src/main/scala/explore/config/ModesTableCommon.scala
@@ -178,17 +178,26 @@ trait ModesTableCommon:
         )
     )
 
-  def tooltipContent(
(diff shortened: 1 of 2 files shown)


explore/src/main/scala/explore/config/ModesTableCommon.scala
```diff
@@ -248,7 +257,7 @@ trait ModesTableCommon:
             val baseText = s"${r.snAt.map(_.single.value).foldMap(_.format)} / exposure"
             tooltipContent(baseText, r.ccdWarnings)
 
-        (if (r.ccdWarnings.nonEmpty)
+        (if (filterWarnings(r.ccdWarnings).nonEmpty)
```
rpiaggio, line 260: You can call `filterWarnings` once after line 241 and pass the result to both invocations of `tooltipContent` and this check in line 260.

PR #5096, opened 2025-09-10
Region Editor for ToOs

toddburnside:
image

### DIFF
--- a/explore/src/main/scala/explore/targeteditor/RegionEditor.scala
+++ b/explore/src/main/scala/explore/targeteditor/RegionEditor.scala
@@ -1,0 +1,120 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package explore.targeteditor
+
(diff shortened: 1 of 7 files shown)


explore/src/main/scala/explore/targeteditor/RegionEditor.scala
```diff
@@ -0,0 +1,120 @@
...
+
+        val decArcView   = props.region.zoom(Region.decArc)
+        val decStartView = decArcView.zoom(Arc.start[Declination])
+        val decEndView   = decArcView.zoom(Arc.end[Declination])
+
+        React.Fragment(
+          FormEnumDropdownView(
+            id = "ra-arc-type".refined,
+            value = raType.withOnMod(at => raArcView.set(at.toRaArc)),
+            label = React.Fragment("RA Arc", HelpIcon("target/main/region.md".refined)),
```
toddburnside, line 66: I don't know what the intended path scheme is for the help files. I just used the same path that was being used for RA/Dec for sidereal targets.

toddburnside:
> Nice. Screenshot?

It's pretty boring, but I added a screenshot. 😄

PR #5135, opened 2025-09-25
Don't request ITC for certain calibration, e.g. twilight

cquiroz:
Twilights have no target info, only coordinates but no SED or brightnesses thus itc calls will fail

This PR makes requesting itc optional and in particular disabled for twilights


toddburnside (approved):
LGTM

PR #5141, opened 2025-09-29
Update aladin to use the official npm lib

cquiroz:


explore/src/main/scala/explore/targeteditor/AladinContainer.scala
```diff
@@ -286,12 +286,6 @@ object AladinContainer extends AladinCommon {
 
         // Use fov from aladin
         fov        <- useState(none[Fov])
-        // Update aladin if coords change
```
rpiaggio, line 289: I guess the new aladin takes care of this now?

rpiaggio (approved):
LGTM

